PR #5, opened 2017-01-19
Zephyr specific updates

rsalveti:
Restructuring the Zephyr build system to make it easier to support new targets.

Also replacing printks with SYS_LOG, as it can be managed via configs (and disabled when needed).

Tested on 96boards Carbon and Nitrogen.


rsalveti:
@d3zd3z mind taking a look?

boot/zephyr/main.c
```diff
@@ -60,11 +60,12 @@ void main(void)
 	}
 
 	printk("Bootloader chain: 0x%x\n", rsp.br_image_addr);
-	vt = (struct vector_table *)(rsp.br_image_addr +
-				     rsp.br_hdr->ih_hdr_size);
+	vt = (struct vector_table *)(rsp.br_image_addr + TEXT_SECTION_OFFSET);
```
d3zd3z: I wouldn't expect this to work. The bootloader should be built with a TEXT_SECTION_OFFSET of 0. Only the programs booted should themselves have it set.

boot/zephyr/main.c
```diff
@@ -60,11 +60,12 @@ void main(void)
...
 	irq_lock();
 	_MspSet(vt->msp);
 
+	printk("Setting vector table to %p\n", vt);
+
 	/* Not all targets set the VTOR, so just set it. */
 	_scs_relocate_vector_table((void *) vt);
```
d3zd3z: Perhaps this can be removed, now that we're going to be setting this properly in Zephyr.

boot/zephyr/targets/frdm_k64f.h
```diff
@@ -27,3 +27,6 @@
 #define FLASH_AREA_IMAGE_1_SIZE		0x20000
 #define FLASH_AREA_IMAGE_SCRATCH_OFFSET	0x60000
 #define FLASH_AREA_IMAGE_SCRATCH_SIZE	0x20000
+
+/* Same offset as used by the Zephyr app (CONFIG_TEXT_SECTION_OFFSET) */
+#define TEXT_SECTION_OFFSET		0x100
```
d3zd3z: Needs to be 0x200 for K64F

boot/zephyr/main.c
d3zd3z: This really needs to be the value read out of the header, which will then work with however it got built. There isn't really a reason to code this value into the bootloader.

boot/zephyr/targets/96b_nitrogen.h
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @file
+ * @brief Bootloader device specific configuration.
+ */
+
+#define FLASH_DRIVER_NAME		"NRF5_FLASH"
+#define FLASH_ALIGN			1
```
d3zd3z: Make sure this is really right. Having it too small can excessively wear the flash device.

boot/zephyr/flash_map.c
```diff
@@ -34,28 +36,22 @@ extern struct device *boot_flash_device;
 static const struct flash_area part_map[] = {
 	{
 		.fa_id = FLASH_AREA_IMAGE_0,
-		.fa_off = 0x20000,
-		.fa_size = 0x20000,
+		.fa_off = FLASH_AREA_IMAGE_0_OFFSET,
```
d3zd3z: I think this is good for now. It is likely that we will bring in the partition table code from mynewt into mcuboot, which will make all of this quite different, though.

boot/zephyr/main.c
rsalveti: This is the vector table of the app itself, right? That offset will depend not on the header, but on CONFIG_TEXT_SECTION_OFFSET, used by the Zephyr app.

Not sure how to best align both things here, which is why I had it defined per device.

boot/zephyr/targets/96b_nitrogen.h
rsalveti: Yeah, same value as used by mynewt, so fine to have that now.

boot/zephyr/flash_map.c
rsalveti: Yeah, I just wanted to improve it a bit enough to remove the hardcoded values. The best to do here is parsing mynewt's partition table, to avoid duplicating info.

boot/zephyr/main.c
d3zd3z: It _only_ depends on the value in the header. I can't think of any reason to define this and not use the header. The header is part of the application.

boot/zephyr/prj.conf
```diff
@@ -11,4 +11,3 @@ CONFIG_MBEDTLS_CFG_FILE="config-boot.h"
 CONFIG_HEAP_MEM_POOL_SIZE=16384
 
 CONFIG_FLASH=y
-CONFIG_SOC_FLASH_STM32F4=y
```
d3zd3z: Did you do something to Zephyr to make this work? When I did it, there were two flash devices defined for the target so Kconfig picked neither.

rsalveti: That's probably because of https://github.com/Linaro/zephyr/commit/942edd44751ce45dd4b979302b648e5d90b68a1b. I'm hoping the upstream version to be merged soon, then we will have the right fix at the right place.

boot/zephyr/main.c
rsalveti: Makes sense need to make sure the header then gets the right value, will revert this change.

rsalveti: Reverted.

d3zd3z:
Looks good to me. I'd like to get a +1 from a mynewt person, because of the image_rsa change.

ccollins476ad:
It looks good to me, too.  Thanks Ricardo.

PR #6, opened 2017-01-23
Fix issues with strict devices and large alignment

d3zd3z:
The frdm-k64f has an 8-byte alignment for flash writes, and the driver is strict about the writes not only happening on alignment boundaries, but must be a multiple of the alignment size in length. Without this, the update silently goes on, but ends up corrupting images, often resulting in repeated swapping on each boot, because none of the status has been written.

The series starts with some Zephyr script changes to make it easier to reproduce this problem. This is followed by a k64f patch that changes the name of the flash device to match the upstream master branch. Following this are a series of 'sim' patches to enable the simulator to test for this situation. Lastly there is the small patch to the bootloader code itself that changes the status updates to always write at least 'align' bytes when writing status values.

This fixes on-target and simulation results for alignment sizes of 1, 2, 4, and 8. Note, however, that there is one powerdown case that fails when the alignment is set to 8. This needs to be investigated separately.


ccollins476ad:
Hi David,

This is great.  Thanks for doing it.

I'll merge it anyway, but what do you think about using a `#define` rather than the hard coded`8` for the buffer size?

Thanks,
Chris

PR #8, opened 2017-01-25
zephyr: fix _scs_relocate_vector_table build break

Hashcode:
In the master branch of Zephyr, the _scs_relocate_vector_table function
was removed and replaced by a static function relocate_vector_table()
(_scs_relocate_vector_table() was also static but was in a header
include so we were able to use it.)

The new function uses CMSIS defines to replace the assembly code
previously used.

We can follow that change here by creating the same CMSIS friendly
function accepting the vector table address.  This will also work for
building mcuboot in versions of Zephyr previous to the build break.

Signed-off-by: Michael Scott <michael.scott@linaro.org>


d3zd3z:
I'm wondering if we shouldn't just remove this code entirely, now that Zephyr sets the vtor for all of the cortex-m targets properly.

Hashcode:
As long as you're ok not being backwards compatible.  I'd say remove the code if we don't need it.

rsalveti:
Yeah, +1 for removing it, if it works fine with the supported targets. Code is new, we can manage the differences in our own branch.

d3zd3z:
Of 'carbon' and 'k64f' it only affects the k64f. I only put that code in there while coming up with the proper fix for k64f.

d3zd3z:
I agree, lets remove this code instead of fixing it.

boot/zephyr/flash_map.c
```diff
@@ -93,9 +93,14 @@ int flash_area_read(const struct flash_area *area, uint32_t off, void *dst,
...
 		     uint32_t len)
 {
+	int rc = 0;
+
 	SYS_LOG_DBG("%s: area=%d, off=%x, len=%x", __func__,
 			area->fa_id, off, len);
-	return flash_write(boot_flash_device, area->fa_off + off, src, len);
+	flash_write_protection_set(boot_flash_device, false);
+	rc = flash_write(boot_flash_device, area->fa_off + off, src, len);
+	flash_write_protection_set(boot_flash_device, true);
```
d3zd3z: I wasn't clear on the docs for this. It sounded like doing the write would clear the write protection. But, this way of using it makes more sense. Does your device leave the flash un-protected until the write protection is re-set?

boot/zephyr/main.c
```diff
@@ -73,7 +81,7 @@ void main(void)
 	SYS_LOG_INF("Setting vector table to %p", vt);
 
 	/* Not all targets set the VTOR, so just set it. */
-	_scs_relocate_vector_table((void *) vt);
+	relocate_vector_table((void *) vt);
```
d3zd3z: Have we determined we want to set the vector table in the bootloader at all? It should be fixed for all Cortex-M targets in Zephyr, and I would argue if any are missing, it should be fixed there, and this code just removed entirely from here.

d3zd3z: The call to _scs_relocate_vector_table() was just a hack I added, and kind of accidentally committed.

boot/zephyr/flash_map.c
mike-scott: The docs say that the device can re-enable write protection after a write, so this method is preferred.

boot/zephyr/main.c
mike-scott: I can rebase the last 2 patches ontop of whatever fix you want to do for the _scs_relocate_vector_table issue.

mike-scott: For reference: https://gerrit.zephyrproject.org/r/gitweb?p=zephyr.git;a=blob;f=include/flash.h;h=1d90a22fd2f0e47bcecb938f06b73ac73f794d57;hb=refs/heads/master#l107

boot/zephyr/flash_map.c
mike-scott: For reference here's the flash call description mentioning that the device re-enables the write protection after each completed call:
https://gerrit.zephyrproject.org/r/gitweb?p=zephyr.git;a=blob;f=include/flash.h;h=1d90a22fd2f0e47bcecb938f06b73ac73f794d57;hb=refs/heads/master#l107

mbolivar: We should fix this upstream in the future, I think.

boot/zephyr/main.c
d3zd3z: I guess what I'm asking is to just remove the _scs_relocate_vector_table line entirely. It really doesn't belong in the bootloader. Zephyr should be setting the VTOR when it starts, so the applications we run should work fine. If there is a problem there, it should be fixed in Zephyr.

mike-scott: I'll update this pull request tomorrow after a test OK K64F.

mike-scott: s/OK/on

d3zd3z:
Looks good to me in this form.

PR #10, opened 2017-01-31
imgtool: Create simple signing tool

d3zd3z:
The imgtool is a simple image signing tool based on the libraries in the
`newt` tool, but appropriate for images that live outside of a Mynewt
project.

This initial version is able to generate, extract public keys from, and
sign ECDSA P-256, P-224, and RSA keys.  The signing tool is also able
to add an image trailer to mark this image for upgrade.


ccollins476ad:
Awesome.  Looks good to me.

PR #12, opened 2017-02-02
Validate slot zero before booting

d3zd3z:
Instead of just checking the upgrade image signature, check the
signature on each boot.  This helps to prevent rogue images being flash
by a means other than the upgrade process.

This feature is controlled by whether BOOTUTIL_VALIDATE_SLOT0 is
defined.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -545,13 +545,13 @@ split_image_check(struct image_header *app_hdr,
 }
 
 static int
-boot_validate_slot1(void)
+boot_validate_slot(int slot)
 {
     const struct flash_area *fap;
     int rc;
     
-    if (boot_data.imgs[1].hdr.ih_magic == 0xffffffff ||
-        boot_data.imgs[1].hdr.ih_flags & IMAGE_F_NON_BOOTABLE) {
+    if (boot_data.imgs[slot].hdr.ih_magic == 0xffffffff ||
+        boot_data.imgs[slot].hdr.ih_flags & IMAGE_F_NON_BOOTABLE) {
 
         /* No bootable image in slot 1; continue booting from slot 0. */
         return -1;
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/loader.c
```diff
@@ -1039,6 +1040,12 @@ boot_go(struct boot_rsp *rsp)
 
     switch (swap_type) {
     case BOOT_SWAP_TYPE_NONE:
+#ifdef BOOTUTIL_VALIDATE_SLOT0
```
ccollins476ad: I guess this is more a nitpick than anything else, but what do you think about using `#if` here rather than `#ifdef`?  Undefined symbols resolve to 0 in preprocessor conditionals, so it works if the macro is not defined.  The plus side is that someone can define the symbol to 0 and the feature won't get enabled.

ccollins476ad: Then again, if your build system defines these symbols to to the empty string, this won't work.  I suppose `#ifdef` is more standard.

ccollins476ad:
Looks good to me.  Ignore my ill-conceived comment!

boot/bootutil/src/loader.c
d3zd3z: Well -DSYMBOL should define it to 1, so it usually works. But ifdef is more common. I'm not sure if you wanted to use the Newt specific configuration macro.

ccollins476ad: Yeah, I think #ifdef is fine.  I just have a memory from long ago of getting "burned" by this.  We were using a config system similar to Mynewt's (header file full of #defines).  I wanted to be sure a feature was disabled, so I defined the corresponding symbol to 0.  Because the code checked if the macro was defined, rather than its actual value, I actually ended up enabling the feature.

As you said, #ifdef is more common, so it is probably best to stick with that.

PR #14, opened 2017-02-06
MCUB-3 Add sample Mynewt boot app.

ccollins476ad:
This commit implements the sample Mynewt boot app.

I'm a little uncertain about the directory structure (apps/boot).  I chose one that makes sense for Mynewt.  I think I am not familiar enough with Zephyr to know how anti-social this is.  Other options are:
* Rename app to mynewt-boot; other OS apps are siblings.
* Create OS-specific top-level directories (mynewt/apps/boot).

I don't think it is a problem to rearrange things when we have a better idea of what we're going for, so I opted for simplicity for now.

### DIFF
--- a/apps/boot/README.md
+++ b/apps/boot/README.md
@@ -1,0 +1,6 @@
+# mcuboot - apps/boot
+
+This sample app implements a boot loader for the Mynewt OS (apache.mynewt.org).
+This app requires the following Mynewt repositories:
+    * @mcuboot (this one)
+    * @apache-mynewt-core
--- a/apps/boot/pkg.yml
+++ b/apps/boot/pkg.yml
@@ -1,0 +1,35 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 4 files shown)


mbolivar:
Hi, I'm doing some related work on logging here:

https://github.com/mbolivar/mcuboot/commits/logging

The idea is to make it possible to log from the core bootutil code when running on an MCU, but for those logging definitions to get compiled out (or potentially do something different) when running in the simulator, etc.

I'd like to know how best to integrate this with the mynewt code, especially if you're interested in having mynewt, zephyr, and 'core bootutil' logging live comfortably alongside. In particular, this commit adds some OS-specific definitions guarded by ifdefs here:

https://github.com/mbolivar/mcuboot/commit/7784609fcd650adf950d1b9e4c37cb36f903ac8c#diff-c3c4a4de3f0dc28dc17c98709523a99a

I can see a few options for how to proceed.

1. Add an `#ifdef __SOME_MYNEWT_DEFINE_HERE__` section that provides mynewt-specific definitions for `BOOT_LOG_LEVEL_xxx` and `BOOT_LOG_xxx`.
2. Rework bootutil_log.h to a common no-op case and separate target-specific headers that get included appropriately according to build configuration
3. Have separate logging calls for bootutil, zephyr, and mynewt code.

Of these, I think 1. and 2. seem best -- thoughts? Between them, 1. seems simplest for now, given that we're not likely to have additional platforms to support soon, to my knowledge. 2. is the best in the long term, and it'd make sense to do that refactor if we wanted to add another platform.

Seem reasonable? Any other suggestions?

Thanks!

d3zd3z:
I don't think it is ideal for the Zephyr Makefile to be at the top, but its build system currently requires this (or it starts writing files outside of the output directory). I think this means that the Zephyr code currently has to live next to the bootutil directory. I'd like to fix this, but I think it involves some Makefile magic work in Zephyr.

d3zd3z:
Oh, and Zephyr isn't going to care about other directories appearing. The only directories will have to be explicitly named in the top level Makefile, or one of the subdirectory Makefiles. I'm guessing that Mynewt will ignore these makefiles, but if we add a third OS that also wants Makefiles, we might have to come up with something quite different in order to build this code.

d3zd3z:
+1, though, I'm ok with this code.

PR #15, opened 2017-02-07
Logging

mbolivar:
This patch series adds a common logging abstraction to mcuboot, which targets can customize to use their logging backends. Currently, a no-op backend for the simulator and a Zephyr-specific backend are provided.

Simulator tests all pass:

```
plop: sim (logging) mbolivar$ ./target/release/bootsim run --device k64f
ERROR:bootsim: 0 out of 577 failed 0.00%
1|plop: sim (logging) mbolivar$ ./target/release/bootsim run --device Stm32f4
ERROR:bootsim: 0 out of 391 failed 0.00%
plop: sim (logging) mbolivar$ ./target/release/bootsim run --device K64fBig
ERROR:bootsim: 0 out of 391 failed 0.00%
```

This may interact with https://github.com/runtimeco/mcuboot/pull/14.

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil_log.h
+++ b/boot/bootutil/include/bootutil/bootutil_log.h
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2017 Linaro Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef H_BOOTUTIL_LOG_H_
+#define H_BOOTUTIL_LOG_H_
+
+#include "ignore.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * When building for targets running Zephyr, delegate to its native
+ * logging subsystem.
+ *
+ * In this case:
+ *
+ * - BOOT_LOG_LEVEL determines SYS_LOG_LEVEL,
+ * - BOOT_LOG_ERR() and friends are SYS_LOG_ERR() etc.
+ * - SYS_LOG_DOMAIN is unconditionally set to "MCUBOOT"
+ */
+#ifdef __ZEPHYR__
+
+#define BOOT_LOG_LEVEL_OFF	SYS_LOG_LEVEL_OFF
+#define BOOT_LOG_LEVEL_ERROR	SYS_LOG_LEVEL_ERROR
+#define BOOT_LOG_LEVEL_WARNING	SYS_LOG_LEVEL_WARNING
+#define BOOT_LOG_LEVEL_INFO	SYS_LOG_LEVEL_INFO
+#define BOOT_LOG_LEVEL_DEBUG	SYS_LOG_LEVEL_DEBUG
+
+/* Treat BOOT_LOG_LEVEL equivalently to SYS_LOG_LEVEL. */
+#ifndef BOOT_LOG_LEVEL
+#define BOOT_LOG_LEVEL CONFIG_SYS_LOG_DEFAULT_LEVEL
+#elif (BOOT_LOG_LEVEL < CONFIG_SYS_LOG_OVERRIDE_LEVEL)
+#undef BOOT_LOG_LEVEL
+#define BOOT_LOG_LEVEL CONFIG_SYS_LOG_OVERRIDE_LEVEL
+#endif
+
+#define SYS_LOG_LEVEL BOOT_LOG_LEVEL
+
+#undef SYS_LOG_DOMAIN
+#define SYS_LOG_DOMAIN "MCUBOOT"
+
+#define BOOT_LOG_ERR(...) SYS_LOG_ERR(__VA_ARGS__)
+#define BOOT_LOG_WRN(...) SYS_LOG_WRN(__VA_ARGS__)
+#define BOOT_LOG_INF(...) SYS_LOG_INF(__VA_ARGS__)
+#define BOOT_LOG_DBG(...) SYS_LOG_DBG(__VA_ARGS__)
+
+#include <logging/sys_log.h>
+
+/*
+ * In other environments, logging calls are no-ops.
+ */
+#else  /* !defined(__ZEPHYR__) */
+
+#define BOOT_LOG_LEVEL_OFF	0
+#define BOOT_LOG_LEVEL_ERROR	1
+#define BOOT_LOG_LEVEL_WARNING	2
+#define BOOT_LOG_LEVEL_INFO	3
+#define BOOT_LOG_LEVEL_DEBUG	4
+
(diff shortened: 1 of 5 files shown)


boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -0,0 +1,86 @@
...
+#elif (BOOT_LOG_LEVEL < CONFIG_SYS_LOG_OVERRIDE_LEVEL)
+#undef BOOT_LOG_LEVEL
+#define BOOT_LOG_LEVEL CONFIG_SYS_LOG_OVERRIDE_LEVEL
+#endif
+
+#define SYS_LOG_LEVEL BOOT_LOG_LEVEL
+
+#ifdef SYS_LOG_DOMAIN
+#undef SYS_LOG_DOMAIN
+#endif
```
ccollins476ad: Just a trivial nitpick: there is no need for the #ifdef here.  The #undef directive is silently ignored if the specified macro is undefined.

boot/bootutil/src/loader.c
```diff
@@ -185,11 +201,17 @@ boot_status_source(void)
              table->bst_magic_scratch   == state_scratch.magic) &&
             (table->bst_copy_done_slot0 == 0    ||
              table->bst_copy_done_slot0 == state_slot0.copy_done)) {
-
-            return table->bst_status_source;
+            uint8_t source = table->bst_status_source;
```
ccollins476ad: This prompts an interesting dilemma - I guess we need some sort of coding style for mcuboot.  The Mynewt coding style (https://github.com/apache/incubator-mynewt-core/blob/master/CODING_STANDARDS.md) differs from Zephyr's.  Notably, in Mynewt:
* Indentation is done with four spaces (no tabs)
* No mixing of variable declarations and statements within a block.

I don't personally mind which style we settle on, but it would be good to be consistent.  Since most of the mcuboot code is already written in the Mynewt style, it might be easiest to stick with that one (but like I said, I don't have a strong opinion).

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -0,0 +1,86 @@
...
+#define BOOT_LOG_LEVEL_OFF	0
+#define BOOT_LOG_LEVEL_ERROR	1
+#define BOOT_LOG_LEVEL_WARNING	2
+#define BOOT_LOG_LEVEL_INFO	3
+#define BOOT_LOG_LEVEL_DEBUG	4
+
+#define BOOT_LOG_ERR(...) ((void)0)
+#define BOOT_LOG_WRN(...) ((void)0)
+#define BOOT_LOG_INF(...) ((void)0)
+#define BOOT_LOG_DBG(...) ((void)0)
```
ccollins476ad: We may run into some "defined but not used" gcc warnings when the log macros expand to (void)0.  Mynewt gets around this problem with some ugly macros: https://github.com/apache/incubator-mynewt-core/blob/master/sys/log/include/log/ignore.h

mbolivar: Sure, I can copy that file into mcuboot as well and use it here, if that's what you mean.

boot/bootutil/src/loader.c
mbolivar: Ah, lovely, a coding standard doc! I didn't find one in tree and wasn't quite sure. I'll fix this up to match the mynewt style and make the other changes.

On a related note, how would you feel about adding a hidden file to the top level that directs emacs to format things according to mynewt style? It'd be called .dir-locals.el.

If that's undesirable, how about a patch adding the same to .gitignore, so I can have one in my clone?

boot/bootutil/include/bootutil/bootutil_log.h
ccollins476ad: Just to clarify, here's an example of the kind of issue I'm thinking of:

```
{
    rc = foo();
    BOOT_LOG_DBG("foo called; status=%d", rc);
}
```

We can't put the call to `foo()` in the macro invocation, because we need the function to get called even if logging is disabled.  We avoid this by assigning to `rc`, but that elicits an "unused variable" warning from gcc if rc is disabled (given the appropriate command line options).

I guess I'm not sure how common this issue will be.  We added the "ignore" macros to Mynewt because it became an issue.  It seems like a bit of a drag to have to add the ignore.h file to this repo.  I don't mind waiting and seeing if it ever becomes an issue.

boot/bootutil/src/loader.c
ccollins476ad: An emacs format file sounds great.

boot/bootutil/include/bootutil/bootutil_log.h
d3zd3z: I'm a little curious what warnings you were getting about `((void)0)`. Assuming you want these in a statement context, how about `do{}while(0)` which is the standard way of doing nothing, but it being a single statement (so doesn't break `if (...) BOOT_LOG_DBG() else ...`

David

boot/bootutil/src/loader.c
sterlinghughes: emacs sounds great, let's use Mynewt style, as the majority of the code is written in that format already.  Let's bring the coding style document into the mcuboot project.  Unless there are any major objections?  My only sticking point is the avoidance of tabs.

d3zd3z: I'm fine with keeping the Mynewt indentation rules. I just have to pay more attention when working on the code that vim gets set correctly.

I wouldn't find emacs rules particularly useful. I'm not sure how polite it is to have a `.dir-locals.el` in a tree. Sometimes people are a little leary of running arbitrary code.

mbolivar: How about putting .dir-locals.el in .gitignore, but including a copy of what to put there in the tree and mentioning it from the README?

I've made the changes to this pull request to keep the bootutil code in line with mynewt style. Can we submit patches bringing boot/zephyr into compliance later on, as a separate PR, since they already use tabs?

sterlinghughes:
For sure.  I’m OK if the OS specific ports also follow their style (up 
to you guys).  I think it’s just about main source code tree.

ccollins476ad:
Just an update - I think this looks good, but I still noticed a few tabs in loader.c.

boot/bootutil/include/bootutil/bootutil_log.h
ccollins476ad: The example above would elicit an unused variable warning.  Here is a complete example:

```
[ccollins@ccollins-mac:~/scratch/c]$ gcc-5 -c unused.c -Wall
unused.c: In function 'main':
unused.c:8:9: warning: variable 'rc' set but not used [-Wunused-but-set-variable]
     int rc;
         ^
[ccollins@ccollins-mac:~/scratch/c]$ cat unused.c
#define FOO(fmt, ...) ((void)0)

int do_proc(void);

int
main(void)
{
    int rc;

    rc = do_proc();
    FOO("rc=%d\n", rc);
    return 0;
}
[ccollins@ccollins-mac:~/scratch/c]$ gcc-5 -c unused.c -Wall
unused.c: In function 'main':
unused.c:8:9: warning: variable 'rc' set but not used [-Wunused-but-set-variable]
     int rc;
         ^
```

mbolivar:
I don't see this patch adding any new tabs to loader.c. Did I miss something?

There are definitely tabs in loader.c, but they were added by d930ec6 ("Validate slot zero before booting"). It seems like a separate patch (in a separate pull request) to get rid of those, no?

ccollins476ad:
Oops!  Right you are.  Sorry about that.  I'll merge it now.

mbolivar:
No worries and thanks for the merge!

PR #19, opened 2017-02-24
Fix slot0 validation

d3zd3z:
commit d930ec69c835b83ba933677d29760ba7605a1158
    Author: David Brown <david.brown@linaro.org>
    Date:   Wed Dec 14 07:59:48 2016 -0700

        Validate slot zero before booting

adds a feature to validate slot 0 before booting it.  However, there
is an error in the logic, and if the magic number is written to an
invalid value, but not all 0xFF, it will consider any image to be
valid.

Fix this logic so that slot zero is always validated.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -580,22 +580,20 @@ boot_validate_slot(int slot)
         return -1;
     }
 
-    /* Image in slot 1 is invalid.  Erase the image and continue booting
-     * from slot 0.
-     */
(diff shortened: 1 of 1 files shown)


ccollins476ad:
Hi David,

It looks good to me.  I just have a few nitpicks:

1. The "Image in slot 1 is invalid..." comment is duplicated in that function.  It seems like the bottom one is more appropriate, but should probably go inside the _if_ block.

2. The second line of the _if_ condition uses tabs for indentation rather than spaces.

Thanks,
Chris

PR #22, opened 2017-02-28
zephyr: disable clock before reset

rsalveti:
Avoid clock related interrupts before jumping into the first image. Safe
to disable as the first application will enable the clock/timer again
once booted.

Signed-off-by: Ricardo Salveti <ricardo.salveti@linaro.org>


boot/zephyr/main.c
```diff
@@ -69,7 +70,8 @@ void main(void)
 	irq_lock();
 	_MspSet(vt->msp);
```
d3zd3z: We probably should have these log and clock calls before the _MspSet.

Did you find specific problems with the clock enabled, since the interrupts should be masked?

d3zd3z:
Change included in #27

PR #24, opened 2017-03-15
Inital x86 support

andrewboie:
I've successfully built MCUBoot as a 3rd stage loader and have been able to use it to validate and boot a Zephyr OS image.


d3zd3z:
https://github.com/apache/incubator-mynewt-newt/blob/master/newt/image/image.go#L472 for the other tool's location for the header skipping.

PR #26, opened 2017-03-16
Makefile: update location of target-specific .conf

andrewboie:
arduino_101.conf was moved up a level, but the toplevel Makefile
was not updated so that it can be located.

Disable some more stuff in the arduino_101 configuration to save ROM space.


Makefile
```diff
@@ -30,7 +30,7 @@ CFLAGS += -DBOOTUTIL_VALIDATE_SLOT0
 BOARD ?= qemu_x86
 
 # Additional board-specific Zephyr configuration
-CONF_FILE += $(wildcard boot/zephyr/targets/$(BOARD).conf)
+CONF_FILE += $(wildcard boot/zephyr/$(BOARD).conf)
```
mbolivar: Whoops, sorry about that!

PR #27, opened 2017-03-16
Nrf52840

mbolivar:
This patch adds support for the Zephyr board named nrf52840_pca10056, also known as the NRF52840-PDK.

In general, we need https://github.com/runtimeco/mcuboot/pull/22 in place for NRF5x to work reliably after chain-loading. I've taken the liberty of rebasing that here, and addressing the review comments (ensuring logging and clock disabling happen before changing to the new main stack pointer). The symptom this fixes is hanging during sleep in the chain-loaded Zephyr image.


boot/zephyr/targets/nrf52840_pca10056.h
```diff
@@ -0,0 +1,18 @@
...
+#define H_TARGETS_NRF52840_PCA10056_
+
+#define FLASH_DRIVER_NAME		CONFIG_SOC_FLASH_NRF5_DEV_NAME
+#define FLASH_ALIGN			4
+#define FLASH_AREA_IMAGE_0_OFFSET	0x08000
+#define FLASH_AREA_IMAGE_0_SIZE		0x7a000
+#define FLASH_AREA_IMAGE_1_OFFSET	0x82000
+#define FLASH_AREA_IMAGE_1_SIZE		0x7a000
+#define FLASH_AREA_IMAGE_SCRATCH_OFFSET	0xfc000
+#define FLASH_AREA_IMAGE_SCRATCH_SIZE	0x01000
```
mbolivar: Given how large the image areas are, @mike-scott brought up that we could make a better tradeoff by increasing the scratch size and shrinking the images. @d3zd3z, any suggestions on what values would tune this well?

d3zd3z: Performance isn't affected that much by the scratch size, however, you will be doing a lot of erases on the scratch area, which reduces the lifetime of the device.  With this configuration, each update will do 122 erases of the scratch sector (for one erase of each data sector). Given that some of the devices are only good for thousands of erases, this could result in a dead part fairly shortly. It might be worth trying to get out of Nordic what the expected lifetime should be.

mbolivar: Hmm, by that argument we already have a problem on Nitrogen, which we need to fix as well.

This is the current situation on Nitrogen:

|                 | Start   | End     | Size (KiB) | User           |
|-----------------|---------|---------|------------|----------------|
| Sectors 0-7     | 0x00000 | 0x07FFF | 32         | mcuboot        |
| Sectors 8-65    | 0x08000 | 0x41FFF | 232        | Image 0        |
| Sectors 66-123  | 0x42000 | 0x7BFFF | 232        | Image 1        |
| Sector 124      | 0x7C000 | 0x7CFFF | 4          | Scratch        |
| Sectors 125-127 | 0x7D000 | 0x7FFFF | 12         | FOTA app state |

At 10,000 erases (which is the figure for the NRF52832 on that board), we get 172 updates before we expect wear on the part. I expect we are close to seeing this on some of our boards, e.g. the ones @embeddedandroid uses for CI.

Subtracting mcuboot and the FOTA app state, we have 117 4-KiB sectors to divide between images and scratch on Nitrogen, and 245 sectors on NRF52840.

For Nitrogen, splitting up 117 sectors into images and a scratch partition sized 1/4 that of the images would be 52 KiB (13 sectors) scratch, 208 KiB (52 sectors) images.

For NRF52840, we'd get 116 KiB (29 sectors) 432 KiB (108 sectors) images.

Seem reasonable to everyone? I'll make the changes if so.

mike-scott: Originally on Zephyr 1.5 the Nitrogen image was much larger than it is now (let's say 190KB)..  but checking it now: 155692 Mar 17 10:09 zephyr.bin  So I think your logic is sound.  Leaves us ~50KB of expansion for apps.

mbolivar:
OK, I've updated the patch series to fix up Nitrogen and adjust the NRF52840-PDK as described here.

Along the way, I ran into https://runtimeco.atlassian.net/browse/MCUB-39 and had to add da29913 to fix flash_area_sectors(), which I think deserves a bit of extra attention.

*Note that this will break the LTD FOTA application unless https://github.com/Linaro/zephyr-fota-hawkbit/pull/69 is merged at the same time*

boot/zephyr/Makefile
```diff
@@ -1,4 +1,4 @@
-subdir-ccflags-y += -DMCUBOOT_TARGET_CONFIG='"$(BOARD).h"'
+subdir-ccflags-y += -DMCUBOOT_TARGET_CONFIG='"$(BOARD).inc"'
```
d3zd3z: Changing the extension of these files will break numerous tools/editors that use the extension to determine the file type.

I would suggest something along the lines of:
* Move target.h into boot/zephyr/include
* Remove boot/zephyr/targets from the include path entirely
* Perhaps move it to boot/zephyr/include/targets
* Have the target files include target/name.h instead of just name.h

If you want to really protect against direct inclusion, they could #error if there is a define not defined, and then define that in target.h before doing the include, but this might be overkill.

mbolivar:
OK, I've moved target.h into boot/zephyr/include and moved all the .inc files so they end in .h.

I left boot/zephyr/targets in the include path; that makes it easier to pull in the SOC family header for NRF5 without extra work in the Makefile. I'd rather not put it under include/, to make it clear it's not meant to be touched.

mkiiskila:
Hi,

This setting really depends on product the consumer is building.

IMHO there’s not going to be that many updates with the devices on the field
(maybe once a 1/2 - 1 year). Maybe every 3 months to begin with.

My recommendation would be to add a comment in the code mentioning
that they should pay attention to it. Along with a number saying how many
upgrades are possible given the current setting.

I imagine many will value available text space over number of updates.

d3zd3z:
Probably what is wanted is to use a larger scratch during development to avoid wearing out the parts, and then use a smaller one for production.

This is a good reason to work on making the partitions easy to change. Right now, on Zephyr, we have to keep numbers in several places to all match. Our goal is to be able to define the partitions in one place and have the data used by the other tools.

PR #28, opened 2017-03-29
Sim updates

d3zd3z:
Some updates to the simulator that make it more useful in a CI environment. Note that there is a failure in some of the test cases (that hits an assert), which will need to be fixed before this can be run through CI successfully.


utzig:
LGTM. Go ahead and merge it!

PR #29, opened 2017-04-04
Fix 8-byte aligned sector failure

utzig:
This fixes issue MCUB-36.

While copying data from slot to slot, if using 8-byte aligned
sectors, not all boot status data falls in the same copy buffer
which might result in inconsistency of state if a reset occurs
before all boot status sectors were copied.


utzig:
DO NOT MERGE!

Still testing in HW.

d3zd3z:
I'm not quite seeing how this helps, though. No matter the size of the copy buffer, the writes happen 8 bytes at a time (on frdm k64f at least).

PR #30, opened 2017-04-10
Add test of multiple random resets while upgrading

utzig:
This adds simulation of the effects of doing multiple resets while upgrade is not completed, which is a failure condition I can easily reproduce while testing on real hardware.

Since I'm not seeding the random generator this will fail intermittently, which for a CI is probably not that great.

### DIFF
--- a/sim/src/main.rs
+++ b/sim/src/main.rs
@@ -9,7 +9,8 @@ extern crate rustc_serialize;
 extern crate error_chain;
 
 use docopt::Docopt;
-use rand::{Rng, SeedableRng, XorShiftRng};
+use rand::{Rng, SeedableRng, XorShiftRng, thread_rng};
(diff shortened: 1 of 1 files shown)


sim/src/main.rs
```diff
@@ -233,8 +234,9 @@ impl RunStatus {
         }
 
         let mut bad = 0;
-        // Let's try an image halfway through.
-        for i in 1 .. total_count {
+        let mut stops: Vec<i32> = (1..total_count).collect();
```
d3zd3z: Actually, thinking about this, this change doesn't really have any affect. The 'flash' is not passed mutably, so is restored to the initial state after each call to try_upgrade. It doesn't really matter what order the individual tests are run in. Did you find problems this way?

PR #34, opened 2017-04-17
Move trailer magic to end

utzig:
DO NOT MERGE.

This is an experiment on putting the trailer magic at the end. Solves issues with copy but doesn't solve random resets. WIP.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -33,12 +33,14 @@
 
 int boot_current_slot;
 
-const uint32_t boot_img_magic[4] = {
+const uint32_t boot_img_magic[] = {
     0xf395c277,
     0x7fefd260,
     0x0f505235,
     0x8079b62c,
 };
+
+const uint32_t BOOT_MAGIC_SZ = sizeof boot_img_magic;
 
 struct boot_swap_table {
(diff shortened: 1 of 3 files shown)


boot/bootutil/src/bootutil_misc.c
```diff
@@ -144,11 +146,11 @@ boot_magic_code(const uint32_t *magic)
...
     int i;
 
-    if (memcmp(magic, boot_img_magic, sizeof boot_img_magic) == 0) {
+    if (memcmp(magic, boot_img_magic, BOOT_MAGIC_SZ) == 0) {
         return BOOT_MAGIC_GOOD;
     }
 
-    for (i = 0; i < 4; i++) {
+    for (i = 0; i < BOOT_MAGIC_SZ; i++) {
         if (magic[i] == 0xffffffff) {
```
d3zd3z: It is interesting that we only compare against the good magic or 0xffffffff. I wonder, in general, about the handling of partial writes to flash. Generally, when a write is interrupted by power loss, the resulting data will contain a combination of:
- The old data, in this case 1 bits.
- The new data, zero bits.
and the combination that is read can be different, even on subsequent reads, but even worse, can change as temperature and voltage change.

utzig:
Closing this one. Relevant portions sent in separate PRs.

PR #41, opened 2017-04-26
Update to status management

utzig:
PLEASE DO NOT MERGE.

Trailer management, most importantly status, is now handled separately
from image sectors handling. This should fix issues with resets that
happen while doing a swap.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -286,12 +286,8 @@ boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state)
     }
 
     rc = boot_read_swap_state(fap, state);
-    if (rc) {
-        flash_area_close(fap);
-        return rc;
-    }
-
-    return 0;
+    flash_area_close(fap);
+    return rc;
 }
 
 int
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -376,7 +376,7 @@ boot_swap_type(void)
              table->bsw_image_ok_slot0  == state_slot0.image_ok)        &&
             (table->bsw_image_ok_slot1  == 0    ||
              table->bsw_image_ok_slot1  == state_slot1.image_ok)) {
-            BOOT_LOG_INF("Swap type: %s\n",
+            BOOT_LOG_INF("Swap type: %s",
                    table->bsw_swap_type == BOOT_SWAP_TYPE_NONE   ? "none"   :
                    table->bsw_swap_type == BOOT_SWAP_TYPE_TEST   ? "test"   :
                    table->bsw_swap_type == BOOT_SWAP_TYPE_PERM   ? "perm"   :
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -166,16 +166,18 @@ uint32_t
 boot_slots_trailer_sz(uint8_t min_write_sz)
 {
     return BOOT_MAGIC_SZ                                                    +
+           /* state for all sectors */
            BOOT_STATUS_MAX_ENTRIES * BOOT_STATUS_STATE_COUNT * min_write_sz +
+           /* copy_done + image_ok */
            min_write_sz * 2;
 }
 
 static uint32_t
 boot_scratch_trailer_sz(uint8_t min_write_sz)
 {
-    return BOOT_MAGIC_SZ                           +
-           BOOT_STATUS_STATE_COUNT * min_write_sz  +
-           min_write_sz * 2;
+    return BOOT_MAGIC_SZ                          +  /* magic */
+           BOOT_STATUS_STATE_COUNT * min_write_sz +  /* state for one sector */
+           min_write_sz;                             /* image_ok */
 }
 
 static uint32_t
@@ -205,6 +207,7 @@ boot_status_off(const struct flash_area *fap)
 static uint32_t
 boot_copy_done_off(const struct flash_area *fap)
 {
+    assert(fap->fa_id != FLASH_AREA_IMAGE_SCRATCH);
     return fap->fa_size - flash_area_align(fap) * 2;
 }
 
@@ -229,10 +232,12 @@ boot_read_swap_state(const struct flash_area *fap,
     }
     state->magic = boot_magic_code(magic);
 
-    off = boot_copy_done_off(fap);
-    rc = flash_area_read(fap, off, &state->copy_done, 1);
-    if (rc != 0) {
-        return BOOT_EFLASH;
+    if (fap->fa_id != FLASH_AREA_IMAGE_SCRATCH) {
+        off = boot_copy_done_off(fap);
+        rc = flash_area_read(fap, off, &state->copy_done, 1);
+        if (rc != 0) {
+            return BOOT_EFLASH;
+        }
     }
 
     off = boot_image_ok_off(fap);
@@ -248,56 +253,31 @@ boot_read_swap_state(const struct flash_area *fap,
  * Reads the image trailer from the scratch area.
  */
 int
-boot_read_swap_state_scratch(struct boot_swap_state *state)
+boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state)
 {
     const struct flash_area *fap;
     int rc;
 
-    rc = flash_area_open(FLASH_AREA_IMAGE_SCRATCH, &fap);
+    switch (flash_area_id) {
+    case FLASH_AREA_IMAGE_SCRATCH:
+    case FLASH_AREA_IMAGE_0:
+    case FLASH_AREA_IMAGE_1:
+        rc = flash_area_open(flash_area_id, &fap);
+        if (rc != 0) {
+            return BOOT_EFLASH;
+        }
+        break;
+    default:
+        return BOOT_EFLASH;
+    }
+
+    rc = boot_read_swap_state(fap, state);
     if (rc) {
-        rc = BOOT_EFLASH;
-        goto done;
+        flash_area_close(fap);
+        return rc;
     }
 
-    rc = boot_read_swap_state(fap, state);
-    if (rc != 0) {
-        goto done;
-    }
-
-    rc = 0;
-
-done:
-    flash_area_close(fap);
-    return rc;
-}
-
-/**
- * Reads the image trailer from a given image slot.
- */
-int
-boot_read_swap_state_img(int slot, struct boot_swap_state *state)
-{
-    const struct flash_area *fap;
-    int area_id;
-    int rc;
-
-    area_id = flash_area_id_from_image_slot(slot);
-    rc = flash_area_open(area_id, &fap);
-    if (rc != 0) {
-        rc = BOOT_EFLASH;
-        goto done;
-    }
-
-    rc = boot_read_swap_state(fap, state);
-    if (rc != 0) {
-        goto done;
-    }
-
-    rc = 0;
-
-done:
-    flash_area_close(fap);
-    return rc;
+    return 0;
 }
 
 int
@@ -316,15 +296,24 @@ boot_write_magic(const struct flash_area *fap)
     return 0;
 }
 
-int
-boot_write_copy_done(const struct flash_area *fap)
+static int
+boot_write_flag(int flag, const struct flash_area *fap)
 {
     uint32_t off;
     int rc;
     uint8_t buf[BOOT_MAX_ALIGN];
     uint8_t align;
 
-    off = boot_copy_done_off(fap);
+    switch (flag) {
+    case BOOT_FLAG_COPY_DONE:
+        off = boot_copy_done_off(fap);
+        break;
+    case BOOT_FLAG_IMAGE_OK:
+        off = boot_image_ok_off(fap);
+        break;
+    default:
+        return BOOT_EFLASH;
+    }
 
     align = hal_flash_align(fap->fa_device_id);
     assert(align <= BOOT_MAX_ALIGN);
@@ -340,26 +329,15 @@ boot_write_copy_done(const struct flash_area *fap)
 }
 
 int
+boot_write_copy_done(const struct flash_area *fap)
+{
+    return boot_write_flag(BOOT_FLAG_COPY_DONE, fap);
+}
+
+int
 boot_write_image_ok(const struct flash_area *fap)
 {
-    uint32_t off;
-    int rc;
-    uint8_t buf[BOOT_MAX_ALIGN];
-    uint8_t align;
-
-    off = boot_image_ok_off(fap);
-
-    align = hal_flash_align(fap->fa_device_id);
-    assert(align <= BOOT_MAX_ALIGN);
-    memset(buf, 0xFF, BOOT_MAX_ALIGN);
-    buf[0] = 1;
-
-    rc = flash_area_write(fap, off, buf, align);
-    if (rc != 0) {
-        return BOOT_EFLASH;
-    }
-
-    return 0;
+    return boot_write_flag(BOOT_FLAG_IMAGE_OK, fap);
 }
 
 int
@@ -371,10 +349,10 @@ boot_swap_type(void)
     int rc;
     int i;
 
-    rc = boot_read_swap_state_img(0, &state_slot0);
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_0, &state_slot0);
     assert(rc == 0);
 
-    rc = boot_read_swap_state_img(1, &state_slot1);
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_1, &state_slot1);
     assert(rc == 0);
 
     for (i = 0; i < BOOT_SWAP_TABLES_COUNT; i++) {
@@ -419,10 +397,9 @@ boot_set_pending(int permanent)
 {
     const struct flash_area *fap;
     struct boot_swap_state state_slot1;
-    int area_id;
     int rc;
 
-    rc = boot_read_swap_state_img(1, &state_slot1);
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_1, &state_slot1);
     if (rc != 0) {
         return rc;
     }
@@ -433,8 +410,7 @@ boot_set_pending(int permanent)
         return 0;
 
     case BOOT_MAGIC_UNSET:
-        area_id = flash_area_id_from_image_slot(1);
-        rc = flash_area_open(area_id, &fap);
+        rc = flash_area_open(FLASH_AREA_IMAGE_1, &fap);
         if (rc != 0) {
             rc = BOOT_EFLASH;
         } else {
@@ -467,7 +443,7 @@ boot_set_confirmed(void)
     struct boot_swap_state state_slot0;
     int rc;
 
-    rc = boot_read_swap_state_img(0, &state_slot0);
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_0, &state_slot0);
     if (rc != 0) {
         return rc;
     }
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -31,6 +31,9 @@
 #include "bootutil/bootutil.h"
 #include "bootutil_priv.h"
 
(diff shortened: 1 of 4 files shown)


utzig:
@d3zd3z Please merge #39 and run CI again for this PR!

d3zd3z:
I've merged the other changes, but I think you'll have to rebase and repush this to get it to rerun (CI doesn't merge with tip, it just runs your tree as-id, this should probably be fixed).

utzig:
@d3zd3z ok, but `PLEASE DO NOT MERGE` is still valid. I'll let you know when it's done.

utzig:
@d3zd3z I think it's OK to merge this. There are some open issues that I can fix later (FIXMEs, etc). This should also enable other PRs to pass.

boot/bootutil/src/loader.c
```diff
@@ -416,8 +417,14 @@ boot_read_status_bytes(const struct flash_area *fap, struct boot_status *bs)
 
     if (found) {
         i--;
-        bs->idx = i / BOOT_STATUS_STATE_COUNT;
-        bs->state = i % BOOT_STATUS_STATE_COUNT;
+        /* FIXME: is this test required? */
```
mbolivar: What's the concern here?

boot/bootutil/src/loader.c
```diff
@@ -488,7 +490,12 @@ boot_write_status(struct boot_status *bs)
     uint8_t buf[8];
     uint8_t align;
 
-    if (bs->idx == 0) {
+    /* NOTE: The first sector copied (that is the last sector on slot) contains
```
mbolivar: Thanks for adding this comment; this had been really confusing me this week.

boot/bootutil/src/loader.c
```diff
@@ -778,6 +785,61 @@ boot_copy_sector(int flash_area_id_src, int flash_area_id_dst,
...
+    struct boot_swap_state swap_state;
+    int rc;
+
+    rc = flash_area_open(flash_area_id, &fap);
+    assert(rc == 0);
+
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_1, &swap_state);
+    assert(rc == 0);
+
+    if (swap_state.image_ok == 0x01) {
```
mbolivar: Could we have macros instead of magic numbers?

boot/bootutil/src/loader.c
utzig: That the resulting values of the division/modulus give the same values that are hardcoded in the `else` so it ends up not being required. It was just a reminder to check later and clean up if that's actually possible.

utzig:
@mbolivar Thanks for the review, I'll fix the issues.

boot/bootutil/src/loader.c
mbolivar: Ah, I see what you mean. Yes, I agree, it should give the same values.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -379,7 +366,13 @@ boot_swap_type(void)
              table->bsw_image_ok_slot0  == state_slot0.image_ok)        &&
             (table->bsw_image_ok_slot1  == 0    ||
              table->bsw_image_ok_slot1  == state_slot1.image_ok)) {
-
+            BOOT_LOG_INF("Swap type: %s\n",
```
mbolivar: The newline is not necessary, btw -- the logging subsystem adds newlines

boot/bootutil/src/loader.c
```diff
@@ -793,58 +855,122 @@ boot_copy_sector(int flash_area_id_src, int flash_area_id_dst,
...
     if (bs->state == 0) {
         rc = boot_erase_sector(FLASH_AREA_IMAGE_SCRATCH, 0, sz);
         assert(rc == 0);
 
         rc = boot_copy_sector(FLASH_AREA_IMAGE_1, FLASH_AREA_IMAGE_SCRATCH,
-                              img_off, 0, sz);
+                              img_off, 0, copy_sz);
         assert(rc == 0);
 
+        if (bs->idx == 0) {
```
mbolivar: I believe this has problems, but perhaps I'm missing something.

(By the way, the commit logs in this series are a little terse -- do you have any pointers to more detailed discussion explaining these changes? It would help me understand what's going on a bit better.)

Anyway, this if statement is going to at least break some targets, e.g. the STM32F401RE-based 96Boards Carbon, which is a Zephyr target. I'm not sure how it's meant to work in general, but it looks to me that there are other issues too.

On STM32F401RE, the sectors have variable sizes. Here is the layout:

| Sector | Start address | End address | Size (KiB) | Size (B) | User           |
|--------|---------------|-------------|------------|----------|----------------|
| 0      | 0x08000000    | 0x08003FFF  | 16         | 0x4000   | mcuboot        |
| 1      | 0x08004000    | 0x08007FFF  | 16         | 0x4000   | mcuboot        |
| 2      | 0x08008000    | 0x0800BFFF  | 16         | 0x4000   | None           |
| 3      | 0x0800C000    | 0x0800FFFF  | 16         | 0x4000   | FOTA app state |
| 4      | 0x08010000    | 0x0801FFFF  | 64         | 0x10000  | None           |
| 5      | 0x08020000    | 0x0803FFFF  | 128        | 0x20000  | Image 1        |
| 6      | 0x08040000    | 0x0805FFFF  | 128        | 0x20000  | Image 2        |
| 7      | 0x08060000    | 0x0807FFFF  | 128        | 0x20000  | Scratch        |

As you can see, the images and scratch each take up exactly one of the three 128KB size sectors at the end of the device flash.

Here's what will happen during a copy:

1. line 881, boot_erase_sector() erases scratch (nothing is lost)
2. line 884, boot_copy_sector() copies image 1 into scratch (nothing is lost)
3. line 896, boot_erase_last_sector_by_id() erases the entire image 0 sector (**data loss; mcuboot cannot revert if image 1 is bad**)

I'm not sure what is supposed to be happening in the call to boot_status_init_by_id() on line 899. It looks like what will happen is:

- Read image 1 trailer's swap state (magic, copy_done, image_ok)
- If image 1's image_ok is set, write image_ok to image 0 trailer
- Write image 0 trailer magic

It seems like the idea is to set up the image 0 trailer to be at least a test swap from image 1, and maybe a permanent swap if image 1 has the image_ok flag set.

But that doesn't make sense to me.

- If the trailer is smaller than the last sector in image 0, **some image 0 data can get permanently erased on line 896**. This is true even on targets where the sector size is smaller than the image area, **not just on 96Boards Carbon**.
- If the trailer is larger than the last sector in image 0, then not enough of flash has been erased in line 896, and **mcuboot might later try to write data in image 0 it has not erased yet**, e.g. writing the magic bytes during the call to boot_status_init_by_id(), depending on the sector size.
- If the trailer is exactly the same size as the last sector of image 0, then things seem like they may be OK, but that's **not a safe assumption** (e.g. it's not the case on any Zephyr target).

What's going on here? Am I missing something?

boot/bootutil/src/loader.c
```diff
@@ -793,58 +855,122 @@ boot_copy_sector(int flash_area_id_src, int flash_area_id_dst,
...
     int rc;
 
     /* Calculate offset from start of image area. */
     img_off = boot_data.imgs[0].sectors[idx].fa_off -
               boot_data.imgs[0].sectors[0].fa_off;
 
+    copy_sz = sz;
+
+    trailer_sz = boot_slots_trailer_sz(boot_data.write_sz);
+    if (boot_data.imgs[0].sectors[idx].fa_off + sz >
```
utzig: `sz` in this function is always a size that is corresponds to the full number of bytes of the amount of sectors to copy (never less than a sector size). You're right that this checks against the first address of the last sector and it was done on purpose. What I wanted to know is if we're copying the last sector. If we are, than I discount the trailer, but of course, this is based on the assumption that `sz` will still be the whole last sector. It also breaks if the fw would occupy more than what trailer size leaves for it but that needs to be handled elsewhere.

boot/bootutil/src/loader.c
utzig: Your analysis looks good but seems to be missing the `if (bs->use_scratch)` tests. `boot_status_init_by_id` initializes a new trailer from an existing one, exactly as you said. But it chooses which one to initialize based on the fact that we are copying the last sector. If we are (as is always the case for stm32f401) it should not erase SLOT0 and instead use scratch for writing a temporary trailer that will only end up in SLOT0 when state is 2.

Btw, I have not really tested on the stm32f401 (only stm32f407) but would be great to test it to be sure if this really works. Unfortunately I don't have the board here with me but I'll be at Runtime's HQ by Thursday and can give it a try.

utzig: > In general, I was surprised to learn this week that mcuboot copies around the image trailers as it swaps.

This was the case before but now there's no more copying of trailer. Trailer is always created (possibly using parts of an existing trailer) by `boot_status_init_by_id`.

> Has any thought been given to having a bootloader state partition that is not copied around during image swap? It seems like it would help avoid some of the issues.

I have not given it much thought because I started on the project just one month ago and was basically trying to fix these issues that were found by doing as few changes as possible. The code ended up being quite complicated. I agree that using a dedicated sector for at least the "status bytes" trailer stuff would be a really good idea. As long as we have a spare sector which hopefully will be the case.

> (By the way, the commit logs in this series are a little terse -- do you have any pointers to more detailed discussion explaining these changes? It would help me understand what's going on a bit better.)

Apart from reading the `design.txt` document, I have not. I'll try to get more background on the idea design when I'm at Runtime later this week.

boot/bootutil/src/loader.c
mbolivar: Ah, that helps a lot, thanks for the clear explanation. What about adding a comment to the code explaining that, since it's a bit "tricky"?

d3zd3z:
I do agree with Marti that it would be helpful to have better commit text for most of these changes.  Normally, I would expect to see in the commit text a justification for applying the path. If we thought of patches having to go through a mailing list, the first line would be the subject of the email, a blank line, and then the rest of the commit text as the body. And both are justifying/convincing the maintainer why they should pull these changes in.

We're fixing a fairly significant issue here, and someone looking back on the history is going to want to know more about it.

mbolivar:
> Your analysis looks good but seems to be missing the if (bs->use_scratch) tests. boot_status_init_by_id initializes a new trailer from an existing one, exactly as you said. But it chooses which one to initialize based on the fact that we are copying the last sector. If we are (as is always the case for stm32f401) it should not erase SLOT0 and instead use scratch for writing a temporary trailer that will only end up in SLOT0 when state is 2.

Ah, you're right, I see that better now, and had missed that.

> 
> Btw, I have not really tested on the stm32f401 (only stm32f407) but would be great to test it to be sure if this really works. Unfortunately I don't have the board here with me but I'll be at Runtime's HQ by Thursday and can give it a try.

I rebased this series onto master and added a commit which does a bunch of logging, then tested it on STM32F401. Here's the code (I'll probably post something like this for master):

https://github.com/mbolivar/mcuboot/tree/update-status-mgmt

Here is the resulting log for a test swap (I have one more patch I'm not showing that's just related to the Zephyr MPU implementation). The breakage I was so sure was going to happen is noticeably absent ;).

```
[MCUBOOT] [INF] boot_go: Preparing to boot.
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 1: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 3: .magic=3, image_ok=0xff
[MCUBOOT] [INF] boot_status_source: image 0: magic=good, copy_done=0xff
[MCUBOOT] [INF] boot_status_source: scratch: magic=unset
[MCUBOOT] [INF] boot_status_source: Boot status source: slot 0
[MCUBOOT] [DBG] boot_read_status_bytes: no boot status found; swap was not in progress
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 1: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 2: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [INF] boot_swap_type: Swap type: test

[MCUBOOT] [DBG] boot_image_check: flash area ID 2 image is valid
[MCUBOOT] [DBG] boot_swap_if_needed: swapping images
[MCUBOOT] [DBG] boot_copy_image: initial boot_status: .idx=0, .state=0
[MCUBOOT] [DBG] boot_swap_sectors: sector info: idx=0, size=0x20000
[MCUBOOT] [DBG] boot_swap_sectors: boot_status: .idx=0, .state=0
[MCUBOOT] [DBG] boot_swap_sectors: img_off=0x0, sz=0x20000, copy_sz=0x1fe6e, bs->use_scratch=1
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 3: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=2/0x0, dst ID/off=3/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_status_init_by_id: flash_area_id=3
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 2: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_write_magic: wrote magic for flash area ID 3
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=0, .state=1 to scratch
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 2: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=1/0x0, dst ID/off=2/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=0, .state=2 to scratch
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 1: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=3/0x0, dst ID/off=1/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=3/0x1fffc, dst ID/off=1/0x1fe7e, sz=0x3
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 3: .magic=1, image_ok=0xff
[MCUBOOT] [DBG] boot_write_magic: wrote magic for flash area ID 1
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=1, .state=0 to image 0
[MCUBOOT] [DBG] boot_swap_if_needed: final swap type was: 2
[MCUBOOT] [DBG] boot_finalize_test_swap: finalizing image 0
[MCUBOOT] [DBG] boot_write_flag: wrote copy_done for flash area ID 1
[MCUBOOT] [INF] main: Bootloader chainload address: 0x20000
[MCUBOOT] [INF] main: Jumping to the first image slot
```

mbolivar:
Here's the log after rebasing onto your latest version (thanks for cleaning up the stray newline):

```
[MCUBOOT] [INF] boot_go: Preparing to boot.
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 1: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 3: .magic=3, image_ok=0xff
[MCUBOOT] [INF] boot_status_source: image 0: magic=good, copy_done=0xff
[MCUBOOT] [INF] boot_status_source: scratch: magic=unset
[MCUBOOT] [INF] boot_status_source: Boot status source: slot 0
[MCUBOOT] [DBG] boot_read_status_bytes: no boot status found; swap was not in progress
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 1: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 2: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [INF] boot_swap_type: Swap type: test
[MCUBOOT] [DBG] boot_image_check: flash area ID 2 image is valid
[MCUBOOT] [DBG] boot_swap_if_needed: swapping images
[MCUBOOT] [DBG] boot_copy_image: initial boot_status: .idx=0, .state=0
[MCUBOOT] [DBG] boot_swap_sectors: sector info: idx=0, size=0x20000
[MCUBOOT] [DBG] boot_swap_sectors: boot_status: .idx=0, .state=0
[MCUBOOT] [DBG] boot_swap_sectors: img_off=0x0, sz=0x20000, copy_sz=0x1fe6e, bs->use_scratch=1
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 3: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=2/0x0, dst ID/off=3/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_status_init_by_id: flash_area_id=3
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 2: .magic=1, copy_done=0xff, image_ok=0xff
[MCUBOOT] [DBG] boot_write_magic: wrote magic for flash area ID 3
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=0, .state=1 to scratch
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 2: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=1/0x0, dst ID/off=2/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=0, .state=2 to scratch
[MCUBOOT] [DBG] boot_erase_sector: flash area ID 1: off=0x0, sz=0x20000
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=3/0x0, dst ID/off=1/0x0, sz=0x1fe6e
[MCUBOOT] [DBG] boot_copy_sector: src ID/off=3/0x1fffc, dst ID/off=1/0x1fe7e, sz=0x3
[MCUBOOT] [DBG] boot_read_swap_state: flash area ID 3: .magic=1, image_ok=0xff
[MCUBOOT] [DBG] boot_write_magic: wrote magic for flash area ID 1
[MCUBOOT] [DBG] boot_write_status: persisting boot_status .idx=1, .state=0 to image 0
[MCUBOOT] [DBG] boot_swap_if_needed: final swap type was: 2
[MCUBOOT] [DBG] boot_finalize_test_swap: finalizing image 0
[MCUBOOT] [DBG] boot_write_flag: wrote copy_done for flash area ID 1
[MCUBOOT] [INF] main: Bootloader chainload address: 0x20000
[MCUBOOT] [INF] main: Jumping to the first image slot
```

boot/bootutil/src/loader.c
```diff
@@ -908,9 +1024,55 @@ boot_copy_image(struct boot_status *bs)
...
+    if (hdr->ih_magic == IMAGE_MAGIC) {
+        copy_size = hdr->ih_hdr_size + hdr->ih_img_size + hdr->ih_tlv_size;
+    }
+
+    hdr = &boot_data.imgs[1].hdr;
+    if (hdr->ih_magic == IMAGE_MAGIC) {
+        size = hdr->ih_hdr_size + hdr->ih_img_size + hdr->ih_tlv_size;
+    }
+
+    if (!size || !copy_size || size == copy_size) {
```
mbolivar: What is `|| size == copy_size` for? Is it because we might have the same image's header copied into both slots at the same time?

What if the two images are the same size? Is it guaranteed that the scratch bytes at the ih_magic offset will be something different than IMAGE_MAGIC when this path is reached?

utzig: > What is || size == copy_size for? Is it because we might have the same image's header copied into both slots at the same time? What if the two images are the same size?

Yeah, I think this is tricky. The comparison really tests if it *could be* the same image (same size). If that's the case the scratch is checked and if there is a valid header there, its size if used (assuming both slots' headers where the same). 

> Is it guaranteed that the scratch bytes at the ih_magic offset will be something different than IMAGE_MAGIC when this path is reached?

The only chance of having invalid data or same data in the slot's headers is when doing the swap operation of the sector that has the header. Since scratch is the first thing erased on a swap I assume that if it has the `IMAGE_MAGIC`, the header swap was in progress. If it doesn't have the `IMAGE_MAGIC` it's ignored and everything proceeds as if both images were actually the same size. I have to give more thought on what could go wrong here.

utzig:
@mbolivar Hey, are you still reviewing this?

mbolivar:
> I started on the project just one month ago and was basically trying to fix these issues that were found by doing as few changes as possible.

Makes sense. FWIW, I'm in favor of merging the changes as I understand them better now; they're a significant improvement. Thanks for all the patient explanations.

PR #45, opened 2017-05-02
[RFC, DO NOT MERGE] Flash map fixups

mbolivar:
There's a bug in boot_go() and its users, which assume that the image offset is equal to its address. This will break with 96b_carbon on recent Zephyr builds, which are using an MPU to make sure that flash addresses use the base at 0x08000000 rather than relying on aliasing at 0x0.

This RFC series adds flash_device_base() to flash_map.h to allow fixing that, and makes other improvements to the API which clarify its usage and save memory (this file is annoying to change since there are multiple versions and implementations, but while we're here, we might as well).

It's not ready to merge; while I verified that I could swap an image on 96b_nitrogen (not Carbon as there are other MPU-related issues to debug as well), there are bugs, the simulator isn't updated, we'd need to propagate changes to the mynewt-core version of flash_map.h, etc.

I'm posting it in hopes of getting some feedback about the approach. If it's welcomed, I'll polish it up for a "real" PR.

### DIFF
--- a/boot/zephyr/flash_map.c
+++ b/boot/zephyr/flash_map.c
@@ -32,26 +32,49 @@
 extern struct device *boot_flash_device;
 
 /*
+ * For now, we only support one flash device.
+ *
+ * Pick a random device ID for it that's unlikely to collide with
+ * anything "real".
+ */
+#define FLASH_DEVICE_ID 100
+#define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS
+
+/*
  * The flash area describes essentially the partition table of the
  * flash.  In this case, it starts with FLASH_AREA_IMAGE_0.
  */
 static const struct flash_area part_map[] = {
     {
         .fa_id = FLASH_AREA_IMAGE_0,
+        .fa_device_id = FLASH_DEVICE_ID,
         .fa_off = FLASH_AREA_IMAGE_0_OFFSET,
         .fa_size = FLASH_AREA_IMAGE_0_SIZE,
     },
     {
         .fa_id = FLASH_AREA_IMAGE_1,
+        .fa_device_id = FLASH_DEVICE_ID,
         .fa_off = FLASH_AREA_IMAGE_1_OFFSET,
         .fa_size = FLASH_AREA_IMAGE_1_SIZE,
     },
     {
         .fa_id = FLASH_AREA_IMAGE_SCRATCH,
+        .fa_device_id = FLASH_DEVICE_ID,
         .fa_off = FLASH_AREA_IMAGE_SCRATCH_OFFSET,
         .fa_size = FLASH_AREA_IMAGE_SCRATCH_SIZE,
     },
 };
+
+int flash_device_base(uint8_t fd_id, uint32_t *ret)
+{
+    if (fd_id != FLASH_DEVICE_ID) {
+        BOOT_LOG_ERR("invalid flash ID %d; expected %d",
+                     fd_id, FLASH_DEVICE_ID);
+        return -EINVAL;
+    }
+    *ret = FLASH_DEVICE_BASE;
+    return 0;
+}
 
 /*
  * `open` a flash area.  The `area` in this case is not the individual
--- a/boot/zephyr/flash_map.c
+++ b/boot/zephyr/flash_map.c
@@ -155,19 +155,8 @@ int flash_area_id_from_image_slot(int slot)
 #define FLASH_AREA_IMAGE_SECTOR_SIZE FLASH_AREA_IMAGE_SCRATCH_SIZE
 #endif
 
-/*
- * Lookup the sector map for a given flash area.  This should fill in
- * `ret` with all of the sectors in the area.  `*cnt` will be set to
- * the storage at `ret` and should be set to the final number of
- * sectors in this area.
- */
-int flash_area_to_sectors(int idx, int *cnt, struct flash_area *ret)
+static int validate_idx(int idx, uint32_t *off, uint32_t *len)
 {
-    uint32_t off;
-    uint32_t len;
-    uint32_t max_cnt = *cnt;
-    uint32_t rem_len;
-
     /*
      * This simple layout has uniform slots, so just fill in the
      * right one.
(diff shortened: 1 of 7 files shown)


mbolivar:
Bug for reference: https://runtimeco.atlassian.net/browse/MCUB-54

mkiiskila:
I have to admit, I'm not a fan of adding flash_device_base(), because it adds an API function.
On the other hand, parameter(s) to hal_system_start() don't really work when flash_area uses relative offsets. So clearly something has to change in the code.

I guess another option would be to have flash_area_XXX() routines to take care of offset conversion.
Which would avoid the API call, but would come with additional computation penalty when relative offsets are being used. Or to pass additional parameters to hal_system_start(), which could then do it's memory accesses with flash_area_read(), followed by jump to start address. Thoughts?

Adding struct flash_sector and it's use is acceptable to me; it is unlikely that we'd have 2nd image slot/scratch area span across different flashes. However, this function is a public function on MyNewt, and we have restrictions regarding API changes (it's much easier to add functions than remove them). Given that, I'd ask you to add a new function prototype and not redefine flash_area_to_sectors().

mbolivar:
Thanks for the feedback.

I'm not sure what you mean by "hav[ing] flash_area_XXX() routines to take care of offset conversion". Do you mean re-doing flash_device_base() to take a flash_area pointer instead? Something like:

```
int flash_area_base(const struct flash_area *area, uint32_t *base);
```

If not, could you please explain what you meant?

I understand compatibility is important since this is a core API. I'll add a new routine named `flash_area_get_sectors()` instead of changing `flash_area_to_sectors()`.

I see MyNewt has a deprecation policy:

https://cwiki.apache.org/confluence/display/MYNEWT/Release+and+Support+Policy#ReleaseandSupportPolicy-BackwardsCompatibility

Can we mark flash_area_to_sectors() as deprecated as part of this work?

mkiiskila:
Hi,

here's some answers (out of order):

Indeed, MyNewt will retire functions; it will take a few releases (in theory, this should translate to about 6 months of real-time of function being called as deprecated).

Here's what I meant by having flash offset computations inside flash_area_XXX():
Let's take first what flash_area would look like:

static const struct flash_area part_map[] = {
    {
        .fa_id = FLASH_AREA_IMAGE_0,
        .fa_off = FLASH_AREA_IMAGE_0_OFFSET + BOOT_FLASH_DEVICE_OFF,
        .fa_size = FLASH_AREA_IMAGE_0_SIZE,
    },
...
and so one, and then in the read function:
int flash_area_read(const struct flash_area *area, uint32_t off, void *dst,
            uint32_t len)
{
    return flash_read(boot_flash_device, area->fa_off + off - BOOT_FLASH_DEVICE_OFF, dst, len);
}

That being said, I'm not sure I have a strong preference of one solution over the other :)

As for the question of flash areas spanning multiple flashes: I don't believe there is anyone at the moment who would need it. I can come up with scenarios which are plausible, but I'm not sure whether anyone would design a system like that.
 
Hope this helps,
M

mbolivar:
Ah, I see; thanks for clarifying.

That would make "fa_off" as stored an "offset from zero", i.e. a real address, rather than an offset. I agree that change would let us keep the code as-is, but I'd argue that it's less correct - or at least, it's a bit misleading.

This decision seems to be between keeping the status quo at the cost of some potentially acceptable issues with the results, versus adding new API and handling its fallout.

Would anyone else like to comment?

boot/zephyr/flash_map.c
```diff
@@ -31,28 +31,45 @@
 
 extern struct device *boot_flash_device;
 
+/* For now, we only support one flash device. */
+#define FLASH_DEVICE_ID 0
```
d3zd3z: I wonder if we should use something non-zero, and unlikely to conflict with something else. That way we would catch instances where different APIs have different ids that are used for different purposes.

boot/zephyr/include/flash_map/flash_map.h
```diff
@@ -43,14 +43,64 @@ extern "C" {
  */
 #include <inttypes.h>
 
+/**
+ * @brief Structure describing an area of memory-mapped flash.
```
d3zd3z: This API should work fine for non memory-mapped flash. I don't think that is where the 'map' in the name comes from, it's more like just an alternate word chosen instead of 'partition'.

We do want to support non memory-mapped flash, and other than jumping to the code, it should currently work. Imagine an external SPI flash to hold slot 1 and scratch so that most of the flash could be used for the app.

boot/zephyr/include/flash_map/flash_map.h
```diff
@@ -43,14 +43,64 @@ extern "C" {
...
+    /**
+     * Size of this sector, in bytes.
+     */
+    uint32_t fs_size;
+};
+
+/*
+ * Retrieve a flash device's base address.
+ */
+int flash_device_base(uint8_t fd_id, uint32_t *ret);
```
d3zd3z: This needs to be able to fail, for the non-memory-mapped flash.  At least describe what the return code would be in that case.  Realistically, nobody should be configuring slot 0 as non memory-mapped, but it'd make debugging easier to catch that case.

mbolivar:
> It might be a little hard to tease out, but I wonder if the improvements to the flash_map stuff could be separated into its own patch.

Do you mean splitting up 7bf58ca somehow? There are 6 patches in the series; I tried to split them up into logical pieces. Which one is combining two different logical changes?

boot/zephyr/flash_map.c
mbolivar: Pick a card, any card :). Sure, I'll make this 0xAA.

boot/zephyr/include/flash_map/flash_map.h
mbolivar: That makes sense; I'll reword this. 

I think that also implies that we need a separate flash_device_base()-like routine which can return error in case the device isn't memory mapped. Keeping the current code, but putting real addresses into fa_off, wouldn't let callers tell the two situations apart, and hilarity could ensue.

boot/zephyr/include/flash_map/flash_map.h
mbolivar: Yup, it's already able to fail; the int return is an error code or zero.

In my defense, I was imitating the rest of this file's lack of detail when commenting prototypes, but I'm happy to do better on this function.

mbolivar:
Now updated to incorporate feedback.

- Add new flash_area_get_sectors(), deprecate but do not remove flash_area_to_sectors()
- Make the zephyr flash device ID a random value that's unlikely to collide
- Fix wording on memory-mapped flash comment
- Document that flash_device_base() can return an error (but don't say which ones, since Zephyr and MyNewt conventions seem to be different)
- Split the patch adding flash_area_get_sectors() from the changes to loader.c which use it

This seems like it's in more agreeable shape. Barring other comments, I'll start polishing this into a real series.

mbolivar:
GitHub doesn't handle rebases well and is not displaying these commits in order. My branch is here:

https://github.com/mbolivar/mcuboot/tree/flash-map-fixups

d3zd3z:
As far as the API's go, I think we should be using offsets and not physical addresses nearly all of the time. For this reason, I prefer what Marti has.

The need for a physical address is something that happens rarely, in this case just be able to boot. Technically we could also use the flash API to get the boot vector, but since we are about to jump to the code, we do need a real address at this point.

d3zd3z:
> Do you mean splitting up 7bf58ca somehow? There are 6 patches in the series; I tried to split them up into logical pieces. Which one is combining two different logical changes?

Sorry, my bad, I was viewing the default github squash-everything-together view instead of the individual patches.

mbolivar:
Closing RFC; will open another PR.

PR #47, opened 2017-05-05
Add logging to the simulator

d3zd3z:


boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -62,9 +62,54 @@ extern "C" {
 #include <logging/sys_log.h>
 
 /*
+ * When built on Linux (e.g. simulator builds), just use printf().
```
mbolivar: Leftovers from previous version.

PR #48, opened 2017-05-05
Add overwrite-only upgrade

d3zd3z:
Implement an overwrite-only upgrade as well as a query to determine at run time if that is available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ CFLAGS += -DBOOTUTIL_SIGN_RSA -DBOOTUTIL_USE_MBED_TLS
 # verification only happens on upgrade.
 CFLAGS += -DBOOTUTIL_VALIDATE_SLOT0
 
+# Enable this option to not use the swapping code and just overwrite
+# the image on upgrade.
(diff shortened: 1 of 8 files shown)


utzig:
Temporarily I think this is ok. But would be interesting to support both upgrade methods in the bootloader and define a semantics for the images to chose which method they prefer. What do you think?

d3zd3z:
> Temporarily I think this is ok. But would be interesting to support both upgrade methods in the bootloader and define a semantics for the images to chose which method they prefer. What do you think?

It would also help testing to be able to have both methods compiled in. However, the main motivation behind writing this code is to make the bootloader code more simple. People are a bit leery of complexity in the bootloader itself.

d3zd3z:
And, there actually are already semantics for the image to choose. If the image trailer has image_ok, the image will be marked as ok after the copy, and the bootloader will not try to revert. Perhaps in normal compilation, that should use this code instead of trying to swap. If anything, the upgrade should be faster that way.

d3zd3z:
Rebased to resolve merge conflicts.

PR #49, opened 2017-05-09
MCUB-54

mbolivar:
This series fixes MCUB-54 on Zephyr targets.

@d3zd3z, I could use some advice on how to port flash_area_get_sectors() to the simulator if you have some time. It's broken as-is, and I don't know Rust.

@mkiiskila, I hope this addresses your feedback and that it's acceptable for porting to MyNewt core.

The series adds new flash_device_base() and flash_area_get_sectors() routines to flash_map.h. If accepted, it needs to be propagated to mynewt as well, as it also ports the core bootutil library to use flash_area_get_sectors().

While updating split_go() to use flash_area_get_sectors(), I discovered it makes use of helper routines meant for boot_go(), which aren't quite right. Things are improved in "bootutil: use flash_area_get_sectors()", but more work is needed (see https://runtimeco.atlassian.net/browse/MCUB-56).

This series is based on https://github.com/runtimeco/mcuboot/pull/45, and incorporates feedback given there.

With these core updates, the Zephyr bootloader now uses flash_device_base() to fix a bug that manifests on STM32 devices depending on the MCU configuration, and warns when flash areas are left open at the end of boot.

### DIFF
--- a/apps/boot/src/boot.c
+++ b/apps/boot/src/boot.c
@@ -47,6 +47,7 @@ int
 main(void)
 {
     struct boot_rsp rsp;
+    uintptr_t flash_base;
     int rc;
 
 #if MYNEWT_VAL(BOOT_SERIAL)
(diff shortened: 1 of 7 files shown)


mbolivar:
I'm going to update this so loader.c continues to use flash_area_to_sectors() by default, but can be switched at compile time to use flash_area_get_sectors().

I also received a patch from @d3zd3z adding simulator support for both, which I will integrate into this series as well.

I am hoping that will be enough to get this merged, as the additional functionality will be present but shouldn't break any existing MyNewt targets.

mbolivar:
Thanks! I'm testing the updated version I described above and will probably have it ready to submit today.

mbolivar:
Will hold off and rebase after #41 is merged, as there are conflicts.

PR #51, opened 2017-05-10
sim: Allow Rust logging level to control C logging

d3zd3z:
Change the C logging code, when in the simulator, to query what the rust
logging level is set to.  This allows the level of logging from the C
code to be set through the environment.  For example

    RUST_LOG=bootsim=info cargo run --release runall

will enable logging at the "info" level for all of the C code as well as
the simulator code.  The C code's logging can be selected specifically
by using bootsim::api instead of just bootsim in the above.

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil_log.h
+++ b/boot/bootutil/include/bootutil/bootutil_log.h
@@ -78,30 +78,48 @@ extern "C" {
 #define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_ERROR
 #endif
 
+int sim_log_enabled(int level);
+
 #if BOOT_LOG_LEVEL >= BOOT_LOG_LEVEL_ERROR
-#define BOOT_LOG_ERR(_fmt, ...)						\
-    do { printf("[ERR] " _fmt "\n", ##__VA_ARGS__); } while (0)
+#define BOOT_LOG_ERR(_fmt, ...)                                         \
+    do {                                                                \
+        if (sim_log_enabled(BOOT_LOG_LEVEL_ERROR)) {                    \
+            printf("[ERR] " _fmt "\n", ##__VA_ARGS__);                  \
+        }                                                               \
+    } while (0)
 #else
 #define BOOT_LOG_ERR(...) IGNORE(__VA_ARGS__)
 #endif
 
 #if BOOT_LOG_LEVEL >= BOOT_LOG_LEVEL_WARNING
-#define BOOT_LOG_WRN(_fmt, ...)                                               \
-    do { printf("[WRN] " _fmt "\n", ##__VA_ARGS__); } while (0)
+#define BOOT_LOG_WRN(_fmt, ...)                                         \
+    do {                                                                \
+        if (sim_log_enabled(BOOT_LOG_LEVEL_WARNING)) {                  \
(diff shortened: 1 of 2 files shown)


boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -74,34 +74,58 @@ extern "C" {
...
 
+/*
+ * The compiled log level determines the minimum level that can be
+ * printed.  Messages at or below this level can be printed, provided
+ * they are also enabled through the Rust logging system, such as by
+ * setting RUST_LOG to bootsim::api=info.
+ */
 #ifndef BOOT_LOG_LEVEL
-#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_ERROR
+#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
```
mbolivar: Pedantically, seems like a separate patch. But not a big deal since I imagine you're one of the main users right now.

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -74,34 +74,58 @@ extern "C" {
 #define BOOT_LOG_LEVEL_INFO	3
 #define BOOT_LOG_LEVEL_DEBUG	4
 
+/*
+ * The compiled log level determines the minimum level that can be
+ * printed.  Messages at or below this level can be printed, provided
```
mbolivar: At or above

sim/src/api.rs
```diff
@@ -35,3 +36,23 @@ fn map_err(err: Result<()>) -> libc::c_int {
         },
     }
 }
+
+/// Called by C code to determine if we should log at this level.  Levels are defined in
+/// bootutil/bootutil_log.h.  This makes the logging from the C code controlled by bootsim::api, so
+/// for example, it can be enabled with something like:
+///     RUST_LOG=bootsim::api=info cargo run --release runall
```
mbolivar: This disagrees with the commit log; which is right?

boot/bootutil/include/bootutil/bootutil_log.h
d3zd3z: I think it belongs with this patch, since if the default level isn't changed, the runtime logging level won't do anything.

I do need to remove the defines that occur in other places, or at least make the defaults more reasonable for on-target.

boot/bootutil/include/bootutil/bootutil_log.h
d3zd3z: It's confusing since larger numbers are lower levels.

mbolivar: Then it should say maximum instead of minimum.

boot/bootutil/include/bootutil/bootutil_log.h
mbolivar: Maybe then the patch to change to info comes first, so the next patch is useful out of the box?

Again, though, I'm not trying to stop the merge with this comment.

sim/src/api.rs
d3zd3z: Hmm, either works, without the ::api it applies to the whole sim, but with it, it applies to just the C code. I'll make them consistent.

PR #52, opened 2017-05-15
Make mcuboot less mynewt style

utzig:
Remove most of mynewt specific stuff to a separate port package. This
should make mcuboot less "mynewt'y" and slightly easier to port to.

- Mynewt specific stuff moved to boot/mynewt.
- Sample app moved from apps/boot to boot/mynewt.
- Use MYNEWT_VAL macro only on mynewt port.
- BOOTUTIL_* and MYNEWT_VAL() usage moved to MCUBOOT_ defines.


boot/zephyr/keys.c
```diff
@@ -61,14 +61,14 @@ const unsigned char root_pub_der[] = {
     0x8b, 0x68, 0x34, 0xcc, 0x3a, 0x6a, 0xfc, 0x53,
     0x8e, 0xfa, 0xc1, };
 const unsigned int root_pub_der_len = 91;
-#else
-#error "No public key available for given signing algorithm."
```
d3zd3z: Why get rid of this error? Link will fail without it.

BTW, there is some desire to have at least the Zephyr make generate this key from a .pem file.

utzig: Ah, I see. mcuboot built on Zephyr is never supposed to be built without signing enabled. While testing I built without signing and had to remove this.

PR #53, opened 2017-05-16
scripts: Initial version of Python imgtool

d3zd3z:
This is the start of a python implementation of imgtool.  This
implements all of the functionality that was missing in the zep2newt.py
tool, namely creation of keypairs, and converting the public version of
these keys into C code.


scripts/imgtool.py
```diff
@@ -0,0 +1,109 @@
...
+    sign.add_argument("-v", "--version", type=decode_version, required=True)
+    sign.add_argument("-H", "--header-size", type=hexint, required=True)
+    sign.add_argument("--pad", type=hexint,
+            help='Pad image to this many bytes, adding trailer magic')
+    sign.add_argument("infile")
+    sign.add_argument("outfile")
+
+    args = parser.parse_args()
+    if args.subcmd is None:
+        print('Must specify a subcommand')
```
mbolivar: file=sys.stderr

scripts/imgtool.py
```diff
@@ -0,0 +1,109 @@
...
+    parser = argparse.ArgumentParser()
+    subs = parser.add_subparsers(help='subcommand help', dest='subcmd')
+
+    keygenp = subs.add_parser('keygen', help='Generate pub/private keypair')
+    keygenp.add_argument('-k', '--key', metavar='filename', required=True)
+    keygenp.add_argument('-t', '--type', metavar='type',
+            choices=['rsa-2048', 'ecdsa-p224', 'ecdsa-p256'],
+            required=True)
+
+    getpub = subs.add_parser('getpub', help='Get public key from keypair')
```
mbolivar: If you end up adding a `file` kwarg to `emit_c()`, having a '-o' option here which put the output into the given file would be great.

scripts/imgtool.py
```diff
@@ -0,0 +1,109 @@
...
+    getpub.add_argument('-k', '--key', metavar='filename', required=True)
+
+    sign = subs.add_parser('sign', help='Sign an image with a private key')
+    sign.add_argument('-k', '--key', metavar='filename', required=True)
+    sign.add_argument("--align", type=alignment_value, required=True)
```
mbolivar: It would be super awesome if this script kept track of these values (alignment, image size, etc.) for all Zephyr targets supported by mcuboot, and accepted a `-b ZEPHYR_BOARD_NAME` instead. That would mean callers wouldn't have to keep track of this stuff. You could still fall back on --align et al if the board was supported out of tree.

Doing this would also let future versions of this script dump this information in machine-readable formats. That could be used to build tooling on top of upstream mcuboot itself, which application build systems could use to build and link Zephyr apps which want to be chain-loaded by mcuboot.

I can help doing this; I've got some of the needed infrastructure already.

scripts/imgtool.py
```diff
@@ -0,0 +1,109 @@
...
+                int(m.group(5)) if m.group(5) else 0,
+                int(m.group(7)) if m.group(7) else 0)
+        return result
+    else:
+        msg = "Invalid version number, should be maj.min.rev+build with later parts optional"
+        raise argparse.ArgumentTypeError(msg)
+
+def args():
+    parser = argparse.ArgumentParser()
+    subs = parser.add_subparsers(help='subcommand help', dest='subcmd')
```
d3zd3z: It sure took some digging to find out. Of the things I like about Python, documentation is not high on that list. It suffers from the document-the-pieces, but forget about the whole, problem.

scripts/imgtool.py
```diff
@@ -0,0 +1,109 @@
...
+
+    keygenp = subs.add_parser('keygen', help='Generate pub/private keypair')
+    keygenp.add_argument('-k', '--key', metavar='filename', required=True)
```
d3zd3z: Within reason. I think 2 positional arguments is fine. Needing 7 gets hard to remember, and I don't find it all that helpful.

scripts/imgtool.py
d3zd3z: Part of the problem is that this script runs at the wrong end of the chain. It would make more sense for this script to be able to extract the information from a particular build. Maybe I should do that, an argument that you point to a built tree (where zephyr.{elf,bin} lives), and it gets it from the generated headers there. That will depend on the DTS patches implementing those headers, though.

scripts/imgtool/keys.py
```diff
@@ -0,0 +1,129 @@
+"""
+Cryptographic key management for imgtool.
+"""
+
+from __future__ import print_function
```
d3zd3z: Yeah, it was an artifact of me writing keys.py before starting the rest of the code, or thinking about it.

Is there a way to bomb quickly if the script is run in python2?

d3zd3z: Wait, this is python, the solution is supposed to be boring: if sys.version[0] < '3'...

scripts/imgtool/keys.py
```diff
@@ -0,0 +1,129 @@
...
+from ecdsa import SigningKey, NIST256p, util
+from pyasn1.type import namedtype, univ
+from pyasn1.codec.der.encoder import encode
+
+class RSAPublicKey(univ.Sequence):
+    componentType = namedtype.NamedTypes(
+            namedtype.NamedType('modulus', univ.Integer()),
+            namedtype.NamedType('publicExponent', univ.Integer()))
+
+class RSA2048():
```
d3zd3z: I thought about the base class, but wasn't sure what advantage it gives? Maybe documentation, but I'm not aware of python actually enforcing anything related to typing, even at runtime.

scripts/imgtool/keys.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @staticmethod
+    def generate():
+        return RSA2048(RSA.generate(2048))
+
+    def export_private(self, path):
+        with open(path, 'wb') as f:
+            f.write(self.key.exportKey('PEM'))
+
+    def emit_c(self):
```
d3zd3z: We also probably want to be able to insert the key into an existing file, rather than just spitting it out and requiring either the user, or makefile work to do it.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
+"""
```
d3zd3z: Does that mainly allow "from imgtool import *"?

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+    def load(cls, path, **kwargs):
+        """Load an image from a given file"""
+        with open(path, 'rb') as f:
+            payload = f.read()
+        obj = cls(**kwargs)
+        obj.payload = payload
+        obj.check()
+        return obj
+
+    def __init__(self, version="0", header_size=IMAGE_HEADER_SIZE, pad=0):
```
d3zd3z: Heh, yeah missed that before creating the SemiSemVersion class.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+                len(self.payload))
+
+    def save(self, path):
+        with open(path, 'wb') as f:
+            f.write(self.payload)
+
+    def check(self):
+        """Perform some sanity checking of the image."""
+        # If there is a header requested, make sure that the image
+        # starts with all zeros.
```
d3zd3z: Yes, it breaks on x86, and I don't seem to be able to even find that code. I can ask Andrew (or just try a build on x86).

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+        self.payload += tlv.get()
+
+    def add_header(self, key):
+        """Install the image header.
+
+        The key is needed to know the type of signature, and
+        approximate the size of the signature."""
+
+        flags = IMAGE_F[key.sig_type()]
+        tlvsz = 4 + 32
```
d3zd3z: Maybe, at least for the 3. I find sometimes single-use definitions instead of numbers can make code even harder to follow. In this case, the 32 should come from something in the SHA256 library about the size of the hash.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+            ) # }
+        assert struct.calcsize(fmt) == IMAGE_HEADER_SIZE
+        header = struct.pack(fmt,
+                IMAGE_MAGIC,
+                tlvsz, # TlvSz
+                0, # KeyId (TODO: allow other ids)
+                0,  # Pad1
+                self.header_size,
+                0, # Pad2
+                # self.header_size + len(self.payload), # ImageSz
```
d3zd3z: The comment was the broken code, and should just be removed. I had to dig through the image verification code to figure out what they were expecting. There is some weird inconsistency with the header, and sizes.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+                self.version.major,
+                self.version.minor or 0,
+                self.version.revision or 0,
+                self.version.build or 0,
+                0) # Pad3
+        self.payload = bytearray(self.payload)
+        self.payload[:len(header)] = header
+
+    def pad_to(self, size, align):
+        """Pad the image to the given size, with the given flash alignment."""
```
d3zd3z: Good idea. Hopefully, soon, Fabio will make the change to the trailer so that none of this code needs to know the alignment, though.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+        tsize = trailer_sizes[align]
+        padding = size - (len(self.payload) + tsize)
+        if padding < 0:
+            raise Exception("Image exceeds requested size + trailer")
```
d3zd3z: Aww, how about
    raise Exception("NO")
yeah, that makes sense, though.

mbolivar: > raise Exception("NO")

I love it!

scripts/imgtool/image.py
mbolivar: Yeah, it allows you to control what happens when someone imports *. If you ever add more code to this init file, those definitions would end up in the namespace of anybody that imports * on this package. Maybe that's not a problem though.

scripts/imgtool/keys.py
mbolivar: I suppose; inserting into an existing file at anything other than the beginning and the end gets messy, though.

(And you could support that in higher layers by passing an io.StringIO to as a file kwarg, then putting the StringIO contents into an existing file in the right spot.)

scripts/imgtool/keys.py
mbolivar: Well, for example, "A class that has a metaclass derived from ABCMeta cannot be instantiated unless all of its abstract methods and properties are overridden."

https://docs.python.org/3.6/library/abc.html#abc.abstractmethod

scripts/imgtool/keys.py
mbolivar: That can go in `__init__.py` at package level, right?

scripts/imgtool.py
mbolivar: At least all of the ARM boards in Zephyr have a DTS, but I take your point: that's less useful for the other arches.

scripts/imgtool.py
mbolivar: Fair enough.

scripts/imgtool/image.py
mbolivar: Here it is: https://github.com/runtimeco/mcuboot/blob/master/scripts/zep2newt.py#L178

scripts/imgtool.py
d3zd3z: How about if we leave it as is for now, and a subsequent patch can add some of this improved functionality, including extracting flash configuration, and editing the key file.

It'd also be nice since there is pending work in MCUboot that should eliminate the need for this code to even know the alignment.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+    0x60, 0xd2, 0xef, 0x7f,
+    0x35, 0x52, 0x50, 0x0f,
+    0x2c, 0xb6, 0x79, 0x80, ])
+
+class TLV():
+    def __init__(self):
+        self.buf = bytearray()
+
+    def add(self, kind, payload):
+        """Add a TLV record.  Kind should be a string found in TLV_VALUES above."""
```
d3zd3z: It will raise a KeyError. This does indicate a programming error, and doesn't come from the user, so I don't see a strong reason to generate a different exception.

scripts/imgtool/image.py
d3zd3z: Since the Go imgtool doesn't do this right, either, how about I do this in a future change, that fixes both of them.

scripts/imgtool.py
mbolivar: Yup, as discussed off-PR, that sounds good to me.

scripts/imgtool/image.py
mbolivar: That makes sense.

scripts/imgtool/image.py
mbolivar: Seems reasonable to get Python to feature parity and then fix bugs.

scripts/imgtool/keys.py
d3zd3z: Hmm, I wasn't aware of the abc stuff. Seems like an awful lot of stuff to pull in for what is essentially two classes, defined next to each other, though.

scripts/imgtool/keys.py
d3zd3z: Typically you'll have a template file, and a generated file, and the keys will get inserted based on a comment that it looks for in the template. Otherwise, you have to build the file using either multiple commands in a makefile, or another python program.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,163 @@
...
+    def save(self, path):
+        with open(path, 'wb') as f:
+            f.write(self.payload)
+
+    def check(self):
+        """Perform some sanity checking of the image."""
+        # If there is a header requested, make sure that the image
+        # starts with all zeros.
+        if self.header_size > 0:
+            if not all(v == 0 for v in self.payload[0:self.header_size]):
```
utzig: I think both above tests could be written more succinctly with:
`if any(v != 0 for v in payload[:self.header_size]):`

mbolivar: Why not just `any(self.payload[:self.header_size])`?

utzig: It would work but somehow, to me at least it doesn't tell the meaning of what it's being done as well. Maybe I'm just not "pythonist" enough! ;-)

mbolivar: PEP 008 is silent on this -- it only has rules about comparing to `None` (always use `x is None` or `x is not None`, never `x == None` etc.) and booleans (use `not x`, not `x == False` etc.).

I prefer using the fact that integers are truthy in the expected/C-like way to write `if x` instead of `if x != 0` (and similarly when using `any`, `all`, etc.) because it's similar style to how comparisons to None and booleans are done, but it's a personal preference.

d3zd3z: I consider truthy integers to be a significant language misfeature. I don't do it in C, either.

mbolivar: Well, given that in Python, just about everything can be interpreted in sensible ways in a boolean context with well-defined methods, I think it's less of a misfeature than in languages which special-case interpreting certain built-in types as booleans. But I still agree it's a matter of preference.

scripts/imgtool/version.py
```diff
@@ -0,0 +1,27 @@
...
+    if m:
+        result = SemiSemVersion(
+                int(m.group(1)) if m.group(1) else 0,
+                int(m.group(3)) if m.group(3) else 0,
+                int(m.group(5)) if m.group(5) else 0,
+                int(m.group(7)) if m.group(7) else 0)
+        return result
+    else:
+        msg = "Invalid version number, should be maj.min.rev+build with later parts optional"
+        raise argparse.ArgumentTypeError(msg)
```
kYc0o: This error gives an exception if `import argparse` is missing on this file.

d3zd3z: Thanks, I've fixed this, as well as disallowing zeros in the patch "imgtool: Better version parsing".

doc/imgtool.md
```diff
@@ -0,0 +1,88 @@
...
+operations that are necessary to manage keys and sign images.  Using
+this script should be preferred to the manual steps described in
+`doc/signed_images.md`.
+
+This program is written for Python3, and has several dependencies on
+Python libraries.  These can be installed using 'pip3' manually:
+
+    pip install --user pycrypto
+    pip install --user pyasn1
+    pip install --user ecdsa
```
kYc0o: I think this should be `pip3 install ...` since the script works only for python3.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+        self.payload += tlv.get()
+
+    def add_header(self, key):
+        """Install the image header.
+
+        The key is needed to know the type of signature, and
+        approximate the size of the signature."""
+
+        flags = IMAGE_F[key.sig_type()]
+        tlvsz = 0
```
mbolivar: This and the next line should just be `tlvsz = TLV_HEADER_SIZE + key.sig_len()`.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+        tlvsz += TLV_HEADER_SIZE + key.sig_len()
+
+        flags |= IMAGE_F['SHA256']
+        tlvsz += 4 + hashlib.sha256().digest_size
```
mbolivar: Use `TLV_HEADER_SIZE` instead of magic number 4.

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+                0) # Pad3
+        self.payload = bytearray(self.payload)
+        self.payload[:len(header)] = header
+
+    def pad_to(self, size, align):
+        """Pad the image to the given size, with the given flash alignment."""
+        tsize = trailer_sizes[align]
+        padding = size - (len(self.payload) + tsize)
+        if padding < 0:
+            msg = "Image size (0x{:x}) + trailer (0x{:x}) exceeds requested size 0x{:x}".format(
```
mbolivar: As discussed off GitHub, calling this the "image size" is misleading, since it has the signature and SHA256 digest TLVs appended already. People can mistake this with the size of their input image (which is what I did at first), especially if they are not familiar with the internals of the mcuboot image format.

Can this error message get cleaned up a bit to make it more clear what happened?

scripts/imgtool/image.py
```diff
@@ -0,0 +1,164 @@
...
+        with open(path, 'rb') as f:
+            payload = f.read()
+        obj = cls(**kwargs)
+        obj.payload = payload
+        obj.check()
+        return obj
+
+    def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE, pad=0):
+        self.version = version or versmod.decode_version("0")
+        self.header_size = header_size or IMAGE_HEADER_SIZE
```
d3zd3z: Except that args.header_size can still be None, so it has to be set here. Maybe I should just take the default out of the argument list, so it is required.

scripts/imgtool/image.py
d3zd3z: I had it that way earlier. I think it is clearer this way, especially as things get moved around. Each thing that is added, adds to the tlvsize as well as adding items.

It is a little unfortunate that we have to even know the size of the tlv before we can make it, since the signature covers the header itself.

scripts/imgtool/keys.py
```diff
@@ -0,0 +1,137 @@
...
+from Crypto.PublicKey import RSA
+from Crypto.Signature import PKCS1_v1_5, PKCS1_PSS
+from ecdsa import SigningKey, NIST256p, util
+import hashlib
+from pyasn1.type import namedtype, univ
+from pyasn1.codec.der.encoder import encode
+
+# By default, we use RSA-PSS (PKCS 2.1).  That can be overridden on
+# the command line to support the older (less secure) PKCS1.5
+sign_rsa_pss = True
```
d3zd3z: Yes, the better way is to remove the v1.5 code entirely. But we have to do something gradually.

I don't know if it is really worth supporting what hopefully will be going away soon.

d3zd3z: The Go code is even worse, but hopefully it can end up with the same ultimate solution.

scripts/imgtool/image.py
mbolivar: do_sign should initialize the header_size kwarg appropriately -- putting that into the library is an abstraction violation.

scripts/imgtool/image.py
mbolivar: Then you should have a `flags = 0 /* ... */ flags |= IMAGE_F[key.sig_type()]` for consistency.

scripts/imgtool/keys.py
mbolivar: If it's going away soon, I'm fine with leaving the hack in.

PR #57, opened 2017-05-31
rsa: Add support for RSA-PSS

d3zd3z:
The PKCS#1 standards, which define RSA signatures, are currently at
version 2.2.  Starting in v2.1, the standard defines a new signature
method RSA-PSS, which has a stronger security proof than the signature
method used in earlier versions.  The standard recommends that RSA-PSS
be used in new designs, instead of the older algorithm.

This patch implements RSA-PSS verification for a specific set of
parameters:

    - RSA-2048
    - SHA256 for both the message digest and the internal hash
    - 32-byte salt
    - 2047 bit message

Although RSA-PSS supports other parameters, due to size constraints,
this verificatino code only supports these specific parameters, and
signatures with other parameters will be considered invalid.

To encourage the use of the more secure algorithm, the default build
configuration is RSA-PSS.  BOOTUTIL_RSA_PKCS1_15 needs to be defined in
order to support the older signature algorithm.


d3zd3z:
Don't pull until tool changes go into imgtool imgtool.py and newt.

utzig:
Looks good!

boot/bootutil/src/image_rsa.c
```diff
@@ -71,6 +106,157 @@ bootutil_parse_rsakey(mbedtls_rsa_context *ctx, uint8_t **p, uint8_t *end)
...
+ * Validate an RSA signature, using RSA-PSS, as described in PKCS #1
+ * v2.2, section 9.1.2, with many parameters required to have fixed
+ * values.
+ */
+static int
+bootutil_cmp_rsasig(mbedtls_rsa_context *ctx, uint8_t *hash, uint32_t hlen,
+  uint8_t *sig)
+{
+    bootutil_sha256_context shactx;
+    uint8_t buf[MBEDTLS_MPI_MAX_SIZE];
```
jbech-linaro: So `buf` here is the `DB`, so I suggest that you rename the variable so it's easier to follow what is going on when comparing this with specs etc.

boot/bootutil/src/image_rsa.c
```diff
@@ -71,6 +106,157 @@ bootutil_parse_rsakey(mbedtls_rsa_context *ctx, uint8_t **p, uint8_t *end)
...
+     * octet in maskedDB are not all equal to zero, output
+     * inconsistent and stop.
+     *
+     * 8emLen - emBits is zero, so there is nothing to test here.
+     */
+
+    /* Step 7.  let dbMask = MGF(H, emLen - hLen - 1). */
+    pss_mgf1(db_mask, &buf[PSS_HASH_OFFSET]);
+
+    /* Step 8.  let DB = maskedDB xor dbMask. */
```
jbech-linaro: It should be:
```
/* Step 8.  let maskedDB = DB xor dbMask. */
```
Right?

boot/bootutil/src/image_rsa.c
```diff
@@ -71,6 +106,157 @@ bootutil_parse_rsakey(mbedtls_rsa_context *ctx, uint8_t **p, uint8_t *end)
...
+        }
+    }
+
+    if (db_mask[PSS_MASK_ONE_POS] != 1) {
+        return -1;
+    }
+
+    /* Step 11. Let salt be the last sLen octets of DB */
+
+    /* Step 12.  Let M' = 0x00 00 00 00 00 00 00 00 || mHash || salt; */
```
jbech-linaro: Where is this done? Is that what you are saying in "step 2" ? I.e., this is the input to the function?

boot/bootutil/src/image_rsa.c
```diff
@@ -71,6 +106,157 @@ bootutil_parse_rsakey(mbedtls_rsa_context *ctx, uint8_t **p, uint8_t *end)
...
 
+#if !MYNEWT_VAL(BOOTUTIL_RSA_PKCS1_15)
+/*
+ * Compute the RSA-PSS mask-generation function.  Assumptions are that
+ * the mask length will be less than 256 * PSS_HLEN, and therefore we
+ * never need to increment anything other than the low byte of the
+ * counter.
+ */
+static void
+pss_mgf1(uint8_t *mask, const uint8_t *hash)
```
d3zd3z: It is called that in the PKCS#1 standard.

boot/bootutil/src/image_rsa.c
```diff
@@ -71,6 +106,157 @@ bootutil_parse_rsakey(mbedtls_rsa_context *ctx, uint8_t **p, uint8_t *end)
...
+
+    if (mbedtls_rsa_public(ctx, sig, buf)) {
+        return -1;
+    }
+
+    /*
+     * PKCS #1 v2.2, 9.1.2 EMSA-PSS-Verify
+     *
+     * emBits is 2048
+     * emLen = ceil(emBits/8) = 256
```
d3zd3z: Hmm, I copied this text out of PKCS #1.

boot/bootutil/src/image_rsa.c
d3zd3z: Again, copied directly out of the spec. However, I modify `buf` instead of the making an additional buffer for the unmasked db. The maskedDB is in the message, and the xoring extracts the DB.

boot/bootutil/src/image_rsa.c
d3zd3z: The code below step 13 performs the Hash(M'). Since the SHA256 code allows multiple updates, there isn't a need to have a separate buffer to hold the prepended zeros, hash, and salt. I can try to clarify that in the comments.

boot/bootutil/src/image_rsa.c
d3zd3z: `buf` is actually `EM` the encoded message. `db_mask` starts as the mask, but then becomes `db` once the xor is finished. I'll try to explain that a little more clearly.

PR #60, opened 2017-06-08
Make header padding optional

d3zd3z:
Zephyr defaults to having zero bytes at the beginning of the binary image as a gap allowing for the header. Other targets do not, so accommodate that.

This is built on top of pr#53 and shouldn't be merged until it goes in. Only the last patch is relevant to this PR.

### DIFF
--- a/doc/imgtool.md
+++ b/doc/imgtool.md
@@ -66,6 +66,7 @@ header and trailer that the bootloader is expecting:
       --align ALIGN
       -v VERSION, --version VERSION
       -H HEADER_SIZE, --header-size HEADER_SIZE
+      --included-header     Image has gap for header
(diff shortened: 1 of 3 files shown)


doc/imgtool.md
```diff
@@ -75,8 +76,10 @@ flash device in question, and the header size.
 
 The header size depends on the operating system and the particular
 flash device.  For Zephyr, it will be configured as part of the build,
-and will be a small power of two.  The generated image should start
-with zero bytes of this length (and the script will check this).
+and will be a small power of two.  By default, the header will be
+prepended to the image.  If `--included-header`` is given, the image
```
utzig: This double closing quote seems strange here...

d3zd3z: Habit from rst. I'll fix it.

PR #62, opened 2017-06-10
Fix broken Mynewt BOOTUTIL_ -> MCUBOOT_ mapping

utzig:
Fix mapping of syscfg.yml BOOTUTIL_* vars to C #defines, which was broken
on most .c files for lacking inclusion of mynewt "config.h"


utzig:
I'm merging this because it's a straight-forward change, which fixes a broken Mynewt build when using signing.

PR #66, opened 2017-06-14
Flash map updates, redux

mbolivar:
This is basically a rewrite of the changes previously merged and then reverted in https://github.com/runtimeco/mcuboot/pull/49.

It gets us 99% of the way to resolving MCUB-54. The remaining portions require an out of tree series of patches in Zephyr to be merged (https://github.com/zephyrproject-rtos/zephyr/pull/205).

The main fix is to introduce, and then use, flash_device_base(). However, while we're changing API that mynewt also uses, I take the opportunity to add flash_area_get_sectors() and deprecate flash_area_to_sectors().

@d3zd3z, this switches to the new flash API by default in the simulator.

@mkiiskila, this should be even better for mynewt: you can take your time and implement flash_device_base() and flash_area_get_sectors() whenever you want. I have left a default implementation of flash_device_base() for mynewt that keeps existing behavior for now, and leave flash_area_to_sectors() as the default in the bootloader as well.

(Zephyr targets use 'real' versions of both).

@jamike, this series plus the Zephyr series I posted above should get you going on STM32F4. You need to change prj.conf in mcuboot to allow flash writes as well.

### DIFF
--- a/boot/zephyr/flash_map.c
+++ b/boot/zephyr/flash_map.c
@@ -40,29 +40,46 @@ extern struct device *boot_flash_device;
 #define FLASH_DEVICE_ID 100
 #define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS
 
+#define FLASH_MAP_ENTRY_MAGIC 0xd00dbeef
+
+struct flash_map_entry {
+    const uint32_t magic;
+    const struct flash_area area;
+    unsigned int ref_count;
+};
+
 /*
  * The flash area describes essentially the partition table of the
  * flash.  In this case, it starts with FLASH_AREA_IMAGE_0.
  */
-static const struct flash_area part_map[] = {
+static struct flash_map_entry part_map[] = {
     {
-        .fa_id = FLASH_AREA_IMAGE_0,
-        .fa_device_id = FLASH_DEVICE_ID,
-        .fa_off = FLASH_AREA_IMAGE_0_OFFSET,
-        .fa_size = FLASH_AREA_IMAGE_0_SIZE,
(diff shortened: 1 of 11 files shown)


mbolivar:
For anyone reviewing the changes: the branch has been rebased heavily, and the GitHub UI is not displaying the commits in their correct order. I recommend fetching it from my fork into a local clone; the commits make a lot more sense that way, and it's the order they'd be merged in.

mbolivar:
In addition: the first three commits, if accepted, should be applied to the mynewt core version of flash_map.h as well. Any advice on how we could make that happen?

utzig:
I'll take a look later in the day or tomorrow.

mbolivar:
v2: fixed commit log for "bootutil: optionally use flash_area_get_sectors()"; no code changes.

boot/zephyr/flash_map.c
```diff
@@ -176,30 +165,44 @@ int flash_area_to_sectors(int idx, int *cnt, struct flash_area *ret)
...
-    }
-
     switch (idx) {
     case FLASH_AREA_IMAGE_0:
-        off = FLASH_AREA_IMAGE_0_OFFSET;
-        len = FLASH_AREA_IMAGE_0_SIZE;
-        break;
+        *off = FLASH_AREA_IMAGE_0_OFFSET;
+        *len = FLASH_AREA_IMAGE_0_SIZE;
+        goto done;
```
d3zd3z: Probably cleaner to just leave these as `break`.

mbolivar: Thanks, done.

PR #70, opened 2017-06-21
Mynewt fixes

utzig:


boot/bootutil/src/loader.c
```diff
@@ -222,6 +222,26 @@ boot_previous_swap_type(void)
...
+        return BOOT_EFLASH;
+    }
+
+    rc = flash_area_read(fap, 0, out_hdr, sizeof *out_hdr);
+    if (rc != 0) {
+        flash_area_close(fap);
+        return BOOT_EFLASH;
+    }
+
+    return 0;
```
d3zd3z: The close doesn't happen in the successful read case.

boot/bootutil/src/loader.c
```diff
@@ -234,11 +234,11 @@ boot_read_header_from_scratch(struct image_header *out_hdr)
 
     rc = flash_area_read(fap, 0, out_hdr, sizeof *out_hdr);
     if (rc != 0) {
-        flash_area_close(fap);
-        return BOOT_EFLASH;
+        rc = BOOT_EFLASH;
     }
 
-    return 0;
+    flash_area_close(fap);
```
d3zd3z: This is good, but we should probably squash it into the previous commit, so that we don't subtly break git bisect.

utzig: Done

PR #71, opened 2017-06-22
zephyr: frdm_k64f: Remove partition defines

d3zd3z:
Now that the partition table has been added to the device tree for the
frdm_k64f in upstream Zephyr, these symbols become redundant defines.
Remove them to fully use the partitions defined in Zephyr.


utzig:
It looks good. What did the other targets not change? What is the process of setting up the partitions for freedom k64f after this PR is merged?

d3zd3z:
In order for partitions to work, the partition table has to be added to the device tree in Zephyr first. If it is added and this isn't updated, the compiler generates warnings about duplicate #defines. If the change goes into mcuboot first, the result will not compile. So, we want to make sure they are merged into Zephyr first. I'll probably do the stm32l475 soon, since I have one, and will keep an eye on definitions that go into other SOCs into Zephyr.

d3zd3z:
As far as how to set up the partitions, they are defined in the device tree within Zephyr in `dts/arm/nxp/nxp_k6x.dtsi`. That way every app that builds with that Zephyr tree will have the same partitions defined.

PR #74, opened 2017-06-26
Remove old 'imgtool' and 'zep2newt.py'

d3zd3z:
Now that script/imgtool.py contains all of the functionality of these
tools, remove the old tools from the tree.  The zep2newt.py was only
able to perform a limited set of signatures.  The 'imgtool' requires a
'Go' toolchain, which has been a hinderance to many users.

Please see `doc/imgtool.md` for directions on how to use this new tool
for image signing.


mbolivar:
ACK

PR #75, opened 2017-06-28
Fixes trying to close an invalid flash map handle

utzig:
This error was catched by Coverity and it happens when a fail occurs
opening a flash map handle, which is not checked by the close
routine.

Right now this only affects Zephyr, but extra checking was added
assuming that in a future Mynewt implementation close could actually
be changed to do something.


utzig:
@d3zd3z Should we use "Rebase and merge" from now on? Doesn't seem to be the default if I just click "Merge pull request"...

PR #76, opened 2017-06-28
New trailer format

utzig:
This implements issue MCUB-14.

Trailer was updated with magic moved to the end and flags now are 8 bytes no matter which is the flash write size. This makes it easier to write external apps the can read/write this data without having the get extra info from the bootloader itself.

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil.h
+++ b/boot/bootutil/include/bootutil/bootutil.h
@@ -40,6 +40,9 @@ extern "C" {
 
 #define BOOT_SWAP_TYPE_FAIL     0xff
 
+#define MAX_FLASH_ALIGN         8
+extern const uint32_t BOOT_MAX_ALIGN;
+
 struct image_header;
 /**
  * A response object provided by the boot loader code; indicates where to jump
@@ -55,6 +58,17 @@ struct boot_rsp {
      */
     uint8_t br_flash_dev_id;
     uint32_t br_image_off;
+};
+
+/* This is not actually used by mcuboot's code but can be used by apps
+ * when attempting to read/write a trailer.
+ */
+struct image_trailer {
+    uint8_t copy_done;
+    uint8_t pad1[MAX_FLASH_ALIGN - 1];
+    uint8_t image_ok;
+    uint8_t pad2[MAX_FLASH_ALIGN - 1];
+    uint8_t magic[16];
 };
 
 /* you must have pre-allocated all the entries within this structure */
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -20,6 +20,7 @@
 #include <assert.h>
 #include <string.h>
 #include <inttypes.h>
+#include <stddef.h>
 
 #include "sysflash/sysflash.h"
 #include "hal/hal_bsp.h"
@@ -43,6 +44,7 @@ const uint32_t boot_img_magic[] = {
 };
 
 const uint32_t BOOT_MAGIC_SZ = sizeof boot_img_magic;
+const uint32_t BOOT_MAX_ALIGN = MAX_FLASH_ALIGN;
 
 struct boot_swap_table {
     /** * For each field, a value of 0 means "any". */
@@ -164,37 +166,25 @@ boot_magic_code(const uint32_t *magic)
 uint32_t
 boot_slots_trailer_sz(uint8_t min_write_sz)
 {
-    return BOOT_MAGIC_SZ                                                    +
-           /* state for all sectors */
+    return /* state for all sectors */
            BOOT_STATUS_MAX_ENTRIES * BOOT_STATUS_STATE_COUNT * min_write_sz +
-           /* copy_done + image_ok */
-           min_write_sz * 2;
+           BOOT_MAX_ALIGN * 2 /* copy_done + image_ok */                    +
+           BOOT_MAGIC_SZ;
 }
 
 static uint32_t
 boot_scratch_trailer_sz(uint8_t min_write_sz)
 {
-    return BOOT_MAGIC_SZ                          +  /* magic */
-           BOOT_STATUS_STATE_COUNT * min_write_sz +  /* state for one sector */
-           min_write_sz;                             /* image_ok */
+    return BOOT_STATUS_STATE_COUNT * min_write_sz +  /* state for one sector */
+           BOOT_MAX_ALIGN                         +  /* image_ok */
+           BOOT_MAGIC_SZ;
 }
 
 static uint32_t
 boot_magic_off(const struct flash_area *fap)
 {
-    uint32_t off_from_end;
-    uint8_t elem_sz;
-
-    elem_sz = flash_area_align(fap);
-
-    if (fap->fa_id == FLASH_AREA_IMAGE_SCRATCH) {
-        off_from_end = boot_scratch_trailer_sz(elem_sz);
-    } else {
-        off_from_end = boot_slots_trailer_sz(elem_sz);
-    }
-
-    assert(off_from_end <= fap->fa_size);
-    return fap->fa_size - off_from_end;
+    assert(offsetof(struct image_trailer, magic) == 16);
+    return fap->fa_size - BOOT_MAGIC_SZ;
 }
 
 int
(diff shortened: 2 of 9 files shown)


d3zd3z:
We also need to update the padding code in imgtool.py. We can remove the --align flag, and simplify the padding to use this new format.

boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -57,6 +60,17 @@ struct boot_rsp {
     uint32_t br_image_off;
 };
 
+/* This is not actually used by mcuboot's code but can be used by apps
```
mbolivar: Shouldn't there be an ioctl() or something for this? What if the format changes in v2?

In general, I think we should consider how to share code between the bootloader and applications a bit. Putting this into an existing header alongside code that applications should not use seems like asking for trouble, a bit. Could we start by having a header which is only for shared definitions?

boot/bootutil/src/bootutil_misc.c
```diff
@@ -43,6 +44,7 @@ const uint32_t boot_img_magic[] = {
 };
 
 const uint32_t BOOT_MAGIC_SZ = sizeof boot_img_magic;
+const uint32_t BOOT_MAX_ALIGN = MAX_FLASH_ALIGN;
```
mbolivar: Why not just use the macro?

boot/bootutil/include/bootutil/bootutil.h
utzig: Well, I agree that API could use this but I'll wait for this one to be merged and send another PR over that one or update the existing PR if it's not merged by then (most probable situation!).

Btw, I added the original `struct` as proposed and added `offsetof` asserts to guarantee the relevante offsets are not changed in the code. That `struct` is supposed to be used by "clients". Creating an include with only "exported" stuff seems like a good idea.

boot/bootutil/src/bootutil_misc.c
utzig: Rust cannot import `#define`s and a `const` variable with `extern` cannot be used to size arrays in a header so this was the best option IMO.

mbolivar: Ah.

mbolivar:
> I'll wait for this one to be merged and send another PR over that one or update the existing PR if it's not merged by then

Will that cross the 1.0 release boundary?

utzig:
> > I'll wait for this one to be merged and send another PR over that one or update the existing PR if it's not merged by then

> Will that cross the 1.0 release boundary?

We had pretty much decided on moving forward leaving the API stuff out of 1.0 on our last meeting but if people (on Zephyr ML) are good with adding the API as is I think we can add it and update the ecosystem later (eg Mynewt's patches need to be officially added/merged, and so on...)

mbolivar:
> We had pretty much decided on moving forward leaving the API stuff out of 1.0 on our last meeting but if people (on Zephyr ML) are good with adding the API as is I think we can add it and update the ecosystem later (eg Mynewt's patches need to be officially added/merged, and so on...)

Sorry, just to confirm what you mean, then: this patch would be part of mcuboot 1.0, and if it becomes API later, that's post-1.0, right?

utzig:
> We also need to update the padding code in imgtool.py. We can remove the --align flag, and simplify the padding to use this new format.

The status bytes still depend on "write size" so I haven't removed the `--align` flag. Should I assume always 8 byte write size? That could error if the padding is small enough, I think...

utzig:
> > We had pretty much decided on moving forward leaving the API stuff out of 1.0 on our last meeting but if people (on Zephyr ML) are good with adding the API as is I think we can add it and update the ecosystem later (eg Mynewt's patches need to be officially added/merged, and so on...)

> Sorry, just to confirm what you mean, then: this patch would be part of mcuboot 1.0, and if it becomes API later, that's post-1.0, right?

What I meant was that this PR here (#76) is for 1.0. PR #61 and any related changes might be postponed.

mbolivar:
>> Sorry, just to confirm what you mean, then: this patch would be part of mcuboot 1.0, and if it becomes API later, that's post-1.0, right?
> What I meant was that this PR here (#76) is for 1.0. PR #61 and any related changes might be postponed.

OK, so mcuboot 1.0 users will all just assume this format. Then we need a major version bump if we're going to change it later. And I guess that means the absence of the magic values which signal mcuboot is passing a parameter to the child means we can assume 1.0 for now, and tell it apart from 2.0 later, right? Just trying to understand how to maintain compatibility across major releases, including 1.0.

d3zd3z:
I thinking trying to have an API into the bootloader that does flash operations is going to be difficult. Some of the flash drivers may want interrupts or clocks, and it would be difficult to manage in the context of a new OS running. We could come up with something that returned information about it.

But, my main goal here was to define this is such a way that it never needs to change. To mark an upgrade, you write 16 magic bytes at the end of a partition. The image_ok is defined here, but everything before that in the trailer could change, and OTA shouldn't need to know about it.

mbolivar:
> But, my main goal here was to define this is such a way that it never needs to change.

I too hope that we never have to change this :). Just trying to make sure there's a path if we ever have to -- which, with a major version change, it seems like we would.

Not sure what you mean about the bootloader doing flash operations, but yes, I had in mind having a query about the layout, not a call into a flash driver (for the reasons you describe).

d3zd3z:
After a discussion with [redacted] today, I think the things we'll need to communicate are:

* Partitions, at least the partitions the bootloader cares about (the slots and scratch).
* Version information.
* Separate version information about the trailer, so know how it should be marked.
* How mcuboot is configured (swap, upgrade only, run slot)
* Which slot was run (needed for OTA with dual slot solution).

Things I'm less sure about:

* Other partitions
* sector map and write alignment (the drive really should know these)
* ???

utzig:
So, what's the verdict? Do we merge this PR and track updates at #61 or is anything else to do here?

mbolivar:
> So, what's the verdict? Do we merge this PR and track updates at #61 or is anything else to do here?

So you mean you want to create the header with exported declarations in #61?

utzig:
> So you mean you want to create the header with exported declarations in [#61](https://github.com/runtimeco/mcuboot/pull/61 "Tracer bullet for bootloader API")?

I'm feeling a little lost here! What does the exported declarations have to do with the on disk trailer format? Also not sure why @d3zd3z added the same comment here and on #61 about the discussion with [redacted], but those suggestions all seem to be related only to what the API will export, aka #61.

d3zd3z:
Yeah, we don't need to worry about the API in this change.

I think it is fine to merge this, at least with the imgtool change to go with it. This is separate from the API, which isn't even fully defined.

mbolivar:
>> So you mean you want to create the header with exported declarations in #61?

> I'm feeling a little lost here! What does the exported declarations have to do with the on disk trailer format? 

My original comment related to the addition of struct image_trailer to boot/bootutil/include/bootutil/bootutil.h.

That part of the patch has a comment which reads "This is not actually used by mcuboot's code but can be used by apps".

I originally said that, if there wasn't going to be an OS/bootloader ioctl() style API for it (and you've since made it clear that having that API is post-1.0, and this patch series is for 1.0), then it seemed like a good idea to keep code that has users outside of mcuboot itself in a separate place than bootutil.h, because mixing up "internals" with exported definitions seems like asking for trouble. It will make the compiler flags for other applications pretty ugly at the least, and risks leaking internal definitions that others will come to rely on, etc.

You responded:

> Creating an include with only "exported" stuff seems like a good idea.

So my turn to feel a little lost, since that comment and that struct are still in bootutil.h. Will that be moved to another place, are you going to keep it in bootutil.h?

utzig:
> So my turn to feel a little lost, since that comment and that struct are still in bootutil.h. Will that be moved to another place, are you going to keep it in bootutil.h?

Ah, that clarifies everything! I think we can add it now and I'll update the ticket for PR #61 mentioning that a new "exported functions" header needs to be created and the trailer struct moved there. Would that be OK?

mbolivar:
> Ah, that clarifies everything! I think we can add it now and I'll update the ticket for PR #61 mentioning that a new "exported functions" header needs to be created and the trailer struct moved there. Would that be OK?

Just to be clear, I'm not against merging this, just saying that I think that struct isn't so useful for external applications until it's in its own header. If you're going to merge as-is then I think that the comment which says it's not for mcuboot is a little strange, but that's not a big deal. If you're going to update this to make a separate header for that struct definition then awesome.

PR #77, opened 2017-06-30
Always validate slot 0 when requested

d3zd3z:
The MCUBOOT_VALIDATE_SLOT0 feature only verifies the signature when
there is no swapping happening.  The assumption was that if there is a
swap being done, the code will verify the signature of slot 1 before
doing the slot.

However, either due to bugs, or intentional trickery, it may be possible
to confuse the code into continuing a swap operation.  If the data is
modified before this, the bootloader can be tricked into booting the
resulting image in slot 0 without having verified the signature.

Fix this by always verifying slot 0's signature before booting it.

JIRA: MCUB-64

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1249,14 +1249,6 @@ boot_go(struct boot_rsp *rsp)
 
     switch (swap_type) {
     case BOOT_SWAP_TYPE_NONE:
-#ifdef MCUBOOT_VALIDATE_SLOT0
-        rc = boot_validate_slot(0);
-        assert(rc == 0);
-        if (rc != 0) {
-            rc = BOOT_EBADIMAGE;
-            goto out;
-        }
-#endif
         slot = 0;
         break;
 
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/loader.c
```diff
@@ -1286,6 +1278,15 @@ boot_go(struct boot_rsp *rsp)
         break;
     }
 
+#ifdef MCUBOOT_VALIDATE_SLOT0
+    rc = boot_validate_slot(0);
```
mbolivar: I'm not sure about this. Doesn't this need to re-read the image headers? boot_validate_slot() just uses the cached headers in boot_data, which were read before any swapping was done.

d3zd3z:
Don't merge yet, until we can create some enhancements to the sim that will allow us to test this.

utzig:
@d3zd3z I'm doing some "final" changes for #81 which affect some of the same lines as this patch. Not sure there's something remaining to be done, otherwise would be glad if it is merged soon!

PR #79, opened 2017-07-06
Simulator refactorings

utzig:
Refactor of main test code
- Some helper structs added to track state
- Add new trailer checking routine
- Add extra test for upgrade+revert with failure
- Misc improvements

### DIFF
--- a/sim/src/main.rs
+++ b/sim/src/main.rs
@@ -110,6 +110,7 @@ fn main() {
 
         let align = args.flag_align.map(|x| x.0).unwrap_or(1);
 
+
         let device = match args.flag_device {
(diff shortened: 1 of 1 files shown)


utzig:
@d3zd3z I still want to apply some changes to the no revert test but might as well open a smaller PR over this one.

d3zd3z:
Overall it looks good. I think it might make sense to make a bunch of these functions methods on the `Image` struct, but that's an easy change to make later.

d3zd3z:
I'm not sure I like that it now doesn't print anything, even with logging on. It'd be nice to at least have an indication that some tests have run.

PR #80, opened 2017-07-06
Simulator refactorings

d3zd3z:
Some improvements to the simulator, including:

- Move the flash simulator to its own crate (to prevent a circular dependency)
- Move the C binding to mcuboot and the building of mcuboot into a new crate (mcuboot-sys).
- Run the image verification code as part of the simulation.
- Allow RSA to be checked (--features sig-rsa).

This does include mbedtls as a submodule. Travis already does the clone with --recursive. Otherwise, it may be necessary to run `git submodule update --init` the first time to bring this in, and possibly `git submodule update` in the future if more are brought in.


sim/mcuboot-sys/build.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::io;
+use std::path::Path;
+
+fn main() {
+    // Feature flags.
+    let sig_rsa = env::var("CARGO_FEATURE_SIG_RSA").is_ok();
+    let sig_ecdsa = env::var("CARGO_FEATURE_SIG_ECDSA").is_ok();
+    let overwrite_only = env::var("CARGO_FEATURE_OVERWRITE_ONLY").is_ok();
+
+    // TODO: Force clang if we are requestion fuzzing.
```
utzig: There's a typo here! You mean to use libfuzzer, right?

sim/mcuboot-sys/build.rs
```diff
@@ -0,0 +1,94 @@
...
+    conf.file("../../boot/bootutil/src/image_validate.c");
+    conf.file("../../boot/bootutil/src/image_rsa.c");
+    conf.file("../../boot/bootutil/src/loader.c");
+    conf.file("../../boot/bootutil/src/caps.c");
+    conf.file("../../boot/bootutil/src/bootutil_misc.c");
+    conf.file("csupport/run.c");
+    conf.include("../../boot/bootutil/include");
+    conf.include("../../boot/zephyr/include");
+    conf.debug(true);
+    conf.flag("-Wall");
```
utzig: Maybe `-Werror` too?

sim/mcuboot-sys/build.rs
d3zd3z: Other than the 'requestion' word. But, I don't really need the comment. When I use libfuzzer it should be obvious that it needs to build with clang, since the fuzzing arguments are clang only.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,28 +1,53 @@
...
+    }
+}
+
 // This isn't meant to call directly, but by a wrapper.
 
 #[no_mangle]
-pub extern fn sim_flash_erase(dev: *mut Flash, offset: u32, size: u32) -> libc::c_int {
-    let mut dev: &mut Flash = unsafe { &mut *dev };
+pub extern fn sim_flash_erase(offset: u32, size: u32) -> libc::c_int {
+    let dev = unsafe { get_flash!() };
```
utzig: Shouldn't those `sim_flash_` functions check for the possibility of receiving `-19` from `get_flash!()` ?

d3zd3z: The -19 will get returned back to C, which gets returned as an error code. It could be checked in hal_flash_, or just returned as an error, which will cause failure. If things do fail.

It might be worth check for an error (any error) in the hal_flash wrappers to detect problems. I don't know if we ever expect to get a flash operation error.

PR #81, opened 2017-07-10
Rework some of the handling of swap errors

utzig:
This improves behavior of swap handling.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -48,97 +48,45 @@ const uint32_t BOOT_MAX_ALIGN = MAX_FLASH_ALIGN;
 
 struct boot_swap_table {
     /** * For each field, a value of 0 means "any". */
-    uint8_t bsw_magic_slot0;
-    uint8_t bsw_magic_slot1;
-    uint8_t bsw_image_ok_slot0;
-    uint8_t bsw_image_ok_slot1;
+    uint8_t magic_slot0;
+    uint8_t magic_slot1;
+    uint8_t image_ok_slot0;
+    uint8_t image_ok_slot1;
 
-    uint8_t bsw_swap_type;
+    uint8_t swap_type;
 };
 
 /**
  * This set of tables maps image trailer contents to swap operation type.
  * When searching for a match, these tables must be iterated sequentially.
+ *
+ * NOTE: the table order is very important. The settings in Slot 1 always
+ * are priority to Slot 0 and should be located earlier in the table.
+ *
+ * The table lists only states where there is action needs to be taken by
+ * the bootloader, as in starting/finishing a swap operation.
  */
 static const struct boot_swap_table boot_swap_tables[] = {
     {
-        /*          | slot-0     | slot-1     |
-         *----------+------------+------------|
-         *    magic | Unset      | Unset      |
-         * image-ok | Any        | Any        |
-         * ---------+------------+------------'
-         * swap: none                         |
-         * -----------------------------------'
-         */
-        .bsw_magic_slot0 =      BOOT_MAGIC_UNSET,
-        .bsw_magic_slot1 =      BOOT_MAGIC_UNSET,
-        .bsw_image_ok_slot0 =   0,
-        .bsw_image_ok_slot1 =   0,
-        .bsw_swap_type =        BOOT_SWAP_TYPE_NONE,
+        .magic_slot0 =      0,
+        .magic_slot1 =      BOOT_MAGIC_GOOD,
+        .image_ok_slot0 =   0,
+        .image_ok_slot1 =   0xff,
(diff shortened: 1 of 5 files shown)


d3zd3z:
The code looks good. However, I tried the "Test imgtool generated FW written to slot0" by itself, and it doesn't provoke the failure, suggesting it isn't exercising the broken code path.

utzig:
@d3zd3z My bad, I left the `image_ok` bit set which didn't trigger the fail. Now it's failing and the fix is being committed just now (PS: I squashed the sim commits for history to look less awful)

utzig:
Ok, now correctly fixed. Initial comment still applies, I still want to audit the `finalize` functions before considering the PR ready for merge.

sim/simflash/src/lib.rs
```diff
@@ -39,6 +39,7 @@ fn ebounds<T: AsRef<str>>(message: T) -> ErrorKind {
     ErrorKind::OutOfBounds(message.as_ref().to_owned())
 }
 
+#[allow(dead_code)]
```
d3zd3z: We could also just remove the function, although it would then be work to figure out how to add it back in if we did want to return errors.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -48,97 +48,45 @@ const uint32_t BOOT_MAX_ALIGN = MAX_FLASH_ALIGN;
...
+ *
+ * NOTE: the table order is very important. The settings in Slot 1 always
+ * are priority to Slot 0 and should be located earlier in the table.
+ *
+ * The table lists only states where there is action needs to be taken by
+ * the bootloader, as in starting/finishing a swap operation.
  */
 static const struct boot_swap_table boot_swap_tables[] = {
     {
-        /*          | slot-0     | slot-1     |
```
mbolivar: This deletion needs a doc/design.txt update too.

utzig: I was just updating it!

boot/bootutil/src/loader.c
```diff
@@ -1257,6 +1229,14 @@ boot_go(struct boot_rsp *rsp)
         if (rc != 0) {
             goto out;
         }
+
+        /*
+         * The following states need image_ok be explicitly set after the
+         * swap was finished to avoid a new revert.
+         */
+        if (swap_type == BOOT_SWAP_TYPE_REVERT || swap_type == BOOT_SWAP_TYPE_FAIL) {
+            boot_set_image_ok();
```
mbolivar: Why isn't it necessary to check for errors here?

boot/bootutil/src/loader.c
```diff
@@ -1268,14 +1248,9 @@ boot_go(struct boot_rsp *rsp)
...
     case BOOT_SWAP_TYPE_PERM:
-        slot = 1;
-        boot_finalize_test_swap();
-        reload_headers = true;
-        break;
-
     case BOOT_SWAP_TYPE_REVERT:
         slot = 1;
-        boot_finalize_revert_swap();
+        boot_set_copy_done();
```
mbolivar: Or here?

utzig:
@mbolivar I added checking for those failures and now mcuboot panics if that happens. Can't decide what to do in those cases so leaving a FIXME. But the only chance of this happening would be a flash error and some way of informing the user of fatal failures needs to be added.

boot/bootutil/src/loader.c
```diff
@@ -1261,11 +1268,19 @@ boot_go(struct boot_rsp *rsp)
...
         slot = 1;
-        boot_set_copy_done();
+        rc = boot_set_copy_done();
+        if (rc != 0) {
+            /* FIXME: this flag is basically used to trigger a revert and
+             * in this specific case a failure would not be fatal and the
+             * image in slot0 could be safely booted, but seems better to
+             * fail now while a better handling is properly thought about.
+             */
+            assert(0);
```
mbolivar: Should this be swap_type = BOOT_SWAP_TYPE_PANIC so it happens in production as well as in testing?

boot/bootutil/src/loader.c
```diff
@@ -1280,8 +1295,8 @@ boot_go(struct boot_rsp *rsp)
 
     case BOOT_SWAP_TYPE_PANIC:
     default:
-        /* TODO: what to do it a fatal error like flash read/write error
-         *       happened?
+        /* FIXME: what to do it a fatal error like flash read/write error
```
mbolivar: How about logging an error and spinning forever? It seems better than jumping into potentially compromised code.

utzig:
I agree, it's the best thing to do. Btw, not every doc update in my PR is for things that I did and I think it's somewhat complicated to keep track of what needs updating for non-obvious design changes, but for bigger things like the new image format we surely should update the docs together with the new feature itself.

PR #83, opened 2017-07-13
This add support for mynewt resigned images

utzig:
`newt` accepts a parameter for the keyid which can be any u8
value. This means that for mynewt keyid is not an index and
the only way to validate the image is to try every key and
fail only if none validates.

Implements MCUB-65.

### DIFF
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -108,6 +108,9 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
     uint8_t buf[256];
     uint8_t hash[32];
     int rc;
+#ifdef APP_mynewt
+    int i;
+#endif
 
 #ifdef MCUBOOT_SIGN_RSA
 #ifdef MCUBOOT_RSA_PKCS1_15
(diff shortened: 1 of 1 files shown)


utzig:
NOTE: conditional compile was used because that change is probably not wanted in other OSs.

boot/bootutil/src/image_validate.c
```diff
@@ -216,10 +219,19 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
         return -1;
     }
 
+#ifndef APP_mynewt
```
mbolivar: Could this be abstracted out into a helper routine which is then called unconditionally? There is starting to be a fair bit of nesting of conditional compilation here, which is getting a bit hard to read and thus audit. (I also wonder if a third RTOS might need its own solution.)

Something like:

```
#if defined(MCUBOOT_SIGN_RSA) || defined(MCUBOOT_SIGN_EC) || defined(MCUBOOT_SIGN_EC256)
#ifndef APP_mynewt
int helper_verify_sig(struct image_header *hdr, uint8_t buf[256], uint8_t hash[32], uint32_t sig_len)
{
     if (hdr->ih_key_id >= bootutil_key_cnt) {
          return -1;
      }
      return bootutil_verify_sig(hash, 32, buf, sig_len, hdr->ih_key_id);
}
#else
int helper_verify_sig(struct image_header *hdr, uint8_t buf[256], uint8_t hash[32], uint32_t sig_len)
{
    int i, rc;
    for (i = 0; i < bootutil_key_cnt; i++) {
        rc = bootutil_verify_sig(hash, sizeof(hash), buf, sig_len, i);
        if (!rc) {
            return rc;
        }
    }
    return 0;
}
#endif
#endif
```

Then call helper_verify_sig() from here.

d3zd3z:
:-1:

utzig:
This will be completely revamped for 1.0 including a new image format. Closing this one.

PR #87, opened 2017-07-20
Update docs

utzig:


doc/design.txt
```diff
@@ -210,15 +210,15 @@ An image trailer has the following structure:
...
      0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    ~                       MAGIC (16 octets)                       ~
-    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     ~                                                               ~
     ~             Swap status (128 * min-write-size * 3)            ~
     ~                                                               ~
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    |   Copy done   |     0xff padding (up to min-write-sz - 1)     ~
+    |           0xff padding (7 octets)             |   Copy done   |
```
mbolivar: Shouldn't this be:

```
    |   Copy done   |           0xff padding (7 octets)             |
```
Similarly for image OK.

doc/design.txt
```diff
@@ -24,7 +24,7 @@
 The Mynewt bootloader comprises two packages:
```
mbolivar: If you're updating this to note that there are multiple ports, how about saying "mcuboot" instead of "The Mynewt bootloader"?

README.md
```diff
@@ -10,16 +10,13 @@ easy software upgrade.
 MCUboot is operating system and hardware independent, and relies on 
 hardware porting layers from the operating system it works with.  Currently
 mcuboot works with both the Apache Mynewt, and Zephyr operating systems, but
-more ports are planned in the future.
+more ports are planned in the future. RIOT is currently supported as a boot
+target with complete port arriving in the near future.
```
mbolivar: "a complete port planned", maybe?

doc/design.txt
```diff
@@ -210,15 +210,15 @@ An image trailer has the following structure:
      0                   1                   2                   3
```
mbolivar: There is nothing wrong with this line, but the GitHub UI won't let me comment on parts of the file that aren't in a diff hunk.

It looks like the "BOOT STATES" section above also needs to be reworked in light of the recent changes as well.

doc/design.txt
```diff
@@ -272,7 +272,7 @@ purposes; they are not actually present in the boot vector.
     -----------------+--------+--------'
      swap: none                        |
     -----------------------------------'
-
```
mbolivar: Sorry, I'm confused about this (and actually always have been).

This section says there are 5 states. The "BOOT STATES" section above says there are only 4. The code in boot_swap_tables now actually only has 3 entries.

What's going on here?

utzig: There are 3 states that need special handling: `test`, `permanent` and `revert` and those are in the tables. There is one additional state which does nothing:  `none`. It just ends up with the image in slot 0 booting. This last one never gets to the swapping routines which employ the swap table you mentioned.

doc/design.txt
utzig: I don't think it does. I know what you mean, but there are two ways of explaining things in the design doc: one more "abstract", which uses words like `pending` and `confirmed` and the practical, real implementation, which uses `magic`, `image_ok` and `copy_done`.

mbolivar: If pending is one of "unset", "temporary", "permanent", and confirmed is one of "set", "unset", then with two slots there are 36 possible initial abstract "BOOT STATES", but only 4 are documented above.

Does that mean that the above text "the four possible states that the device can be in" really means something along the lines of "the four permitted states that mcuboot recognizes as valid boot states"?

Will the other 32 states reliably trigger a "panic" swap type in the new update code?

utzig: You're doing permutation, while it would be more correct to consider it like a tree of states, `if magic is unset do nothing`, `if magic is set check the other flag` and so on and so forth.

doc/design.txt
mbolivar: Something is still fishy here. In the paragraph just above this, it says "It is better to map all the possible vector states to the three states described above via a set of tables".

However:

- As discussed above, this is not "all possible" vector states
- 3 != 4

Further, "State I" and "State V" here both lead to the "none" swap type, which is two "states", not "one additional state" as you wrote in the above comment.

I still think that this document is not consistent with itself as a result; what am I missing?

doc/design.txt
utzig: > Will the other 32 states reliably trigger a "panic" swap type in the new update code?

Nope, a weird state should either be `NONE` or `FAIL` which end up with somewhat similar behavior.

doc/design.txt
```diff
@@ -298,20 +298,24 @@ purposes; they are not actually present in the boot vector.
...
     -----------------+--------+--------'
              pending |        |        |
           confirmed  |        |   X    |
     -----------------+--------+--------'
      swap: revert (test image running) |
     -----------------------------------'
 
+Note: copy-done is only used to determine if the current image was the result
+of a swap operation or if it was written by the provided imgtool, in which case
+no revert is performed. All other states don't need it.
```
mbolivar: How about using "Any" lines to show that instead?

utzig: Done!

doc/design.txt
utzig: > 3 != 4 (and there are four states listed in the BOOT STATES section)

I agree with this and I think the sentence should be ""It is better to map all the possible vector states to the *four* states described...".

doc/design.txt
mbolivar: > You're doing permutation, while it would be more correct to consider it like a tree of states, if magic is unset do nothing, if magic is set check the other flag and so on and so forth.

I disagree with this assessment.

I'm enumerating all of the states and asking what the boot outcome would be in each one, which is a perfectly correct thing to do when considering "possible" states.

A decision tree representation is equivalent to a complete enumeration, which can be more compactly represented if many outcomes are equivalent due to some components being "don't care", such as your example of "if magic is unset do nothing".

mbolivar: >> Will the other 32 states reliably trigger a "panic" swap type in the new update code?

> Nope, a weird state should either be NONE or FAIL which end up with a somewhat similar behavior.

Statement like this worry me.

If this is the design document for how the bootloader is supposed to make decisions, and we're still saying that states not listed here (despite the document claiming this is all "possible" states) "should" result in an outcome which is "either [...] NONE or FAIL", which is "somewhat similar behavior", it means that the design has not been completely thought through.

doc/design.txt
utzig: > Further, "State I" and "State V" here both lead to the "none" swap type, which is two "states", not "one additional state" as you wrote in the above comment.

Right, "State I" and "State V" are really **two** states in the flash. But both of them end up being `NONE` in the code.

README.md
```diff
@@ -34,11 +31,15 @@ For more information in the source, here are some pointers:
...
-## Joining 
+## Joining
 
-Developers welcome!  To join in the discussion, please join the developer mailing list: 
+Developers welcome!
 
-http://lists.runtime.co/mailman/listinfo/dev-mcuboot_lists.runtime.co
+* Our developer mailing list:
+  http://lists.runtime.co/mailman/listinfo/dev-mcuboot_lists.runtime.co
+* Our Slack channel: https://runtimeco.slack.com/
```
d3zd3z: I believe Slack requires an invitation to join. Not sure if it would be obvious to people that they need to send email and ask for an invite.

doc/design.txt
```diff
@@ -231,27 +231,27 @@ then min-write-size is 2, and so on.
...
 
-3. Copy done: A single byte indicating whether the image in this slot is
+2. Copy done: A single byte indicating whether the image in this slot is
 complete (0x01=done; 0xff=not done).
 
-4. Image OK: A single byte indicating whether the image in this slot has been
+3. Image OK: A single byte indicating whether the image in this slot has been
 confirmed as good by the user (0x01=confirmed; 0xff=not confirmed).
 
+4. MAGIC: The following 16 bytes, written in host-byte-order:
```
d3zd3z: `host-byte-order`? Shouldn't it be written in target byte order, since that what will be checking it?

release-notes.txt
```diff
@@ -0,0 +1,19 @@
+mcuboot 0.9 - Release Notes
```
d3zd3z: Do we want to put this document in either markdown or RST. This formatting looks like RST, and naming it that would make it format nicely when viewed in github.

doc/design.txt
utzig: I think I agree with you. I'll give it a try on mapping this in a less loose way...

mbolivar: > I'm enumerating all of the states and asking what the boot outcome would be in each one, which is a perfectly correct thing to do when considering "possible" states.
> 
> A decision tree representation is equivalent to a complete enumeration

Further, the document above is not described as a decision tree, but rather an enumeration of states. Perhaps if that's what the code is really doing, then reworking the "BOOT STATES" section so it's described a decision tree is what is needed? Then we can compare that to how the concrete "vector states" (with their "Any" values) map to that tree, to convince ourselves things work the way we mean them to.

doc/design.txt
mbolivar: > Right, "State I" and "State V" are really two states in the flash. But both of them end up being NONE in the code.

Well, they're two "vector states", which both lead to the same "swap state" or "swap result" ("vector state" being an input to the bootloader, "swap state" being an output or at least an intermediate output).

You bring up a good point in the earlier sections discussion that this isn't really a listing of states so much as it is a sequential listing of what the bootloader is checking. Perhaps that is a better way to write this section too?

doc/design.txt
utzig: Ah, I haven't seen that but it confuses me anyway. `host-byte-order`, `target-byte-order`, `network-byte-order`, etc all seem bad. I think it should be either `little-endian` or `big-endian`.

utzig:
Design doc will be updated before official release. Tracked by issue: https://runtimeco.atlassian.net/browse/MCUB-67

PR #91, opened 2017-07-21
Fix newtmgr

utzig:
This fixes `newtmgr` when building `mynewt` images to be run on `mcuboot`. README file also added.


boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h
```diff
@@ -16,10 +16,10 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-#ifndef __BOOT_CONFIG_H__
-#define __BOOT_CONFIG_H__
+#ifndef MCUBOOT_CONFIG_H
```
d3zd3z: I can see needing to make the name more unique, but why remove the underscores?

PR #93, opened 2017-07-26
Fix mcuboot version in Mynewt repository

utzig:


d3zd3z:
Instead, going to do this as a single commit before the tag, that will set this version, as well as one in the README. This way the tag will land on a commit.

PR #95, opened 2017-07-26
Minor zephyr fixes in since rc1

d3zd3z:
Two minor fixes. One to be able to flash the full.bin target. The other needed in recent Zephyr trees to allow writes to the flash device.

### DIFF
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -12,6 +12,7 @@ CONFIG_MBEDTLS_CFG_FILE="config-boot.h"
 CONFIG_HEAP_MEM_POOL_SIZE=16384
 
 CONFIG_FLASH=y
+CONFIG_MPU_ALLOW_FLASH_WRITE=y
 
 ### Disable Bluetooth by default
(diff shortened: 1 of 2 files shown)


boot/zephyr/prj.conf
```diff
@@ -12,6 +12,7 @@ CONFIG_MBEDTLS_CFG_FILE="config-boot.h"
 CONFIG_HEAP_MEM_POOL_SIZE=16384
 
 CONFIG_FLASH=y
+CONFIG_MPU_ALLOW_FLASH_WRITE=y
```
mbolivar: This is option hasn't been merged into mainline Zephyr yet. Is this meant as an RFC?

d3zd3z: It doesn't hurt anything to put the option here, so we're not tied to the timing. As long as the option name doesn't change in Zephyr, we should be good. It gives us a little more chance of not needing additional changes.

mbolivar: A bit dirty :), but that makes sense.

Hopefully your patch doesn't get bikeshedded too much longer.

PR #104, opened 2017-08-04
Design doc fixups

mbolivar:
This continues the work in MCUB-67. The substantial change is deleting the 'pending' / 'confirmed' state space in favor of one defined in terms of 'swap types'.

### DIFF
--- a/doc/design.txt
+++ b/doc/design.txt
@@ -110,12 +110,15 @@ region of disk with the following properties:
     (1) An area can be fully erased without affecting any other areas.
     (2) A write to one area does not restrict writes to other areas.
 
-The boot loader uses the following flash areas:
+The boot loader uses the following flash area IDs:
 
 #define FLASH_AREA_BOOTLOADER                    0
 #define FLASH_AREA_IMAGE_0                       1
 #define FLASH_AREA_IMAGE_1                       2
 #define FLASH_AREA_IMAGE_SCRATCH                 3
+
+The bootloader area contains the bootloader image itself. The other areas are
+described in subsequent sections.
 
 *** IMAGE SLOTS
 
--- a/doc/design.txt
+++ b/doc/design.txt
@@ -141,78 +141,58 @@ even when it is reset during the middle of an image swap. For this reason, the
 rest of the document describes its behavior when configured to swap images
 during an upgrade.
 
-*** BOOT STATES
+*** BOOT SWAP TYPES
 
-Logically, you can think of a pair of values associated with each image slot:
-pending and confirmed.  On startup, the boot loader determines the state of the
-device by inspecting each pair of values.  These values have the following
-meanings:
+When the device first boots under normal circumstances, there is an up-to-date
+firmware image in slot 0, which mcuboot can validate and then chain-load. In
+this case, no image swaps are necessary. During device upgrades, however, new
+candidate images are present in slot 1, which mcuboot must swap into slot 0
+before booting as discussed above.
 
-* pending: Indicates whether the image should be used on the next reboot; can
-  hold one of three values:
-    " " (unset):     Don't use image on next boot
-    "T" (temporary): Use image on next boot; absent subsequent confirm command,
-                     revert to original image on second reboot.
-    "P" (permanent): Use image on next boot and all subsequent boots
+Upgrading an old image with a new one by swapping can be a two-step process. In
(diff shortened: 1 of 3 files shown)


mbolivar:
I fixed some issues with the commit log in the last patch in v2. No changes to the patches themselves.

doc/design.txt
```diff
@@ -222,21 +218,31 @@ An image trailer has the following structure:
     |                       MAGIC (16 octets)                       |
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 
-These records are at the end of each image slot.  The offset immediately
-following such a record represents the start of the next flash area.
+The offset immediately following such a record represents the start of the next
```
d3zd3z: Hmm. This sentence is still somewhat awkward, and I think you've removed the part that is easier to understand. The current wording sounds like a good way to describe it for a standard document, but not really a design document. How about something like:

"The image trailer is placed such that it butts up against the end of the partition. For example, if the following slot begins at offset 'n', the MAGIC field will be at offset 'n-16'."

doc/design.txt
```diff
@@ -255,18 +261,15 @@ confirmed as good by the user (0x01=confirmed; 0xff=not confirmed).
 
 *** IMAGE TRAILERS
 
-At startup, the boot loader determines which of the above four states the
-device is in by inspecting the image trailers.  When using the term "image
-trailers" what is meant is the aggregate information provided by both image
-slot's trailers.
+At startup, the boot loader determines the boot swap type by inspecting the
+image trailers.  When using the term "image trailers" what is meant is the
```
d3zd3z: I think this sentence really just indicates that the following description was awkward. If we take it out, do things still make sense? If not, I think we should fix the descriptions below, rather than try to define what plural means.

doc/design.txt
mbolivar: Good idea. The next thing might not be a 'slot', so I'll use 'flash area'.

doc/design.txt
mbolivar: I've tried to improve this paragraph.

mbolivar:
v3 posted accounting for review comments from @d3zd3z. Diff from v2:

```
diff --git a/doc/design.txt b/doc/design.txt
index 8894304..27facc4 100644
--- a/doc/design.txt
+++ b/doc/design.txt
@@ -218,8 +218,9 @@ image trailer. An image trailer has the following structure:
     |                       MAGIC (16 octets)                       |
 
-The offset immediately following such a record represents the start of the next
-flash area.
+Each image trailer is placed at the very end of its flash area. For example, if
+the following flash area begins at offset 'n', the MAGIC field of the image
+trailer starts at offset 'n-16', the 'Image OK' byte is at offset 'n-24', etc.
 
 Note: "min-write-size" is a property of the flash hardware.  If the hardware
 allows individual bytes to be written at arbitrary addresses, then
@@ -261,9 +262,10 @@ confirmed as good by the user (0x01=confirmed; 0xff=not confirmed).
 
 *** IMAGE TRAILERS
 
-At startup, the boot loader determines the boot swap type by inspecting the
-image trailers.  When using the term "image trailers" what is meant is the
-aggregate information provided by both image slot's trailers.
+At startup, mcuboot determines the boot swap type by inspecting the image
+trailers. If a previous boot was interrupted, some of the contents of the image
+trailers may be in the swap flash area as well, as described in more detail
+below in IMAGE SWAPPING and RESET RECOVERY.
 
 The image trailers records are structured around the limitations imposed by flash
 hardware.  As a consequence, they do not have a very intuitive design, and it
```

doc/design.txt
```diff
@@ -280,10 +283,7 @@ higher priority when testing the image trailers.
             image-ok | Any    | Unset  |
            copy-done | Any    | Any    |
     -----------------+--------+--------'
-             pending |        |   T    |
-          confirmed  |   X    |        |
-    -----------------+--------+--------'
-     swap: test                        |
+     result: BOOT_SWAP_TYPE_TEST       |
```
utzig: I changed the same thing on my last doc updates, but decided to rollback before finishing the commit! :-)

mbolivar:
@utzig are you saying I should squash all of the design doc changes into one commit? I made the commits separate logical changes on purpose so they'd be easier to review.

PR #108, opened 2017-08-22
New image format

mkiiskila:
Removed tlv_size from image header.
Public key used in signing indicated by a separate TLV before signature itself.
TLV area started by it's own header, which contains info about the length of the TLV area.


boot/bootutil/src/image_validate.c
```diff
@@ -87,6 +87,29 @@ bootutil_img_hash(struct image_header *hdr, const struct flash_area *fap,
...
+bootutil_find_key(uint8_t *keyhash, int8_t keyhash_len)
+{
+    bootutil_sha256_context sha256_ctx;
+    int i;
+    const struct bootutil_key *key;
+    uint8_t hash[32];
+
+    for (i = 0; i < bootutil_key_cnt; i++) {
+        key = &bootutil_keys[i];
+        bootutil_sha256_init(&sha256_ctx);
```
d3zd3z: I wonder if we should come up with a way of computing these hashes at build time to compare that way?

boot/bootutil/include/bootutil/image.h
```diff
@@ -28,34 +28,38 @@ extern "C" {
...
  */
-#define IMAGE_TLV_SHA256            1	/* SHA256 of image hdr and body */
-#define IMAGE_TLV_RSA2048           2	/* RSA2048 of hash output */
-#define IMAGE_TLV_ECDSA224          3   /* ECDSA of hash output */
-#define IMAGE_TLV_ECDSA256          4   /* ECDSA of hash output */
+#define IMAGE_TLV_KEYHASH         0x01  /* hash of the public key */
+
+#define IMAGE_TLV_SHA256          0x10	/* SHA256 of image hdr and body */
+
+#define IMAGE_TLV_RSA2048         0x20	/* RSA2048 of hash output */
```
d3zd3z: I think we should make the TLV's fully describe the signature algorithm. RSA2048 is ambiguous, since we support both PKCS#1 v1.5, an PSS signatures. Distinguishing these at this point would: 1. give more meaningful errors if the signature mismatched how the code was built, and 2. allow both to be added to an image to match either configuration of the bootloader.

2. Is not really supposed to be done, but knowing the type of signature could also help external tools.

Perhaps IMAGE_TLV_RSA2048-PKCS1V15 and IMAGE_TLV_RSA2048_PSS, or something like that.

mkiiskila: As per discussion, I was going to remove the pcks1v15 version and just keep the PSS.
I'll include PSS in the define name.

boot/bootutil/src/image_validate.c
mkiiskila: I did wonder about that myself earlier, but ended up computing them at runtime here. Good thing about current code is that there is no need to change existing files containing keys, and no need to have special tooling around the key file generation (besides openssl).
Downside is the extra code space taken by having these calls, of course.

mkiiskila: Extra code space taken by these calls is 36 bytes on Atmel SAMD21 (cortex-m0). Maybe this is not worth worrying about?

d3zd3z: I was more concerned about the time, but SHA256 is pretty fast, even on these CPUs.

boot/bootutil/src/image_validate.c
```diff
@@ -205,7 +205,7 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
             }
         }
 #ifdef MCUBOOT_SIGN_RSA
-        if (tlv.it_type == IMAGE_TLV_RSA2048) {
+        if (tlv.it_type == IMAGE_TLV_RSA2048_PSS) {
```
d3zd3z: This should probably be squashed into the previous commit,

boot/bootutil/include/bootutil/image.h
```diff
@@ -64,28 +68,28 @@ struct image_version {
...
+/** Image TLV info. All fields in little endian. */
+struct image_tlv_info {
+    uint16_t it_magic;
+    uint16_t it_tlv_tot;  /* size of TLV area (including tlv_info header) */
 };
 
 /** Image trailer TLV format. All fields in little endian. */
 struct image_tlv {
-    uint8_t  it_type;   /* IMAGE_TLV_[...]. */
+    uint8_t  it_type;     /* IMAGE_TLV_[...]. */
```
d3zd3z: Since we're already changing things, why don't we make the type field 16-bits instead of 8. It seems kind of silly to have padding there, and then have a very small tag field that could easily overflow.

mkiiskila: That we could do.

d3zd3z:
A bit of a cleanup, and move code to new PR #113.

PR #113, opened 2017-09-01
New image format code.

d3zd3z:
This is a bit of a cleanup from #108. With the changes from @utzig it passes all of the simulation tests.


scripts/imgtool.py
```diff
@@ -26,7 +26,10 @@ def do_keygen(args):
 
 def do_getpub(args):
     key = keys.load(args.key)
-    key.emit_c()
+    if args.rust:
```
mbolivar: Instead of adding --rust, the args should gain a --type which defaults to 'c' and has choices=['c','rust'], just in case any more formats are needed later.

boot/bootutil/include/bootutil/image.h
```diff
@@ -37,12 +37,7 @@ struct flash_area;
  * Image header flags.
  */
 #define IMAGE_F_PIC                      0x00000001 /* Not supported. */
-#define IMAGE_F_SHA256                   0x00000002 /* Hash TLV is present */
```
mbolivar: Asking for trouble here. What about making these "reserved, must be zero" instead?

boot/bootutil/include/bootutil/image.h
```diff
@@ -52,6 +47,7 @@ struct flash_area;
 /*
  * Image trailer TLV types.
  */
+#define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
```
mbolivar: Same thing here; re-using old values given that 0.9 was a release feels like asking for trouble. Why does this have to be number one?

sim/src/tlv.rs
```diff
@@ -117,3 +117,41 @@ impl TlvGen {
         result
     }
 }
+
+/* Autogenerated by imgtool.py, do not edit. */
```
mbolivar: Autogenerated code has no place in the same file as non-autogenerated code. This isn't a .emacs file ;)

boot/bootutil/src/image_validate.c
```diff
@@ -88,6 +88,53 @@ bootutil_img_hash(struct image_header *hdr, const struct flash_area *fap,
...
 /*
+ * Currently, we only support being able to verify one type of
+ * signature, because there is a single verification function that we
+ * call.  List the type of TLV we are expecting.  If we aren't
+ * configured for any signature, don't define this macro.
+ */
+#if defined(MCUBOOT_SIGN_RSA)
+#    define EXPECTED_SIG_TLV IMAGE_TLV_RSA2048
+#    define EXPECTED_SIG_LEN(x) ((x) == 256) /* 2048 bits */
+#    if defined(MCUBOOT_SIGN_EC) || defined(MCUBOOT_SIGN_EC256)
```
mbolivar: This is a bit hard to read and might get harder if more signature types are supported.

What about:

```
#if (defined(MCUBOOT_SIGN_RSA) + defined(MCUBOOT_SIGN_EC) + defined(MCUBOOT_SIGN_EC256)) > 1
#error Multiple signature types not yet supported
#endif
```

Then you can just write:

```
#if defined(MCUBOOT_SIGN_RSA)
...
#elif defined(MCUBOOT_SIGN_EC)
... EXPECTED_SIG_TLV, EXPECTED_SIG_LEN for each case go here...
#elif defined(MCUBOOT_SIGN_EC256)
...
#endif
```

boot/bootutil/src/image_validate.c
```diff
@@ -148,81 +169,81 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
...
-            sha_off = off + sizeof(tlv);
-        }
-#ifdef MCUBOOT_SIGN_RSA
-        if (tlv.it_type == IMAGE_TLV_RSA2048) {
-            if (tlv.it_len != 256) { /* 2048 bits */
+            rc = flash_area_read(fap, off + sizeof(tlv), buf, sizeof hash);
+            if (rc) {
+                    return rc;
+            }
+            if (memcmp(hash, buf, sizeof(hash))) {
```
mbolivar: Just curious as I know this memcmp() was here before this patch, but should this be using a constant time comparison?

boot/bootutil/src/image_validate.c
```diff
@@ -148,81 +169,81 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
...
-        if (tlv.it_type == IMAGE_TLV_ECDSA256) {
-            if (tlv.it_len < 72) { /* oids + 2 * 32 bytes */
+            rc = flash_area_read(fap, off + sizeof(tlv), buf, tlv.it_len);
+            if (rc) {
                 return -1;
             }
-            sig_off = off + sizeof(tlv);
-            sig_len = tlv.it_len;
-        }
+            rc = bootutil_verify_sig(hash, sizeof(hash), buf, tlv.it_len, key_id);
```
mbolivar: nitpick: line length is too long after indentation change

boot/bootutil/src/image_validate.c
```diff
@@ -148,81 +169,81 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
...
+                    return -1;
+            }
+
+            sha256_valid = 1;
+#ifdef EXPECTED_SIG_TLV
+        } else if (tlv.it_type == IMAGE_TLV_KEYHASH) {
+            /*
+             * Determine which key we should be checking.
+             */
+            if (tlv.it_len > 32) {
```
mbolivar: I'm not complaining about this patch since the problem exists elsewhere in the file, but it would be good to replace the magic number 32s throughout with SHA256_HASH_SIZE or so.

mbolivar: And actually, shouldn't this be `if (tlv.it_len != 32)`?

boot/bootutil/src/image_validate.c
```diff
@@ -148,81 +169,81 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
...
-        if (tlv.it_type == IMAGE_TLV_ECDSA224) {
-            if (tlv.it_len < 64) { /* oids + 2 * 28 bytes */
+            rc = flash_area_read(fap, off + sizeof tlv, buf, tlv.it_len);
+            if (rc) {
+                return rc;
+            }
+            key_id = bootutil_find_key(buf, tlv.it_len);
+            /*
+             * The key may not be found, which is acceptable.  There
+             * can be multiple signatures, each preceeded by a key.
```
mbolivar: nitpick: "preceded" typo

boot/bootutil/src/image_validate.c
```diff
@@ -148,81 +169,81 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
...
+            rc = flash_area_read(fap, off + sizeof(tlv), buf, sizeof hash);
+            if (rc) {
+                    return rc;
+            }
+            if (memcmp(hash, buf, sizeof(hash))) {
+                    return -1;
+            }
+
+            sha256_valid = 1;
+#ifdef EXPECTED_SIG_TLV
```
mbolivar: I like this change. Moving the various supported signature TLVs out of this function makes it easier to read.

boot/bootutil/include/bootutil/image.h
mbolivar: I see you plan on handling all the renumbering by changing the magic later.

d3zd3z: It still leaves a window when the format doesn't match the magic. Another possibility would be to make the magic a third not-to-be-used value. But, is this overkill? It will only affect someone doing a bisect, since these patches will all need to go in together.

Short of changing the magic to completely different values at each patch, it still isn't ever quite right.

sim/src/tlv.rs
d3zd3z: What do you propose? The intent is to show that this data is generated from the public key and pasted in? It's already inserted just like this in the keys.c file.

boot/bootutil/src/image_validate.c
d3zd3z: No reason to use a constant time comparison, as nothing is revealed. All operations performed in the bootloader are based on public keys, and no information would be gained by a timing attack.

boot/bootutil/src/image_validate.c
d3zd3z: The original implementation allowed the hash to be less than the full size of the hash itself. You could, for example, only provide 8 bytes of the key hash, which would still be plenty to uniquely identify a hash, as long as the space of keys involved is less than 2^32.

utzig:
@d3zd3z Please merge from https://github.com/utzig/mcuboot/tree/swap-size-to-trailer

Those changes save the calculated swap size to the trailer, enabling the calculation to happen only before beginning a swap upgrade.

I only tested on the sim, but will proceed with testing on HW now.

sim/src/tlv.rs
d3zd3z: I've moved the key into its own file, and used `include!("path")`.

boot/bootutil/src/image_validate.c
d3zd3z: I'm not sure we have an official coding standard we are complying with, but it doesn't need to be the Linux convention. I personally would rather see an 82 character line than have to awkwardly split the line.

PR #115, opened 2017-09-04
Add travis' validation of "Signed-off-by:" line

utzig:
This adds validation of "Signed-off-by:" line.

### DIFF
--- a/scripts/check-signed-off-by.sh
+++ b/scripts/check-signed-off-by.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,10 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-MAIN_BRANCH=master
+# this retrieves the merge commit created by GH
+parents=(`git log -n 1 --format=%p HEAD`)
 
-# ignores last commit because travis/gh creates a merge commit
-commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD | tail -n +2)
+if [[ "${#parents[@]}" -ne 2 ]]; then
+  echo "This PR's merge commit is missing a parent!"
+  exit 1
+fi
+
+from="${parents[0]}"
+into="${parents[1]}"
+commits=$(git show -s --format=%h ${from}..${into})
 
 has_commits=false
 for sha in $commits; do
--- a/scripts/check-signed-off-by.sh
+++ b/scripts/check-signed-off-by.sh
@@ -15,7 +15,7 @@
 MAIN_BRANCH=master
 
 # ignores last commit because travis/gh creates a merge commit
-commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD~1)
+commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD | tail -n +2)
 
 has_commits=false
 for sha in $commits; do
--- a/scripts/check-signed-off-by.sh
+++ b/scripts/check-signed-off-by.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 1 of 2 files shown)


scripts/check-signed-off-by.sh
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+MAIN_BRANCH=master
+
+# ignores last commit because travis/gh creates a merge commit
+commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD | tail -n +2)
```
d3zd3z: How about
```
commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD~1)
```
since git can do the parent thing and stuff.

scripts/check-signed-off-by.sh
```diff
@@ -0,0 +1,28 @@
...
+
+for sha in $commits; do
+  actual=$(git show -s --format=%B ${sha} | sed '/^$/d' | tail -n 1)
+  expected="Signed-off-by: $(git show -s --format="%an <%ae>" ${sha})"
```
d3zd3z: I wonder if we should also check the committer (%cn %ce). After github rebases, it'll just be github, but it would be nice to check this for when people apply other's patches, they should also be adding the Signed-off-by line.

scripts/check-signed-off-by.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+MAIN_BRANCH=master
+
+# ignores last commit because travis/gh creates a merge commit
+commits=$(git log --format=%h ${MAIN_BRANCH}..HEAD | tail -n +2)
+
+for sha in $commits; do
+  actual=$(git show -s --format=%B ${sha} | sed '/^$/d' | tail -n 1)
```
d3zd3z: The Sob is not required to be the last thing in the commit. In fact, I think this would've failed on a few of mine, because when pulling someone's patch, I put my Sob after theirs, so the author's was the second to the last, and mine was before that. Also, JIRA, and other footer lines could come after the Sob.

scripts/check-signed-off-by.sh
utzig: Done!

scripts/check-signed-off-by.sh
utzig: I changed to check every line looking for it.

scripts/check-signed-off-by.sh
utzig: And undone!

utzig: It fails with `~1` and it was failing before for me with `^`. I suppose it has todo with the latest PR commit that GH creates which might not have the right parent<-> association. Not sure.

utzig:
Regarding the comment suggesting to add commiter checking as well as author:

```
[utzig@inspiron mcuboot]$ git show -s --format="%cn <%ce>" cb1bb48
David Brown <davidb@davidb.org>
[utzig@inspiron mcuboot]$ git show -s --format="%an <%ae>" cb1bb48 
David Brown <david.brown@linaro.org>
```

I'm not sure it would be a great idea at the moment. I would prefer to work on it later in a new patch (if at all).

scripts/check-signed-off-by.sh
d3zd3z: Have you found an example where this is failing? I would argue that tail +2 is doing the wrong thing, and HEAD~1 is doing the right thing. I'm curious what is going on with that.

d3zd3z:
You should only ever see the davidb.org commit after github rebases it. It is using my primary account for the rebase, as far as it is concerned.

You should always be ok adding the committer check, and when added, it will usually be after the author Sob.

scripts/check-signed-off-by.sh
utzig: https://travis-ci.org/runtimeco/mcuboot/builds/272715788

utzig:
Added check for commiter as well, let's see how it goes.

scripts/check-signed-off-by.sh
d3zd3z: My suspicion here is that Travis only makes a merge if it has to. If the tree can be fast forwarded, it will be. Your "tail" version is then ignoring the latest commit in the series, and my suggestion ignores all of them. Neither is correct.

I think we need to figure out either how to make travis use merge --no-ff, or be much smarter in what we check for.

```
git log --pretty=raw ${MAIN_BRANCH}..
exit 1
```
in a travis build might be revealing, do it where it needs to be rebased, and where it doesn't.

PR #118, opened 2017-09-07
[RFC] Dual image boot on ARMv8M - Rebased

tamasban:
Hi all,

I'm Tamas Ban, new in MCUBoot, and new in open source as well. I worked together with mattot01(Mate Toth Pal) at ARM on MCUBoot. The goal was that MCUBoot be able to authenticate two image on ARMv8M arch. We do code modification in the bootutil and in the simulator as well. We added the necessary BSP to Zephyr. It was tested on MPS2 board and in simulator environment as well.

Please take a look and review it!

Best regards,
Tamas


sim/src/main.rs
```diff
@@ -896,9 +1189,25 @@ pub struct ImageHeader {
     _pad3: u32,
 }
 
-impl AsRaw for ImageHeader {}
+impl AsRaw for SmallImageHeader {}
+
+#[repr(C)]
+pub struct LargeImageHeader {
+    small_header: SmallImageHeader,
+    _pad0: u32,
```
d3zd3z: `_pad: [u32; 8],` would be a little cleaner, when writing the values, use `[0; 8]` for the value of all zeros.

samples/zephyr/hello1/mps2_an505.overlay
```diff
@@ -0,0 +1,9 @@
+/* DTS overlay file.
```
d3zd3z: Did this file get left around after the rebase? This directory no longer exists.

samples/zephyr/hello2/mps2_an505.overlay
```diff
@@ -0,0 +1,9 @@
+/* DTS overlay file.
```
d3zd3z: Did this file get left around after a rebase? This directory no longer exists.

utzig:
@tamban01 Could you point me to the ARMv8 documentation related to this changes?

tamban01:
These docs can be good starting points:
[whitepaper](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&cad=rja&uact=8&ved=0ahUKEwiV7JyIisbWAhUhs1QKHZE_A8EQFggzMAE&url=https%3A%2F%2Fcommunity.arm.com%2Fcfs-file%2F__key%2Ftelligent-evolution-components-attachments%2F01-2142-00-00-00-00-66-90%2FWhitepaper-_2D00_-ARMv8_2D00_M-Architecture-Technical-Overview.pdf&usg=AFQjCNFkYimgCaaXBkmEpJi5nFtWRLBz-Q)

TRM:
http://infocenter.arm.com/help/topic/com.arm.doc.100230_0002_00_en/cortex_m33_trm_100230_0002_00_en.pdf

boot/bootutil/src/loader.c
```diff
@@ -1247,116 +1297,104 @@ boot_go(struct boot_rsp *rsp)
...
+     * bootable image:
+     *          - primary slot(image always executes from here)
+     *          - secondary slot(partitons for new image during sw upgrade)
+     *          - scratch area(used during image swapping)
+     * The flash_map_entry structure is initilized at the beginning of the
+     * loop to contain the addreses of the corresponding flash partitions,
+     * which belongs the current image.
+     */
+    for (img_idx = 0 ; img_idx < img_cnt; img_idx++) {
+        /* Initialize flash partitoin map */
```
agross-linaro: nit.  partition <sic>

boot/zephyr/flash_map.c
```diff
@@ -43,44 +43,59 @@ extern struct device *boot_flash_device;
...
+        part_map[1].area.fa_size = FLASH_AREA_IMAGE_1_SIZE;
+
+        part_map[2].magic = FLASH_MAP_ENTRY_MAGIC;
+        part_map[2].area.fa_id = FLASH_AREA_IMAGE_SCRATCH;
+        part_map[2].area.fa_device_id = FLASH_DEVICE_ID;
+        part_map[2].area.fa_off = FLASH_AREA_IMAGE_SCRATCH_OFFSET;
+        part_map[2].area.fa_size = FLASH_AREA_IMAGE_SCRATCH_SIZE;
     }
-};
+#if BOOT_IMAGE_COUNT > 1
```
agross-linaro: for n partitions, perhaps a loop with a macro to assign all this stuff would suffice.

sim/mcuboot-sys/csupport/run.c
```diff
@@ -229,4 +257,45 @@ int flash_area_get_sectors(int fa_id, uint32_t *count,
     return 0;
 }
 
-void flash_partition_map_init(uint8_t img_idx) {}
+// Set up the current_flash_areas structure.
```
agross-linaro: use C style comment blocks.  /*..... */

d3zd3z:
> utzig: Could you point me to the ARMv8 documentation related to this changes?

I think deeper than this is that this change really has nothing to do with v8m, in particular.  Yes, v8m needs to be able to boot two images, but there is nothing precluding another platform from having two images. Maybe there are two mcus that share the same flash.

boot/zephyr/flash_map.c
d3zd3z: I'm not sure how the loop would expand to macros either based on names, or with numbers in the names? The loop is a runtime thing, the names are compile time.

boot/zephyr/targets/96b_carbon.h
```diff
@@ -33,3 +33,6 @@
  * sector in size.
  */
 #define FLASH_AREA_IMAGE_SECTOR_SIZE	0x20000
+
+/* The number of bootable images by MCUBoot */
+#define BOOT_IMAGE_COUNT	1
```
mbolivar: Individual target files are going away:

https://github.com/runtimeco/mcuboot/pull/114
https://github.com/runtimeco/mcuboot/pull/140

This information is not target-specific, I don't believe. Where else can this information go? Perhaps it could be added to the Zephyr device tree, the way we have moved the flash layout? Or perhaps it can be a top-level mcuboot configuration?

My feeling is that the Zephyr image is going to need to be modified to describe the new flash partition layout in device tree.

boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -76,7 +77,7 @@ struct image_trailer {
 };
 
 /* you must have pre-allocated all the entries within this structure */
-int boot_go(struct boot_rsp *rsp);
+int boot_go(int img_cnt, struct boot_rsp *rsp);
```
mbolivar: This API change is at the heart of the new work here.

There is a design document in the doc/ directory. It is harder to review these changes without a corresponding patch to that document which declares the high-level assumptions being made.

This series should include a design doc update as well, to make sure that the code changes correspond to the high-level proposal.

boot/bootutil/src/loader.c
```diff
@@ -1247,116 +1297,104 @@ boot_go(struct boot_rsp *rsp)
...
+        flash_partition_map_init(img_idx);
+
+        /* Open boot_data image areas for the duration of this call. */
+        for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
```
mbolivar: The language around "images" versus "slots" is getting confused for the reader here; it seems like there are still three "slots" for each of the two "images", but all six of these are "flash areas". This should be cleared up in the design doc.

boot/bootutil/src/loader.c
```diff
@@ -1247,116 +1297,104 @@ boot_go(struct boot_rsp *rsp)
...
+        /* Attempt to read an image header from each slot. */
+        rc = boot_read_image_headers();
         if (rc != 0) {
             goto out;
         }
 
-        /*
-         * The following states need image_ok be explicitly set after the
-         * swap was finished to avoid a new revert.
+        /* If the image slots aren't compatible, no swap is possible.  Just boot
```
mbolivar: I'm not sure this makes sense in a multi-image context. We could install an upgrade in one "image", but fail to do so in another if the boot slots aren't compatible. This would lead to a partial upgrade, which I don't believe is what you want -- am I missing something? It seems like we need to make sure that success or failure to install upgrades should happen more uniformly across the image areas, or perhaps be made configurable (though fewer configuration knobs would be better IMO).

In general, I think the semantics around partial upgrade need to be specified clearly and the error handling behavior described.

boot/bootutil/src/loader.c
```diff
@@ -1221,21 +1221,71 @@ boot_swap_if_needed(int *out_swap_type)
...
+ * Finalize some administration work, after swapping was done:
+ *  - set copy done flag on demand
+ *  - decide wether image headers must be reloaded or not
+ *
+ * @param swap_type         The type of the previously done swapping operation
+ * @param reload_headers    Flag to indicates do we need to reload the image's
+ *                          headers
+ */
+static void
+boot_finalize_swap(int swap_type, bool *reload_headers)
```
mbolivar: This can't be correct without also knowing the image ID number, unless I am missing something.

boot/bootutil/src/loader.c
```diff
@@ -1247,116 +1297,104 @@ boot_go(struct boot_rsp *rsp)
...
-        BOOT_LOG_ERR("panic!");
-        assert(0);
-
-        /* Loop forever... */
-        while (1) {}
-    }
-
-#ifdef MCUBOOT_VALIDATE_SLOT0
-    if (reload_headers) {
+        boot_finalize_swap(swap_type, &reload_headers);
```
mbolivar: How is this correct without knowing which image to finalize the swap for?

boot/bootutil/src/loader.c
```diff
@@ -1247,116 +1297,104 @@ boot_go(struct boot_rsp *rsp)
...
-    rc = boot_validate_slot(0);
-    assert(rc == 0);
-    if (rc != 0) {
-        rc = BOOT_EBADIMAGE;
-        goto out;
-    }
-#else
-    (void)reload_headers;
+#ifdef MCUBOOT_VALIDATE_SLOT0
+        rc = boot_validate_slot(0);
```
mbolivar: Same here, not clear how this can be correct without specifying which image's slot 0 is being validated.

boot/zephyr/flash_map.c
mbolivar: If we're just sticking with two images (not N), as was discussed, BOOT_IMAGE_COUNT should become BOOT_DUAL_IMAGE or so.

boot/zephyr/flash_map.c
```diff
@@ -43,44 +43,59 @@ extern struct device *boot_flash_device;
 #define FLASH_MAP_ENTRY_MAGIC 0xd00dbeef
 
 struct flash_map_entry {
-    const uint32_t magic;
-    const struct flash_area area;
+    uint32_t magic;
```
mbolivar: These could be kept const if you initialized the array at build time by checking the build configuration and having different static initializers in the two cases.

You could also use a Zephyr SYS_INIT routine to initialize this before main() gets control as a runtime alternative that would still mean you don't need to change APIs shared with Mynewt.

But the const declarations are nice; it'd be better to keep them.

boot/zephyr/include/flash_map/flash_map.h
```diff
@@ -95,6 +95,14 @@ struct flash_sector {
 };
 
 /*
+ * Initialize the flash partition map description array according to
+ * the image index. Three memory partition belongs to each bootable
+ * image: sw_slot_0(primary slot); sw_slot_1(secondary_slot);
+ * scratch_area.
+ */
+void flash_partition_map_init(uint8_t img_idx);
```
mbolivar: This will break the build for mynewt. The bootutil code is the core code shared across the operating systems. This change only affects Zephyr's copy of flash_map.h. Thus, this API change would need to be discussed with the other supported RTOSes.

Is it possible to implement this without adding this function at all? I believe so; see comments elsewhere.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -7,4 +7,8 @@
 #define FLASH_AREA_IMAGE_1 2
 #define FLASH_AREA_IMAGE_SCRATCH 3
 
+#define FLASH_AREA_IMAGE_2 FLASH_AREA_IMAGE_0
```
mbolivar: Same story, this is just a Zephyr-specific copy of a Mynewt header; such a change would need to be discussed across OSes

boot/bootutil/src/loader.c
mbolivar: Ah, I see this is using a global variable now... this should be changed

utzig:
@tamasban Could you/we close this PR?

PR #119, opened 2017-09-07
Add Tinycrypt to tree

d3zd3z:
Instead of relying on Tinycrypt being present, and at a specific version, in each target OS, include a snapshot in our own tree. This is a snapshot of a slightly older version of Tinycrypt that works with MCUboot.


d3zd3z:
This passes my sanity tests on Zephyr, both RSA and ECDSA p256 work.

PR #131, opened 2017-09-11
Updates mynewt port to use bundled tinycrypt

utzig:
This removes the dependency on tinycrypt lib provided by mynewt and switches to the one bundled in mcuboot. Also fixes a bug where tc was only being used when eliptic curve 256 was selected as sign
algorithm.


boot/bootutil/pkg.yml
```diff
@@ -39,5 +39,5 @@ pkg.deps:
     - "@apache-mynewt-core/sys/defs"
     - "@apache-mynewt-core/sys/flash_map"
 
-pkg.deps.BOOTUTIL_SIGN_EC256:
-    - "@apache-mynewt-core/crypto/tinycrypt"
+pkg.deps.BOOTUTIL_USE_TINYCRYPT:
```
d3zd3z: This does make a bit more sense, although currently, we do only need tinycrypt if we are using EC256 signatures.

utzig: Ah, I suppose tinycrypt could be used for EC (224) too, correct?

PR #140, opened 2017-09-21
Zephyr: generic target config

mbolivar:
This series allows mcuboot to obtain all of its target-specific information from the Zephyr RTOS, instead of including target-specific headers. However, it requires some out of tree Zephyr patches (https://github.com/zephyrproject-rtos/zephyr/pull/1594), so it can't be merged yet. It's being posted for review and testing while the Zephyr patches are being considered.

It adds two patches (863cc17 and 0ae943b) to https://github.com/runtimeco/mcuboot/pull/114, so that needs to get merged first (along with the Zephyr patches) before this series can be considered for merge.


mike-scott:
Looks like the Zephyr side of this was merged.

mbolivar:
> Looks like the Zephyr side of this was merged.

Not quite -- the nRF series was merged, but the series this PR depends on, https://github.com/zephyrproject-rtos/zephyr/pull/1594, is not merged.

I've updated this to keep up with Zephyr 1594, but this unfortunately currently means nRF devices won't use flash page layout support, since the patch which selects CONFIG_FLASH_PAGE_LAYOUT for nRF devices was NAKed by Nordic. @nvlsianpu will propose an alternative approach which will allow us to enable that within mcuboot instead of zephyr.

mbolivar:
@d3zd3z this is ready for review.

I heard from @nvlsianpu that he was going to post a series that adds a Kconfig tree to the mcuboot application. That will be needed to select CONFIG_FLASH_PAGE_LAYOUT on that series, as my proposal to select it in Zephyr was rejected. Until that's available, Nordic targets will generate the 'falling back on hacks' warning, even though their driver supports page layout APIs.

The warnings on NXP targets are valid; that flash driver hasn't been augmented with page support upstream.

I still think this is worth merging now; though it adds warnings, the behavior is at worst unchanged on existing targets, and it's improved on STM targets.

d3zd3z:
When I try building on NXP, I get an "#error "Target support is incomplete; cannot build mcuboot"", I wonder how we should indicate that we depend on specific versions of Zephyr.

PR #143, opened 2017-10-02
Remove obsolete `signed_images.md` docs

d3zd3z:
This document is out of date, and irrelevant with the existence of
`imgtool.py`.  Remove it entirely to avoid confusion.  Key generation
and converting public keys to C should be done using this tool.


d3zd3z:
Perhaps we should have a comment about where newt gets involved in the imgtool docs. Since, it will be for signing, and imgtool is just used for key creation and pubkey extraction.

utzig:
Maybe updating the README-mynewt file with a new section on signing, that includes the newt commands and links to imgtool key generation docs.

d3zd3z:
Instead of removing this file, work on password protection for private keys will end up resulting in a rewrite of this document. Closing this PR for now.

PR #144, opened 2017-10-05
Log error message when slot fails validation

utzig:
This is helpful when testing an image that is not signed or signed with an invalid key.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -590,6 +590,7 @@ boot_validate_slot(int slot)
              * continue booting from slot 0.
              */
         }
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/loader.c
```diff
@@ -590,6 +590,7 @@ boot_validate_slot(int slot)
              * continue booting from slot 0.
              */
         }
+        BOOT_LOG_ERR("Image in slot%d is not valid!", slot);
```
mbolivar: I'd put a space after "slot", but this seems fine to me

utzig: Added the space!

PR #146, opened 2017-10-05
Avoid ec256 pubkey memcpy

utzig:
This is a optional fix that removes a TODO avoids a memcpy.

### DIFF
--- a/boot/bootutil/src/image_ec256.c
+++ b/boot/bootutil/src/image_ec256.c
@@ -42,46 +42,49 @@ static const uint8_t ec_secp256r1_oid[] = MBEDTLS_OID_EC_GRP_SECP256R1;
  * Parse the public key used for signing.
  */
 static int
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/image_ec256.c
```diff
@@ -65,23 +65,20 @@ tinycrypt_import_key(uint8_t *pubkey, uint8_t *cp, uint8_t *end)
...
         return -7;
     }
 
     if (len != 2 * NUM_ECC_BYTES + 1) {
         return -8;
     }
-    if (cp[0] != 0x04) {
+    if (*cp[0] != 0x04) {
         return -9;
     }
```
d3zd3z: Perhaps we should add 1 to *cp here before returning, although I guess it doesn't make all that much difference. But it makes a little more sense to do it here, since here is where we are looking for the 0x04.

Unrelated to this change, I suppose we should document somewhere that we only support uncompressed public keys.

utzig: Will do!

PR #147, opened 2017-10-05
Add faster copy/erase to overwrite-only mode

utzig:
This add a new option to overwrite-only mode that enables copy/erase of
only the amount of sectors that are required to store the source image.

This is enabled by default when overwrite-only mode is used.

MCUB-70


utzig:
@d3zd3z @mbolivar This seems good enough that I left it enabled by default for Mynewt/Zephyr/Simulator.

There's one possible "issue" remaining, which is when erasing slot 1 after the copy is done, both the first sector (header) and last (trailer) are erased. This is done because if the trailer is left un-erased, the image_ok/magic will still be there if a new image without padding/trailer is written (as is the case with mynewt), triggering a new update. It's possible that a reset between both erases would trigger this unwanted behaviour but it seems quite hard to fix for something that will hardly ever happen.

d3zd3z:
Do we need this to be an option, or should we just make this the default behavior?  We always do this with swapping.

utzig:
I was not sure so I left as an option, imagining someone might some day want to actually erase all sectors (for security reasons IMO!).

mbolivar:
I'm +1 to leave it an option.

PR #149, opened 2017-10-19
Correction to swap procedure design documentation

arewhyan:
The swap procedure as documented doesn't match the code, and it would
loose the contents of slot1[index] as it's erased before it is copied
anywhere. Also, add the missing word 'to'.

Signed-off-by: Ryan C Johnson <ryan.johnson@flex.com>


utzig:
LGTM

utzig:
Thanks Ryan!

PR #150, opened 2017-10-20
Zephyr: Use mbed TLS allocator

d3zd3z:
Instead of the overly-simplistic allocator in Zephyr, use the still
simplistic allocator in mbed TLS.

On K64f, this saves 848 bytes of text, 44 bytes of data, and 208 bytes
of bss.


boot/zephyr/os.c
```diff
@@ -24,29 +24,21 @@
...
-    size_t total = nelem * size;
-    void *buf = k_malloc(total);
-    if (buf) {
-        memset(buf, 0, total);
-    }
-    return buf;
-}
+/*
+ * This is the heap for mbed TLS.  The value needed depends on the key
+ * size and algorithm used.  For RSA-2048, 6144 bytes seems to be
```
mbolivar: Looks good to me, but how was this tested? Did you run STACK_ANALYZE() in zephyr or something like that?

boot/zephyr/prj-p256.conf
```diff
@@ -11,8 +11,8 @@ CONFIG_MBEDTLS_CFG_FILE="config-asn1.h"
 # CONFIG_TINYCRYPT_ECC_DSA is not set
 # CONFIG_TINYCRYPT_SHA256 is not set
 
-### mbedTLS wants a heap
-CONFIG_HEAP_MEM_POOL_SIZE=16384
+### mbedTLS has its own heap
+# CONFIG_HEAP_MEM_POOL_SIZE is not set
```
mbolivar: Not familiar with this API -- is it necessary to have an "is not set" line, or can this be removed from the prj.conf completely?

d3zd3z: I wasn't quite sure what the best way was. Leaving it out causes it to be "unchanged" from any default. The "is not set" will explicitly unset it, even if there is a default. I did this to try and make sure to remove any heap if it were to become set in Zephyr's default.

boot/zephyr/os.c
d3zd3z: I couldn't get any of the heap analysis code in mbed TLS to work (it didn't even compile, and didn't work once I fixed those problems). I ended up doing kind of a binary search, and being a bit generous.

PR #151, opened 2017-11-06
Test framework update

d3zd3z:
Major cleanup of the simulator test cases. This includes avoiding globals in the tests, which allows multi-threaded tests as long as we protect the actual invocation of `boot_go` with a mutex.

The tests can now be run with `cargo test`. The test executable is still present, but the executable may become something different at some point.


mbolivar:
Looks good to me, but I think the bootsim readme should be updated to match the travis changes.

PR #152, opened 2017-11-07
port serial recovery capability  to zephyr target

nvlsianpu:
This patch introduced serial bootloader functionality ported
from mynewt targets tree.

For achieving this following changes were applied:
- added tinycbor and cborattr library for zephyr. Tinycbor lib from mynewt is imported because it is already modified for support boot-serial module.
- added serial adapter module for support serial communication for zephyr 
- Modified boot_serial module for using, zephyr-os modules
  (crc driver, mbedtls-base64 library) and the zephyr serial adapter module
  introduced recently.
- Added service of boot serial recovery mode to main.
- Adapted the input parser to using static buffers.
- Added defines for map to memory allocation API of zephyr.
- Added Kconfig for zephyr-mcuboot project for configuring it in easy way.

<b>remarks:</b>
- For using serial recovery feature flash region for bootloader must be extended which is introduced in zephyr PR https://github.com/zephyrproject-rtos/zephyr/pull/4765
- console driver can't use uart along with serial recovery. It is possible to use segger RTT instead.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -16,8 +16,17 @@ menuconfig MCUBOOT_SERIAL
 	prompt "MCUBOOT serial recovery"
 	default n
 	select REBOOT
+	select UART_INTERRUPT_DRIVEN
+	select SERIAL
 	help
 	  Enable serial recovery feature in mcuboot.
+
+config BOOT_MAX_LINE_INPUT_LEN
+	int "cmd line max length" 
+	default 512
+	depends on MCUBOOT_SERIAL
+	help
+	  Maximum length of commands transported over the serial protocol
 
 config BOOT_SERIAL_DETECT_PORT
 	string
--- a/Kconfig
+++ b/Kconfig
@@ -18,6 +18,7 @@ menuconfig MCUBOOT_SERIAL
 	select REBOOT
 	select UART_INTERRUPT_DRIVEN
 	select SERIAL
+	select NEWLIB_LIBC
 	help
 	  Enable serial recovery feature in mcuboot.
 
@@ -31,7 +32,7 @@ config BOOT_MAX_LINE_INPUT_LEN
 config BOOT_SERIAL_DETECT_PORT
 	string
 	prompt "GPIO Port to trigger serial recovery mode"
-	default CONFIG_GPIO_NRF5_P0_DEV_NAME
+	default GPIO_0 if SOC_FAMILY_NRF5
 	depends on MCUBOOT_SERIAL
 	help
 	  Serial recovery detect port
@@ -39,7 +40,8 @@ config BOOT_SERIAL_DETECT_PORT
 config BOOT_SERIAL_DETECT_PIN
 	int
 	prompt "Pin to trigger serial recovery mode"
-	default 0
+	default 11 if BOARD_NRF52840_PCA10056
+	default 13 if BOARD_NRF52_PCA10040
 	depends on MCUBOOT_SERIAL
 	help
 	  Serial recovery detect pin
--- a/Kconfig
+++ b/Kconfig
@@ -1,0 +1,44 @@
+# Copyright (c) 2017 Linaro Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+mainmenu "MCUBOOT configuration"
+
+config ZEPHYR_BASE
+	string
+	option env="ZEPHYR_BASE"
+
+source "$ZEPHYR_BASE/Kconfig.zephyr"
+
+menuconfig MCUBOOT_SERIAL
+	bool
+	prompt "MCUBOOT serial recovery"
+	default n
+	select REBOOT
+	help
+	  Enable serial recovery feature in mcuboot.
+
+config BOOT_SERIAL_DETECT_PORT
+	string
+	prompt "GPIO Port to trigger serial recovery mode"
+	default CONFIG_GPIO_NRF5_P0_DEV_NAME
+	depends on MCUBOOT_SERIAL
+	help
+	  Serial recovery detect port
+
+config BOOT_SERIAL_DETECT_PIN
+	int
+	prompt "Pin to trigger serial recovery mode"
+	default 0
+	depends on MCUBOOT_SERIAL
+	help
+	  Serial recovery detect pin
+
+config BOOT_SERIAL_DETECT_PIN_VAL
+	int "Boot Pin trigger value"
+	default 0
+	range 0 1
+	depends on MCUBOOT_SERIAL
+	help
+	  The input pin logic value which triggers the boot serial recovery mode.
--- a/boot/boot_serial/src/CMakeLists.txt
+++ b/boot/boot_serial/src/CMakeLists.txt
@@ -1,0 +1,6 @@
+zephyr_sources(boot_serial.c)
+
+#mbedtls-base64 library header
+zephyr_include_directories($ENV{ZEPHYR_BASE}/ext/lib/crypto/mbedtls/include)
+
+zephyr_include_directories(${BOOT_DIR}/bootutil/include)
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -24,24 +24,35 @@
 
 #include "sysflash/sysflash.h"
 
+#ifdef __ZEPHYR__
+#include <misc/reboot.h>
+#include <misc/byteorder.h>
+#include <misc/__assert.h>
+#include <flash.h>
+#include <crc16.h>
+#include <serial_adapter/serial_adapter.h>
+
+#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
+#include "bootutil/bootutil_log.h"
+#include "mbedtls/base64.h"
+#else
 #include <bsp/bsp.h>
-
-#include <flash_map/flash_map.h>
-#include <hal/hal_flash.h>
 #include <hal/hal_system.h>
-
 #include <os/endian.h>
-#include <os/os.h>
-#include <os/os_malloc.h>
 #include <os/os_cputime.h>
-
 #include <console/console.h>
+#include <crc/crc16.h>
+#include <base64/base64.h>
+#endif /* __ZEPHYR__ */
 
 #include <tinycbor/cbor.h>
 #include <tinycbor/cbor_buf_reader.h>
 #include <cborattr/cborattr.h>
-#include <base64/base64.h>
-#include <crc/crc16.h>
+
+#include <flash_map/flash_map.h>
+#include <hal/hal_flash.h>
+#include <os/os.h>
+#include <os/os_malloc.h>
 
 #include <bootutil/image.h>
 
@@ -49,6 +60,20 @@
 #include "boot_serial_priv.h"
 
 #define BOOT_SERIAL_OUT_MAX	48
+
+#ifdef __ZEPHYR__
+/* mbedtls-base64 lib encodes data to null-terminated string */
+#define BASE64_ENCODE_SIZE(in_size) ((((((in_size) - 1) / 3) * 4) + 4) + 1)
+
+#define CRC16_INITIAL_CRC       0       /* what to seed crc16 with */
+#define CRC_CITT_POLYMINAL 0x1021
+
+#define ntohs(x) sys_be16_to_cpu(x)
+#define htons(x) sys_cpu_to_be16(x)
+
+static char in_buf[CONFIG_BOOT_MAX_LINE_INPUT_LEN + 1];
+static char dec_buf[CONFIG_BOOT_MAX_LINE_INPUT_LEN + 1];
+#endif
 
 static uint32_t curr_off;
 static uint32_t img_size;
@@ -229,6 +254,7 @@ bs_upload(char *buf, int len)
     curr_off += img_blen;
 
 out:
+    BOOT_LOG_INF("RX: 0x%x", rc);
     cbor_encoder_create_map(&bs_root, &bs_rsp, CborIndefiniteLength);
     cbor_encode_text_stringz(&bs_rsp, "rc");
     cbor_encode_int(&bs_rsp, rc);
@@ -264,8 +290,13 @@ bs_reset(char *buf, int len)
     cbor_encoder_close_container(&bs_root, &bs_rsp);
 
     boot_serial_output();
+#ifdef __ZEPHYR__
+    k_sleep(250);
+    sys_reboot(SYS_REBOOT_COLD);
+#else
     os_cputime_delay_usecs(250000);
     hal_system_reset();
+#endif
 }
 
 /*
@@ -288,9 +319,9 @@ boot_serial_input(char *buf, int len)
 
     buf += sizeof(*hdr);
     len -= sizeof(*hdr);
-
     bs_writer.bytes_written = 0;
     cbor_encoder_init(&bs_root, &bs_writer, 0);
+
 
     /*
      * Limited support for commands.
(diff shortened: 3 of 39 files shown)


nvlsianpu:
testing:
So far I have tested patch on nrf52_pca10040 and nrf52840_pca10056 targets;

For prepare the image use imgtool form mcuboot dirtectory, example of cmd: ```./scripts/imgtool.py sign  -k ./root-rsa-2048.pem -H 0x200 --included-header --version 3.0 --align 8 ../zephyr/samples/basic/blinky/outdir/nrf52_pca10040/zephyr.bin zephyr_img_nrf52.bin```

For testing the upload it is need to compile newtmgr application form master (unless released one doesn't expose -e option for image upload subcommand. Expected command is:  ```<executable_dir>/newtmgr image upload -e <signed_image-path> -c <your_serial_profile_name>```

nvlsianpu:
@utzig @carlescufi

carlescufi:
Note: the TinyCBOR we included here is temporary. Runtime is currently working on upstreaming the changes to the TinyCBOR official repository, and once that happens we will be able to use the TinyCBOR implementations in the respective RTOSs, since we will include a copy of the upstream TinyCBOR inside the ext/ folder in Zephyr.

Kconfig
```diff
@@ -0,0 +1,58 @@
...
+	int "cmd line max length" 
+	default 512
+	depends on MCUBOOT_SERIAL
+	help
+	  Maximum length of commands transported over the serial protocol
+
+config BOOT_SERIAL_DETECT_PORT
+	string
+	prompt "GPIO Port for triger serial recovery mode"
+	default CONFIG_GPIO_NRF5_P0_DEV_NAME
```
utzig: Should not this config be BSP specific somehow?

nvlsianpu: It is configuration for the zephyr application which in this case is mcuboot. It shouldn't be a part of zephyr-project sources as this peculiar configuration.

nvlsianpu:
I prepared version with cmake build enabled - here are the branch https://github.com/nvlsianpu/mcuboot/tree/add_serial_zephyr_cmake.

nvlsianpu:
@utzig @mbolivar @d3zd3z @carlescufi @SebastianBoe I reworked this for using zephyr cmake build system.

boot/zephyr/tinycbor/include/tinycbor/assert_p.h
```diff
@@ -0,0 +1,29 @@
...
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+** OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+** THE SOFTWARE.
+**
+****************************************************************************/
+
```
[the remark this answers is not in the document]
utzig: This comes from mainstream tinycbor. If we want to add header guards we should PR the main project first instead of making even more modifications than we already did. We should strive to be in sync...

boot/boot_serial/src/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+zephyr_sources(boot_serial.c)
+zephyr_include_directories(./)
```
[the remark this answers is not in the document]
nvlsianpu: boot_serial_priv.h

boot/zephyr/tinycbor/include/tinycbor/assert_p.h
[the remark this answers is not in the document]
nvlsianpu: Whole tinycbor library here ia temporary hack (see carles comment above https://github.com/runtimeco/mcuboot/pull/152#issuecomment-342554984). Threa are already works in progress for provide `good enaught` unstreamed tinycbor for zephyr, but we won't to wait for it. I would keep this library as is because it doesn't make any value to update it. I can add guards in mcuboot tinycbor copy if you wan (?)

Kconfig
carlescufi: Or maybe there should not be a default at all here? This needs to be properly configured when building MCUboot with serial recovery. Instead I would fail early if this is an empty string?

boot/zephyr/include/os/os_heap.h
```diff
@@ -20,19 +20,34 @@
...
 #include <stddef.h>
+#endif // __ZEPHYR__
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#ifdef __ZEPHYR__
+
+#define os_malloc k_malloc
```
carlescufi: is this for mbedTLS or for TinyCBOR?

boot/boot_serial/src/CMakeLists.txt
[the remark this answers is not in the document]
SebastianBoe: AFAICT all users of boot_serial_priv.h use "" relative path includes, so they don't need it on the include path.

```
sebo@mach:~/repos/mcuboot$ find . -name boot_serial_priv.h
./boot/boot_serial/src/boot_serial_priv.h
sebo@mach:~/repos/mcuboot$ ack boot_serial_priv
boot/boot_serial/test/src/boot_test.h
37:#include "boot_serial_priv.h"

boot/boot_serial/test/src/boot_test.c
34:#include "boot_serial_priv.h"

boot/boot_serial/src/boot_serial.c
49:#include "boot_serial_priv.h"
```

boot/zephyr/tinycbor/include/tinycbor/assert_p.h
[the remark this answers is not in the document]
SebastianBoe: >  I can add guards in mcuboot tinycbor copy if you wan (?)
No, that's not necessary then, was not aware that this was forked code.

Kconfig
nvlsianpu: I would convert it to:
```default CONFIG_GPIO_NRF5_P0_DEV_NAME if SOC_FAMILY_NRF5```
?

File: boot/zephyr/include/os/os_heap.h (hunk 4 above)
Comment (nvlsianpu) [reply boot/zephyr/include/os/os_heap.h]: This was for mbedTLS, thanks for point this because thanks to #150 it is unused code.

File: Kconfig [hunk 5]
```diff
@@ -20,9 +20,18 @@ menuconfig MCUBOOT_SERIAL
 	prompt "MCUBOOT serial recovery"
 	default n
 	select REBOOT
+	select UART_INTERRUPT_DRIVEN
+	select SERIAL
 	help
 	  Enable serial recovery feature in mcuboot.
 
+config BOOT_MAX_LINE_INPUT_LEN
+	int "cmd line max length"
```
d3zd3z: There is a trailing space on this line.

d3zd3z:
Applying these patches, and running `make test-good-rsa`, then `pyocd-flashtool -ce`, and `make flash_boot` results in:
```
***** USAGE FAULT *****
  Executing thread ID (thread): 0x20002608
  Faulting instruction address:  0x0
  Illegal use of the EPSR
```

boot/boot_serial/src/boot_serial.c
```diff
@@ -339,8 +370,14 @@ boot_serial_output(void)
     bs_hdr->nh_len = htons(len);
     bs_hdr->nh_group = htons(bs_hdr->nh_group);
 
+#ifdef __ZEPHYR__
+    crc =  crc16((u8_t *)bs_hdr, sizeof(*bs_hdr), CRC_CITT_POLYMINAL, CRC16_INITIAL_CRC,
```
carlescufi: FYI:https://github.com/zephyrproject-rtos/zephyr/pull/5098/files

carlescufi: Actually, you should use the one in the header and then you'll get the faster version once that PR is merged:

https://github.com/zephyrproject-rtos/zephyr/blob/master/include/crc16.h#L47

nvlsianpu:
@d3zd3z 

> Applying these patches, and in samples/zephyr, running make test-good-rsa, then pyocd-flashtool -ce, and make flash_boot results in:

I tested this on nrf52_pca10040 and this test works well (apart from that image_ok.bin file is missing). Unfortunately I don't have K64F which I think you used. Could you investigate what had happened? Could you run this test for the master?

boot/boot_serial/src/boot_serial.c
nvlsianpu: The problem here is that crc is calculated form two parts, so 1st crc is initial value for 2nd calculation and 1st calculation is without padding. Current and pointed implementation of ciit crc make such calculation impossible.
But maybe ther is a mathematical feature of CRC that you know, but I don't know.

boot/zephyr/include/os/os_heap.h
nvlsianpu: loader.c fille also call malloc so I would keep these definitions.

nvlsianpu:
@d3zd3z Updated - I check that now frdm_k64f *.hex output for default configuration is the same as without this patch. The serial mode is disabled by default.

boot/zephyr/main.c
```diff
@@ -101,6 +107,29 @@ void main(void)
             ;
     }
 
+#ifdef CONFIG_MCUBOOT_SERIAL
+
+    struct device *detect_port;
+    u32_t detect_value;
+
+    detect_port = device_get_binding(CONFIG_GPIO_NRF5_P0_DEV_NAME);
```
mbolivar: I think it's OK to start out the merge like this since the initial users are all on nRF devices, but there is a better way to do this than putting SoC specific CONFIG option names in the main sources of the file which I think should be fixed in the future.

The basic idea is that, instead of putting CONFIG_GPIO_NRF5_P0_DEV_NAME in the source itself, you should pick a name that is mcuboot-specific and use it here, like this:

```
#define SERIAL_DETECT_GPIO_DEVICE "mcuboot-serial-gpio"

...

detect_port = device_get_binding(SERIAL_DETECT_GPIO_DEVICE);
```

Then, you can use SoC or device-specific .conf file fragments to set CONFIG_GPIO_NRF5_P0_DEV_NAME to "mcuboot-serial-gpio".

Here is an example:

https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/drivers/led_ws2812/src/main.c

Note how SPI_DEV_NAME is just "ws2812_spi". Then, board-specific .conf file fragments configure the appropriate device:

https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/drivers/led_ws2812/boards/96b_carbon.conf

To eliminate duplication, these fragments are merged into the final configuration along with the top-level prj.conf:

https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/drivers/led_ws2812/CMakeLists.txt#L1

(Perhaps DT could be used as well, but I don't know how.)

boot/zephyr/prj.conf
```diff
@@ -13,6 +13,7 @@ CONFIG_MBEDTLS_CFG_FILE="config-boot.h"
 
 CONFIG_FLASH=y
 CONFIG_MPU_ALLOW_FLASH_WRITE=y
+#CONFIG_FLASH_PAGE_LAYOUT=y
```
mbolivar: Debug code, should be deleted, fine to fix in a follow-up series.

boot/zephyr/tinycbor/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(src)
+zephyr_compile_definitions(-DWITHOUT_OPEN_MEMSTREAM)
+zephyr_include_directories(include)
```
mbolivar: Missing newline at end of file, fine to fix in a follow-up

boot/zephyr/tinycbor/src/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+zephyr_sources(cbor_buf_reader.c cborencoder.c cborerrorstrings.c cborparser_dup_string.c cbortojson.c open_memstream.c cbor_buf_writer.c cborencoder_close_container_checked.c cborparser.c cborpretty.c)
```
mbolivar: missing newline, ok to fix in a follow up

Kconfig
```diff
@@ -0,0 +1,66 @@
...
+	option env="PROJECT_BASE"
+
+source "$ZEPHYR_BASE/Kconfig.zephyr"
+
+# this is shadow entry for selections making
+config MCUBOOT_FORCE_CONFIG_SELECTION
+	bool
+	default y
+	# avoid MCUB-90 bug on nRF5x targets
+	select FLASH_PAGE_LAYOUT if SOC_FLASH_NRF5
```
carlescufi: Why does this need to be selected here instead of being done in Zephyr's own Kconfig files?

grandcat: Actually would safe a lot of hassle (currently) :+1:

mbolivar: This came up a while back; I believe @nvlsianpu may have been concerned about the additional size this added to the flash library:

https://github.com/zephyrproject-rtos/zephyr/pull/1594#pullrequestreview-64524847

nvlsianpu: @carlescufi I was intended to avoid forcing to use this extension in general (for all zephyr nRF5 targets).
I wouldn't add mcuboot dependency into zephyr (that is what you mean?).

boot/zephyr/CMakeLists.txt
```diff
@@ -168,3 +171,12 @@ if (NEED_TINYCRYPT)
   target_sources(app PRIVATE "${TINYCRYPT_DIR}/source/sha256.c")
   target_sources(app PRIVATE "${TINYCRYPT_DIR}/source/utils.c")
 endif()
+
+if (CONFIG_MCUBOOT_SERIAL)
+zephyr_sources(${BOOT_DIR}/zephyr/serial_adapter.c)
+
+add_subdirectory(${BOOT_DIR}/boot_serial ./boot/boot_serilal)
```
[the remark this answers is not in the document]
nvlsianpu: This haw I described modules for zephyr build in these many files is dictating by zephyr approach to this: one module/directory - one CMakeLists.txt For sure this is not requirement for mcuboot so it could be reworked later.

SebastianBoe:
The one-to-one mapping between directories and build script files was a
requirement in KBuild,
but it does not exist in CMake. It is up to the developer how many
CMakeLists.txt files he finds
it appropriate to use.

Sebastian Bøe

boot/zephyr/main.c
nvlsianpu: @mbolivar  - I agree - lest merge this like is now and rework it later - thank for tutorial.

Kconfig
nvlsianpu: After discussion with Carles I will remove this, and I will make FLASH_PAGE_LAYOUT selected by default for nRF5 targets in zephyr.

boot/zephyr/main.c
mbolivar: @nvlsianpu :+1: np

boot/zephyr/cborattr/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(src)
+zephyr_include_directories(include)
```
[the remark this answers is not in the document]
nvlsianpu: All pointed missing newlines were added - I don't know why it is invisible in conversation view.

boot/zephyr/main.c
mbolivar: So looking at this series again and some of the discussion, I think it's almost there, strangely. This series adds BOOT_SERIAL_DETECT_PORT in the Kconfig, sets it to CONFIG_GPIO_NRF5_P0_DEV_NAME, but then uses the nRF value here instead of the Kconfig value... why? 

Anyway, the cleanup commit should just use CONFIG_BOOT_SERIAL_DETECT_PORT here. You can add board-specific config fragments as described in my earlier comments to replace the "default foo if BOARD_NRFxxx" entries for BOOT_SERIAL_DETECT_PIN. Other changes can be made at SoC or board level with some mix of Kconfig edits, DT changes, and board config fragments.

SebastianBoe:
LGTM.


(I don't know how to remove my "requested changes" status)

d3zd3z:
This tests ok on Zephyr frdmk64. I haven't tried testing the serial part.

PR #153, opened 2017-11-14
Zephyr: port to cmake

mbolivar:
@carlescufi @nvlsianpu FYI

This is an initial port of the Zephyr bootloader port to CMake.

@d3zd3z, not sure why, but I need the patch https://github.com/runtimeco/mcuboot/commit/d742ec88afd85a9f93d2decad392fc73e8f5866b to get this to build when I run cmake as "cmake -DCONF_SIGNATURE_TYPE=ECDSA_P256 .." from my build directory.

Does this seem legit to you? The linker error I get without that is in the commit message for that patch.

This series deletes all of the random Zephyr-specific build system files from around the tree and concentrates all the instructions in a single boot/zephyr/CMakeLists.txt. I think that's much cleaner and friendlier to other RTOS.

If this looks good (and defining exit() is OK -- though I'll add some logging before final submission), I'll port the samples and zephyr-testplan.md


nvlsianpu:
@mbolivar I will look at this patch in nearest Thursday.

mbolivar:
> @mbolivar I will look at this patch in nearest Thursday.

Thanks!

By the way, I retried building the bootloader like this:

```
make BOARD=nrf52_blenano2 CONF_SIGNATURE_TYPE=ECDSA_P256 -j8
```

With Zephyr commit 8bffcda54 ("doc: usermode: iterative refinements"), and it breaks with the same issue (exit() is not defined).

So the definiton of exit() indeed appears to be needed. I'm going to re-spin this as a non-RFC with that definition provided.

boot/zephyr/os.c
```diff
@@ -42,3 +42,10 @@ void os_heap_init(void)
 {
     mbedtls_memory_buffer_alloc_init(mempool, sizeof(mempool));
 }
+
+void exit(int status)
+{
+	/* Why? */
```
d3zd3z: This is unlikely to be the correct fix :-)

This is a consequence of me moving the allocator to use the mbed_tls allocator. I did this with the RSA code, making a change to the config file, and there likely needs to be a change made to the config file used for ECDSA. I'll try to fix this today.

utzig:
LGTM

mbolivar:
@d3zd3z I've added your patch to the beginning of the series as a dependency since it's not merged yet.

This is ready for review. Note that I added a fix to the core bootloader (bcbd284) to help the tests pass, and added a Zephyr-specific script for making the testing more convenient.

mbolivar:
> Note that I added a fix to the core bootloader (bcbd284) to help the tests pass

@utzig heads up on this one

carlescufi:
@SebastianBoe could you take a quick look?

mbolivar:
> I've added your patch to the beginning of the series as a dependency since it's not merged yet.

That's now merged, so I've rebased.

SebastianBoe:
LGTM.

mbolivar:
Thanks for the review, @SebastianBoe.

I caught a copy/paste error in a comment which I've fixed. Diff:

https://gist.github.com/mbolivar/ac2e16f2a2ed8984f43992143da5c71c

README-zephyr.rst
```diff
@@ -31,16 +31,18 @@ Building the bootloader itself
...
-building a bit easier.  It assumes that the mcuboot tree is next to,
-at the same level, as the zephyr source tree.  It takes a single
-argument, which is the target to build. This must be a Zephyr board
-which supports mcuboot.
+To build mcuboot, create a build directory in boot/zephyr, and build
+it as usual:
+
+$ cd boot/zephyr
+$ mkdir build && cd build
+$ cmake -DBOARD=your_board ..
```
carlescufi: for consistency with the existing Zephyr doc I would recommend using `-DBOARD=<board>`

samples/zephyr/hello-world/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# frdm_k64f.
+if (NOT DEFINED BOARD)
+  set(BOARD frdm_k64f)
+endif()
+
+# The default top-level application configuration is prj.conf.
+# You can place additional board-specific files in boards/${BOARD}.conf,
+# and they will be merged into the configuration along with prj.conf.
+macro(set_conf_file)
+  if(EXISTS               ${APPLICATION_SOURCE_DIR}/boards/${BOARD}.conf)
```
carlescufi: extra tab here?

boot/zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,168 @@
+# CMakeLists.txt for building mcuboot as a Zephyr project
+#
+# Copyright (c) 2017 Open Source Foundries Limited
+#
+# SPDX-License-Identifier: Apache-2.0
+
```
carlescufi: do you need a cmake_miminum_version() here? the boilerplate from Zephy will do a check, but there might be something you're doing before that that it's not supported in older versions.

samples/zephyr/hello-world/CMakeLists.txt
mbolivar: copy/paste issue
https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/drivers/led_lpd8806/CMakeLists.txt

boot/zephyr/CMakeLists.txt
mbolivar: That's not a recommended practice in the application development primer, but I can add one in if it's a good idea (though everything above is pretty simple list, set, etc.). I'm not using anything that isn't documented in the 3.8.2 CMake reference manual, so I'll use that to match Zephyr.

mbolivar:
@carlescufi updated, thanks for the review!

nvlsianpu:
@mbolivar @SebastianBoe Basing on this patch I prepared changes to  [zephyr serial recover PR](https://github.com/runtimeco/mcuboot/pull/152)
branch: https://github.com/nvlsianpu/mcuboot/tree/add_serial_zephyr_cmake, cmake descryption are here [cmake diff](https://github.com/nvlsianpu/mcuboot/compare/4155b089769d69741870492323ff60ce90366fc6...add_serial_zephyr_cmake)

PR #156, opened 2017-11-20
Fix repository for v1.0.0

kYc0o:
When using the release version of mcuboot for mynewt I realised (thanks to @utzig) that the release v1.0.0 wasn't up-to-date in repository.yml.

This PR adds it and makes it the latest.


utzig:
Thanks for the fix!

PR #158, opened 2017-11-21
Password support for private keys

d3zd3z:
This series changes a few things:
- Private keys are now written in PKCS#8. It can still read the old format keys, but 'Newt' will have to be modified to support PKCS#8 in order to merge this, because it will now only write the new format.
- PKCS1.5 is removed entirely.
- It adds support for password protected private keys. The `-p` argument to `keygen` will prompt for a password, and any time a password protected key file is used, it will prompt for the password.
- The dependencies change, now needing only cryptography instead of several crypto libraries.


scripts/imgtool/keys/__init__.py
```diff
@@ -0,0 +1,62 @@
+"""
```
jbech-linaro: What is this with a file called `__init__.py`, I haven't seen that before.

scripts/imgtool.py
```diff
@@ -1,13 +1,27 @@
...
 from imgtool import version
 import sys
 
 def gen_rsa2048(args):
-    keys.RSA2048.generate().export_private(args.key)
+    if args.password:
+        while True:
+            passwd = getpass.getpass("Enter key passphrase: ")
+            passwd2 = getpass.getpass("Reenter passphrase: ")
+            if passwd == passwd2:
```
utzig: This is OK but maybe in the future would be interesting checking for a minimum size, and maybe using `passwordmeter` or some other entropy measuring library...

scripts/imgtool/keys/__init__.py
utzig: Every package needs one and it works similarly to `__init__` in classes. When a package is imported `__init__.py` is run.

scripts/imgtool/keys/ecdsa_test.py
```diff
@@ -0,0 +1,99 @@
...
+        k.key.public_key().verify(
+                signature=sig,
+                data=buf,
+                signature_algorithm=ec.ECDSA(SHA256()))
+
+        # Modify the message to make sure the signature fails.
+        self.assertRaises(InvalidSignature,
+                k.key.public_key().verify,
+                signature=sig,
+                data=b'This is thE message',
```
utzig: `thE` ?

scripts/imgtool/keys/ecdsa.py
```diff
@@ -0,0 +1,102 @@
...
+        self.key = key
+
+    def shortname(self):
+        return "ecdsa"
+
+    def _unsupported(self, name):
+        raise ECDSAUsageError("Operation {} requires private key".format(name))
+
+    def _get_public(self):
+        return self.key
```
utzig: I would have gone for using `self.key` directly below.

utzig:
Are you planning to run the tests on travis-ci?

scripts/imgtool/keys/ecdsa.py
d3zd3z: The method is overridden below, and returns something different. Using self.key wouldn't work there.

scripts/imgtool/keys/ecdsa_test.py
d3zd3z: That is the error that causes the signature to fail. I was trying to make it clear it was a single bit failure.

d3zd3z:
Could possibly run the tests with travis, although I'm not real sure how they are even supposed to be run. I didn't try very hard to run the `unittest` discovery code, but initially, it didn't work.

scripts/imgtool/keys/__init__.py
mbolivar: (Though note the requirement for `__init__.py` was dropped in Python 3)

scripts/imgtool/keys/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    except ValueError:
+        # This seems to happen if the key is a public key, let's try
+        # loading it as a public key.
+        pk = serialization.load_pem_public_key(
+                raw_pem,
+                backend=default_backend())
+
+    if isinstance(pk, RSAPrivateKey):
+        if pk.key_size != 2048:
+            raise Exception("Unsupport RSA key size: " + pk.key_size)
```
mbolivar: should be "Unsupported", here and below.

scripts/imgtool/keys/general.py
```diff
@@ -0,0 +1,35 @@
...
+            print("0x{:02x},".format(b), end='', file=file)
+        print("\n" + trailer, file=file)
+        if len_format is not None:
+            print(len_format.format(len(encoded)), file=file)
+
+    def emit_c(self, file=sys.stdout):
+        self._public_emit(
+                header="const unsigned char {}_pub_key[] = {{".format(self.shortname()),
+                trailer="};",
+                indent="\t",
```
mbolivar: Please use four spaces for consistency with mcuboot style

scripts/imgtool.py
```diff
@@ -1,15 +1,33 @@
...
+                break
+            print("Passwords do not match, try again")
+
+        # Password must be bytes, assume UTF-8.
+        return passwd.encode('utf-8')
```
mbolivar: Why not sys.stdin.encoding? Where else does getpass get its input from?

(Same comment applies elsewhere)

scripts/imgtool.py
```diff
@@ -1,15 +1,33 @@
...
 def gen_rsa2048(args):
-    keys.RSA2048.generate().export_private(args.key)
+    passwd = get_password(args)
+    keys.RSA2048.generate().export_private(path=args.key, passwd=passwd)
+
 def gen_ecdsa_p256(args):
-    keys.ECDSA256P1.generate().export_private(args.key)
+    passwd = get_password(args)
+    keys.ECDSA256P1.generate().export_private(args.key, passwd=passwd)
 def gen_ecdsa_p224(args):
```
mbolivar: Missing newline nit

scripts/imgtool.py
d3zd3z: The "correct" answer is to file a bug against `getpass`. It should return a byte-string, and not be decoding/reencoding a string.

It is pretty reasonable to assume the user's terminal is UTF-8. There are too many other things that wouldn't be working for them if that weren't the case, including all of Go.

scripts/imgtool/keys/general.py
d3zd3z: This is per pep8, continuations lines should be indented more than one indent level.

mbolivar: I meant replacing the literal string "\t" with four spaces.

scripts/imgtool.py
mbolivar: > The "correct" answer is to file a bug against getpass. It should return a byte-string, and not be decoding/reencoding a string.

No, a password is inherently a human-defined string. That means Unicode, not bytes. Just because cryptography.hazmat.primitives.serialization.BestAvailableEncryption wants bytes doesn't change that.

So, I read the getpass.py source code for Python 3.6.

On Unix, getpass tries:

- TextIOWrapper(FileIO('/dev/tty')).readline(), which returns str
- sys.stdin.readline(), which returns str

On Windows, it uses msvcrt.getwch() (https://docs.python.org/3/library/msvcrt.html#msvcrt.getwch), which returns "a Unicode value". I'm assuming that starts with the wint_t returned from conio.h's _getwch() under the covers.

So getting a Unicode value back from getpass() makes sense.

> It is pretty reasonable to assume the user's terminal is UTF-8.

I'm not sure what the terminal has to do with anything, especially since two out of the three possible input sources for getpass() are not (necessarily) Unix terminals.

The real issue here seems to be that the cryptography module wants a byte string argument to BestAvailableEncryption. I assume you ran into the `raise ValueError("Password must be 1 or more bytes.")` in serialization.py. That has nothing to do with the user's terminal as far as I can tell.

> There are too many other things that wouldn't be working for them if that weren't the case, including all of Go.

That seems like a non-sequitur. Go is not a mandatory mcuboot dependency in the Zephyr port. What are you saying?

At any rate, I think you really should reconsider this in favor of something like sys.getdefaultencoding(). That way we both get what we seem to want: the value stays utf-8 usually, but the code isn't making assumptions. Remember the Zen of Python:

"In the face of ambiguity, refuse the temptation to guess."

mbolivar: > At any rate, I think you really should reconsider this in favor of something like sys.getdefaultencoding().

And perhaps filing a bug against cryptography, not getpass :wink:

utzig:
@d3zd3z running `getpub` passing the wrong password, throws the exception below:

```
Enter key passphrase: 
Traceback (most recent call last):
  File "/home/utzig/src/mynewt/mcuboot/scripts/imgtool/keys/__init__.py", line 26, in load
    backend=default_backend())
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/primitives/serialization.py", line 20, in load_pem_private_key
    return backend.load_pem_private_key(data, password)
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1015, in load_pem_private_key
    password,
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1234, in _load_key
    self._handle_key_loading_error()
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1263, in _handle_key_loading_error
    raise ValueError("Bad decrypt. Incorrect password?")
ValueError: Bad decrypt. Incorrect password?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./imgtool.py", line 129, in <module>
    args()
  File "./imgtool.py", line 126, in args
    subcmds[args.subcmd](args)
  File "./imgtool.py", line 54, in do_getpub
    key = load_key(args)
  File "./imgtool.py", line 51, in load_key
    return keys.load(args.key, passwd)
  File "/home/utzig/src/mynewt/mcuboot/scripts/imgtool/keys/__init__.py", line 39, in load
    backend=default_backend())
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/primitives/serialization.py", line 24, in load_pem_public_key
    return backend.load_pem_public_key(data)
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1041, in load_pem_public_key
    self._handle_key_loading_error()
  File "/usr/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1292, in _handle_key_loading_error
    raise ValueError("Could not deserialize key data.")
ValueError: Could not deserialize key data.
```

scripts/imgtool.py
d3zd3z: Actually, after some thought, I will argue that this should _always_ be using UTF-8. The fact that the user may have different encodings at different times only argues more strongly that we should always use a consistent encoding when using this input as a password.

Let's say a user starts out with Shift-JIS encoding. As long as getpass does things right, the result will end up as a Unicode string within python, because in Python 3, strings are always Unicode. Later, they may change their configuration to use a UTF-8-based input system. The string should still come into python as Unicode.

If we used the user's encoding, the password bytes would be different. But, if we consistently use UTF-8, we'll get the same password bytes, no matter how the user input it.

d3zd3z:
Any thoughts on handling the bad password? Unfortunately, PEP-5 doesn't have an hmac and no real way to determine if a password is correct. I could trap ValueError (which is really vague) and just hope that means the password is incorrect. The message from the exception is actually fairly useful.

I think the first error is because the padding is incorrect on the resulting decryption (which happens slightly less than 255/256 times). The second error is when it tries to decode the decrypted key, anyway, and is unable to decode it. Neither of these is definitive. Openssl doesn't return a more specific error about that padding because that tends to result in opening a padding oracle attack (not the case here, though)

utzig:
What about something like this?

```
-    except ValueError:
+    except ValueError as e:
         # This seems to happen if the key is a public key, let's try
         # loading it as a public key.
-        pk = serialization.load_pem_public_key(
-                raw_pem,
-                backend=default_backend())
+        if "deserialize key data" in str(e):
+            pk = serialization.load_pem_public_key(
+                    raw_pem,
+                    backend=default_backend())
+        else:
+            raise e
```

d3zd3z:
I've left the last four unsquashed (and un-signed-off-by). Once we're happy with them, I'll squash them into the other commits before merge.

utzig:
@d3zd3z LGTM

scripts/imgtool.py
mbolivar: Alright, that's fair. The comment you've added is a good explanation for the updated rationale.

d3zd3z:
This is just squashing plus a single change, in RSA: Use 32-byte salt with PSS, I missed two other instances of the salt size, which I also set to 32. Now the tests pass.

mbolivar:
> This is just squashing plus a single change, in RSA: Use 32-byte salt with PSS, I missed two other instances of the salt size, which I also set to 32. Now the tests pass.

Cool, thanks for catching that.

PR #159, opened 2017-11-24
Do not assert on status write fails

utzig:
TLDR: this changes how mcuboot handles write fails to the status area (trailer metadata), allowing writes to fail as long as writes to the image data do not fail, and slot 0 validation is enabled. The usual behavior of aborting on any write failure is retained if slot 0 validation is disabled.

Beyond the changes to the swap write routines, some new stuff was added to the simulator to enable more thorough testing of this change:

* Enable/disable setting to check writes to non-erased areas;
* Allow declaring `bad memory` regions;
* Enable catching asserts.

Both of those are used by new tests. The assert stuff is used by a new test that was added to ensure that write failures with validation of slot 0 disabled are caught without breaking the test (it is expected to assert!). The bad memory "simulation" is used to force writes to fail based on address, size and rate of failure.

JIRA: MCUB-31

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -34,6 +34,12 @@
 #define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
 #include "bootutil/bootutil_log.h"
 
+#ifdef __BOOTSIM__
+#undef assert
+void sim_assert(int, const char *test, const char *, unsigned int, const char *);
+#define assert(x) sim_assert((x), #x, __FILE__, __LINE__, __func__)
+#endif
+
 int boot_current_slot;
 
 const uint32_t boot_img_magic[] = {
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -34,12 +34,6 @@
 #define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
 #include "bootutil/bootutil_log.h"
 
-#ifdef __BOOTSIM__
-#undef assert
-void sim_assert(int, const char *test, const char *, unsigned int, const char *);
-#define assert(x) sim_assert((x), #x, __FILE__, __LINE__, __func__)
-#endif
-
 int boot_current_slot;
 
 const uint32_t boot_img_magic[] = {
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -49,6 +49,18 @@ void sim_assert(int, const char *test, const char *, unsigned int, const char *)
 
 static struct boot_loader_state boot_data;
 
+#ifdef MCUBOOT_VALIDATE_SLOT0
+static int boot_status_fails = 0;
+#define BOOT_STATUS_ASSERT(x)                \
+    do {                                     \
+        if (x) {                             \
+            boot_status_fails++;             \
+        }                                    \
+    } while (0)
+#else
+#define BOOT_STATUS_ASSERT(x) assert(x)
+#endif
+
 struct boot_status_table {
     /**
      * For each field, a value of 0 means "any".
@@ -392,6 +404,8 @@ boot_read_status_bytes(const struct flash_area *fap, struct boot_status *bs)
     uint8_t status;
     int max_entries;
     int found;
+    int found_idx;
+    int invalid;
     int rc;
     int i;
 
@@ -399,6 +413,8 @@ boot_read_status_bytes(const struct flash_area *fap, struct boot_status *bs)
     max_entries = boot_status_entries(fap);
 
     found = 0;
+    found_idx = 0;
+    invalid = 0;
     for (i = 0; i < max_entries; i++) {
         rc = flash_area_read(fap, off + i * BOOT_WRITE_SZ(&boot_data),
                              &status, 1);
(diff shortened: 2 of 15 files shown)


utzig:
@mbolivar @d3zd3z I am still running some sanity tests on hw but so far so good, I don't want to change anymore code here unless as result of review issues!

utzig:
@d3zd3z Fixed the GLIBC dependency.

sim/mcuboot-sys/csupport/run.c
```diff
@@ -238,8 +238,15 @@ int flash_area_get_sectors(int fa_id, uint32_t *count,
 void sim_assert(int x, const char *assertion, const char *file, unsigned int line, const char *function)
 {
     if (!(x)) {
-        c_catch_asserts
-            ? c_asserts++
-            : __assert_fail(assertion, file, line, function);
+        if (c_catch_asserts) {
```
d3zd3z: Should we fold this into the patch that adds sim_assert? Right now, we have a bisection regression on MacOS.

d3zd3z:
Hmm, not sure where my comment went. I think the Fix glibc thing should get folded into the change that adds the code. Otherwise, we have a git bisection problem on MacOS.

utzig:
@d3zd3z squashed the mentioned commits!

boot/bootutil/src/bootutil_misc.c
```diff
@@ -17,7 +17,12 @@
  * under the License.
  */
 
+#ifdef __BOOTSIM__
```
mbolivar: In general I don't like this sort of thing; we already have quite a bit of ifdeffery in mcuboot, and it makes it harder to audit/review. The fact that this definition has to happen in multiple files is also a sign of trouble.

Can you just build in the simulator with -Dassert=sim_assert or something like that? (Maybe adding an -include "place-where-sim_assert-is-declared.h" to the compile flags as well to get the declaration.)

Barring that, I think adding a header like bootutil_log.h for assertions, which abstracts the runtime environment in the same way, would be better -- especially since BOOT_STATUS_ASSERT() gets added later in the series.

mbolivar: Hmm... and now I see even later in the series, the behavior changes even more.

Can you please clean this up a bit and squash the commits around so it's easier to review? The behavior of asserts seems to change a few times in this series, and it's not clear to me where this is going.

boot/bootutil/src/loader.c
```diff
@@ -48,6 +48,18 @@ void sim_assert(int x);
 
 static struct boot_loader_state boot_data;
 
+#ifdef MCUBOOT_VALIDATE_SLOT0
```
mbolivar: I admit I'm nervous about the correctness of this patch. No specific issues, it's just making code that's already a bit complex even more so and have more ifdefs.

I think that any failures here are probably a sign of bad hardware, and that mcuboot in general should fail fast, especially in critical code like this part.

Is there a use case driving this?

mbolivar: > Is there a use case driving this?

E.g. if it's just to work around write failures caused randomly by 75037fd ("Add configurable simulation of bad memory writes"), I think that either the simulation should disable those errors in certain areas that "shouldn't" fail, or the errors should be taken seriously here.

mbolivar:
@utzig not sure why my reviews are showing up as "outdated"

d3zd3z:
Github seems to be getting excessively confused by this particular series. Let's make sure everything gets address, even if it means manually confirming before doing a merge.

boot/bootutil/src/loader.c
utzig: > I think that any failures here are probably a sign of bad hardware, and that mcuboot in general should fail fast, especially in critical code like this part.
>
> Is there a use case driving this?

This was an agreed on behavior to use in a meeting and thankfully (for my fuzzy memory!) it was written down here: https://runtimeco.atlassian.net/browse/MCUB-31

```
We will examine the code and make sure: 
If it is a data write failure then panic 
If it is a status failure then ignore
```
Also I am pretty sure this came from someone not from Runtime, @d3zd3z might remember the rationale.

> E.g. if it's just to work around write failures caused randomly by 75037fd ("Add configurable simulation of bad memory writes"), I think that either the simulation should disable those errors in certain areas that "shouldn't" fail, or the errors should be taken seriously here.

I think this is exactly what it does! The random fails are configured by address/size, so I leave it disabled for most tests apart from the ones (two if remember correctly) that I want to inspect behavior of failure handling. And for those I only mark as bad area the status area inside the trailer leaving most of the flash memory unaffected and behaving as fail on every error.

mbolivar: > This was an agreed on behavior to use in a meeting and thankfully (for my fuzzy memory!) it was written down here: https://runtimeco.atlassian.net/browse/MCUB-31

Alright, if this has been discussed and agreed, I'm OK with being nervous in my little corner instead ;). Thanks for the pointer.

utzig:
I squashed the 3 commits that dealt with assert simulation, etc. I am not particularly happy with the results but well... will try fixing the "abstraction" issue now to try to finish this.

utzig:
@mbolivar A slightly better abstraction for the `assert` stuff in the simulator could be like this: https://gist.github.com/utzig/89554e63a66ddb6fbfebac557cb5ec84

I am not sure how much of an improvement it is, because I am not particularly unhappy with the way it was done, but this one at least moves responsibilities to more "appropriate" places and fixes the copy/paste.

mbolivar:
@utzig that's indeed better, but now every file and header has to make sure that every include (and transitive includes) pulls in assert.h before bootsim.h, right? That seems like a recipe for trouble :smiley: .

Having something specific like BOOT_STATUS_ASSERT() is nice, since it too avoids the copy/paste, but has the additional advantage that it doesn't matter what order headers get pulled in. That's kind of what I was trying to say in this earlier comment:

> adding a header like bootutil_log.h for assertions, which abstracts the runtime environment in the same way, would be better -- especially since BOOT_STATUS_ASSERT() gets added later in the series.

With an mcuboot-specific name, whenever you need an assert that behaves in a predictable way, you can use that.

utzig:
@mbolivar Ah I see, but `BOOT_STATUS_ASSERT` had a specific purpose. I tried hard not to change every `assert` in mcuboot to let's say `ASSERT`. Maybe changing the ones that I actually expect to fail in tests than!

utzig:
@mbolivar This commit makes `ASSERTS` explicit and better encapsulated: https://github.com/runtimeco/mcuboot/pull/159/commits/9b9379f273a239d846a1fb15086b7972d917151b

utzig:
@mbolivar Anything else remaining for approval?

PR #160, opened 2017-11-29
Fix for gcc warnings

utzig:
Fixes warnings treated as errors when upgrading gcc crate to 0.3.54. Warnings are due to non-used variables and comparison between signed and unsigned.

### DIFF
--- a/boot/bootutil/src/image_rsa.c
+++ b/boot/bootutil/src/image_rsa.c
@@ -268,7 +268,7 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, int slen,
     end = cp + *bootutil_keys[key_id].len;
 
     rc = bootutil_parse_rsakey(&ctx, &cp, end);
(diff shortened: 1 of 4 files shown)


boot/bootutil/src/image_rsa.c
```diff
@@ -268,7 +268,7 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, int slen,
     end = cp + *bootutil_keys[key_id].len;
 
     rc = bootutil_parse_rsakey(&ctx, &cp, end);
-    if (rc || slen != ctx.len) {
+    if (rc || (size_t)slen != ctx.len) {
```
mbolivar: Why not just take a size_t?

utzig: The main reason for not doing it here is that by changing the interface of `bootutil_verify_sig` I would need to update all validation algorithms too (not just RSA). In #161 I add ecdsa validation to the simulator so there I have better testing coverage to do this.

PR #161, opened 2017-12-05
Add ecdsa testing in sim

utzig:
This adds ECDSA testing to the simulator.


utzig:
@mbolivar The suggested change in #160 was added to this PR with the following commit:

https://github.com/runtimeco/mcuboot/pull/161/commits/035191dc4a3202dda3cab066a850f8c23f7fe187

sim/mcuboot-sys/build.rs
```diff
@@ -37,10 +38,23 @@ fn main() {
...
+        conf.include("mbedtls/include");
+        conf.include("../../ext/tinycrypt/lib/include");
+
+        conf.file("../../boot/zephyr/keys.c");
+
+        conf.file("../../ext/tinycrypt/lib/source/utils.c");
+        conf.file("../../ext/tinycrypt/lib/source/sha256.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc_dsa.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc_platform_specific.c");
```
d3zd3z: If you're adding files from a new directory, you need to add a call to `walk_dir` below for that directory, or it will not see changes to those files as a dependency.

sim/mcuboot-sys/src/c.rs
```diff
@@ -60,5 +71,9 @@ mod raw {
 
         pub static BOOT_MAGIC_SZ: u32;
         pub static BOOT_MAX_ALIGN: u32;
-    }
+
+        pub fn ecdsa256_sign_(privkey: *const u8, hash: *const u8,
+                              hash_len: libc::c_uint,
+                              signature: *mut u8) -> libc::c_int;
+}
```
d3zd3z: This brace should be indented to line up with the `extern "C" {` above, so 4 spaces.

sim/mcuboot-sys/build.rs
utzig: fixed

sim/mcuboot-sys/src/c.rs
utzig: fixed

utzig:
Btw, I also dropped `std=c99` and squashed the review issues into the suitable commits.

boot/zephyr/include/config-boot.h
```diff
@@ -35,8 +35,9 @@
 #endif
 
 /* TODO: Configure this between app and target.  Really, we want the
- * config to come from the app. */
+ * config to come from the app.
 #define CONFIG_BOOT_VERIFY_RSA_SIGNATURE
+ */
```
d3zd3z: Let's just remove all of these lines

sim/mcuboot-sys/build.rs
```diff
@@ -24,6 +24,7 @@ fn main() {
     }
 
     if sig_rsa {
+        conf.define("CONFIG_BOOT_VERIFY_RSA_SIGNATURE", None);
```
d3zd3z: Let's get rid of this define entire. Instead, change the use of it in the code to MCUBOOT_SIGN_RSA.

sim/mcuboot-sys/build.rs
```diff
@@ -83,6 +83,10 @@ fn main() {
     conf.debug(true);
     conf.flag("-Wall");
     conf.flag("-Werror");
+    /// FIXME: travis-ci still uses gcc 4.8.4 which defaults to std=gnu90.
```
d3zd3z: These comments need to be with just two slashes "//". "///" are doc comments that go only before certain declarations. This generates a warning on the nightly compiler.

utzig:
@mbolivar 

> I'm still seeing 11d654f ("Build in std=c99 mode") in this series, so it doesn't look like that got "dropped".

Yes, that's rigtht. I dropped it but then travis started failing, because it is still using gcc 4.8.4 for our setup, and it defaults to `std=gnu90`, so I had to add it back. Maybe after 1.1 is released we can take some time to update travis and drop this for good.

mbolivar:
> Yes, that's rigtht. I dropped it but then travis started failing, because it is still using gcc 4.8.4 for our setup, and it defaults to std=gnu90, so I had to add it back. Maybe after 1.1 is released we can take some time to update travis and drop this for good.

Ah, OK, I get it. Thanks for explaining.

sim/mcuboot-sys/build.rs
utzig: Fixed.

sim/mcuboot-sys/build.rs
utzig: Done.

utzig:
@d3zd3z anything remaining?

PR #162, opened 2017-12-05
Add mbed asn1 parser

utzig:
WIP - This is done but needs rebase on #161.

This bundles the asn1 parser from mbed-tls into mcuboot, which allows adding EC crypto (tinycrypt based) functionality for target OSes that don't bundle mbed-tls.

JIRA: MCUB-76


utzig:
@mbolivar @d3zd3z Commit a926a1a adds the bundled ASN1 parser support for Zephyr (breaks the requirement on Zephyr's embedded mbed-tls). I tested Zephyr (as well as Mynewt) with both RSA and EC256 to check nothing was broken, but I am not well versed on CMake as well as the new Zephyr build system, so not sure the way I did it is the best way.

NOTE: This should be easier to review once #161 is merged (rebasing this one).

sim/mcuboot-sys/build.rs
```diff
@@ -53,8 +53,7 @@ fn main() {
         conf.file("../../ext/tinycrypt/lib/source/ecc_dsa.c");
         conf.file("../../ext/tinycrypt/lib/source/ecc_platform_specific.c");
 
-        conf.file("mbedtls/library/bignum.c");
-        conf.file("mbedtls/library/asn1parse.c");
+        conf.file("../../ext/mbedtls/src/asn1parse.c");
```
d3zd3z: Changing this directory needs to be accompanied by a similar change near the end of the function, where it calls `walk_dir()`, otherwise changes in these files won't cause rebuilds.

utzig:
@d3zd3z I squashed the 3 mynewt related commits in to one, dropped the `CONFIG_*RSA*` and added  the missing `walk_dir`!

sim/mcuboot-sys/build.rs
```diff
@@ -92,6 +91,7 @@ fn main() {
 
     walk_dir("../../boot").unwrap();
     walk_dir("../../ext/tinycrypt/lib/source").unwrap();
+    walk_dir("../../ext/mbedtls/src").unwrap();
```
d3zd3z: This shouldn't go into 'src' since it will miss header files. Just call
```
walk_dir("../../ext/mbedtls").unwrap()
```
The tinycrypt one is broken as well, but can be fixed in a different series.

ext/mbedtls/pkg.yml
```diff
@@ -0,0 +1,24 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
```
d3zd3z: There is a stray space at the end of this line.

utzig: done. this is a copy/paste, there are other pkg.yml all around with the same whitespace...

sim/mcuboot-sys/build.rs
utzig: done

utzig:
@d3zd3z Anything still blocking approval?

d3zd3z:
Looks good. We still should fix the `walk_dir()` for tinycrypt, but in a separate patch series.

PR #166, opened 2017-12-08
zephyr: Port mcuboot to Cortex-M0 thargets whitch don't support VTOR in HW

miyatsu:
Some arch need special treatment when jump to slot0, the zephyr already
done that by define CONFIG_IS_BOOTLOADER.

Set it as default in prj.conf and prj-p256.conf.


miyatsu:
I just find this commit have a bug, which I don't know how to fix it.

When I try to make the mcuboot after run "cmake -DBOARD=xxx", the shell output with error message below:

[ 90%] Linking C static library libzephyr.a
[ 94%] Built target zephyr
[ 95%] Linking C executable zephyr_prebuilt.elf
/opt/zephyr-sdk/sysroots/x86_64-pokysdk-linux/usr/libexec/arm-zephyr-eabi/gcc/arm-zephyr-eabi/6.2.0/real-ld: zephyr_prebuilt.elf section `noinit' will not fit in region `SRAM'
/opt/zephyr-sdk/sysroots/x86_64-pokysdk-linux/usr/libexec/arm-zephyr-eabi/gcc/arm-zephyr-eabi/6.2.0/real-ld: section .intList VMA [0000000020030000,00000000200300e3] overlaps section noinit VMA [000000002002e6b8,00000000200316b7]
/opt/zephyr-sdk/sysroots/x86_64-pokysdk-linux/usr/libexec/arm-zephyr-eabi/gcc/arm-zephyr-eabi/6.2.0/real-ld: region `SRAM' overflowed by 6280 bytes
collect2: error: ld returned 1 exit status
zephyr/CMakeFiles/zephyr_prebuilt.dir/build.make:101: recipe for target 'zephyr/zephyr_prebuilt.elf' failed
make[2]: *** [zephyr/zephyr_prebuilt.elf] Error 1
CMakeFiles/Makefile2:566: recipe for target 'zephyr/CMakeFiles/zephyr_prebuilt.dir/all' failed
make[1]: *** [zephyr/CMakeFiles/zephyr_prebuilt.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

When I discard this commit, and run "make menuconfig" to re-enable CONFIG_IS_BOOTLOADER manually, then run "make", everything is fine without any error.

I've tryed add CONFIG_BOOTLOADER_SRAM_SIZE=0 and CONFIG_BOOTLOADER_SRAM_SIZE=CONFIG_SRAM_SIZE, the error output is still there.

Some times the 'noinit' will be replaced with ‘bss’.

For novice programmer (me), I think it may need you guys to check this out. Forgive me for my poor English.

miyatsu:
Hi @mbolivar 
Oh, I forgot that. Sorry..
You can check one of my PR in zephyr repo first, all my work are explained here.
https://github.com/zephyrproject-rtos/zephyr/pull/4536

mbolivar:
Hi @miyatsu, thanks for your reply! I read through the comment history in Zephyr pull request https://github.com/zephyrproject-rtos/zephyr/pull/4536.

It seems like what you are trying to do is support mcuboot on Cortex M0 targets, where the vector table is fixed at address 0x00000000 because there is no VTOR register.

It also seems like @nvlsianpu has suggested that you implement it like this:

1. Have mcuboot remember the address of the second image's vector table, which mcuboot computes on ARM targets in do_boot().

2. Before mcuboot switches to the child image, configure the mcuboot IRQ handlers (which are the ones installed at the fixed vector table address, since mcuboot has to run first) to "forward" to the second image's IRQ table. "Forward" means getting the Zephyr IRQ handler addresses from the vector table and calling them.

3. Jump to the child by setting the stack pointer to vt->msp and calling vt->reset.

4. When an interrupt is received now, the mcuboot IRQ handler calls the Zephyr IRQ handler instead.

Do I understand correctly what you are trying to do?

If so (and I'm just guessing here since I don't see anyone saying this), then I think you are trying to enable CONFIG_IS_BOOTLOADER so that the mcuboot IRQ handlers configured in step 2 and called in step 4 can access global variables defined by mcuboot in SRAM even after calling the child in step 3.

Is that right?

If so, I think that CONFIG_IS_BOOTLOADER should only be selected when it is really needed -- that is, when CONFIG_ARMV6_M=y. Perhaps you can add something like this (untested) to the top-level Kconfig:

```
if ARMV6_M

config IS_BOOTLOADER
	default y

endif
```

You are going to need to conditionally compile this code no matter what, because the extra forwarding is not necessary on other targets.

Then you will need to figure out how big you need to set CONFIG_BOOTLOADER_SRAM_SIZE. You can do this by compiling mcuboot and using the linker errors you have already seen to pick a better value. You can add that default setting to the Kconfig as well.

Make sense?

Thanks for your efforts to port mcuboot to M0!

mbolivar:
> Then you will need to figure out how big you need to set CONFIG_BOOTLOADER_SRAM_SIZE. You can do this by compiling mcuboot and using the linker errors you have already seen to pick a better value. You can add that default setting to the Kconfig as well.

After that you'll need to add the actual code which forward the IRQs and test it all out, of course :).

miyatsu:
Lunch time~, I will read your comment layer.

Actually I'm now using STM32F4 on my daily work recently, probably don't have much free time to finish that PR. But, my boss once promised me when I finish all the office job in my hand, he will give me a few days free time to concentrate on porting mcuboot to ARM Cortex-M0. I hope I can do it fast, hahah.

mbolivar:
> But, my boss once promised me when I finish all the office job in my hand, he will give me a few days free time to concentrate on porting mcuboot to ARM Cortex-M0. I hope I can do it fast, hahah.

Good luck! If any of my comments were unclear, please let me know.

Also pinging @nvlsianpu and @carlescufi on this as an FYI.

nvlsianpu:
I'm agree with remarks for @mbolivar. Whatever this PR will change, it must waiting once corresponding support will be available in zephyr (mean zephyrproject-rtos/zephyr#4536 or other solution).

d3zd3z:
So, one question I have is what happens on reset? Is the SRAM cleared, or is there some other way for the handler to know to not forward the interrupts. This is important when the image is being upgraded and power is lost or a reset happens, that we don't try jumping into the corrupt image.

carlescufi:
This should instead use `CONFIG_IS_MCUBOOT`, which is required to enable certain features in Zephyr whenever we're building Zephyr as MCUboot. I have fixed the SRAM issues there: https://github.com/zephyrproject-rtos/zephyr/pull/5789

carlescufi:
I think we can close this in favor of https://github.com/runtimeco/mcuboot/pull/183

PR #167, opened 2017-12-08
zephyr: samples: Fix possible unbootable without bootloader

miyatsu:
It may cause the text section linked into wrong place when the
application do not support bootloader.

### DIFF
--- a/samples/zephyr/hello-world/dts.overlay
+++ b/samples/zephyr/hello-world/dts.overlay
@@ -5,6 +5,8 @@
 
 / {
 	chosen {
+#if defined(CONFIG_IMG_MANAGER)
 		zephyr,code-partition = &slot0_partition;
+#endif
 	};
 };


mbolivar:
Just to be clear, I think unless I'm misunderstanding your purpose that this pull request should be closed.

If you're looking for a hello world sample that does not assume it's booted by mcuboot, please use the one in the Zephyr tree: https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/hello_world

miyatsu:
Hi @mbolivar thank you for the comment.

I'm try to run mcuboot on my application, but, sometimes I don't want the application to support bootloader. For test purpose or some other reasons I guess?.

To do that I can simply disable the CONFIG_IMG_MANAGER in Kconfig or application config file. However, the dts.overlay file is still working, which will make the application code to be load at address of SLOT0 at link stage (defined in board spec .dts) rather than load the application to the start of flash address.

mbolivar:
While I can understand why that is convenient for you, it makes the application itself more complicated for everyone else to read.

One of the main reasons we include this application is to provide the smallest example we can for starting an application which uses mcuboot. The application itself doesn't actually do anything useful. It just prints hello world. So I don't think there's a good reason to add extra features like "boots directly without mcuboot", because the application isn't doing any other useful things except being loaded by mcuboot.

If you want to use an application which isn't loaded by mcuboot, then you can use the hello world which is already provided by Zephyr and which I linked to above.

So I'm sorry but this is not a patch which mcuboot should accept in my opinion.

miyatsu:
Haha, you are right about it. I'm noticed that the dtc support conditional compilation, it is convenient for me to flash application without write any other scripts by add this commit.

Thank you take your time to review this.

PR #168, opened 2017-12-12
Describe the tradeoffs of scratch size

utzig:
JIRA: MCUB-74

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -138,7 +138,45 @@ upgrades, but must be configured at build time to choose one of these two
 strategies.
 
 In addition to the two image slots, the boot loader requires a scratch area to
-allow for reliable image swapping.
+allow for reliable image swapping. The scratch area must have a size that is
+enough to store at least the largest sector that is going to be swapped. Many
+devices have small equally sized flash sectors, eg 4K, while others have variable
+sized sectors where the largest sectors might be 128K or 256K, so the scratch
+must be big enough to store that. The scratch is only ever used when swapping
+firmware, which means only when doing an upgrade. Given that, the main reason
+for using a larger size for the scratch is that flash wear will be more evenly
+distributed, because a single sector would be written twice the number of times
+than using two sectors, for example. To evaluate the ideal size of the scratch
+for your use case the following parameters are relevant:
+
+* the ratio of image size / scratch size
+* the number of erase cycles supported by the flash hardware
+
+The image size is used (instead of slot size) because only the slot's sectors
+that are actually used for storing the image are copied. The image/scratch ratio
+is the number of times the scratch will be erased on every upgrade. The number
+of erase cycles divided by the image/scratch ratio will give you the number of
+times an upgrade can be performed before the device goes out of spec.
+
+```
+num_upgrades = number_of_erase_cycles / (image_size / scratch_size)
+```
+
+Let's assume, for example, a device with 10000 erase cycles, an image size of
+150K and a scratch of 4K (usual minimum size of 4K sector devices). This would
+result in a total of:
+
+`10000 / (150 / 4) ~ 267`
+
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -138,7 +138,17 @@ upgrades, but must be configured at build time to choose one of these two
...
-allow for reliable image swapping.
+allow for reliable image swapping. The scratch area must have a size that is
+enough to store at least the largest sector that is going to be swapped. Many
+devices have small linear sized flash sectors, eg 4K, while others have variable
+sized sectors where the largest sectors might be 128K or 256K, so the scratch
+must be big enough to store that. The scratch is only ever used when swapping
+firmware, which means only when doing an upgrade. Given that, there are two main
+reasons for using a larger size for the scratch. One is that the speed of an
+upgrade is proportional to the size, because larger size allows for less swap
+operations. Second is that flash wear will be more evenly distributed, because
```
mbolivar: Overall I like this change. I think that it would be an improvement to quantify the wear on the flash. People's biggest question is always "is this big enough", and a quantitative description in terms of how much it will reduce the lifetime of the flash will help them make their own decisions.

Here is an example where I described the tradeoff for nRF5:

https://github.com/zephyrproject-rtos/zephyr/pull/1307#discussion_r136318044

Do you think you can include something like that, with concrete numbers or a formula?

utzig: Done!

docs/design.md
```diff
@@ -138,7 +138,47 @@ upgrades, but must be configured at build time to choose one of these two
...
+sectors for example. To evaluate the ideal size of the scratch for your use
+case the following parameters are relevant:
+
+* the ratio of image size / scratch size
+* the number of erase cycles supported by the flash hardware
+
+The image size is used (instead of slot size) because only the slot's sectors
+that are actually used for storing the image are copied. The image/scratch ratio
+is the amount of times the scratch will be erased on every upgrade. The number
+of erase cycles divided by the image/scratch rate will give you the amount of
```
mbolivar: s/rate/ratio/

mbolivar: I'll pile on "number of times" here and on the previous line, instead of "amount of times" since there is already a typo to fix

docs/design.md
```diff
@@ -138,7 +138,47 @@ upgrades, but must be configured at build time to choose one of these two
...
+150K and a scratch of 4K (usual minimum size of 4K sector devices). This would
+result in a total of:
+
+`10000 / (150 / 4) ~ 267`
+
+Increasing the scratch to 16K would give us:
+
+`10000 / (150 / 16) ~ 1067`
+
+There is no *best* ratio, as the right is use-case dependent. Factors to
```
mbolivar: "right size"

docs/design.md
```diff
@@ -138,7 +138,47 @@ upgrades, but must be configured at build time to choose one of these two
...
 
 In addition to the two image slots, the boot loader requires a scratch area to
-allow for reliable image swapping.
+allow for reliable image swapping. The scratch area must have a size that is
+enough to store at least the largest sector that is going to be swapped. Many
+devices have small equally sized flash sectors, eg 4K, while others have variable
+sized sectors where the largest sectors might be 128K or 256K, so the scratch
+must be big enough to store that. The scratch is only ever used when swapping
+firmware, which means only when doing an upgrade. Given that, there are two main
+reasons for using a larger size for the scratch. One is that the speed of an
```
d3zd3z: Is this true, though?  Assuming an image that is decently larger than the swap, the same amount of flash is going to be erased and written, whether the swap is 1 sector or more than one. Yes there are fewer swap operations with a larger scratch, but those operations take longer, since the speed concerns (write and erase time) are proportional to the size of the area written or erased.

utzig: You're right, fixed!

PR #175, opened 2017-12-27
Correct the path of zephyr sample dts file

easonxiang:
frdm_k64f.dts has been moved to boards/arm/frdm_k64f/

Signed-off-by: Dong Xiang <dong.xiang@linaro.org>


utzig:
Thanks for the fix!

PR #183, opened 2018-01-23
zephyr: Set IS_MCUBOOT for Zephyr builds

carlescufi:
Enable the new IS_MCUBOOT Kconfig option on Zephyr builds and move the Kconfig file to boot/zephyr


nvlsianpu:
LGTM

utzig:
Yup, looks good to me, but I will add some zephyr reviewers to confirm!

carlescufi:
Would it be better instead to do `select IS_MCUBOOT`
in the root `Kconfig` ?

nvlsianpu:
I think yes - Advantage would be that it will be forced to be selected.

carlescufi:
See https://github.com/zephyrproject-rtos/zephyr/pull/5789

carlescufi:
@utzig @nvlsianpu I've replaced it with a "mandatory" `MCUBOOT` option

carlescufi:
I'm going to close this because it adds a circular dependency between Zephyr and MCUboot, and I would like to avoid that

PR #189, opened 2018-01-25
kconfig: Make MCUBOOT_VALIDATE_SLOT0 configurable through Kconfig

SebastianBoe:
RFC for approach to resolve https://github.com/runtimeco/mcuboot/issues/186

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -37,7 +37,15 @@
 #define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
 #include "bootutil/bootutil_log.h"
 
-#ifdef MCUBOOT_MYNEWT
+#if defined(USE_MCUBOOT_CONFIG)
+#include <mcuboot_config.h>
+#endif
+
+// For legacy-reasons MYNEWT uses the define 'MCUBOOT_MYNEWT' and the
+// header 'mcuboot_config/mcuboot_config.h' to specify it's
+// mcuboot_config.h file. Eventually it will use the OS-agnostic
+// 'USE_MCUBOOT_CONFIG' define and 'mcuboot_config.h' header.
+#if defined(MCUBOOT_MYNEWT)
 #include "mcuboot_config/mcuboot_config.h"
 #endif
 
--- a/boot/zephyr/include/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config.h
@@ -1,0 +1,22 @@
+#ifndef H_MCUBOOT_CONFIG_
+#define H_MCUBOOT_CONFIG_
+
+// This header file acts as an adapter layer from Kconfig-style
+//
+//   '#define CONFIG_foo'
+//
+// to MCUBoot-style
+//
+//   '#define foo'
+//
+// macros. This file is manually maintained for now.
+//
+// The adapter layer keeps MCUBoot semi-agnostic to Zephyr and
+// MyNewt's configuration systems. See MyNewt's adapter in it's own
+// mcuboot_config.h file.
+
+#ifdef CONFIG_MCUBOOT_VALIDATE_SLOT0
+#define MCUBOOT_VALIDATE_SLOT0
+#endif
+
+#endif /* H_MCUBOOT_CONFIG_ */
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -26,22 +26,6 @@ if (NOT DEFINED CONF_SIGNATURE_TYPE)
   set(CONF_SIGNATURE_TYPE RSA)
 endif()
 
-# If CONF_VALIDATE_SLOT0 is set, the bootloader attempts to validate
-# the signature of slot0 every boot.  This adds the signature check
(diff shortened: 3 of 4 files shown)


boot/bootutil/src/loader.c
```diff
@@ -43,7 +43,7 @@
 
 static struct boot_loader_state boot_data;
 
-#if defined(MCUBOOT_VALIDATE_SLOT0) && !defined(MCUBOOT_OVERWRITE_ONLY)
+#if defined(CONFIG_MCUBOOT_VALIDATE_SLOT0) && !defined(MCUBOOT_OVERWRITE_ONLY)
```
mbolivar: The problem with this approach is that bootutil is the common library that implements mcuboot functionality, and cannot contain any Zephyr-specific code. (The Zephyr bootloader app is in boot/zephyr/, and you've already seen the Zephyr samples/zephyr).

SebastianBoe:
> Can you change this so the Zephyr CONFIG values cause the existing MCUBOOT_xxx defines to be made in bootutil instead?

I can change this so that boot/bootutil/src/loader.c uses MCUBOOT_xxx defines by introducing an adapter layer that does

#define MCUBOOT_xxx CONFIG_MCUBOOT_xxx

not sure where the adapter layer should be plugged in, perhaps akin to here: https://github.com/SebastianBoe/mcuboot/blob/10155f47524ff6877eb0ef7499620ba693de79c7/boot/bootutil/src/loader.c#L41

SebastianBoe:
> Can you change this so the Zephyr CONFIG values cause the existing MCUBOOT_xxx defines to be made in bootutil instead?

Done.

boot/zephyr/include/mcuboot_config.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
```
mbolivar: I know this is an RFC, but can you please use include guards for the real submission?

boot/bootutil/src/loader.c
```diff
@@ -37,8 +37,10 @@
 #define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
 #include "bootutil/bootutil_log.h"
 
-#ifdef MCUBOOT_MYNEWT
+#if defined(MCUBOOT_MYNEWT)
```
mbolivar: I like the approach, but I don't like having the mynewt OS specific stuff in here and would rather not add more for Zephyr, especially since we need to remember to include it in every file that needs it.

Can you add a `-include mcuboot_config.h` to the compiler invocations for Zephyr?

With that change, I like the approach and would have no objections.

SebastianBoe: I wouldn't recommend using "-include".

The problem with "-include" is that it complicates the build, this complexity again causes:

Extra build script code to be needed.

IDE's and other build system SW to struggle (E.g. CLion did not support -include for a while, and CMake will remove -include flags when there are multiple of them)

Programmers can no longer manually follow #include's to find the preprocessor macros used.

To avoid OS-specific code in here one thing one could do is something like this:

```
#if defined(USE_MCUBOOT_CONFIG)
#include <mcuboot_config.h>
#endif
```

But I will use -include if the project maintainers want it.

carlescufi:
LGTM, @mbolivar are you OK with this?

carlescufi:
Note that I've moved `Kconfig` to `boot/zephyr` in another PR

boot/bootutil/src/loader.c
mbolivar: Interesting; I am surprised to learn this. Since Zephyr uses `-include` and `-imacros` to pull in `${AUTOCONF_H}`, I assumed there was no problem with it. Is this something that will be addressed in Zephyr eventually?

Anyway, I'm happy to defer to your experience here and use `USE_MCUBOOT_CONFIG`. If mynewt wants to switch to it later, then the OS-specific code here for that can be removed.

mbolivar:
One more thing -- can you please keep samples/zephyr/Makefile up to date in the non-RFC version? It uses the CONF_ variables to configure the build in various situations which are used in release testing.

See docs/testplan-zephyr.md for more details (and you can try samples/zephyr/run-tests.sh to check that things are working, though the Makefile is frdm_k64f specific at the moment).

mbolivar:
@SebastianBoe BTW, I've got a new config option to mass-erase flash in my queue to send upstream:

https://github.com/mbolivar/mcuboot/tree/mass-erase

I'll wait until after this is merged and re-do it in terms of Kconfig if you think you'll have time to work on this more soon.

boot/bootutil/src/loader.c
SebastianBoe: > I assumed there was no problem with it. Is this something that will be addressed in Zephyr eventually?

Yes, it's on my TODO list to get rid of it.

SebastianBoe:
@mbolivar : Thank you for your comments.

It is good to know for the future of MCUBoot-Zephyr integration that the approach
is OK.

Unfortunately I don't have the time to take this from an RFC to a PR, nor do I know when I will have the time. Perhaps some time later this year.

Closing the RFC as accepted.

PR #192, opened 2018-01-29
zephyr: boot: Use abs. path instead of rel. path for dts overlay

### LINKED ISSUE #187: The build fails when CMake is invoked from outside of the build directory
Hi,

I am observing that 

`mkdir build && cd build && cmake ..`

works.

But 

`cmake -Bbuild -H.`

doesn't for mcuboot/boot/zephyr

SebastianBoe:
When a relative path was used the build would fail when CMake was
invoked with the -B -H flags. Using an absolute path for the
dts.overlay flags fixes #187.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -118,7 +118,7 @@ endif()
 
 # This is necessary to ensure mcuboot is linked into, and fits inside,
 # the boot partition.
-set(DTC_OVERLAY_FILE dts.overlay)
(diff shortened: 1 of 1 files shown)


carlescufi:
LGTM, I assume this is for the .dts mix-in drop folder?

SebastianBoe:
> LGTM, I assume this is for the .dts mix-in drop folder?

No, that happens to be unrelated. This is to fix #187

carlescufi:
>No, that happens to be unrelated. This is to fix #187

Oh I see. Baader-Meinhof effect I guess.

PR #197, opened 2018-01-30
zephyr: try to mass erase when flashing

mbolivar:
It's recommended practice to mass erase the flash chip before
reflashing mcuboot. Add a configuration option for Zephyr which does
this by default on some Zephyr runner backends which can support
this (pyOCD and dfu-util). Additional runners can be added as needed (and possible).

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -57,6 +57,24 @@ if (NOT DEFINED CONF_UPGRADE_ONLY)
   set(CONF_UPGRADE_ONLY NO)
 endif()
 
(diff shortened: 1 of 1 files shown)


mbolivar:
@carlescufi @SebastianBoe FYI.

Being able to do something similar for nrfjprog would be nice. However, I'm not sure how to use nrfjprog to flash the signed binary at a random offset afterwards, so I haven't attempted using a board based on it with mcuboot in an upstreamable way.

I do have a Python script based on pynrfjprog that uses pynrfjprog.API.API.erase_page() in a loop over the range needed by the signed binary with hard-coded sector sizes, but that won't go upstream. Any input on the best way to handle this?

PR #198, opened 2018-01-30
requirements.txt and imgtool .hex support 

carlescufi:


carlescufi:
@mbolivar @d3zd3z @utzig @nvlsianpu FYI

scripts/imgtool/image.py
```diff
@@ -69,8 +72,17 @@ class Image():
     @classmethod
     def load(cls, path, included_header=False, **kwargs):
         """Load an image from a given file"""
-        with open(path, 'rb') as f:
-            payload = f.read()
+        ext = os.path.splitext(path)[1][1:]
```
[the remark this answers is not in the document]
carlescufi: Actually now that you mention this in addition to your comment I need to handle the case where the filename has no extension.

scripts/imgtool/image.py
```diff
@@ -69,8 +72,17 @@ class Image():
...
+        if ext == "hex":
```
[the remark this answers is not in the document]
carlescufi: Yes, I will create an `INTEL_HEX_EXTENSION`

scripts/imgtool/image.py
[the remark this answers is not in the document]
utzig: `os.path.splitext` returns an empty string for no extension so it should work... (I hope!)

scripts/imgtool/image.py
```diff
@@ -69,8 +72,17 @@ class Image():
...
+            ih = IntelHex(path)
```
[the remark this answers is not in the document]
carlescufi: I guess it relies on `open` throwing an exception: https://github.com/bialix/intelhex/blob/master/intelhex/__init__.py#L199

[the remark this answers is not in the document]
carlescufi: Yes, `open` will throw a `FileNotFoundError` if there's no file there. Regarding invalid contents, it will raise an exception itself: https://github.com/bialix/intelhex/blob/master/intelhex/__init__.py#L118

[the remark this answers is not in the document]
utzig: And for file format errors:

https://github.com/bialix/intelhex/blob/master/intelhex/__init__.py#L1246

Maybe catching `ValueError` and `IntelHexError` would be enough handling but not sure what's the best way to do it. (I personally would catch those and print the exception `.msg`).

@d3zd3z Suggestions?

scripts/imgtool/image.py
```diff
@@ -69,8 +73,17 @@ class Image():
     @classmethod
     def load(cls, path, included_header=False, **kwargs):
         """Load an image from a given file"""
-        with open(path, 'rb') as f:
-            payload = f.read()
+        ext = os.path.splitext(path)[1][1:].lower()
```
carlescufi: This works fine if the filename has no extension, since `ext` will be an empty string

carlescufi:
@utzig comments so far addressed

scripts/imgtool/image.py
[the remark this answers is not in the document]
carlescufi: I don't think we catch exceptions consistently in the rest of the tool?

scripts/imgtool/image.py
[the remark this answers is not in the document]
carlescufi: Yes, it does. I just tested it :) We will of course default to binary if there's no extension

utzig:
I would prefer to not throw an exception under normal use, even though there are many places in the tool that are doing that right now, we should go in the direction of doing less of it!

scripts/requirements.txt
```diff
@@ -0,0 +1,2 @@
+cryptography
+intelhex
```
mbolivar: Nitpick: this should go into the next patch ideally. It's fine by me to merge as-is though.

scripts/imgtool/image.py
```diff
@@ -81,21 +94,30 @@ def load(cls, path, included_header=False, **kwargs):
         obj.check()
         return obj
 
-    def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE, pad=0):
+    def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE, pad=0,
+                 hex_base_addr=None):
```
mbolivar: Why is this `hex_base_addr`? You might imagine loading the base address from an ELF image too. Is there a reason for this not to just be `base_addr`, and have `load()` fill it in only for `hex` right now?

scripts/imgtool/image.py
```diff
@@ -81,21 +94,30 @@ def load(cls, path, included_header=False, **kwargs):
...
                 self.version,
                 self.header_size,
+                self.hex_base_addr or "N/A",
                 self.pad,
                 len(self.payload))
 
     def save(self, path):
-        with open(path, 'wb') as f:
-            f.write(self.payload)
+        if self.hex_base_addr:
```
mbolivar: This also doesn't seem so future-proof - could we have a "format" parameter to the constructor that defaults to "bin" and can be "hex" instead? That can be saved as an attribute and used here.

scripts/imgtool/image.py
carlescufi: Agreed!

scripts/imgtool/image.py
carlescufi: Yep, that is better than what I have now

scripts/requirements.txt
carlescufi: Done

mbolivar: I'm still seeing it in this same patch... ?

carlescufi: Sorry, "done locally". Pushed now with all your comments addressed.

scripts/imgtool/image.py
d3zd3z: It does rely on [1:] in python not failing on an empty string (which was surprising to me).

scripts/imgtool.py
```diff
@@ -1,5 +1,6 @@
 #! /usr/bin/env python3
 #
+# Copyright 2018 Nordic Semiconductor ASA
```
d3zd3z: You can't really justify adding a copyright statement to a file you aren't otherwise modifying.

scripts/imgtool/image.py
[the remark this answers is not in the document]
d3zd3z: We could catch exceptions at the top level, and try to print something friendly. Honestly, I hate it when programs do that, though, it makes them difficult to debug. I'd much rather just see the exception, even with the stack trace, for something like a corrupt file.

Now "normal" things like a file not found, I think we could catch and print a nice message over.

scripts/imgtool/image.py
```diff
@@ -69,8 +74,18 @@ class Image():
     @classmethod
     def load(cls, path, included_header=False, **kwargs):
         """Load an image from a given file"""
-        with open(path, 'rb') as f:
-            payload = f.read()
+        ext = os.path.splitext(path)[1][1:].lower()
+        if ext == INTEL_HEX_EXT:
+            ih = IntelHex(path)
+            payload = ih.tobinarray()
+            segments = ih.segments()
```
d3zd3z: Any chance that the linker might actually generate multiple segments? What about code and initialized data?

scripts/imgtool/image.py
```diff
@@ -81,21 +96,32 @@ def load(cls, path, included_header=False, **kwargs):
...
         return obj
 
-    def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE, pad=0):
+    def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE, pad=0,
+                 base_addr=None, format=BIN_EXT):
         self.version = version or versmod.decode_version("0")
         self.header_size = header_size or IMAGE_HEADER_SIZE
+        self.base_addr = base_addr
         self.pad = pad
+        self.format = format
```
d3zd3z: Instead of setting and then checking a string, why not create two save methods, one for bin, and one for hex, and set self format to the desired method. self.save would then just invoke self.format(self, path)?

scripts/requirements.txt
d3zd3z: Why would it go in the next patch? This is the patch that needs intelhex in order to run?

scripts/imgtool/image.py
mbolivar: This is legit:

```
The slice of s from i to j is defined as the sequence of items with index k such that i <= k < j. If i or j is greater than len(s), use len(s).
```

https://docs.python.org/3/library/stdtypes.html#typesseq-common

scripts/requirements.txt
mbolivar: @d3zd3z you're looking at the corrected revision of the series. It was previously in the patch that introduced requirements.txt.

scripts/imgtool/image.py
mbolivar: That seems equivalently future-proof but I have no objections to it

scripts/imgtool/image.py
nvlsianpu: Is it hard to support multiple segments? I don't know about any case which will require multiple segments - but i can imagine that a zephyr ARMv8 future target might  generates separate segments: 1st for secure and 2nd for non-secure part

scripts/imgtool.py
carlescufi: Yes, that was an oversight. Will remove.

scripts/imgtool/image.py
carlescufi: I have never seen multiple segments before in .hex files. Signing them separately would require deciding first on a format for multiple images and their signatures, which is outside the scope of this PR. What I can do is fail instead of warn because multiple segments at this point are unhandled.

scripts/imgtool/image.py
carlescufi: I can do that yes. I don't have a strong preference for either.

scripts/imgtool/image.py
nvlsianpu: understand this - I seen multiple segments many times - but it was for complex systems with extensive usage of MPU. 
I agree - this is out of scope.

scripts/imgtool/image.py
carlescufi: Might as well do it properly and use inheritance

carlescufi:
@utzig @mbolivar @d3zd3z @nvlsianpu Updated with inheritance and removed the use of segments, using `minaddr()` instead which is a much better fit.

scripts/requirements.txt
d3zd3z: Ahh, thanks.

PR #227, opened 2018-02-06
zephyr: Fix type mismatch warning when link with new libc

miyatsu:
Fix this by add explicit cast.

### DIFF
--- a/boot/zephyr/flash_map.c
+++ b/boot/zephyr/flash_map.c
@@ -254,7 +254,7 @@ struct layout_data {
  * flash_area_get_sectors(). A lot of this can be inlined once
  * flash_area_to_sectors() is removed.
  */
-static int flash_area_layout(int idx, int *cnt, void *ret,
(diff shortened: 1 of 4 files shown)


boot/zephyr/flash_map.c
```diff
@@ -355,6 +355,6 @@ int flash_area_get_sectors(int idx, uint32_t *cnt, struct flash_sector *ret)
 {
     struct layout_data data;
 
-    return flash_area_layout(idx, cnt, ret, get_sectors_cb, &data);
+    return flash_area_layout(idx, (int*)cnt, ret, get_sectors_cb, &data);
```
mbolivar: Can you add the cast in flash_area_to_sectors(), and change the prototype of flash_area_layout()? That would be cleaner since flash_area_to_sectors() is the "old style" routine, and struct layout_data already wants an unsinged value to assign to ret_len and pull from ret_idx.

PR #228, opened 2018-02-06
zephyr: mbedtls: Fix compile error no such file or directory

miyatsu:
Upstream zephyr will automatically link mbedtls when CONFIG_MBEDTLS is
enabled.

If user chose to use ECDSA_P256 as signature type, will get a compile
error message in file ${ZEPHYR_BASE}/ext/lib/crypto/mbedtls/zephyr_init.c:
    fatal error: config-asn1.h: No such file or directory

Fix this by disable CONFIG_MBEDTLS  when ECDSA_P256 signature type is
selected.


SebastianBoe:
Hi, I don't understand this fix.

I don't think it makes sense to set CONFIG_MBEDTLS=n and keep other CONFIG_MBEDTLS options in the project configuration.

If the intention is to not use MBEDTLS, couldn't one just delete these lines?

CONFIG_MBEDTLS=y
CONFIG_MBEDTLS_BUILTIN=n		  
CONFIG_MBEDTLS_CFG_FILE="config-asn1.h"

Or are we trying to do something else here?

SebastianBoe:
I can confirm that deleting the three lines of MBEDTLS config also resolves the issue.

I don't understand what they were doing there in the first place though, so I don't know if it is safe to remove them.

miyatsu:
@SebastianBoe Yes, I've double checked just now. Its working well without all three lines.

miyatsu:
@SebastianBoe CONFIG_MBEDTLS_CFG_FILE="config-asn1.h" may needed. MCUboot use this line to config its own mbedtls.

utzig:
@SebastianBoe We have only bundled the asn1 parser from mbedTLS in MCUBoot. What the configuration was trying to achieve was to add this external (from the perspective of Zephyr) lib with a very simple config (`config-asn1.h`) that only enables the asn1parser. I don't think removing those lines would be the correct way to fix the issue. The error seems to be related to a missing include directory. It fails here:

https://github.com/zephyrproject-rtos/zephyr/blob/master/ext/lihttps://github.com/zephyrproject-rtos/zephyr/blob/master/ext/lib/crypto/mbedtls/zephyr_init.c#L19b/crypto/mbedtls/zephyr_init.c#L19

And the directory that includes `config-asn1.h` is added here:

https://github.com/runtimeco/mcuboot/blob/master/boot/zephyr/CMakeLists.txt#L209

I don't really know much about CMake so not sure yet why this includes are not being used by Zephyr's external mbedtls...

utzig:
@SebastianBoe I tried making a little patch for the Zephyr mbedtls cmake build which would allow for our current build system in mcuboot. Do you think this would be acceptable:

https://github.com/zephyrproject-rtos/zephyr/compare/master...utzig:allow-custom-external-mbedtls?expand=1

PS: I have not tried adding `-DMBEDTLS_CFG_FILE` but that would be the next approach...

SebastianBoe:
Not as-is.

You would need to also expand the Kconfig sources to allow users to be able to select an app-provided MBEDtls implementation.

```
choice
	prompt "Select implementation"

```

You can't piggy-back on MBEDTLS_BUILTIN or MBEDTLS_LIBRARY as they already have well-defined semantics.

There might be more work needed as well if you wanted to go down that path. Would need to do some more review work to find all the corners ...

utzig:
@SebastianBoe I don't think I want to follow that path but thanks for the explanations. @miyatsu Just tested your proposed changes and they work OK for me so merging this (sorry for the delay!). Btw, there are some warning because `calloc/free` are not found when building (used by `mbedtls_(calloc|free)`). Where those warnings already there before?

PR #231, opened 2018-02-12
Change all doc links to use relative names

utzig:
Github pages assumes that links to files using relative names point to the same location so this should work both when accessing through the github browser and mcuboot.com


carlescufi:
@utzig you can use "Fixes #<Issue>" in the PR description to close an issue automatically

PR #238, opened 2018-02-15
bootutil: loader: fix BOOT_STATUS_ASSERT macro

jfischer-phytec-iot:
BOOT_STATUS_ASSERT increments the variable if the macro argument
is evaluated as true, which is incorrect behavior compared to the
ASSERT macro.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -47,7 +47,7 @@ static struct boot_loader_state boot_data;
 static int boot_status_fails = 0;
 #define BOOT_STATUS_ASSERT(x)                \
     do {                                     \
-        if (x) {                             \
+        if (!(x)) {                          \
             boot_status_fails++;             \
         }                                    \
     } while (0)


boot/bootutil/src/loader.c
```diff
@@ -47,7 +47,7 @@ static struct boot_loader_state boot_data;
 static int boot_status_fails = 0;
 #define BOOT_STATUS_ASSERT(x)                \
     do {                                     \
-        if (x) {                             \
+        if (!x) {                            \
```
utzig: This macro expansion would result in `!rc == 0` under some circumstances, so always use extra parenthesis as in `(!(x))`. This is an interesting catch because it was supposed to break under simulator testing. Although it looks right to me, I will redo some testing of this stuff because the simulator needs to break if it's wrong and AFAIR it was happening before (at least under development).

jfischer-phytec-iot: I came across it while trying out mcuboot/samples/zephyr/run-tests.sh

```MCUBOOT] [INF] main: Starting bootloader
[MCUBOOT] [INF] boot_status_source: Image 0: magic=unset, copy_done=0xff, image_ok=0xff
[MCUBOOT] [INF] boot_status_source: Scratch: magic=unset, copy_done=0xe, image_ok=0xff
[MCUBOOT] [INF] boot_status_source: Boot source: slot 0
[MCUBOOT] [INF] boot_swap_type: Swap type: test
[MCUBOOT] [WRN] boot_copy_image: 3 status write fails performing the swap
[MCUBOOT] [INF] main: Bootloader chainload address offset: 0x20000
[MCUBOOT] [INF] main: Jumping to the first image slot0***** BOOTING ZEPHYR OS v1.10.99 - BUILD: Feb 15 20*
Hello World from hello2 on frdm_k64f!```

PR #244, opened 2018-02-26
Add mynewt logging

utzig:
DO NOT MERGE YET.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1436,7 +1436,7 @@ boot_go(struct boot_rsp *rsp)
      * the magic number on the image is OK.
      */
     if (boot_data.imgs[0].hdr.ih_magic != IMAGE_MAGIC) {
-        BOOT_LOG_ERR("bad image magic 0x%x", boot_data.imgs[0].hdr.ih_magic);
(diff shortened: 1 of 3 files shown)


boot/bootutil/src/loader.c
```diff
@@ -1436,7 +1436,7 @@ boot_go(struct boot_rsp *rsp)
      * the magic number on the image is OK.
      */
     if (boot_data.imgs[0].hdr.ih_magic != IMAGE_MAGIC) {
-        BOOT_LOG_ERR("bad image magic 0x%x", boot_data.imgs[0].hdr.ih_magic);
+        BOOT_LOG_ERR("bad image magic 0x%lx", (unsigned long)boot_data.imgs[0].hdr.ih_magic);
```
d3zd3z: If you want the "official" way (and very ugly), you can include <inttypes.h>, and do:

    BOOT_LOG_ERR("bad image magic 0x%" PRIx32, boot_data.imgs[0].hdr.ih_magic);

although even that assumes the field is defined with `uint32_t`. I don't really have a preference, and either is fairly gross.

utzig:
Finally was able to assure this is completely functional and compatible with `boot_serial`!

PR #249, opened 2018-03-14
boot_serial: Fix missing reponse to image erase command

MariuszSkamra:
MCUmgr prior Image Upload command sends Image Erase command which
response is missing in boot_serial_input handler.
This results in inability to upload the image due to command timeout.
This patch fixes this issue by adding empty response to Image Erase
command.

This solution is loosely based on
commit 6c819b1304a8e360ea256e40860275c5a9184bde
Author: Marko Kiiskila <marko@runtime.io>
Date:   Mon Aug 28 16:41:42 2017 -0700

    boot_serial; access uart driver directly instead of going through console.

from apache-mynewt-core.

Signed-off-by: Mariusz Skamra <mariusz.skamra@codecoup.pl>

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -269,11 +269,16 @@ out:
 }
 
 /*
- * Console echo control. Send empty response, don't do anything.
+ * Console echo control/image erase. Send empty response, don't do anything.
  */
 static void
-bs_echo_ctl(char *buf, int len)
+bs_empty_rsp(char *buf, int len)
 {
+    cbor_encoder_create_map(&bs_root, &bs_rsp, CborIndefiniteLength);
+    cbor_encode_text_stringz(&bs_rsp, "rc");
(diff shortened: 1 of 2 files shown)


carlescufi:
CC @nvlsianpu

utzig:
Code looks good, I just want to give it a try before approval/merge!

nvlsianpu:
Hi @MariuszSkamra.

Probably you are aware of -e flag which cause that newtmgr/mcumgr won't send erase command. So use '>newtmgr image upload -e <image-file.bin>-c <profile>'
From functionality point of view this patch didn't add a value functionality - I think rather this should be  explained in a documentation.

MariuszSkamra:
@nvlsianpu 
I was not aware of it. I checked now, and there's '-e' parameter.
So if image erase command won't be sent using this parameter, this patch is not needed.

nvlsianpu:
Anyway - thank for rising this! I should have added this to the doc at some point.

MariuszSkamra:
it was well hidden ^^

PR #251, opened 2018-03-15
imgtool: Add 'create' alias for 'sign'

### LINKED ISSUE #240: Creating an unsigned image with imgtool requires using the "sign" subcommand
Creating an unsigned image with imgtool requires using the "sign" subcommand. Either a new subcmd needs to be added, or the "sign" subcommand needs to renamed to "create".

d3zd3z:
The imgtool's `sign` command also works without a key, but it doesn't
sign, only appends a hash.  Add a `create` alias to this command so that
this usage makes more sense.

Fixes #240

### DIFF
--- a/scripts/imgtool.py
+++ b/scripts/imgtool.py
@@ -122,8 +122,11 @@ def args():
     getpub.add_argument('-k', '--key', metavar='filename', required=True)
     getpub.add_argument('-l', '--lang', metavar='lang', default='c')
 
-    sign = subs.add_parser('sign', help='Sign an image with a private key')
-    sign.add_argument('-k', '--key', metavar='filename')
+    sign = subs.add_parser('sign',
+            help='Sign an image with a private key (or create an unsigned image)',
+            aliases=['create'])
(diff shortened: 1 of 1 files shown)


scripts/imgtool.py
```diff
@@ -122,7 +122,8 @@ def args():
     getpub.add_argument('-k', '--key', metavar='filename', required=True)
     getpub.add_argument('-l', '--lang', metavar='lang', default='c')
 
-    sign = subs.add_parser('sign', help='Sign an image with a private key')
+    sign = subs.add_parser('sign', help='Sign an image with a private key',
+            aliases=['create'])
```
SebastianBoe: I think a user would need some more documentation to understand this.

The help text for 'create' would now state that this would sign an image, wouldn't it?

mbolivar: I agree that just a little more help text would be nice.

nvlsianpu: I have the same opinion.

utzig:
LGTM

PR #252, opened 2018-03-20
Restore functionality on nRF devices

mbolivar:
Testing with the latest Zephyr is showing that CONFIG_SOC_FLASH_NRF5 is no longer set. This manifests in the following error message:

[MCUBOOT] [ERR] main: Flash device not found

Instead of relying on Zephyr to enable individual SOC flash drivers based on the global CONFIG_FLASH value, just explicitly select CONFIG_SOC_FLASH_NRF5 in a new device-specific MCUBOOT_DEVICE_NRF hidden config. This fixes the problem.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -17,6 +17,7 @@ config MCUBOOT_DEVICE_SETTINGS
 	default y
         # CPU options
 	select MCUBOOT_DEVICE_CPU_CORTEX_M0 if CPU_CORTEX_M0
+	select MCUBOOT_DEVICE_NRF if SOC_FAMILY_NRF
         # Enable flash page layout if available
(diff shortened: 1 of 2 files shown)


carlescufi:
Note that we will also rename the flash driver from NRF5 to NRF soon

mbolivar:
> Note that we will also rename the flash driver from NRF5 to NRF soon

Looks like that is the cause of the problem: that rename was partially done already in Zephyr commit f49150cab686ae3. That's possibly breaking samples upstream as well

carlescufi:
I now realize what the issue is and have posted a PR to Zephyr instead: https://github.com/zephyrproject-rtos/zephyr/pull/6721/files

mbolivar:
Closing this to be addressed in Zephyr

PR #254, opened 2018-03-21
zephyr: CMakeLists: allow users to specify DTC_OVERLAY_FILE fragments

mbolivar:
Allow the user to add additional device tree overlays onto the default one that sets up the code partition.


carlescufi:
Can we merge this?

PR #255, opened 2018-03-26
samples: zephyr: Use BOOTLOADER_MCUBOOT

carlescufi:
Instead of manually setting the text section offset and the overlay
file, use the now readily available BOOTLOADER_MCUBOOT Kconfig option to
provide that automatically.


carlescufi:
CC @mbolivar @d3zd3z @utzig @nvlsianpu

PR #256, opened 2018-03-26
boot_serial: zephyr: Use new base64 library

carlescufi:
Instead of relying on mbedTLS' base64 codec, use the new Zephyr base64
library.


carlescufi:
CC @nvlsianpu @utzig

nvlsianpu:
all is done as it should be done :D

PR #259, opened 2018-03-29
Fix boot_serial to follow flash write alignment

utzig:
This fixes an issue found on stm32f3 where `newtmgr` after writing an odd sized packet, would hardfault on the next write due to trying to write an odd address which is not valid on stm32f3 flash.


utzig:
This reapplies in `boot_serial` a fixed applied some days ago on `imgmgr` in Mynewt (https://github.com/apache/mynewt-core/pull/963).

PR #260, opened 2018-03-29
Fix boot_serial size of allocated buffers

utzig:
This fixes an issue where while doing an image upload `newtmgr`/`mcumgr` would send packets that were impossible do base64decode in a 128 byte buffer.

### DIFF
--- a/boot/mynewt/src/main.c
+++ b/boot/mynewt/src/main.c
@@ -41,11 +41,11 @@
 #include "bootutil/image.h"
 #include "bootutil/bootutil.h"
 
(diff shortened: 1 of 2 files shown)


utzig:
Btw, I had `mtu=256` for the serial port used for this particular board. Does this mean that if I increase `mtu` to `512` it would also break even with this change applied?

utzig:
Ok, confirmed that `mtu` was the issue. I will maintain the buffer at `256` bytes but added some documentation as well addressing this "limitation".

PR #261, opened 2018-04-02
Add click handling of cli options

utzig:
This switches CLI parameter handling to click for much better UX.

### DIFF
--- a/scripts/imgtool.py
+++ b/scripts/imgtool.py
@@ -14,137 +14,160 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
+import click
 import getpass
 from imgtool import keys
 from imgtool import image
-from imgtool import version
-import sys
+from imgtool.version import decode_version
 
-def get_password(args):
-    if args.password:
-        while True:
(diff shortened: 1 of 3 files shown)


mbolivar:
I'm not against switching to click, but just want to make sure we're aware of issues with doing so as proposed:

- the -h option goes away (at least in my experience), users need to say --help
- this patch mixes in unrelated changes which would be more cleanly done as separate patches

utzig:
@mbolivar You're right about `-h` but I handled it on the PR so both `--help` and `-h` work! And there really are unrelated changes, sorry... this PR was meant mostly as a proposal so far. Thanks for the feedback! ;-)

utzig:
Btw, I just added support for adding aliases, with `create` aliased to `sign`. I am listing aliases on `help` as well:

```
$ ./imgtool.py 
Usage: imgtool.py [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  create  Create a signed or unsigned image
  getpub  Get public key from keypair
  keygen  Generate pub/private keypair
  sign    Create a signed or unsigned image
```

PR #263, opened 2018-04-03
doc: zephyr: Extend info about flashing

carlescufi:
Add precisions to the information regarding flashing MCUboot and Zephyr
application images in order to make users aware of mass-erase scenarios
that might bite them.


utzig:
What's an example of device that does mass erase on `make flash`?

carlescufi:
@utzig right now the Nordic chips. We will try to change this, but we need additional patches and I think it's a good idea to document that this can be the case for some chips.

carlescufi:
@utzig I think we can merge this now

PR #268, opened 2018-04-12
Replace mcuboot flash_map by zephyr flash_map

nvlsianpu:
This patch touch only zephyr builds of mcuboot.
The patch introduce usage of zephyr flas_map module instead
of mcuboot zephyr-only implementation. Unused flash_area_to_sectors
API of former flash_map was removed as well.
Size of sector-status-update-map entry is now defined thanks to the
minimum write size supported by the flash driver.

hal_flash_align calls were replaced by flash_area_align calls.

For avoid ambiguity former zephyr-only files flash_map.*
were renamed to flash_map_extended.*


boot/boot_serial/test/src/boot_test.c
```diff
@@ -29,7 +29,11 @@
 #include "crc/crc16.h"
 #include "testutil/testutil.h"
 #include "hal/hal_flash.h"
+#ifdef __ZEPHYR__
+#include "flash_map/flash_map_extended.h"
```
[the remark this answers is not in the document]
nvlsianpu: thank for review - shouldn't touch mynewt build thanks to postprocesor, but <b>right</b>, it touch the code, will try to rework this.

nvlsianpu:
I checked that mynewt has been compiling successful  wit this patch as well. I still have to fix test (probably need some help, but firs will try to recognize problem by myself).

utzig:
I will wait for CI fixes before I review this.

mbolivar:
> I will wait for CI fixes before I review this.

Looking at the latest version, I think this still needs re-work to avoid adding the `__ZEPHYR__` ifdeffery in bootutil as discussed during this week's meeting, so that seems safe.

nvlsianpu:
As was discuses hal_flash_align calls were replaced by flash_area_align calls for both Zephyr and MyNewt. flash_map.h was added to the simulator c-support files as now missing in the boot/zephyr sub-directories.

boot/zephyr/flash_map_legacy.c
```diff
@@ -101,25 +58,27 @@ int flash_area_to_sectors(int idx, int *cnt, struct flash_area *ret)
...
 int flash_area_get_sectors(int idx, uint32_t *cnt, struct flash_sector *ret)
 {
-    uint32_t off;
-    uint32_t len;
+    const struct flash_area *fa;
     uint32_t max_cnt = *cnt;
     uint32_t rem_len;
 
-    if (flash_area_get_bounds(idx, &off, &len)) {
+    if (flash_area_open(idx, &fa))  {
```
mbolivar: Please be mindful to close anything that gets opened, including in the error paths.

boot/zephyr/main.c
```diff
@@ -138,7 +136,7 @@ void main(void)
 
     BOOT_LOG_INF("Bootloader chainload address offset: 0x%x",
                  rsp.br_image_off);
-    zephyr_flash_area_warn_on_open();
```
mbolivar: This was a useful feature that let us notice whenever open/close calls were mismatched; why was it removed?

I know close is currently a no-op but with names like open and close we really should not assume it will remain that way, IMO.

nvlsianpu: As you noticed the close is no-op. And therefore Zephyr doesn't implement longing feature for non-existing operation. Think we can re-implement similar feature when there will be a need for that.

In current code it is still possible to add code for counting open/close operation together - it will
require implementation of wrappers for flash_area_open/close into another mcuboot shim which I don't like.

nvlsianpu:
> I will wait for CI fixes before I review this.

@utzig  review this on your convenience.

boot/bootutil/src/loader.c
```diff
@@ -318,8 +318,8 @@ boot_write_sz(void)
      * on what the minimum write size is for scratch area, active image slot.
      * We need to use the bigger of those 2 values.
      */
-    elem_sz = hal_flash_align(boot_img_fa_device_id(&boot_data, 0));
-    align = hal_flash_align(boot_scratch_fa_device_id(&boot_data));
```
utzig: `boot_scratch_fa_device_id` is not used anymore, please remove the definition. Probably ok to remove `boot_img_fa_device_id` as well, and expand the only use of it (on `loader.c`).

boot/mynewt/flash_map_backend/pkg.yml
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) 2018 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+pkg.name: boot/mynewt/flash_map_backend
+pkg.description: Flash_map API helper reference.
+pkg.author: "Nordic Semiconductor ASA <dev@runtimeco.github.io/mcuboot/>"
```
utzig: Pleasse add a proper email here, or leave it out.

boot/zephyr/flash_map_extended.c
```diff
@@ -0,0 +1,46 @@
...
+#include <zephyr.h>
+#include <flash.h>
+
+#include "target.h"
+
+#include <flash_map_backend/flash_map_backend.h>
+#include <hal/hal_flash.h>
+#include <sysflash/sysflash.h>
+
+#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
```
utzig: Please remove this line, eventually #273 will be merged and I prefer not to rework this file after that...

boot/zephyr/main.c
utzig: I suggest opening an issue to re-implement when `close` is not a "no-op" anymore and moving.

nvlsianpu: The issue created here: https://github.com/runtimeco/mcuboot/issues/280

nvlsianpu: @mbolivar can you approve?

utzig:
@mbolivar Any feedback on latest changes?

PR #273, opened 2018-04-26
Allow log level configuration

utzig:
This PR removes the hardcoded per file log level configuration, and adds `syscfg` configurability for Mynewt with `OFF` as default. Zephyr's default stays at 3 (INFO) to maintain previous per file behaviour. Also don't initialize UART and supporting stuff on Mynewt when not required (either boot_serial or logs).

The main reason for this change is to allow for smaller bootloaders. Since 1.1.0, MCUboot has grown from less than 16KB to more than 18KB for a RSA validated image on Mynewt. The reasons were the UART initialization after `boot_serial` support was added and the amount of BOOT_LOG_ messages (strings in FLASH).


boot/boot_serial/src/boot_serial.c
```diff
@@ -24,7 +24,6 @@
 
 #include "sysflash/sysflash.h"
 
-#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
```
mbolivar: I don't agree with this change.

MCUboot is close to its flash size on Zephyr targets, and being able to control log levels on a per-translation-unit basis is the best tool I know of to manage this, as extra logging calls carry nontrivial flash penalties.

utzig: Well, one can always add those lines back and have the previous functionality. It's just not enforced in `HEAD` now.

utzig:
@mbolivar You mean like adding a per file "ifdefery" as in:

```
#ifdef LOADER_C_LOG_LEVEL
#define BOOT_LOG_LEVEL LOADER_C_LOG_LEVEL
#endif
#include <bootutil_log/bootutil_log.h>
```

an so on and so forth... is that what you have in mind?

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -135,16 +135,21 @@ int sim_log_enabled(int level);
  */
 #elif defined(MCUBOOT_MYNEWT)
 
+#include <mcuboot_config/mcuboot_config.h>
```
mbolivar: The switch to Kconfig (https://github.com/runtimeco/mcuboot/pull/272) was careful to make sure all the mcuboot_config.h options are documented in the template file, and that they don't depend on any other macros/values in the system. That is, bootutil depends on mcuboot_config.h, and not the other way around.

This PR has two problems from that perspective:

- It adds MCUBOOT_LOG_LEVEL without documentation in the template file
- The new macro gets its values from macros (BOOT_LOG_LEVEL_xxx) defined in bootutil

Can this PR please be reworked to manage the dependencies in a cleaner way?

I will add a follow-up with a design I considered in #272 but postponed due to lack of time.

mbolivar: > I will add a follow-up with a design I considered in #272 but postponed due to lack of time.

Here is what I was thinking:

- mcuboot_config.h would have a single additional config option added and documented, named `MCUBOOT_HAVE_LOGGING`
- If that option is defined, then the platform guarantees that bootutil can `#include mcuboot_config/mcuboot_logging.h`; otherwise, that file may not be included
- If mcuboot_logging.h is present, it must:
  - define MCUBOOT_LOG_LEVEL_{OFF,ERROR,...DEBUG} appropriately for the platform
  - detect the value of MCUBOOT_LOG_LEVEL if it is defined, and give it a sane platform-specific default if it is undefined
  - define MCUBOOT_LOG_{ERR,WRN,...} appropriately for the platform and the current MCUBOOT_LOG_LEVEL
- This file (bootutil_log.h) can then include mcuboot_config.h and check `MCUBOOT_HAVE_LOGGING`
   - if MCUBOOT_HAVE_LOGGING is defined, use the platform definitions, as in: `#define BOOT_LOG_LEVEL_{OFF,ERROR,...} MCUBOOT_LOG_LEVEL_{OFF,ERROR,...}` etc., and `#define BOOT_LOG_{ERR,WRN,...} MCUBOOT_LOG_{ERR,WRN,...}`, etc.
   - if absent, use stub implementations which ignore their arguments
- The rest of the library can then just include bootutil.h and use BOOT_LOG_xxx safely without worrying about the details.

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -135,16 +135,21 @@ int sim_log_enabled(int level);
...
-#define BOOT_LOG_LEVEL_DEBUG    4
+#define BOOT_LOG_LEVEL_OFF      1
+#define BOOT_LOG_LEVEL_ERROR    2
+#define BOOT_LOG_LEVEL_WARNING  3
+#define BOOT_LOG_LEVEL_INFO     4
+#define BOOT_LOG_LEVEL_DEBUG    5
 
 #ifndef BOOT_LOG_LEVEL
-#define BOOT_LOG_LEVEL BOOT_LOG_LEVEL_INFO
+#define BOOT_LOG_LEVEL MCUBOOT_LOG_LEVEL
```
mbolivar: It seems like what happened is that a patch was added on top of the existing series which moved it closer to what I was proposing.

However, this series still has the problems I named, and what I meant by "re-work" was that the patches themselves would be changed using something like an interactive git rebase so that the problems would be avoided by structuring the series in a different way.

For example, this patch still has the same problems:

- introduces MCUBOOT_LOG_LEVEL, apparently to be obtained from mcuboot_config.h, without documenting it in the template file
- values of this macro still depend on other values that don't come from the config file

That's not what I had proposed -- in my proposal, that the only addition to mcuboot_config.h would be MCUBOOT_HAVE_LOGGING. The log level defines (and MCUBOOT_LOG_LEVEL) would be handled by mcuboot_logging.h when it existed. This avoids the problems named earlier.

utzig:
@mbolivar I rewrote the commit history to make it more logical (I hope!) with added template docs and with a simpler implementation (less macros, etc). Caveat here is that github shows commits by date order, not in the history (as per rebase) order.

utzig:
@mbolivar It was updated last week, any feedback?

boot/zephyr/prj.conf
```diff
@@ -1,5 +1,6 @@
 CONFIG_CONSOLE_HANDLER=y
-CONFIG_SYS_LOG=y
+CONFIG_SYS_LOG=n
```
mbolivar: I think as a matter of design:

- if zephyr's sys_log logging is enabled, its mcuboot_config.h should automatically enable MCUboot's logs
- if logging is disabled, its mcuboot_config.h should make sure MCUboot's logs are disabled

That way there is no need for an additional BOOT_HAVE_LOGGING option, which can then simply be deleted. MCUboot logging will "just work" based on Zephyr logging using the configuration here, which users can override using the config options they are already used to. Zephyr logging is pretty confusing, I think, so adding extra options should be avoided if possible.

If desired, this series could instead add a `config BOOT_DEFAULT_LOG_LEVEL` which would control the MCUboot-specific default log level, leaving the syslog default level alone for other Zephyr subsystems used by MCUboot which may be affected by the default.

utzig:
@mbolivar fixed layering violation, license dropped.

utzig:
And yet another week goes by...

utzig:
@mbolivar @d3zd3z @carlescufi If nobody gives me reason to stop, I will merge this by the end of today.

PR #281, opened 2018-05-10
Fix leftovers from image_header v1

### LINKED ISSUE #279: image.py - needs update for image header
Image header structure in bootloader and image.py are not in sync.
image.py does not hold the loadaddr, and holding TlvSz.

utzig:
This removes the old "TLV size" and "keyId" fields that used to exist in the old image header format and updates it to use `load_addr`. Also fixes the name of reserved fields to index from 1.

This fixes #279 

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -80,11 +80,11 @@ struct image_header {
     uint32_t ih_magic;
     uint32_t ih_load_addr;
     uint16_t ih_hdr_size; /* Size of image header (bytes). */
-    uint16_t _pad2;
+    uint16_t _pad1;
     uint32_t ih_img_size; /* Does not include header. */
     uint32_t ih_flags;    /* IMAGE_F_[...]. */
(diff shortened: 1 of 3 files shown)


utzig:
Sorry, just added another commit to fix the `create` command. This is not related to the PR itself but it's a very simple change that does not deemed a new PR. Did not expect a review so fast! :-P

utzig:
@carlescufi Not sure what you mean by "do the same". One commits fixes the issue #279, which is some image header v1 leftovers and the other commits fixes the `imcage create` functionality.

carlescufi:
>@carlescufi Not sure what you mean by "do the same".

I mean why do we need an alias for `sign` called `create`. Both achieve the same result don't they?

utzig:
> I mean why do we need an alias for sign called create. Both achieve the same result don't they?

They do exactly the same thing! The reason was described in #240, but the gist is: `sign` also works without a key file (aka a non-signed image) which was a bit confusing so we added a `create` alias that works the same but would be more "intuitive" for users.

PR #289, opened 2018-06-04
Fix calloc/free warnings building Zephyr with ECSDA

utzig:
Signed-off-by: Fabio Utzig <utzig@apache.org>

PR #290, opened 2018-06-05
ext: mbedtls: Enable PLATFORM_MEMORY to avoid warnings

carlescufi:
calloc and free are actually not used in our usage of mbedtls. In order
to get rid of warnings when building it, enable the
MBEDTLS_PLATFORM_MEMORY macro which will then rely on the system to
provide them for us.


carlescufi:
Another approach to #289

utzig:
This looks much better, will test here, but I don't think you should change de default signature method.

carlescufi:
No, this was committed by mistake. I will fix it as soon as I come back home

carlescufi:
@utzig fixed now

PR #292, opened 2018-06-05
 Add --slot-size to allow size checking without pad

### LINKED ISSUE #241: Add --slot-size argument to imgtool
When imgtool is run with the --pad option, it checks that the image + trailer will fit into the slot.

However, that's the only way to get it to output an error if the image is too big. The image's size and whether or not to insert a pad are orthogonal; mixing them in this way prevents users from being able to check with imgtool that their signed images fit properly into slots unless they build padded images, which is not recommended.

To resolve this, imgtool should gain a --slot-size option to specify the total slot size, and --pad should just be a boolean flag that pads the image out to the --slot-size value (if given, and an error raised if not).

That way, users can use --slot-size to check for the trailer fitting without having to generate a padded image.

utzig:
Add the new `--slot-size` and make `--pad` a bool flag, to allow checking that firmware fits in the slot without overflowing into the trailer region even when no padding was requested.

Fixes #241

### DIFF
--- a/scripts/imgtool.py
+++ b/scripts/imgtool.py
@@ -116,8 +116,10 @@ class BasedIntParamType(click.ParamType):
 @click.argument('infile')
 @click.option('-M', '--max-sectors', type=int,
               help='When padding allow for this amount of sectors (defaults to 128)')
-@click.option('--pad', type=BasedIntParamType(),
-              help='Pad image to this many bytes, adding trailer magic')
+@click.option('--pad', default=False, is_flag=True,
+              help='Pad image to --slot-size bytes, adding trailer magic')
+@click.option('-S', '--slot-size', type=BasedIntParamType(), required=True,
+              help='Size of the slot where the image will be written')
 @click.option('--included-header', default=False, is_flag=True,
               help='Image has gap for header')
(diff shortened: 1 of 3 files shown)


carlescufi:
CC @nvlsianpu

scripts/imgtool.py
```diff
@@ -116,8 +116,10 @@ def convert(self, value, param, ctx):
 @click.argument('infile')
 @click.option('-M', '--max-sectors', type=int,
               help='When padding allow for this amount of sectors (defaults to 128)')
-@click.option('--pad', type=BasedIntParamType(),
-              help='Pad image to this many bytes, adding trailer magic')
+@click.option('--pad', default=False, is_flag=True,
+              help='Pad image to --slot-size bytes, adding trailer magic')
+@click.option('-S', '--slot-size', type=BasedIntParamType(), required=True,
```
nvlsianpu: This is required. I assume that intentionally (so need a lite more knowledge for a newcomer). I think it should be still worth to make this optional.

utzig: Yes, it is intentional, making it optional just means nobody will used it.

nvlsianpu: OK - reasonable, good job

PR #293, opened 2018-06-05
Clean up mbedTLS configuration

carlescufi:


boot/zephyr/CMakeLists.txt
```diff
@@ -104,8 +104,20 @@ if(CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256)
   target_sources(app PRIVATE "${TINYCRYPT_DIR}/source/sha256.c")
   target_sources(app PRIVATE "${TINYCRYPT_DIR}/source/utils.c")
 
+  # When building for ECDSA, we use our own copy of mbedTLS, so the Zephyr
+  # one must not be enabled or the MBEDTLS_CONFIG_FILE macros will collide.
+  if(DEFINED CONFIG_MBEDTLS)
```
carlescufi: Note that this happens when you build with the default `prj.conf` and then use `menuconfig` to switch from RSA to ECDSA. This is a consequence of Kconfig parsing and cannot be fixed by modifying the Kconfig tree. So instead we give a clear error message to the user.

carlescufi: The only real fix here is to use 2 different `.conf` files:
* `prj-rsa.conf`
* `prj-ecdsa.conf`

utzig: It was two files before, I think it makes more sense.

carlescufi: OK, this is still relevant if we have 2 files in case someone goes and sets `CONFIG_MBEDTLS=y` in the `prj-ecdsa.conf`, so I can submit a PR after this one to split into 2

utzig:
@carlescufi building on Mynewt with ECDSA:

```
...
Error: In file included from repos/mcuboot/boot/bootutil/src/image_ec256.c:27:
repos/mcuboot/ext/mbedtls/include/mbedtls/oid.h:27:10: fatal error: config.h: No such file or directory
 #include "config.h"
          ^~~~~~~~~~
compilation terminated.
...
```

carlescufi:
@utzig I suspected that. Mynewt should really provide its own config file, but that's OK. I will remove the commit that does that and push.

utzig:
@carlescufi Mynewt provides its own config if using RSA (with external mbed-TLS). The whole point of having a local `config.h` is that is is used by our bundled mbed-TLS (which includes only the ASN1 parser).

PR #295, opened 2018-06-08
zephyr: cmake: Fix a typo in file CMakeLists.txt

miyatsu:
Replace "bur" with "but".


utzig:
Your commit has `Jerry Ding` as author name which does not match the signed-off-by. Maybe we could relax this "name" requirement and use only email, but for now please fix.

miyatsu:
@utzig Sure, I'm now in my office. I will let you know when I fix this in my home.

miyatsu:
@utzig Its done, please review.

PR #296, opened 2018-06-11
zephyr: Default to RSA .pem file in config fragment

carlescufi:
In order to provide a pleasant out-of-the-box experience to users,
default to the RSA .pem file so that users do not get a cryptic error
when building with the default .conf file in upstream.


carlescufi:
cc @nvlsianpu

utzig:
The main reason I have not left it configured by default was that it would overwrite possible edits the user had done by itself. But I don't oppose this.

nvlsianpu:
LGTM

PR #303, opened 2018-06-25
Update with 1.2.0 release notes

utzig:
DO NOT MERGE

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -3,6 +3,38 @@
 - Table of Contents
 {:toc}
 
+## Version 1.2.0
+
+The 1.2.0 release of MCUboot brings a lot of fixes/updates, where much of the
+changes were on the boot serial functionality and imgtool utility. There are
+no breaking changes in MCUBoot functionality, but some of the CLI parameters
+in imgtool were changed (either removed or added or updated).
+
+### About this release
+
+- imgtool accepts .hex formatted input
+- Logging system is now configurable
+- Most Zephyr configuration has been switched to Kconfig
+- Build system accepts .pem files in build system to autogenerate required
+  key arrays used internally
+- Zephyr build switched to using built-in flash_map and TinyCBOR modules
+- Serial boot has substantially decreased in space usage after refactorings
+- Serial boot build doesn't require newlib-c anymore on Zephyr
+- imgtool updates:
+  + "create" subcommand can be used as an alias for "sign"
+  + To allow imgtool to always perform the check that firmware does not
+    overflow the status area, `--slot-size` was added and `--pad` was updated
+    to act as a flag parameter.
+  + `--overwrite-only` can be passed if not using swap upgrades
+  + `--max-sectors` can be used to adjust the maximum amount of sectors that
+    a swap can handle; this value must also be configured for the bootloader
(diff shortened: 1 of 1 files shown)


utzig:
@carlescufi @nvlsianpu @d3zd3z Please check if you have anything to add here!

@d3zd3z If `imgtool` changed in a way that breaks previous usage is this considered a breaking change of MCUBoot and we should release 2.x.x or is it OK to make it 1.2.0?

Btw, I am testing a small patch for Mynewt serial boot which I might have to PR later, but apart from that I did a some thorough testing of all functionality on Mynewt and basic sanity testing on Zephyr (thanks to @nvlsianpu for serial boot testing on Zephyr!) and all looks good.

d3zd3z:
I'm not real sure what we should do about the version number. I guess it depends on what we consider the API to be. If it includes imgtool, then yes this would be breaking.

But, I'd be more inclined to say it should still be 1.2 because we aren't changing the image format, which is kind of the API that matters. I wonder if @mbolivar has any comments on this, coming from a practical point of view.

docs/release-notes.md
```diff
@@ -3,6 +3,34 @@
...
+The 1.2.0 release of MCUboot brings a lot of fixes/updates, where much of the
+changes were on the boot serial functionality and imgtool utility. There are
+no breaking changes in MCUBoot functionality, but some of the CLI parameters
+in imgtool were changed (either removed or added or updated).
+
+### About this release
+
+- imgtool accepts .hex formatted input
+- Logging system is now configurable
+- Zephyr build system switched to Kconfig; accepts .pem files in build system
```
carlescufi: You mean CMake right?

docs/release-notes.md
```diff
@@ -3,6 +3,34 @@
 - Table of Contents
 {:toc}
 
+## Version 1.2.0
+
+The 1.2.0 release of MCUboot brings a lot of fixes/updates, where much of the
+changes were on the boot serial functionality and imgtool utility. There are
```
carlescufi: If you mention boot serial in here I think we should have an entry for it in "About this release"

docs/release-notes.md
```diff
@@ -3,6 +3,34 @@
...
+  + "create" subcommand can be used as an alias for "sign"
+  + To allow imgtool to always perform the check that firmware does not
+    overflow the status area, `--slot-size` was added and `--pad` was updated
+    to act as a flag parameter.
+  + `--overwrite-only` can be passed if not using swap upgrades
+  + `--max-sectors` can be used to adjust the maximum amount of sectors that
+    a swap can handle; this value must also be configured for the bootloader
+  + `--pad-header` substitutes `--included-header` with reverted semantics,
+    so it's not required for firmware built by Zephyr build system
+
```
nvlsianpu: zephyr footprint was reduced and  source code simplified thanks to:
- zephyr build switched to using flash_map and tinycborr modules build in zephyr-project.
- zephyr serial recovery build doesn't require build with newlib-c anymore

utzig: Added this with small changes, thanks!

docs/release-notes.md
utzig: Now there is!

PR #304, opened 2018-06-25
Deinit timer after serial load is finished

utzig:
This fixes an issue where the bootloader is activating a timer that is later not used by the installed target image causing a non handled interrupt fault.


utzig:
@mkiiskila This bug happens on architectures that don't support tickless, specifically for me on STM32. I build the bootloader without `OS_SCHEDULING`, where it uses a timer for some boot serial control, and the written image is a full OS with `OS_SCHEDULING` enabled (using SysTick) and no timer enabled in the config. Since the timer stays activated it faults just after starting the image in slot0.

PR #311, opened 2018-07-10
Update sim ecdsa testing to use ring signing

### LINKED ISSUE #287: Update simulator to use ring ec-p256 signing
Ring just got it's initial ecdsa256 signing support with this commit:

https://github.com/briansmith/ring/commit/e5a4fe96f77e6eda71b901f5bf790d5db49670d9

Once a new version is released, the simulator can be updated to use ring directly for the ecdsa test, instead of relying on the error prone tinycrypt calls.

utzig:
This removes the previous unsafe tinycrypt binding for testing ecdsa and uses the new ecdsa sign support in ring 0.13 (currently in alpha).

NOTE: Since ring does not support rfc5915 (and probably won't?), the key is converted to PKCS#8 using OpenSSL (when running on travis-ci).

Fix #287


utzig:
Asking for early reviews but I'm not sure I want to merge this before stable 0.13 comes out. ECDSA signing with Ring still seems slower than signing with Tinycrypt right now!

carlescufi:
@utzig I am really not qualified to review this. Can @d3zd3z take this one perhaps?

d3zd3z:
This doesn't build for me, I think you need to add the root-ec-p256-pkcs8.pem to the git repo, along side the others.

d3zd3z:
In other words, check in the pkcs8 file, don't just have it only work for travis.

utzig:
I am not sure it would be a good idea to merge this because it's still unstable in ring. I think not much work is going into this feature at the moment.

d3zd3z:
Sure, even if we don't merge it, it'd be nice to have it working in the pull request, though.

utzig:
@d3zd3z I think this is OK to merge now. It was updated with ring 0.14.1, which breaks the old API, and will also require small updates to #391. Also there's a a buffer overflow fix, please review again!

PR #314, opened 2018-07-11
imgtool: Adjust base_addr when injecting header

schultetwin:
Adjusts the "base address" in an intel hex file if using "--pad-header."

If adding a header to a .hex file (using the command line "--pad-header"), the hex file gets shifted by HEADER_SIZE. This may cause hard coded addresses to be off by "header-size", making the firmware unusable. Instead, adjust the base address by the header size, so the existing firmware in the hex file is loaded to the proper addresses.

I'm not sure I've thought through all use cases here. But I've found the using --pad-header results in an invalid firmware image when passing in a hex file.

### DIFF
--- a/docs/imgtool.md
+++ b/docs/imgtool.md
@@ -79,7 +79,11 @@ flash device.  For Zephyr, it will be configured as part of the build,
 and will be a small power of two.  By default, the Zephyr build system will
 already prepended a zeroed header to the image.  If another build system is
 in use that does not automatically add this zeroed header, `--pad-header` can
-be passed and the `--header-size` will be added by imgtool.
+be passed and the `--header-size` will be added by imgtool. If `--pad-header`
+is used with an Intel Hex file, `--header-size` bytes will be subtracted from
+the load address (in Intel Hex terms, the Extended Linear Address record) to
+adjust for the new bytes prepended to the file. The load address of all data
+existing in the file should not change.
 
 The `--slot-size` argument is required and used to check that the firmware
 does not overflow into the swap status area (metadata). If swap upgrades are
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -79,6 +79,9 @@ class Image():
 
         # Add the image header if needed.
         if pad_header and obj.header_size > 0:
+            if obj.base_addr:
+                # Adjust base_addr for new header
+                obj.base_addr -= obj.header_size
             obj.payload = (b'\000' * obj.header_size) + obj.payload
 
         obj.check()


utzig:
Should not the "hex" generation be considering the offset instead of making adjustments later?

schultetwin:
I don't think so, because I think then the headerless hex image is invalid. (By headerless hex image, I mean the hex file generated by my build system before I run "imgtool sign" on it.) But I'm not sure I understand the question, so here's an example.

Let's say I plan on placing the image header at 0x08020000 (with a 32 byte header), and have the actual code start at 0x08020020. These addresses are codified in my linker script, and after generating the elf file from my build system, all code is placed between 0x08020020 and 0x080210000. Nothing exists between 0x080200000 and 0x08020020 because I plan on using "--pad-header" later. Using objcopy, I convert that elf file to a hex file. This headerless hex file will have a base address of 0x08020020 because nothing exists before that address.  It should not have a base address of 0x08020000 because that would make the hex file incorrect. All the code, if loaded, would be shifted by 32. If I sign the image, with "--pad-header", the imgtool script will prepend 32 bytes to the front, and therefore must change the base address of the hex file. If it doesn't, then all the bytes will again by off by 32 bytes.

Maybe another example is with signing a binary file. When loading a binary file, there's a address parameter of where to load. Therefore your load command would be:

(gdb) load headerless_binary.bin **0x080200020**
(gdb) load signed_binary.bin **0x080200000**

That change in load address must be reflected internal to the hex file. (only when using "--pad-header")

I could also be thinking about this wrong. Let me know.

carlescufi:
@lemrey could you take a look at this one?

utzig:
@schultetwin That was a good explanation, I think it makes sense and after looking at the hex handling code I believe this is required and has to be merged, but I'll leave the final decision to the experts! (DISCLAIMER: I never use `.hex` myself)

utzig:
@lemrey @carlescufi I think this is OK, if there's no feedback by the end of the day I will interpret it as good for merging.

schultetwin:
Thank you all!

utzig:
I was able to give this some testing now, and it seems a bit weird making it work on Zephyr. To correctly locate everything in flash (more specifically the vector table), I had to change the `.dts` updating the location of the slot (to add space for later header padding) and of course the text section offset (to not pad a header) as in:

```
slot0_partition: partition@c000 {
    label = "image-0";
    reg = <0x0000C200 0x000069000>;   # <-- had to force "wrong?" location for slot
};
```

and `CONFIG_TEXT_SECTION_OFFSET=0`

This generates a `.hex` with `base_addr == 0xc200` which after this patch will correctly be located at `0xc000` and it should work on most boards. It gets somewhat weirder when the target board has fixed locations for sections like in the Kinetis line where there is flash config at `0x400`. This would need to be located at `0x200` in the linker script to correctly relocate when using `--pad-header`.

I am not sure why your build system is so different from the usual Zephyr standard (allowing it to already pad the area where the header will be). While it looks that something like this solution must be done when `--pad-header` is used, IMO we should leave this here for some more time waiting for more feedback. Or maybe I am missing something?

schultetwin:
@utzig Thank you for taking some time to review. I'm not too familiar with the Zephyr build system, but looked through it, and think I understand. Hopefully I can address your questions!

**Why Different then Zephyr**
As for why our build system is different, our code base is not MyNewt or Zephyr. Instead, we ported mcuboot to our own code base. Unfortunately, our code base is not at this time open source, so I cannot share it. That being said, I don't think the issue fixed by this patch is specific to our code base, but rather for any code base using the "--pad-header" flag with .hex files.

**Why Use --pad-header**
One option would be to not use the "--pad-header" flag (just get rid of it all together). I could certainly convert our code base to do that, similar to the Zephyr/mcuboot integration. However, I'd rather not do that because when debugging in GDB, the read-only sections of the ELF file and the programmed memory on the microcontroller do not match due to the header added by the ```imgtool``` script. For example:

```
(gdb) compare-sections
MISMATCH in .text section -> This is due to the header added by the imgtool script.
```

compare-sections is a nice sanity check to make sure the symbols from the elf file you're using match the firmware that you're debugging. Not the end of the world if it's not usable, but definitely a nice to have. Especially when working with non-reliable loaders.

**Issue with fixed location data (i.e. Kinetis flash config at 0x400)**
The example with Kinetis chips and needing to program flash config at a fixed location is not something I thought of when making this patch. Thank you for the use case! However, I don't think the proposed solution (locating the flash config at ```0x200``` in the linker script) is needed. I actually think this patch fixes the described offset problem.

After applying the patch in this PR to ```imgtool```, the modified ```imgtool``` both appends ```HEADER_SIZE``` bytes to the front of the image and subtracts ```HEADER_SIZE``` from the ```base_addr```. For example, if there's a hex file with data placed at ```0x400```, and then ```imgtool``` adds a ```0x200``` byte header to the front of said data, the new load address will be ```0x200```, but because there was an additional ```0x200``` bytes added, the data previously loaded at ```0x400``` will still be loaded at ```0x400```.

**Waiting More Feedback**
I'm happy to wait for more feedback. However, I'd recommend the "--pad-header" flag raise an error when used with a ".hex" file, as that is likely to produce incorrect results. Also, I'd appreciate a timeline on the feedback, as opposed to leaving this open indefinitely. Thank you!

Once again, thank you for reviewing! I appreciate it!

utzig:
@schultetwin Alright! Would you mind also adding a commit updating `docs/imgtool.md` with this behavior? There's already has a sentence refering to `--pad-header` and `--header-size` so a small update there should be enough. I will merge afterwards. Thanks!

schultetwin:
Done! Thank you!

PR #317, opened 2018-07-20
DO NOT MERGE!

tamasban:
Design proposal for supporting multiple image update in MCUBoot.

Signed-off-by: Tamas Ban <tamas.ban@arm.com>


docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+   uint8_t iv_minor;
+   uint16_t iv_revision;
+   uint32_t iv_build_num;
+};
+
+struct image_dependency {
+    uint32_t image_id;  /* For simplicity could be just numbering from 0 to
+                         * IMAGE_CNT
+                         */
+    struct image_version image_min_version;  /* Indicates at minimum which
```
nvlsianpu: What about image revert case, when one image (reverted one) require version form a range (probably exact version).
For simplicity it is possibile be satisfied by adding dependency on both depended image.

But in case when one image is provided external (so we can't modify it's) dependency TLV. And The second image has to keep exact range dependency.

Of course exact SHA works here pretty well, similar as addition of image_max_version field.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+    - Both images are to be updated, dependent images (one or both) are
+      missing
+    - Only one image was updated but it must be reverted, dependent image is
+      already in primary slot
+    - Both images was updated, one of them must be reverted, no further revert
+      is needed due to dependency
+    - Both images was updated, one of them must be reverted, further revert of
+      other image is needed due to dependency
+    - Both images was updated, both must be reverted on their own
+  - Dual image update, with abort:
```
nvlsianpu: how the `abort`  is forced at all?

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  - It makes updates more complex, in some cases. But if API change is a rear
+    event and usually independent updates can be done then this approach is
+    feasible.
+
+# Unresolved questions
+  - How the multiple image update conflicts with split image update?
+  - How to combine with rollback protection? After a revert restrict old image
+    functionality to a minimal subset to be able to receive new fixed version or
+    run some diagnostic, but main functionality is disabled.
+  - It was assumed that the result of self test, which sets the `image_ok` flag,
```
nvlsianpu: I doubt in sense of implementation independent image_ok for each image. A image set works well or not. Of course i expected that image_0 can not set image_ok flag for image_1 - what i mean is that analogue for image_ok field for multi-image solution should be conjunction of all image_ok values.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  - Ensure atomic update when both images must be updated in one go, and
+    firmware cannot be started before updating of both images has properly
+    finished.
+  - In case one of the images needs to be reverted, it might be necessary to
+    revert the other one as well due to API incompatibilities.
+  - Flash will be more fragmented, as two memory slots are needed for each
+    image.
+  - Whether to use shared scratch area between images or use dedicated scratch
+    area per image?
+  - Whether to support image revert at all?
```
utzig: Using what is called "overwrite-only" mode is probably much simpler, it doesn't require `scratch`, `swap status`, etc. It basically just overwrites slot0 from slot1 (starting from zero on interruption).

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+
+Mapping of image slots to flash areas might be need to be improved or changed a
+bit, because currently it is a hard coded mapping, with applying an offset.
+
+# Shared scratch
+With the increased partitioning of flash, the overhead of unused flash areas
+caused by padding are also increasing. In order to spare as much flash space as
+possible it could be good approach to share one scratch area among images. The
+swap status info could be saved to image's primary slot or to the common
+scratch.
```
utzig: `scratch` has to be able to fit the largest sector on the device. This is not a problematic for ortogonal small sector size devices (anything below 4K). But it turns very complicated on many of the STM32s that have 128K sized sectors as well as non-linear distribution of sizes. I tend to think it would be much easier to allow for swap status to remain on the origin slot (both secure and non-secure).

It may be that future devices will tend to go with small sector sizes and linear distribution which would allow us to move swap area to its own partition making this much easier, but I would suggest this being another future task not related to the dual image support.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+    needs to contain two key pairs, which public part must be built-in the
+    bootloader.
+  - Signature verification must be changed because signature covers the
+    dependency entries in the manifest and not just the image hash.
+
+# Proposed changes in tools
+  - Scripts has to be modified to create the dependency TLV entries in the
+    image manifest
+  - Signature generation must be changed to take into account the dependency
+    entries in the image manifest
```
utzig: For validating this it is essential to also update the simulator to test dual image support. Once we have fully agreed on the proposal and all doubts are resolved, I can help with this part of the work.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+    functionality to a minimal subset to be able to receive new fixed version or
+    run some diagnostic, but main functionality is disabled.
+  - It was assumed that the result of self test, which sets the `image_ok` flag,
+    is independent from the other image. In theory it means if both image was
+    updated, but one of them must be reverted then after the revert the
+    previously successful self test is expected to be still successful
+    regardless that other image was reverted. This might be not the case in
+    practice.
+  - Does it depend anyway on the RTOS (Mynewt or Zephyr)?
+  - Serial recovery with multiple images.
```
utzig: This will require updating `mcumgr` and `newtmgr` and could probably be done later in another task. You could check that `BOOT_SERIAL` cannot be defined together with `DUAL_IMAGE_SUPPORT` for now.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  - When API is changing between images then the first image which is updated
+    it must be able to handle both version of the API to can interact with still
+    old version of other image. Then other image is updated it, it needs to
+    handle only the new version of API.
+  - It makes updates more complex, in some cases. But if API change is a rear
+    event and usually independent updates can be done then this approach is
+    feasible.
+
+# Unresolved questions
+  - How the multiple image update conflicts with split image update?
```
utzig: I believe this is used only by `Mynewt` and some of the plumbing to make this work is still only available on the `Mynewt` tree. You should not have to worry much with this...

nvlsianpu:
FYI @pizi-nordic @ioannisg

docs/multiple_image_support_design_proposal.md
d3zd3z: I wouldn't imaging a device with large sectors that also needs multiple images. At least we should document this to try and discourage this kind of configuration.

docs/mcuboot_multiple_image_boot_flow.txt
```diff
@@ -0,0 +1,191 @@
...
+participant "Util/Misc"
+participant Loader
+participant Flash
+participant Verifier
+
+== Low level startup ==
+
+== Main ==
+
+note over Main: Device init:\n\
```
d3zd3z: This is a little odd to have lines ending with "\n\" in a text file. Can we just remove these?

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+# Summary
+Firmware for ARMv8-M MCUs are often released in two separate binary images:
+  - secure code
+  - non-secure code
+
+Currently MCUBoot can only handle single firmware image. MCUBoot could be
+extended to handle the secure and non-secure binaries in one single blob and do
+the upgrade of the images in one round.
+This single blob approach has some drawbacks:
+  - It always needs to update both images even if one of them has not changed.
```
d3zd3z: This isn't that much of an issue if the upgrade process uses delta compression.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  - In multi vendor scenario it requires to share the firmware images among the
+    parties. Because the concatenated firmware must be signed by that vendor,
+    who has access to the private key.
```
d3zd3z: This would require different parts of the image to have separate signatures, not necessarily that the upgrade happen independently.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  | Header             |   <- struct image_header
+  +--------------------+
+  | Payload            |
+  +--------------------+
+  | Manifest           |
+  | +----------------+ |
+  | | Manifest header| |   <- struct image_tlv_info
+A | +----------------+ |
+  | | SHA256 hash    | |   <- hash of header and payload (struct image_tlv)
+B | +----------------+ |
```
d3zd3z: Thinking about this a bit more, I think we should make the boundary of 'B' more explicit in the manifest header. It should be very clear which part is covered by the signature, and which part is the signature. It is a fairly minor change, basically putting two size fields in the manifest header. The idea is that on start, the code would verify the signature, and then after that rely on all of the previous manifest information being valid (at least until reboot).

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+    - `NONE` or `<PERM | TEST>`
+    - `REVERT` or `NONE`
+
+# Testing in simulator
+Simulator environment must be improved to able to test possible scenarios during
+multiple image update, focusing on abort and revert scenarios and on the usage
+of common scratch area.
+
+# Testing on hardware
+Testing can be done on ARMv8-M based platform i.e.: MPS2_AN521 or Musca A1.
```
d3zd3z: Unfortunately, Musca A1 isn't a great target to test this, since it runs out of SRAM. It does have the ability to directly execute out of the serial flash, but this shouldn't be used for anything other than copying mcuboot to RAM. The execution directly from flash runs at about 1/75 regular speed (RSA signature check takes around 10 seconds).

Support for Musca A1 is kind of a separate proposal, and we indeed have other targets that boot from flash loading the code into SRAM.

Qemu is also a reasonable choice. I'm not sure how much has been merged back into mainline, but there is v8m support on qemu.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+# Alternative solution
+Use only one secondary area and upgrade the images one-by-one after each other.
+Size of secondary area must be as big as the biggest image primary area.
+Benefits:
+  - Reduced the complexity of the code, because this is basically the currently
+    used process with single image update.
+  - Reduced flash fragmentation
+  - Better utilization of flash
+Drawbacks:
+  - When API is changing between images then the first image which is updated
```
d3zd3z: I'm not sure this is a reasonable restriction, and makes APIs difficult to maintain. I think we're best trying to find the full, correct, solution.

docs/multiple_image_support_design_proposal.md
d3zd3z: I think we need to think deeply about what this `image_ok` flag even means. I have been pretty clear in the past that I feel the whole thing is misguided, and that, in principle, having an image try to determine itself if it is ok is an unsolvable problem. Having two images, and two notions of "ok" makes this even more complicated. Getting it wrong can result in unnecessary reverts, which can be worse than not reverting at all (especially if a security vulnerability is found). Also, on v8m, the non-secure image won't have the ability to write to it's flag, so that would require a secure call.

docs/multiple_image_support_design_proposal.md
d3zd3z: What does it mean, then to revert one of the images? What if the upgrade was atomic, we then need to make sure that both are reverted.

Honestly, I think we should revisit the auto-rollback requirement, and perhaps determine if that is needed, or even should be implemented in the multiple-image case.

docs/mcuboot_multiple_image_boot_flow.txt
tamasban: This is needed to have a good-looking rendered figure from the test description. I use plantUML plug-in Eclipse.

docs/multiple_image_support_design_proposal.md
tamasban: How this delta upgrade works in general? Does the update client copy the active image to slot_1 and then apply the downloaded diffs on that to create the new image? Is delta upgrade approach widely used? It gives extra complexity to the update client, but on the other hand reduce the size of the downloaded image...

docs/multiple_image_support_design_proposal.md
tamasban: I agree this solution makes harder to maintain APIs. I mentioned here just to cover each solution idea. It can be feasible if API is stable between images and there are strict storage space constrains.

docs/multiple_image_support_design_proposal.md
tamasban: If I got it right then this slow performance happens if MCBUoot is executed from flash. Do you know how much time does it take if MCUBoot copied to SRAM and executed from there, but images still in flash?

I'm checking the status of armV8m support in QEMU.

docs/multiple_image_support_design_proposal.md
tamasban: Ok, thanks.

docs/multiple_image_support_design_proposal.md
tamasban: Thanks!

docs/multiple_image_support_design_proposal.md
tamasban: My view on the `image_ok` flag:
- The goal of `image_ok` is not to determine that all the feature of the device is working as expected. The built-in-self-test(BIST) should only cover whether the device are still upgradable and accessible with the new firmware as well. This might mean: checking flash access to secondary slot, checking remote server connectivity, etc.
- I believe that from operator point of view one of  the worst case which can happen with a remote IoT device if the connectivity with the device has lost due to any reason, which can be due to send a faulty/not fully tested firmware for upgrade. Communication stacks, protocols or meshed networks  are quite complex thing. I can imagine this can happen more frequently than expected.
- I think if the BIST implemented well then a revert cannot be unnecessary because its goal to ensure high availability of the device.
- How can an operator handle the scenario if due to a faulty update the connection to device has lost? To send a technician to recover the devices via serial port can be costly.
- The owner of firmware should decide whether to implement BIST or not. If not then he relies on his test coverage.
- I see a value from customer point of view to support this feature(`image_ok` & auto rollback) in MCUBoot. 
- Further this feature can be turned-off or skipped several ways in MCUBoot if customer does not want to use it:
    - permanent upgrade: the `image_ok` flag has set in advanced by update client. Even this can be vary from image to image. If big changes was done between versions then use the auto-rollback. If mandatory update must be done to fix vulnerability then might use permanent upgrade. 
    - build MCUBoot with MCUBOOT_OVERWRITE_ONLY define. 
- To have independent `image_ok` flag for multiple images might be too excessive.  But they still can be handled as one logical `image_ok` flag. If one of the `image_ok` flag is NOK then those images can be reverted which were updated in the previous run. 
- During boot up the first firmware image which determines a faulty BIST result can initiate the restart of the device. In this way the `image_ok` flag of the other updated images are still NOK. 
- On v8m if the embedded flash has a single controller which has access to the whole flash then anyway that flash controller must have only secure access. So not only the setting of the `image_ok` flag requires a secure call, then the update client must be split between S and NS. Where S part is responsible to program the new firmware to the secondary slot.
- What is your opinion to restrict the functionality of secure firmware to bare minimum in case of a revert? In this case only the update functionality would be available to download the new fixed image. This can be useful to avoid the exploit of a vulnerability.

docs/multiple_image_support_design_proposal.md
tamasban: Multi vendor scenario:
- Let's consider this scenario: There is an external vendor who provides the S binary, and there is the device owner who produces the NS binary and has access to the device to update both firmware. 
- I see two possible scenario how can this be handled.
- Scenario 1: Vendor shares the firmware with device owner. Device owner can verify the image authenticity at reception with PKI. Then device owner prepares the received S image for upgrade, which means putting header on it, concatenating TLV and dependency entry, calculating new hash and signature with his own private key (which public part is deployed in the device). Then send it to the device to upgrade. Also NS image must be prepared similar way. Might be different key pair is used to sign the S and NS image.
- Scenario 2: The bootloader contains two public keys. One of them is the corresponding part of the vendor's private key and used to verify the raw S image. The other is the corresponding part of the device owner's private key and used to verify the overhead of the S image and the NS image. In this case, the vendor will only provide the raw S binary with a signature, but the device owner can not verify it at reception, because the corresponding public key is only programmed to the device. Then the device owner still needs to prepare the image to be MCUBoot compatible: put header, include signature as TLV, could add dependency entry. In this case this additional data must also be protected, so there must be a secondary signature which also covers these items.
- Might the vendor can prepare the MCUBoot compatible  S image in some degree, add header (take into account at signing), and provide his own signature.. But probably the vendor does not add the device owner's dependency entry to the image in advanced, because he provides the firmware to many parties who requires different version number in the dependency entry.

```
  +--------------------+
  | Header             |   <- struct image_header
  +--------------------+
  | Payload            |
  +--------------------+
  | Manifest           |
  | +----------------+ |
  | | Manifest header| |   <- struct image_tlv_info
A | +----------------+ |
  | | SHA256 hash    | |   <- hash of header and payload (struct image_tlv)
B | +----------------+ |
  | | Keyhash        | |   <- Hash of vendor public key
  | +----------------+ |
  | | Signature-A-B  | |   <- Signature generated by vendor
C | +----------------+ |
  | | Dependency     | |   <- Device owner dependency entry
D | +----------------+ |
  | | Keyhash        | |   <- Hash of device owner public key
  | +----------------+ |
  | | Signature-C-D  | |   <- Signature generated by device owner
  | +----------------+ |
  +--------------------+
```

docs/multiple_image_support_design_proposal.md
tamasban: According to Peter Maydell v8m support has merged to master. 

We did measurement on Musca A1, if MCUBoot was executed from SRAM:
 - default setting for QSPI access: 1.9s
 - enhanced QSPI access: 1.6s

The measurement covers the complete MCUBoot execution: hash and signature checking, copying image to SRAM, hash checking again.

docs/multiple_image_support_design_proposal.md
tamasban: That's right. But then somehow must be described in header or TLV section that which part of the image is covered by a given signature.

docs/multiple_image_support_design_proposal.md
tamasban: Good idea. Are the two size field pointing to the borders of the signature protected area?

docs/multiple_image_support_design_proposal.md
tamasban: But still if MCUBoot built with overwrite-only mode the scratch area is reserved just not touched at all, because it is coming from the Zephyr memory config.

docs/multiple_image_support_design_proposal.md
tamasban: Whether the upgrade is atomic or not it depends on dependency entries. Atomic update, when new images must upgraded and reverted together, it can be enforced with strict dependencies. But if the dependencies are loose, there were no API change, so it remains the old value. Then still both image can be updated, and if the BIST is failing on one of them in that case only the faulty image will be reverted. 

My original assumption was that the image's BIST result are independent from the other image, but this might be not the best approach. BIST should test the working system and the flags must be in sync:
- If both image was updated then after BIST both `image_ok` flag should be OK, or NOK.
- If only one of the image was updated then only the corresponding `image_ok` must be set, and the other should not changed.

docs/multiple_image_support_design_proposal.md
nvlsianpu: As I understand scenario 1 assumes that mcuboot know only device owner public key? Looks like scenario 2 provide better security level (as mcuboot can verify independently Secure an Non-Secure image), and it is more flexible because by using the same keys it is possible to bring it almost to the 1st scenario.

nvlsianpu: cc @frkv @oyvindronningstad

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+
+*Note: IETF manifest proposal suggest to use exact hash value instead of version
+number to describe dependency*
+
+In the manifest section can be more then one dependency entry, but in practice
+if the platform only supports two individual images then there can be maximum
+one entry which reflects to the other image. I.e.: Non-secure image can be
+dependent on secure image and vice versa.
+
+Dependency is a mutual relation. If one image depends on an other then that
```
oyvindronningstad: When is this true? If an S firmware is dependent on an NS firmware, I would consider that a security issue. Also, e.g. a network stack doesn't care who is using it.

docs/multiple_image_support_design_proposal.md
oyvindronningstad: I think we can assume that vendors who want to sign their own package will also sign their dependencies.  They will want to sign the complete description of the state needed to run their software. In my head, as stated below, dependencies are _not_ mutual (at the least not for the multiple signers case), which I guess simplifies this problem.

docs/multiple_image_support_design_proposal.md
```diff
@@ -0,0 +1,549 @@
...
+  | | Signature      | |   <- signature from A - B(struct image_tlv)
+  | +----------------+ |
+  +--------------------+
+```
+
+# Flash partitioning
+In order to support independent update of multiple images the flash must be
+partitioned further to arrange two slot for each image.
+```
++--------------------+
```
oyvindronningstad: Just out of curiosity, would the following be infeasible in any real-world case?

    +--------------------+
    | MCUBoot            |
    +--------------------+
          ~~~~~              <- memory might be not contiguous
    +--------------------+
    | Image 0            |
    | primary   area     |
    +--------------------+
          ~~~~~              <- memory might be not contiguous
    +--------------------+
    | Image 1            |
    | primary   area     |
    +--------------------+
          ~~~~~              <- memory might be not contiguous
    +--------------------+
    | Image 0            |
    | secondary area     |
    +--------------------+
    | Image 1            |
    | secondary area     |
    +--------------------+
    | Scratch            |
    +--------------------+

docs/multiple_image_support_design_proposal.md
frkv: This method of adding multi-image support seems somewhat similiar to IETF SUIT handling of nested manifests. Any reason why IETF SUIT method isn't considered as an alternative to a home-spun serialization? I don't see that this will scale if you get into more complex means of supporting multi image updates. As examples of a case I would like to see supported is doing a pass-through update to an external chip (banked in one of the slots) IETF SUIT supports this with the concept "ComponentIdentifier"

nvlsianpu: @frkv - I would be not so critic - I have got in my mind iterative approach to the multiple image support - we can leave of simpler solution at the beginning, we can always redevelop it further.
Also IETF SUIT is still under discussion.

oyvindronningstad: That sounds good. Hopefully the work on this can somehow identify shortcomings in the SUIT standard. Do you by chance have more details written down about the eligibility of SUIT, and why it was kept out of this iteration?

oyvindronningstad:
I went through the Information Model for IETF SUIT (https://datatracker.ietf.org/wg/suit/documents/), and quickly identified some security requirements that seem relevant, but that are not addressed:

 - 3.3.2.  Security Requirement MFSR2: Vendor, Device-type Identifiers
 - 3.3.6.  Security Requirement MFSR4b: Authenticated Storage Location
 - 3.3.9.  Security Requirement MFSR4e: Authenticated precursor images
 - 3.3.10.  Security Requirement MFSR4f: Authenticated Vendor and Class IDs

docs/multiple_image_support_design_proposal.md
tamasban: Why a security issue, pls. elaborate a bit? 

On armv8M architecture some functionalities are separated between S ans NS world. But S world provide some service to the NS. They are communicating with a defined API, which by default introduce a dependency between them:
- If API provided by S has changed then NS also must be updated. NS which supports older version of API might cannot collaborate with S, which supports newer API version.
- Also the SG(secure gateway) instruction are placed to a fixed place as part of the S image. If they are moved in memory between S version then a new NS image must be compiled. 

But you are right it can be the case when S and NS image can be updated independently. If there is no change in the interface, only internal changes were made.

Application also depends on network stack if its API changed.

docs/multiple_image_support_design_proposal.md
tamasban: I'm not an expert in flash technology, but considering the following I do not see an issue:
- If flash controller have access to all memory are
- If size of image 0/1  primary and secondary area has the same size
- If sector layout of image 0/1 primary and secondary area is the same

docs/multiple_image_support_design_proposal.md
tamasban: If the memory is not contiguous then this probably not feasible.

docs/multiple_image_support_design_proposal.md
tamasban: @nvlsianpu Yes, scenario 1 assumes that only the device owner public key is present in the device. But there is another key pair shared between SW vendor and device owner which is used to verify the SW component at reception.

 @oyvindronningstad Dependencies are not always mutual, but can be a case when they are (API or ABI changes between images). Regrading what to willing to sign a SW vendor can be vary among vendors. I assumes that a vendor who has many customers does not want to prepare and sign his SW product to meet all his customers needs. If images are not dependent on each other or we can assume that interface between them is quite mature, then it simplifies a lot of thing. 

The main goal of dependency handling/checking to verify whether an upgrade can be  started, all required SW components is in place. Second use case is during an image revert (question that this feature is needed at all) identify which other images must be reverted as well due to dependency.

@frkv  Originally in MCUBoot the TLV fields are used to describe the image's metadata.  I think IETF SUIT as an alternative for this. I think there is a separate discussion whether how and which level to support IETF SUIT in MCUBoot(I am not invovled). But currently it is not there, therefore I built my design proposal to the top of an existing feature.
What do you mean by: "pass-through update to an external chip"?

I will take a look to IETF SUIT to be more familiar. 

Do you have any preference/opinion on regarding to support auto revert feature?

docs/multiple_image_support_design_proposal.md
oyvindronningstad: You are right, S firmwares can safely depend on NS firmwares, though of course care must be taken whenever data is provided by the NS.

To be clear, I disagree with the following paragraph as a blanket statement

> Dependency is a mutual relation. If one image depends on an other then that
other image also depend on the first one. Therefore in case of dependency there
must be one dependency entry in both images manifest in order to ensure proper
updates.

To me it looks like you're saying that, as a rule, all dependencies shall be reciprocated. This doesn't make sense to me.

I'm not saying circular dependencies don't have valid use cases. I'm saying there's a difference between one-way and two-way dependencies, and this distinction should be leveraged when it is useful to do so.

A natural (to me) interpretation of versions is that the FW version strictly means "API version", meaning that a FW has a dependency on another FW iff it actively calls on that FW. Circular dependencies have only a minor role here IMO, in cases where two FWs actively use each other's API. A network stack is a clear case of one-way dependency, but I realize that in ARMv8 scenarios there is a larger chance of circular dependencies happening.

docs/multiple_image_support_design_proposal.md
oyvindronningstad: > If sector layout of image 0/1 primary and secondary area is the same.

By a sector, do you mean the smallest area that can be erased?

What sort of weird flash layouts actually exist out there?

tamasban: Yes use sector in this context: smallest erasable area.
Some STM32 flash has mixed layout not all sector has the same size. There are big and small sector on the flash.

oyvindronningstad: Ok, thanks for the explanation.

docs/multiple_image_support_design_proposal.md
sigvartmh: Does MCUBoot support delta compression? is it in the road map?

utzig: It is not on the roadmap. So how does delta compression work if the upgrade is interrupted halfway through?

sigvartmh: As far as I can understand you don't start the upgrade before you have received the full delta and when you do the delta upgrade you store the progress successively. 
dfu_delta 1

utzig: Ok, let me rephrase it a bit... suppose you are going to upgrade one sector and you have the delta. You read the sector to RAM, apply the delta changes in RAM, and now you erase the sector because you want to write the upgrade data. Now for some reason the device resets, and when you reboot you have the erased sector and some delta to something that does not exist anymore. I suppose you could write the applied delta changes to scratch as well, as well as metadata to know which sector is being upraded,etc, so it might be feasible but might be a somewhat less straight forward than it might seem at first!

utzig:
@tamasban I think you can close this one!

PR #322, opened 2018-07-26
Initialize watchdog when running on Mynewt

utzig:
On Mynewt some flash drivers on very slow devices tickle the watchdog to avoid possible hangs; those were also trying to tickle the watchdog in mcuboot when performing an upgrade or serial boot, without initializing the watchdog, causing an invalid memory write and reset issue.

### DIFF
--- a/boot/mynewt/src/main.c
+++ b/boot/mynewt/src/main.c
@@ -213,6 +213,11 @@ main(void)
 
     hal_bsp_init();
 
+#if !MYNEWT_VAL(OS_SCHEDULING) && MYNEWT_VAL(WATCHDOG_INTERVAL)
+    rc = hal_watchdog_init(MYNEWT_VAL(WATCHDOG_INTERVAL));
+    assert(rc == 0);
+#endif
+
 #if defined(MCUBOOT_SERIAL) || defined(MCUBOOT_HAVE_LOGGING)
     /* initialize uart without os */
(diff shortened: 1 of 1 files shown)


mkiiskila:
Check the size of the bootloader on nrf52 when you build with that enabled. I think it's going to be 1.5k bigger, thanks to division operation done within watchdog_init().

mkiiskila:
Also, I vaguely recall that on some platforms you would not be able to adjust tickle period once watchdog had been enabled. I might be wrong about that, however. Comments @wes3 ?

utzig:
Another option would be to decide if tickling is required by testing for `MYNEWT_VAL(OS_SCHEDULING)` in the flash driver itself and close this one.

wes3:
@mkiiskila @utzig I hope I am answering the correct question here, but Marko is indeed correct about the watchdog on the nrf52. Here is the spec wording:

The watchdog must be configured before it is started. After it is started, the watchdog’s configuration registers, which comprise registers CRV, RREN, and CONFIG, will be blocked for further configuration.
The watchdog can be reset from several reset sources, see Reset behavior on page 83.
When the device starts running again, after a reset, or waking up from OFF mode, the watchdog
configuration registers will be available for configuration again.

wes3:
I should have mentioned that I am not sure about other MCU's.

PR #334, opened 2018-09-24
Test erased flash with flash_area_read_is_empty()

utzig:
Mynewt has recently added an encrypted flash layer driver, that runs transparently on any flash, handling reads and writes, and bypassing other flash operations to the HW driver. As a result of this change, checking for erased data cannot be done by read + compare to `erased_val` but need to be routed to an empty check on the lower level. To do this Mynewt added a new `flash_map` function called `flash_area_read_is_empty` which checks for erased blocks (and reads/decrypts the data as well).
    
This commit uses `flash_area_read_is_empty` to determine if magic, flags and swap status are erased. For Zephyr/sim commits were added previously that mimic this functionality by simply doing the read/compare.

PS: This might be required on very slow MCUs under Mynewt: https://github.com/apache/mynewt-core/pull/1423

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -97,34 +97,17 @@ static const struct boot_swap_table boot_swap_tables[] = {
     (sizeof boot_swap_tables / sizeof boot_swap_tables[0])
 
 static int
-boot_magic_decode(const struct flash_area *fap, const uint32_t *magic)
+boot_magic_decode(const uint32_t *magic)
 {
-    size_t i;
-    uint8_t erased_val;
-
     if (memcmp(magic, boot_img_magic, BOOT_MAGIC_SZ) == 0) {
         return BOOT_MAGIC_GOOD;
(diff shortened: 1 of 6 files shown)


mkiiskila:
Works ok without the PR mentioned.

utzig:
@mkiiskila Yes, this is "independent" of the mentioned PR, but to get the full update cycle working, including `image upload` on slow MCUs, that change is also required. I will remove the dependency and add reviewers.

d3zd3z:
This change looks good, itself. I'm a little concerned with this is_empty functionality, though. There are some devices that when written with 0xFF, will read back 0xFF, but can't be written again without an erase. An example would be a device that has an internal ECC on the data. We should be ok in MCUboot since we don't ever write 0xFF to the areas we are testing for this. But it isn't a general mechanism that can tell is data is erased.

nvlsianpu:
I think we will add the check-empty API to to Zephyr as well (as it is more effective) in near feature.

PR #337, opened 2018-09-25
Fix the wrong array size for the magic array in bootutil_misc

### LINKED ISSUE #335: Wrong size of magic word in bootutil_misc.c
As @danielinux noted on slack the size of magic array in both `boot_read_swap_state()` and `boot_read_swap_size()` is 4x as big as required, since BOOT_MAGIC_SZ is defined as

```
const uint32_t boot_img_magic[] = {
    0xf395c277,
    0x7fefd260,
    0x0f505235,
    0x8079b62c,
};

const uint32_t BOOT_MAGIC_SZ = sizeof boot_img_magic;
```
Evaluates to 4 bytes* 4 = 16 which gives `uint32_t magic[BOOT_MAGIC_SZ]` an array of 16 so 4*16=64 bytes.

https://github.com/runtimeco/mcuboot/blob/master/boot/bootutil/src/bootutil_misc.c#L223
https://github.com/runtimeco/mcuboot/blob/master/boot/bootutil/src/bootutil_misc.c#L281

Assuming magic should fit the whole boot_img_magic then I would suggest changing it to `uint32_t magic[BOOT_MAGIC_SZ/sizeof(uint32_t)]`. Or you could do as suggested in slack and change it to `uint8_t magic[BOOT_MAGIC_SZ]` as suggested by @utzig on slack which will give 1 byte * 16 = 16 bytes array. I don't know which is more preferred?

sigvartmh:
This will fix #335 by defining the array size of the magic word and
give it to the relevant functions.


utzig:
This was a correct implementation, btw!

PR #338, opened 2018-09-26
Fix setting confirm flag

rymanluk:
It was impossible to set confirm flag due to incorrect comparison.
In this point of time state->image_ok is already decoded and contains
one of following values:

If BOOT_FLAG_UNSET then we are ok to set a confirm flag.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -571,7 +571,7 @@ boot_set_confirmed(void)
         goto done;
     }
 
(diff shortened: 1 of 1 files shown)


utzig:
Looks correct to me, also the `copy_done` above has the same "bug". I'll just check first if it's straight-forward to add testing of this change to the sim.

nvlsianpu:
Commit message looks like it missing a list of values. we don't use this in zephyr - that why I didn't encounter this bug.

PR #339, opened 2018-09-26
Fix copy_done setting

utzig:
This compliments the previous change on `image_ok` and also removes `flash_area_erased_val` call that is not needed anymore.


utzig:
I will merge the previous PR and rebase over it.

@rymanluk Funny thing is that on my local tests the behavior is correct even without applying those changes, which clearly it should not!

PR #340, opened 2018-09-27
[imgtool] Add big endian support

schultetwin:
Add big endian support to imgtool so that it can be used on big endian targets.

Signed-off-by: Mark Schulte <mschulte@lyft.com>

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -50,18 +50,26 @@ boot_magic = bytes([
     0x35, 0x52, 0x50, 0x0f,
     0x2c, 0xb6, 0x79, 0x80, ])
 
+STRUCT_ENDIAN_DICT = {
+        'little': '<',
+        'big':    '>'
+}
+
 class TLV():
-    def __init__(self):
+    def __init__(self, endian):
         self.buf = bytearray()
+        self.endian = endian
 
     def add(self, kind, payload):
         """Add a TLV record.  Kind should be a string found in TLV_VALUES above."""
-        buf = struct.pack('<BBH', TLV_VALUES[kind], 0, len(payload))
+        e = STRUCT_ENDIAN_DICT[self.endian]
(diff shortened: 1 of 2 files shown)


utzig:
What target will this generate images for?

schultetwin:
Cortex-m4 targets, at the very least, can be switched into a "big endian" mode:

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/I1835.html
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/I1835.html

In my particular use case, I'm using a cortex-r4 processor that is booted by default into big endian mode.

scripts/imgtool/image.py
```diff
@@ -129,7 +138,7 @@ def check(self):
                 raise Exception(msg)
 
     def sign(self, key):
-        self.add_header(key)
+        self.add_header(key, self.endian)
```
utzig: There's no need to pass `self.endian` here since it's already data from the object.

scripts/imgtool/image.py
```diff
@@ -151,17 +160,18 @@ def sign(self, key):
             sig = key.sign(bytes(self.payload))
             tlv.add(key.sig_tlv(), sig)
 
-        self.payload += tlv.get()
+        self.payload += tlv.get(self.endian)
 
-    def add_header(self, key):
+    def add_header(self, key, endian):
```
utzig: No need for extra param (check comment above)

scripts/imgtool/image.py
```diff
@@ -139,7 +148,7 @@ def sign(self, key):
         sha.update(self.payload)
         digest = sha.digest()
 
-        tlv.add('SHA256', digest)
+        tlv.add('SHA256', digest, self.endian)
```
utzig: I suggest you add endianness to the TLV constructor, so there's not need to supply it on `add` and `get`.

PR #344, opened 2018-10-09
zephyr: fix flash_area_read_is_empty

nvlsianpu:
Implementation of flash_area_read_is_empty introduced recently
uses deprecated hal_flash_read API which caused build failure.

This patch uses zephyr's native flash_area_read() API which
solves the bug.


nvlsianpu:
@utzig Question is whether it is possible to replace hal_flash_read() in simulator as well? Does it make any sense?

utzig:
@nvlsianpu Following the discussion in the meeting, yes, we should remove those. These are relics from the time David built the simulator and Mynewt and the bootloader used those two abstract layers. MCUBoot does not use `hal_flash_*` anymore so let's get rid of that!

PR #348, opened 2018-10-13
Travis speed ups

utzig:
Because the amount of features that the simulator is testing is growing, the test time is taking a long time. This distributes simulator's tests over three travis machines to decrease test time.

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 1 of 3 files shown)


scripts/run_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+
+EXIT_CODE=0
+all_features="sig-rsa sig-ecdsa overwrite-only validate-slot0 enc-rsa enc-kw"
+
+if [[ $FEATURES =~ "none" ]]; then
+  cargo test
+  rc=$? && [ $rc -ne 0 ] && EXIT_CODE=$rc
+fi
+
+for feature in $all_features; do
```
d3zd3z: Although the current test only tests the features one at a time, it does make sense to test some of them together. We probably at least still allow that, whereas this will restrict us to only enabling a single feature at a time.

utzig: Yes, might be a good idea to be able to provide multiple features per run, will do that!

utzig:
@d3zd3z Done, now multiple features can be run together. Btw, one actual compile error was caught so I left it commented out in `.travis.yml` to fix later.

utzig:
So, is it ok? Can it be merged? I am waiting to add some integration testing of `imgtool`.

PR #352, opened 2018-11-01
boot: zephyr: fix compiler warnings

lemrey:
- Updated board configuration
- Added missing  `sys_clock_disable` declaration; funny, this functions is not declered in any header.

Also, Zephyr has deprecated SYS_LOG; this means warnings :)
I plan to add a commit to migrate to the new logger.


lemrey:
@nvlsianpu please, have a look.

boot/zephyr/main.c
```diff
@@ -48,6 +48,8 @@ struct arm_vector_table {
     uint32_t reset;
 };
 
+extern void sys_clock_disable();
```
sigvartmh: Move this up where other external functions are defined like `os_heap_init(void)` is declared  
https://github.com/runtimeco/mcuboot/blob/1469310756c600d71bd6914d9bf2e07b9b077111/boot/zephyr/main.c#L43 

Also add `void` to the declaration so that you won't get conflicting definitions.

Also @nvlsianpu do they want the `extern` declaration for external functions? Functions are external by default in C but it's nice to declare in the program as it makes it more readable so you know these functions are defined in another file. Forward declarations from header files are even better as then you know the contributing file but I can see why this maybe is not desirable for MCUBoot.

lemrey: Since it is only used within that `#if defined(CONFIG_ARM)` directive, it could be wrong to `extern` it unconditionally, e.g. it might not be compile then. Also, function declarations are not `extern` by default, unless they are declared in a header file, this one is not so the compiler will emit a warning unless it's declared `extern`.

sigvartmh: Ah yeah that makes sense ignore my above comment on placement. I agree that the external specifier is nice I just wonder if that's what they want in the project since they didn't use it for the other external functions. It has no real effect due to how the project is compiled.

nvlsianpu: should be" `extern void sys_clock_disable(void);`
Function is available once CONFIG_SYSTEM_CLOCK_DISABLE == y.
Don't care about platform which doesn't implement this function.

@sigvartmh extern is ok here

sigvartmh: Then I would like https://github.com/runtimeco/mcuboot/blob/1469310756c600d71bd6914d9bf2e07b9b077111/boot/zephyr/main.c#L43 to be declared as extern also.

mike-scott:
LGTM.   I had a local version of the 1st patch and @lemrey beat me to submitting it.

PR #355, opened 2018-11-08
boot: zephyr: add MCUBOOT config

lemrey:
Add a hidden MCUBOOT config entry to mark a project as MCUBOOT.
It is useful when other parts of the system need to be aware that they are, in fact, the bootloader.

@nvlsianpu


nvlsianpu:
Can you point where it is used in zephyr?

lemrey:
Used by https://github.com/zephyrproject-rtos/zephyr/pull/11210.

PR #356, opened 2018-11-14
fix zephyr target: corrupted build after zephyr PR #11180

nvlsianpu:
Zephyr target was corrupted as recntly zephyr's device tree
started adding DT_ prefix in generated labels.

This path aligns flash name macro used.


nvlsianpu:
^^ Added the link.

PR #359, opened 2018-11-19
Multi flash

utzig:


utzig:
@d3zd3z If you can do some review of the rust code here I'd be glad (you can skip the `bootutil` commit because it's not 100% done).

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,55 +1,94 @@
...
+
+lazy_static! {
+    static ref FLASH: Mutex<FlashMap> = {
+        Mutex::new(HashMap::new())
+    };
+}
+
+lazy_static! {
+    /// Maintain a table of [device_id -> (align, erased_val)]
+    static ref FLASH_PARAMS: Mutex<HashMap<u8, (u8, u8)>> = {
```
d3zd3z: I'm kind of on the fence as to whether this map should map to a struct with two fields rather than a general tuple. It does seem to currently only be used below, so isn't that great of an issue. The compiler will generate the same code with a struct defined.

utzig: Oh yeah, it definitely improves a lot the legibility; I squashed the fix into the original commit. Thanks!

carlescufi:

boot/bootutil/src/loader.c
```diff
@@ -316,42 +316,87 @@ boot_write_sz(void)
...
+            sz0 += boot_img_sector_size(&boot_data, 0, i);
+            /* guarantee that multiple sectors of slot1 fit into slot0 */
+            if (smaller == 2) {
+                BOOT_LOG_WRN("Cannot upgrade: slots have non-compatible sectors");
+                return 0;
+            }
+            smaller = 1;
+            i++;
+        } else {
+            sz1 += boot_img_sector_size(&boot_data, 1, j);
```
nvlsianpu: Why not try to find LCM instead?

utzig: How would LCM ensure that for a linear scan all sectors in some slot fit inside all sectors in the other slot?

nvlsianpu: So I see two class of flash devices - these witch regular flash pages layout (which mean that each flash page has the same size) and these which irregular flash pages layout (which mean that each flash page might have different size).

regular layout is common for serial flashes, both layout types might be expected for SoC embedded flashes.

So I understand why you implement this this way.

utzig: Yes, I think an even more common case is to simply have 2 flash devices where one has 4KB sector sizes, and the other might have 8KB or any different from 4KB size. But this should also suit STM32's with large sectors and an external SPI flash with small linear sectors.

nvlsianpu: Maybe it would be worth to describe reason behind that implementation i documentation, for  "future generations" of course, that will show how smart we were when irregular flash layout becomes a kind of dinosaurs bones.

boot/bootutil/src/loader.c
```diff
@@ -316,40 +316,92 @@ boot_write_sz(void)
...
+                BOOT_LOG_WRN("Cannot upgrade: not all sectors fit inside scratch");
+                return 0;
+            }
+            smaller = sz0 = sz1 = 0;
+        }
+    }
+
+    if (i != num_sectors_0 || j != num_sectors_1 || slot0_sz != slot1_sz) {
+        BOOT_LOG_WRN("Cannot upgrade: slots are not compatible");
+        return 0;
```
nvlsianpu: So we still expected that slots will be of the same size. What about situation where slot1 is a little bigger than slot0? Then we can just limit image size to slot0 size and still been capable to do the update,

utzig: I think I commented somewhere that I would prefer to add this later. Honestly I am not sure why someone would ever do that, the flash might be any size but you should be configuring your `flashmap` to have same sized slots. Anyway my proposal for future work was not so much about slot's sizes, but to check compatibility only up to the amount of sectors required to fit both slot's images (so slot's sizes don't really matter much).

nvlsianpu: This was just remark, note. For some SoC and cases that might required. This would be easy to add once needed by someone. I'm nordic so I have no business in the discuses enchantment.

boot/bootutil/src/loader.c
```diff
@@ -1294,14 +1359,27 @@ boot_copy_image(struct boot_status *bs)
...
 
-    size = 0;
+    slot0_size = 0;
+    slot1_size = 0;
     last_sector_idx = 0;
+    /* not used in the swap routines, only here to match slot0's and
+     * slot1's sizes
+     */
+    last_idx_slot1 = 0;
     while (1) {
```
nvlsianpu: I don't understand what happens in code below.

utzig: Yeah, I don't think this is very intuitive. I added another comment just above the loop, should help in clarifying.

markus-becker-tridonic-com:
@utzig is there an example nrf52840_pca10056.dts with slot1 and scratch on the external QSPI flash?

utzig:
@markus-becker-tridonic-com Unfortunately I never tried using the external flash on Zephyr, and I am not sure it is actually supported. For Mynewt it is available with a just few `syscfg` changes.

markus-becker-tridonic-com:
@utzig there is https://github.com/zephyrproject-rtos/zephyr/issues/9873 and https://github.com/zephyrproject-rtos/zephyr/pull/7561 which might add support for that chip.

PR #362, opened 2018-11-30
Add minimum size validation for header parameter

utzig:
Throw exception on values that are less than the minimum acceptable size of 32 bytes.


utzig:
Can someone review?

PR #364, opened 2018-12-03
Add boostrapping from slot1 support

utzig:
This adds a new option that allows copying slot0, from the contents of slot1, if slot0 is found out to be erased and not validated, and the contents of slot1 are validated.

This mechanism basically enables a device to "bootstrap" from the contents of an external flash (as slot1) that has a valid image, given that the only thing flashed in the internal flash is the bootloader.


mkiiskila:
This works with my tests. Code looks ok as well.

There's some zephyr RSA-KW changes in here too; not sure if those were included intentionally?

utzig:
> There's some zephyr RSA-KW changes in here too; not sure if those were included intentionally?

That is a config for mbedTLS which is used by the simulator when building with both RSA signing and AES Keywrapping at the same time. This was not tested before so I fixed the related issues and added a test.

carlescufi:

utzig:
This feature is explained on the initial text (just updated), but apart from small refactorings the only change when not enabled is that it renames the boot copy functions, Previously there were 2 `boot_copy_image` with different contents, one run with swap, one with overwrite only. I renamed the swap version to `boot_swap_image`. The reason for renaming is that with bootstrapping enabled the `boot_copy_image` is also used for providing the bootstrapping feature.

Have been testing this for a few weeks on Mynewt and today on Zephyr as well, and it has been working OK. I am not 100% sure of security implications, but leaving the option disabled should have none.

Since this was already approved, I will leave it here for one more day or so, if anyone has comments to add, please add soon. Btw, the feature itself is basically a single commit (what would really need review), other commits add more testing support, etc.

boot/bootutil/src/loader.c
```diff
@@ -715,10 +743,10 @@ boot_validate_slot(int slot, struct boot_status *bs)
     }
 
     hdr = boot_img_hdr(&boot_data, slot);
-    if (boot_magic_is_erased(flash_area_erased_val(fap), hdr->ih_magic) ||
-            hdr->ih_flags & IMAGE_F_NON_BOOTABLE) {
+    if (boot_check_header_erased(slot) == 0 || (hdr->ih_flags & IMAGE_F_NON_BOOTABLE)) {
```
hakonfam: There is no distinction between an error message and a "header not erased".
Currently, it looks like `boot_check_header_erased` has a "tristate" return value, `0` for `true`, `-1` for `false` and `BOOL_EFLASH` for error.
Would it be better to only return the error state, and have  "is erased" set through an output argument?
This way the error code could be checked separately in a better way than the current return scheme.

boot/bootutil/src/loader.c
```diff
@@ -690,18 +690,46 @@ split_image_check(struct image_header *app_hdr,
...
+    uint8_t *p = (uint8_t *)data;
+    for (i = 0; i < len; i++) {
+        if (val != p[i]) {
             return 0;
         }
     }
     return 1;
 }
 
+static int
```
hakonfam: Missing docs.

boot/bootutil/src/loader.c
```diff
@@ -715,10 +743,10 @@ boot_validate_slot(int slot, struct boot_status *bs)
     }
 
     hdr = boot_img_hdr(&boot_data, slot);
-    if (boot_magic_is_erased(flash_area_erased_val(fap), hdr->ih_magic) ||
-            hdr->ih_flags & IMAGE_F_NON_BOOTABLE) {
+    if (boot_check_header_erased(slot) == 0 || (hdr->ih_flags & IMAGE_F_NON_BOOTABLE)) {
         /* No bootable image in slot; continue booting from slot 0. */
-        return -1;
+        rc = -1;
+        goto out;
```
hakonfam: IMO it is better to have several calls to `flash_area_close` than to introduce `goto`.
It uses the same amount of code lines, and has a simpler code flow to read, since the return is immediately afterwards, and one can jump to the next function.

boot/bootutil/src/loader.c
```diff
@@ -1511,9 +1547,34 @@ boot_swap_if_needed(int *out_swap_type)
...
             break;
+#ifdef MCUBOOT_BOOTSTRAP
+        case BOOT_SWAP_TYPE_NONE:
+            /*
+             * Header checks are done first because they are inexpensive.
+             * Since overwrite-only copies starting from offset 0, if
+             * interrupted, it might leave a valid header magic, so also
+             * run validation on slot0 to be sure it's not OK.
+             */
+            if (boot_check_header_erased(0) == 0 ||
```
hakonfam: Again, a potential error message is choked here.

boot/bootutil/src/loader.c
```diff
@@ -690,18 +690,46 @@ split_image_check(struct image_header *app_hdr,
...
 static inline int
-boot_magic_is_erased(uint8_t erased_val, uint32_t magic)
+boot_data_is_setto(uint8_t val, void *data, size_t len)
 {
     uint8_t i;
-    for (i = 0; i < sizeof(magic); i++) {
-        if (erased_val != *(((uint8_t *)&magic) + i)) {
+    uint8_t *p = (uint8_t *)data;
+    for (i = 0; i < len; i++) {
+        if (val != p[i]) {
```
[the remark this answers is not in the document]
utzig: This compares a memory area to a value, it's not its job to test where the memory is located. And for our specific use here it is comparing data that is in RAM already.

boot/bootutil/src/loader.c
```diff
@@ -690,18 +690,46 @@ split_image_check(struct image_header *app_hdr,
     return 0;
 }
 
+/*
+ *
```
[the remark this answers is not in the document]
utzig: Done.

boot/bootutil/src/loader.c
```diff
@@ -690,18 +690,46 @@ split_image_check(struct image_header *app_hdr,
...
+ */
 static inline int
-boot_magic_is_erased(uint8_t erased_val, uint32_t magic)
+boot_data_is_setto(uint8_t val, void *data, size_t len)
```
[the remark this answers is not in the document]
utzig: `I_dont_like_very_long_names_with_lots_of_underscores` but done!

boot/zephyr/Kconfig
```diff
@@ -71,6 +71,14 @@ config BOOT_UPGRADE_ONLY
 	  swapping them.  This prevents the fallback recovery, but
 	  uses a much simpler code path.
 
+config BOOT_BOOTSTRAP
+	bool "Boostrap erased slot0 from slot1"
+	default n
```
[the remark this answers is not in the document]
utzig: I will skip all Zephyr/Kconfig suggestions, because this is the style already used which was initially added by Zephyr commiters, so if the idea is to change it or clean it up, someone should send a PR later cleaning up all options.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
utzig: Done.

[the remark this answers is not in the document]
utzig: And by Done, I mean I changed it to return `bool`. Returning int is mainly a traditional C thing, no particular reason for it.

boot/bootutil/src/loader.c
utzig: This is a good point, but if I were to document, I would almost replicate the function's name. Also it uses the in memory boot data which is something that I wouldn't explain in a doc header so in the end this is too specific to bootloader itself, not an API of something that anyone would ever use.

boot/bootutil/src/loader.c
utzig: This might be true right now. But if there were other clean up steps, it would not be true anymore, and instead of changing everything one would just insert the new clean up step at the exit label. One reason why this is such a common pattern in C. As a matter of style I actually like the way it is done now.

boot/bootutil/src/loader.c
utzig: This is a good point... I decided to change it to `-1` here because if makes no difference. What we want is a bool behaviour. Returning `BOOT_EFLASH` always meant some flash driver error, which many times is not tested, but still has a semantic meaning if there comes a day where something can be done about it.

utzig:
@hakonfam Thanks for reviewing, sorry for having to merge before approval but I really needed this today!

hakonfam:
> @hakonfam Thanks for reviewing, sorry for having to merge before approval but I really needed this today!

Np, I was offline for the holidays :)

PR #368, opened 2018-12-10
[DNM]Intel S1000 support

rgundi:
These patchsets enable the mcubootloader support for Intel S1000 board which is based on Xtensa architecture. Please note that XIP is currently not supported for Xtensa in Zephyr and hence the images in the slot need to be copied to SRAM for this to run.

DNM until : zephyrproject-rtos/zephyr#11850 PR will be merged


boot/zephyr/main.c
```diff
@@ -76,6 +76,53 @@ static void do_boot(struct boot_rsp *rsp)
...
+    void *start;
+
+    BOOT_LOG_INF("br_image_off = %x\n", rsp->br_image_off);
+    BOOT_LOG_INF("ih_hdr_size = %x\n", rsp->br_hdr->ih_hdr_size);
+
+    /* Copy from the flash to HP SRAM */
+    copy_img_to_SRAM(0, rsp->br_hdr->ih_hdr_size);
+
+    /* Jump to entry point */
+    start = (void *)(0xBE030000 + rsp->br_hdr->ih_hdr_size);
```
utzig: Would be better to add a `#define` for this address (also used in the copy to RAM function).

utzig:
It looks OK to me, I'll leave it for the Zephyr commiters to approve/merge.

boot/zephyr/boards/intel_s1000_crb.conf
```diff
@@ -0,0 +1,44 @@
...
+
+### We never want Zephyr's copy of tinycrypt.  If tinycrypt is needed,
+### MCUboot has its own copy in tree.
+# CONFIG_TINYCRYPT is not set
+# CONFIG_TINYCRYPT_ECC_DSA is not set
+# CONFIG_TINYCRYPT_SHA256 is not set
+
+CONFIG_I2C=y
+CONFIG_SPI=y
+CONFIG_FLASH=y
```
nvlsianpu: is there flash driver for this device or not? If yes then DT_FLASH_DEV_NAME should be available now or should been maid available prior this patch merge.

boot/zephyr/main.c
```diff
@@ -108,11 +108,13 @@ void main(void)
 
     os_heap_init();
 
+#ifdef DT_FLASH_DEV_NAME
```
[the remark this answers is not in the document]
nvlsianpu: se now that rather need to do the same for DT_SPI_NOR_DRV_NAME

boot/zephyr/flash_map_extended.c
```diff
@@ -15,14 +15,13 @@
 
 #include "bootutil/bootutil_log.h"
 
-/*
- * For now, we only support one flash device.
- *
- * Pick the SoC Flash driver ID.
- */
+#if defined(DT_FLASH_DEV_NAME)
```
nvlsianpu:
```suggestion
#if !defined(CONFIG_XTENSA) && defined(DT_FLASH_DEV_NAME)
...
#elif defined(CONFIG_XTENSA) && defined(DT_SPI_NOR_DRV_NAME)

...
#else
#error 
#endif
```

nvlsianpu: Reason behind suggestion is that we expect coexistence a few flash driver in mcuboot soon.

boot/zephyr/include/target.h
```diff
@@ -33,8 +33,7 @@
 /*
  * Sanity check the target support.
  */
-#if !defined(DT_FLASH_DEV_NAME) || \
```
nvlsianpu: Instead pleas add sanity check for case:
!defined(CONFIG_XTENSA) && defined(DT_FLASH_DEV_NAME)
defined(CONFIG_XTENSA) && defined(DT_SPI_NOR_DRV_NAME)

boot/zephyr/main.c
```diff
@@ -76,6 +76,53 @@ static void do_boot(struct boot_rsp *rsp)
...
 }
+
+#elif defined(CONFIG_XTENSA)
+
+static void copy_img_to_SRAM(int slot, unsigned int hdr_offset)
+{
+    const struct flash_area *fap;
+    int area_id;
+    int rc;
+    unsigned char *dst = (unsigned char *)0xBE030000 + hdr_offset;
```
nvlsianpu: doses this magic value not possible to being get from somewhere in DT or Kconfig?

boot/zephyr/main.c
```diff
@@ -76,6 +76,53 @@ static void do_boot(struct boot_rsp *rsp)
...
+    BOOT_LOG_INF("Copying image to SRAM");
+
+    area_id = flash_area_id_from_image_slot(slot);
+    rc = flash_area_open(area_id, &fap);
+    if (rc != 0) {
+	BOOT_LOG_ERR("flash_area_open failed with %d\n", rc);
+        goto done;
+    }
+
+    rc = flash_area_read(fap, hdr_offset, dst, 0x30000);
```
nvlsianpu: the same as above - is this a partition size fetch-able form DT?

boot/zephyr/boards/nrf52840_pca10059.conf
```diff
@@ -28,3 +28,5 @@ CONFIG_USB_MASS_STORAGE=n
 
 # Logging
 CONFIG_SYS_LOG_DEFAULT_LEVEL=1
+
+CONFIG_MPU_ALLOW_FLASH_WRITE=y
```
nvlsianpu: Add to Kconfig (here: https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/zephyr/Kconfig#L12)
`select MPU_ALLOW_FLASH_WRIT if ARM_MPU`
so will be enabled for every ARM-MPU equipped device

boot/zephyr/main.c
rgundi: I have explored this option of fetching it from DT but it doesn't seem to be straight forward. It's possible to get it from Kconfig but would really like to do it from DT. Can that be deferred for a while? I can add a macro for now.

boot/zephyr/main.c
rgundi: Yes. This is. I will fetch it from DT. thanks for pointing it out.

boot/zephyr/boards/nrf52840_pca10059.conf
rgundi: Done. much cleaner way to fix it. Thanks.

boot/zephyr/main.c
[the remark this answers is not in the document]
rgundi: agreed.

boot/zephyr/boards/intel_s1000_crb.conf
rgundi: using DT_SPI_NOR_DRV_NAME for s1000 instead of DT_FLASH_DEV_NAME. In my opinion, we should have 2 classes of names, one for on-chip flash (DT_FLASH_DEV_NAME) and one for off-chip flash (DT_SPI_NOR_DRV_NAME) which sits on the SPI bus.

boot/zephyr/flash_map_extended.c
rgundi: Done.

boot/zephyr/include/target.h
rgundi: Done.

boot/zephyr/boards/intel_s1000_crb.conf
nvlsianpu: all right

boot/zephyr/dts.overlay
```diff
@@ -13,5 +13,13 @@
 	chosen {
 		zephyr,code-partition = &boot_partition;
 	};
+
+#ifdef CONFIG_SOC_INTEL_S1000
```
nvlsianpu: better add  dedicated overlay for this board than put Kconfi postprocesor macro. Zephyr is running away from using kconfig inside of DT.

nvlsianpu: obviously need to modify cmakelists.txtx as well

boot/zephyr/main.c
rgundi: done

boot/zephyr/dts.overlay
nvlsianpu: @rgundi ^^

rgundi: @nvlsianpu addressed.

nvlsianpu: I've got error while was building this:
```
-- Loading /home/anpu/codedev/zephyr/boards/xtensa/intel_s1000_crb/intel_s1000_crb.dts as base
-- Overlaying /home/anpu/codedev/zephyr/dts/common/common.dts
-- Overlaying /home/anpu/codedev/mcuboot/boot/zephyr/dts.overlay
-- Overlaying /home/anpu/codedev/mcuboot/boot/zephyr/boards/intel_s1000_crb.overlay
intel_s1000_crb.dts_compiled: ERROR (path_references): /chosen: Reference to non-existent node or label "boot_partition"

ERROR: Input tree has errors, aborting (use -f to force output)
```

rgundi: Not sure why you are getting that error. I tested building for 3 different boards viz. intel_s1000_crb, frdm_k64f and nrf52840_pca10056. It builds fine on all three.
Did you try rebasing?

rgundi: oh.. i get it. You are getting that error because you don't have the corresponding Zephyr changes. There's a PR under review in Zephyr (https://github.com/zephyrproject-rtos/zephyr/pull/11850). Please pull that into your zephyr tree and then compile. It should pass.

PR #369, opened 2018-12-13
Introduce multi-image boot

davidvincze:
This change replaces the slot 0/1 terminology with primary/secondary
slot and replaces FLASH_AREA_IMAGE_0/1 with
FLASH_AREA_IMAGE_PRIMARY/SECONDARY. This naming convention
may be more understandable, fits better to MCUs with multiple images
and it is an architecture agnostic alternative as well.

Signed-off-by: David Vincze <david.vincze@arm.com>

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright (c) 2019 Arm Limited.
+ */
+
 #ifndef H_IMAGE_
 #define H_IMAGE_
 
@@ -76,6 +80,7 @@ struct flash_area;
 #define IMAGE_TLV_ED25519           0x24   /* ed25519 of hash output */
 #define IMAGE_TLV_ENC_RSA2048       0x30   /* Key encrypted with RSA-OAEP-2048 */
 #define IMAGE_TLV_ENC_KW128         0x31   /* Key encrypted with AES-KW-128 */
+#define IMAGE_TLV_DEPENDENCY        0x40   /* Image depends on other image */
 
 struct image_version {
     uint8_t iv_major;
@@ -84,16 +89,24 @@ struct image_version {
     uint32_t iv_build_num;
 };
 
+struct image_dependency {
+    uint8_t image_id;                       /* Image index (from 0) */
+    struct image_version image_min_version; /* Indicates at minimum which
+                                             * version of firmware must be
+                                             * available to satisfy compliance
+                                             */
+};
+
 /** Image header.  All fields are in little endian byte order. */
 struct image_header {
     uint32_t ih_magic;
     uint32_t ih_load_addr;
-    uint16_t ih_hdr_size; /* Size of image header (bytes). */
-    uint16_t _pad1;
-    uint32_t ih_img_size; /* Does not include header. */
-    uint32_t ih_flags;    /* IMAGE_F_[...]. */
+    uint16_t ih_hdr_size;           /* Size of image header (bytes). */
+    uint16_t ih_protect_tlv_size;   /* Size of protected TLV area (bytes). */
+    uint32_t ih_img_size;           /* Does not include header. */
+    uint32_t ih_flags;              /* IMAGE_F_[...]. */
     struct image_version ih_ver;
-    uint32_t _pad2;
+    uint32_t _pad1;
 };
 
 /** Image TLV header.  All fields in little endian. */
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -172,13 +172,12 @@ boot_magic_off(const struct flash_area *fap)
 int
 boot_status_entries(const struct flash_area *fap)
 {
-    switch (fap->fa_id) {
-    case FLASH_AREA_IMAGE_PRIMARY:
-    case FLASH_AREA_IMAGE_SECONDARY:
-        return BOOT_STATUS_STATE_COUNT * BOOT_STATUS_MAX_ENTRIES;
-    case FLASH_AREA_IMAGE_SCRATCH:
+    if (fap->fa_id == FLASH_AREA_IMAGE_SCRATCH) {
         return BOOT_STATUS_STATE_COUNT;
-    default:
+    } else if ((fap->fa_id == FLASH_AREA_IMAGE_PRIMARY) ||
+               (fap->fa_id == FLASH_AREA_IMAGE_SECONDARY)) {
+        return BOOT_STATUS_STATE_COUNT * BOOT_STATUS_MAX_ENTRIES;
+    } else {
         return BOOT_EBADARGS;
     }
 }
@@ -301,16 +300,14 @@ boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state)
     const struct flash_area *fap;
     int rc;
 
-    switch (flash_area_id) {
-    case FLASH_AREA_IMAGE_SCRATCH:
-    case FLASH_AREA_IMAGE_PRIMARY:
-    case FLASH_AREA_IMAGE_SECONDARY:
+    if (flash_area_id == FLASH_AREA_IMAGE_SCRATCH ||
+        flash_area_id == FLASH_AREA_IMAGE_PRIMARY ||
+        flash_area_id == FLASH_AREA_IMAGE_SECONDARY) {
         rc = flash_area_open(flash_area_id, &fap);
         if (rc != 0) {
             return BOOT_EFLASH;
         }
-        break;
-    default:
+    } else {
         return BOOT_EBADARGS;
     }
 
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -741,3 +741,38 @@ done:
     flash_area_close(fap);
     return rc;
 }
+
+#if (BOOT_IMAGE_NUMBER > 1)
+/**
+ * Check if the version of the image is not older than required.
+ *
+ * @param req         Required minimal image version.
+ * @param ver         Version of the image to be checked.
+ *
+ * @return            0 if the version is sufficient, nonzero otherwise.
+ */
+int
+boot_is_version_sufficient(struct image_version *req,
+                           struct image_version *ver)
+{
+    if (ver->iv_major > req->iv_major) {
+        return 0;
+    }
+    if (ver->iv_major < req->iv_major) {
+        return BOOT_EBADVERSION;
+    }
+    /* The major version numbers are equal. */
+    if (ver->iv_minor > req->iv_minor) {
+        return 0;
+    }
+    if (ver->iv_minor < req->iv_minor) {
+        return BOOT_EBADVERSION;
+    }
+    /* The minor version numbers are equal. */
+    if (ver->iv_revision < req->iv_revision) {
+        return BOOT_EBADVERSION;
+    }
+
+    return 0;
+}
+#endif /* BOOT_IMAGE_NUMBER > 1 */
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright (c) 2019 Arm Limited.
+ */
+
 #include <assert.h>
 #include <string.h>
 #include <inttypes.h>
@@ -194,7 +198,7 @@ boot_status_off(const struct flash_area *fap)
 }
 
 uint32_t
-boot_swap_type_off(const struct flash_area *fap)
+boot_swap_info_off(const struct flash_area *fap)
 {
     return fap->fa_size - BOOT_MAGIC_SZ - BOOT_MAX_ALIGN * 3;
 }
@@ -232,6 +236,7 @@ boot_read_swap_state(const struct flash_area *fap,
 {
     uint32_t magic[BOOT_MAGIC_ARR_SZ];
     uint32_t off;
+    uint8_t swap_info;
     int rc;
 
     off = boot_magic_off(fap);
@@ -245,14 +250,19 @@ boot_read_swap_state(const struct flash_area *fap,
         state->magic = boot_magic_decode(magic);
     }
 
-    off = boot_swap_type_off(fap);
-    rc = flash_area_read_is_empty(fap, off, &state->swap_type,
-            sizeof state->swap_type);
+    off = boot_swap_info_off(fap);
+    rc = flash_area_read_is_empty(fap, off, &swap_info, sizeof swap_info);
     if (rc < 0) {
         return BOOT_EFLASH;
     }
+
+    /* Extract the swap type and image number */
+    state->swap_type = BOOT_GET_SWAP_TYPE(swap_info);
+    state->image_num = BOOT_GET_IMAGE_NUM(swap_info);
+
     if (rc == 1 || state->swap_type > BOOT_SWAP_TYPE_REVERT) {
         state->swap_type = BOOT_SWAP_TYPE_NONE;
+        state->image_num = 0;
     }
 
     off = boot_copy_done_off(fap);
@@ -494,14 +504,18 @@ boot_write_image_ok(const struct flash_area *fap)
  * resume in case of an unexpected reset.
  */
 int
-boot_write_swap_type(const struct flash_area *fap, uint8_t swap_type)
+boot_write_swap_info(const struct flash_area *fap, uint8_t swap_type,
+                     uint8_t image_num)
 {
     uint32_t off;
+    uint8_t swap_info;
 
-    off = boot_swap_type_off(fap);
-    BOOT_LOG_DBG("writing swap_type; fa_id=%d off=0x%x (0x%x), swap_type=0x%x",
-                 fap->fa_id, off, fap->fa_off + off, swap_type);
-    return boot_write_trailer_byte(fap, off, swap_type);
+    BOOT_SET_SWAP_INFO(swap_info, image_num, swap_type);
+    off = boot_swap_info_off(fap);
+    BOOT_LOG_DBG("writing swap_info; fa_id=%d off=0x%x (0x%x), swap_type=0x%x"
+                 " image_num=0x%x",
+                 fap->fa_id, off, fap->fa_off + off, swap_type, image_num);
+    return boot_write_trailer_byte(fap, off, swap_info);
 }
 
 int
@@ -648,7 +662,7 @@ boot_set_pending(int permanent)
             } else {
                 swap_type = BOOT_SWAP_TYPE_TEST;
             }
-            rc = boot_write_swap_type(fap, swap_type);
+            rc = boot_write_swap_info(fap, swap_type, 0);
         }
 
         flash_area_close(fap);
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -48,13 +48,14 @@ extern "C" {
 
 struct flash_area;
 
-#define BOOT_EFLASH     1
-#define BOOT_EFILE      2
-#define BOOT_EBADIMAGE  3
-#define BOOT_EBADVECT   4
-#define BOOT_EBADSTATUS 5
-#define BOOT_ENOMEM     6
-#define BOOT_EBADARGS   7
+#define BOOT_EFLASH      1
+#define BOOT_EFILE       2
+#define BOOT_EBADIMAGE   3
+#define BOOT_EBADVECT    4
+#define BOOT_EBADSTATUS  5
+#define BOOT_ENOMEM      6
+#define BOOT_EBADARGS    7
+#define BOOT_EBADVERSION 8
 
 #define BOOT_TMPBUF_SZ  256
 
@@ -240,6 +241,10 @@ int boot_read_swap_size(uint32_t *swap_size);
 int boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
                        const uint8_t *enckey);
 int boot_read_enc_key(uint8_t slot, uint8_t *enckey);
+#endif
+#if (BOOT_IMAGE_NUMBER > 1)
+int boot_is_version_sufficient(struct image_version *req,
+                               struct image_version *ver);
 #endif
 
 /*
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -134,6 +134,12 @@ struct boot_swap_state {
     uint8_t image_num;  /* Boot status belongs to this image */
 };
 
+#ifdef MCUBOOT_IMAGE_NUMBER
+#define BOOT_IMAGE_NUMBER          MCUBOOT_IMAGE_NUMBER
+#else
+#define BOOT_IMAGE_NUMBER          1
+#endif
+
 #define BOOT_MAX_IMG_SECTORS       MCUBOOT_MAX_IMG_SECTORS
 
 /*
@@ -302,23 +308,21 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
     int num_sectors = BOOT_MAX_IMG_SECTORS;
     int rc;
 
-    switch (flash_area) {
-    case FLASH_AREA_IMAGE_PRIMARY:
+    if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
                                    state->imgs[BOOT_PRIMARY_SLOT].sectors);
         state->imgs[BOOT_PRIMARY_SLOT].num_sectors = (size_t)num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SECONDARY:
+
+    } else if (flash_area == FLASH_AREA_IMAGE_SECONDARY) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
                                    state->imgs[BOOT_SECONDARY_SLOT].sectors);
         state->imgs[BOOT_SECONDARY_SLOT].num_sectors = (size_t)num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SCRATCH:
+
+    } else if (flash_area == FLASH_AREA_IMAGE_SCRATCH) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
                                    state->scratch.sectors);
         state->scratch.num_sectors = (size_t)num_sectors;
-        break;
-    default:
+    } else {
         return BOOT_EFLASH;
     }
 
@@ -350,24 +354,19 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
     size_t *out_num_sectors;
     int rc;
 
-    switch (flash_area) {
-    case FLASH_AREA_IMAGE_PRIMARY:
-        num_sectors = BOOT_MAX_IMG_SECTORS;
+    num_sectors = BOOT_MAX_IMG_SECTORS;
+
+    if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
         out_sectors = state->imgs[BOOT_PRIMARY_SLOT].sectors;
         out_num_sectors = &state->imgs[BOOT_PRIMARY_SLOT].num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SECONDARY:
-        num_sectors = BOOT_MAX_IMG_SECTORS;
+    } else if (flash_area == FLASH_AREA_IMAGE_SECONDARY) {
         out_sectors = state->imgs[BOOT_SECONDARY_SLOT].sectors;
         out_num_sectors = &state->imgs[BOOT_SECONDARY_SLOT].num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SCRATCH:
-        num_sectors = BOOT_MAX_IMG_SECTORS;
+    } else if (flash_area == FLASH_AREA_IMAGE_SCRATCH) {
         out_sectors = state->scratch.sectors;
         out_num_sectors = &state->scratch.num_sectors;
-        break;
-    default:
-        return -1;
+    } else {
+        return BOOT_EFLASH;
     }
 
     rc = flash_area_get_sectors(flash_area, &num_sectors, out_sectors);
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -124,6 +124,7 @@ struct boot_status {
  *      (`MCUBOOT_ENC_IMAGES`).
  */
 
+extern uint8_t current_image;
 extern const uint32_t boot_img_magic[4];
 
 struct boot_swap_state {
@@ -202,7 +203,7 @@ struct boot_loader_state {
         const struct flash_area *area;
         boot_sector_t *sectors;
         size_t num_sectors;
-    } imgs[BOOT_NUM_SLOTS];
+    } imgs[BOOT_IMAGE_NUMBER][BOOT_NUM_SLOTS];
 
     struct {
         const struct flash_area *area;
@@ -210,6 +211,7 @@ struct boot_loader_state {
         size_t num_sectors;
     } scratch;
 
+    uint8_t swap_type[BOOT_IMAGE_NUMBER];
     uint8_t write_sz;
 };
 
@@ -245,20 +247,22 @@ int boot_read_enc_key(uint8_t slot, uint8_t *enckey);
  */
 
 /* These are macros so they can be used as lvalues. */
-#define BOOT_IMG_AREA(state, slot) ((state)->imgs[(slot)].area)
+#define BOOT_IMG(state, slot) ((state)->imgs[current_image][(slot)])
+#define BOOT_IMG_AREA(state, slot) (BOOT_IMG(state, slot).area)
 #define BOOT_SCRATCH_AREA(state) ((state)->scratch.area)
 #define BOOT_WRITE_SZ(state) ((state)->write_sz)
+#define BOOT_SWAP_TYPE(state) ((state)->swap_type[current_image])
 
 static inline struct image_header*
 boot_img_hdr(struct boot_loader_state *state, size_t slot)
 {
-    return &state->imgs[slot].hdr;
+    return &BOOT_IMG(state, slot).hdr;
 }
 
 static inline size_t
 boot_img_num_sectors(struct boot_loader_state *state, size_t slot)
 {
-    return state->imgs[slot].num_sectors;
+    return BOOT_IMG(state, slot).num_sectors;
 }
 
 static inline size_t
@@ -273,7 +277,7 @@ boot_scratch_num_sectors(struct boot_loader_state *state)
 static inline uint32_t
 boot_img_slot_off(struct boot_loader_state *state, size_t slot)
 {
-    return state->imgs[slot].area->fa_off;
+    return BOOT_IMG(state, slot).area->fa_off;
 }
 
 static inline size_t boot_scratch_area_size(struct boot_loader_state *state)
@@ -287,7 +291,7 @@ static inline size_t
 boot_img_sector_size(struct boot_loader_state *state,
                      size_t slot, size_t sector)
 {
-    return state->imgs[slot].sectors[sector].fa_size;
+    return BOOT_IMG(state, slot).sectors[sector].fa_size;
 }
 
 /*
@@ -298,8 +302,8 @@ static inline uint32_t
 boot_img_sector_off(struct boot_loader_state *state, size_t slot,
                     size_t sector)
 {
-    return state->imgs[slot].sectors[sector].fa_off -
-           state->imgs[slot].sectors[0].fa_off;
+    return BOOT_IMG(state, slot).sectors[sector].fa_off -
+           BOOT_IMG(state, slot).sectors[0].fa_off;
 }
 
 static inline int
@@ -310,13 +314,13 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
 
     if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
-                                   state->imgs[BOOT_PRIMARY_SLOT].sectors);
-        state->imgs[BOOT_PRIMARY_SLOT].num_sectors = (size_t)num_sectors;
+                                   BOOT_IMG(state, BOOT_PRIMARY_SLOT).sectors);
+        BOOT_IMG(state, BOOT_PRIMARY_SLOT).num_sectors = (size_t)num_sectors;
 
     } else if (flash_area == FLASH_AREA_IMAGE_SECONDARY) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
-                                   state->imgs[BOOT_SECONDARY_SLOT].sectors);
-        state->imgs[BOOT_SECONDARY_SLOT].num_sectors = (size_t)num_sectors;
+                                 BOOT_IMG(state, BOOT_SECONDARY_SLOT).sectors);
+        BOOT_IMG(state, BOOT_SECONDARY_SLOT).num_sectors = (size_t)num_sectors;
 
     } else if (flash_area == FLASH_AREA_IMAGE_SCRATCH) {
         rc = flash_area_to_sectors(flash_area, &num_sectors,
@@ -335,15 +339,15 @@ static inline size_t
 boot_img_sector_size(struct boot_loader_state *state,
                      size_t slot, size_t sector)
 {
-    return state->imgs[slot].sectors[sector].fs_size;
+    return BOOT_IMG(state, slot).sectors[sector].fs_size;
 }
 
 static inline uint32_t
 boot_img_sector_off(struct boot_loader_state *state, size_t slot,
                     size_t sector)
 {
-    return state->imgs[slot].sectors[sector].fs_off -
-           state->imgs[slot].sectors[0].fs_off;
+    return BOOT_IMG(state, slot).sectors[sector].fs_off -
+           BOOT_IMG(state, slot).sectors[0].fs_off;
 }
 
 static inline int
@@ -357,11 +361,11 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
     num_sectors = BOOT_MAX_IMG_SECTORS;
 
     if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
-        out_sectors = state->imgs[BOOT_PRIMARY_SLOT].sectors;
-        out_num_sectors = &state->imgs[BOOT_PRIMARY_SLOT].num_sectors;
+        out_sectors = BOOT_IMG(state, BOOT_PRIMARY_SLOT).sectors;
+        out_num_sectors = &BOOT_IMG(state, BOOT_PRIMARY_SLOT).num_sectors;
     } else if (flash_area == FLASH_AREA_IMAGE_SECONDARY) {
-        out_sectors = state->imgs[BOOT_SECONDARY_SLOT].sectors;
-        out_num_sectors = &state->imgs[BOOT_SECONDARY_SLOT].num_sectors;
+        out_sectors = BOOT_IMG(state, BOOT_SECONDARY_SLOT).sectors;
+        out_num_sectors = &BOOT_IMG(state, BOOT_SECONDARY_SLOT).num_sectors;
     } else if (flash_area == FLASH_AREA_IMAGE_SCRATCH) {
         out_sectors = state->scratch.sectors;
         out_num_sectors = &state->scratch.num_sectors;
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright (c) 2019 Arm Limited.
+ */
+
 #ifndef H_BOOTUTIL_PRIV_
 #define H_BOOTUTIL_PRIV_
 
@@ -106,7 +110,7 @@ struct boot_status {
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  *  |                      Swap size (4 octets)                     |
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
- *  |   Swap type   |           0xff padding (7 octets)             |
+ *  |   Swap info   |           0xff padding (7 octets)             |
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  *  |   Copy done   |           0xff padding (7 octets)             |
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
@@ -127,9 +131,25 @@ struct boot_swap_state {
     uint8_t swap_type;  /* One of the BOOT_SWAP_TYPE_[...] values. */
     uint8_t copy_done;  /* One of the BOOT_FLAG_[...] values. */
     uint8_t image_ok;   /* One of the BOOT_FLAG_[...] values. */
+    uint8_t image_num;  /* Boot status belongs to this image */
 };
 
 #define BOOT_MAX_IMG_SECTORS       MCUBOOT_MAX_IMG_SECTORS
+
+/*
+ * Extract the swap type and image number from image trailers's swap_info
+ * filed.
+ */
+#define BOOT_GET_SWAP_TYPE(swap_info)    ((swap_info) & 0x0F)
+#define BOOT_GET_IMAGE_NUM(swap_info)    ((swap_info) >> 4)
+
+/* Construct the swap_info field from swap type and image number */
+#define BOOT_SET_SWAP_INFO(swap_info, image, type)  {                          \
+                                                    assert((image) < 0xF);     \
+                                                    assert((type)  < 0xF);     \
+                                                    (swap_info) = (image) << 4 \
+                                                                | (type);      \
+                                                    }
 
 /*
  * The current flashmap API does not check the amount of space allocated when
@@ -194,7 +214,7 @@ int boot_magic_compatible_check(uint8_t tbl_val, uint8_t val);
 uint32_t boot_trailer_sz(uint8_t min_write_sz);
 int boot_status_entries(const struct flash_area *fap);
 uint32_t boot_status_off(const struct flash_area *fap);
-uint32_t boot_swap_type_off(const struct flash_area *fap);
+uint32_t boot_swap_info_off(const struct flash_area *fap);
 int boot_read_swap_state(const struct flash_area *fap,
                          struct boot_swap_state *state);
 int boot_read_swap_state_by_id(int flash_area_id,
@@ -204,7 +224,8 @@ int boot_write_status(struct boot_status *bs);
 int boot_schedule_test_swap(void);
 int boot_write_copy_done(const struct flash_area *fap);
 int boot_write_image_ok(const struct flash_area *fap);
-int boot_write_swap_type(const struct flash_area *fap, uint8_t swap_type);
+int boot_write_swap_info(const struct flash_area *fap, uint8_t swap_type,
+                         uint8_t image_num);
 int boot_write_swap_size(const struct flash_area *fap, uint32_t swap_size);
 int boot_read_swap_size(uint32_t *swap_size);
 #ifdef MCUBOOT_ENC_IMAGES
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright (c) 2019 Arm Limited.
+ */
+
 #include <assert.h>
 #include <stddef.h>
 #include <inttypes.h>
@@ -78,12 +82,20 @@ bootutil_img_hash(struct image_header *hdr, const struct flash_area *fap,
     }
 #endif
 
-    /*
-     * Hash is computed over image header and image itself. No TLV is
-     * included ATM.
-     */
+    /* Hash is computed over image header and image itself. */
     hdr_size = hdr->ih_hdr_size;
     size = hdr->ih_img_size + hdr_size;
+
+#if (MCUBOOT_IMAGE_NUMBER > 1)
+    /* If dependency TLVs are present then the TLV info header and the
+     * dependency TLVs are also protected and have to be included in the hash
+     * calculation.
+     */
+    if (hdr->ih_protect_tlv_size != 0) {
+        size += hdr->ih_protect_tlv_size;
+    }
+#endif
+
     for (off = 0; off < size; off += blk_sz) {
         blk_sz = size - off;
         if (blk_sz > tmp_buf_sz) {
@@ -212,7 +224,6 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
     }
 
     /* The TLVs come after the image. */
-    /* After image there are TLVs. */
     off = hdr->ih_img_size + hdr->ih_hdr_size;
 
     rc = flash_area_read(fap, off, &info, sizeof(info));
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1606,6 +1606,198 @@ out:
 }
 #endif /* !MCUBOOT_OVERWRITE_ONLY */
 
+#if (BOOT_IMAGE_NUMBER > 1)
+/**
+ * Check the image dependency whether it is satisfied and modify
+ * the swap type if necessary.
+ *
+ * @param dep               Image dependency which has to be verified.
+ *
+ * @return                  0 on success; nonzero on failure.
+ */
+static int
+boot_verify_single_dependency(struct image_dependency *dep)
+{
+    struct image_version *dep_version;
+    size_t dep_slot;
+    int rc;
+
+    /* Determine the source of the image which is the subject of
+     * the dependency and get it's version. */
+    dep_slot = (boot_data.swap_type[dep->image_id] != BOOT_SWAP_TYPE_NONE) ?
+                BOOT_SECONDARY_SLOT : BOOT_PRIMARY_SLOT;
+    dep_version = &boot_data.imgs[dep->image_id][dep_slot].hdr.ih_ver;
+
+    rc = boot_is_version_sufficient(&dep->image_min_version, dep_version);
+    if (rc != 0) {
+        /* Dependency not satisfied.
+         * Modify the swap type to decrease the version number of the image
+         * (which will be located in the primary slot after the boot process),
+         * consequently the number of unsatisfied dependencies will be
+         * decreased or remain the same.
+         */
+        switch (BOOT_SWAP_TYPE(&boot_data)) {
+        case BOOT_SWAP_TYPE_TEST:
+        case BOOT_SWAP_TYPE_PERM:
+            BOOT_SWAP_TYPE(&boot_data) = BOOT_SWAP_TYPE_NONE;
+            break;
+        case BOOT_SWAP_TYPE_NONE:
+            BOOT_SWAP_TYPE(&boot_data) = BOOT_SWAP_TYPE_REVERT;
+            break;
+        default:
+            break;
+        }
+    }
+
+    return rc;
+}
+
+/**
+ * Read all dependency TLVs of an image from the flash and verify
+ * one after another to see if they are all satisfied.
+ *
+ * @param slot              Image slot number.
+ *
+ * @return                  0 on success; nonzero on failure.
+ */
+static int
+boot_verify_all_dependency(uint32_t slot)
+{
+    const struct flash_area *fap;
+    struct image_header *hdr;
+    struct image_tlv_info info;
+    struct image_tlv tlv;
+    struct image_dependency dep;
+    uint32_t off;
+    uint32_t end;
+    bool dep_tlvs_found = false;
+    int rc;
+
+    rc = flash_area_open(flash_area_id_from_image_slot(slot), &fap);
+    if (rc != 0) {
+        rc = BOOT_EFLASH;
+        goto done;
+    }
+
+    hdr = boot_img_hdr(&boot_data, slot);
+    /* The TLVs come after the image. */
+    off = hdr->ih_hdr_size + hdr->ih_img_size;
+
+    /* The TLV area always starts with an image_tlv_info structure. */
+    rc = flash_area_read(fap, off, &info, sizeof(info));
+    if (rc != 0) {
+        rc = BOOT_EFLASH;
+        goto done;
+    }
+
+    if (info.it_magic != IMAGE_TLV_INFO_MAGIC) {
+        rc = BOOT_EBADIMAGE;
+        goto done;
+    }
+    end = off + info.it_tlv_tot;
+    off += sizeof(info);
+
+    /* Traverse through all of the TLVs to find the dependency TLVs. */
+    for (; off < end; off += sizeof(tlv) + tlv.it_len) {
+        rc = flash_area_read(fap, off, &tlv, sizeof(tlv));
+        if (rc != 0) {
+             rc = BOOT_EFLASH;
+             goto done;
+         }
+
+        if (tlv.it_type == IMAGE_TLV_DEPENDENCY) {
+            if (!dep_tlvs_found) {
+                dep_tlvs_found = true;
+            }
+
+            if (tlv.it_len != sizeof(dep)) {
+                rc = BOOT_EBADIMAGE;
+                goto done;
+            }
+
+            rc = flash_area_read(fap, off + sizeof(tlv), &dep, tlv.it_len);
+            if (rc != 0) {
+                rc = BOOT_EFLASH;
+                goto done;
+            }
+
+            /* Verify dependency and modify the swap type if not satisfied. */
+            rc = boot_verify_single_dependency(&dep);
+            if (rc != 0) {
+                /* Dependency not satisfied. */
+                goto done;
+            }
+
+            /* Dependency satisfied, no action needed.
+             * Continue with the next TLV entry.
+             */
+        } else if (dep_tlvs_found) {
+            /* The dependency TLVs are contiguous in the TLV area. If a
+             * dependency had already been found and the last read TLV
+             * has a different type then there are no more dependency TLVs.
+             * The search can be finished.
+             */
+            break;
+        }
+    }
+
+done:
+    flash_area_close(fap);
+    return rc;
+}
+
+/**
+ * Verify whether the image dependencies in the TLV area are
+ * all satisfied and modify the swap type if necessary.
+ *
+ * @return                  0 if all dependencies are satisfied,
+ *                          nonzero otherwise.
+ */
+static int
+boot_verify_single_image_dependency(void)
+{
+    size_t slot;
+
+    /* Determine the source of the dependency TLVs. Those dependencies have to
+     * be checked which belong to the image that will be located in the primary
+     * slot after the firmware update process.
+     */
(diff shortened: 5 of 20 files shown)


utzig:
I haven't looked at the PR yet, but somehow I feel like `slot0`/`slot1` conveys about the same info as `primary`/`secondary`. What would make more of a difference would be calling them something like `main_slot`/`upgrade_slot` or similar terminology...

davidvincze:
You are right and I understand your point. However, before I lose my fingers :D would you ask others what
their opinions are to reach a consensus, please. After that if it's still requested then I'll be happy to modify it.

mkiiskila:
Hi David!

my opinion would be to not change the name at all. For better or worse, this current terminology is wide-spread outside this repository. This PR as it stands now would break all those who depend on the existing names. To prevent disruption, we'd have to add #define FLASH_AREA_IMAGE_PRIMARY FLASH_AREA_IMAGE_0 in some header file included by all. And in the end that would just make things more confusing. This repo uses one name, and all the others use the old name.

Another point I have for keeping the existing terms is that for MCUs which execute from RAM only, image slot 0/1 as names are a better choice. We'll want the benefits of signing/encryption for those platforms too, so I imagine such bootloader will be built from components found from within this repository as well.

davidvincze:
Personally I would prefer the `main_slot`, `upgrade_slot` terminology. As I see it know, I don't think it is necessary to number each image slot. Of course there should be a flash_area_id, but at a certain level of boot logic we should hide it and use `main_slot`, `upgrade_slot` with a `current_image` variable which could be used to perform a mapping between the slot and flash_area_id as well.

utzig:
So, the current terminology seems OK to me (under the current circumstances), even if it requires a bit of familiarity, which basically means reading one document. The point of this PR is to work on better terminology for the time when we have multiple image support. I would personally favour leaving this PR for later, and working on it together with the actual support for multiple image support. Working on this now seems a bit of "putting the cart before the horse".

tamasban:
We are opening soon more PR's with the multi image support. This terminology change can serve as base of those. I think it makes the coming code more understandable. We can wait to conclude but the rest of our changes already use this new terminology. So to push the multiple image code with old/current terminology would cause extra effort for us.

nvlsianpu:
I personally prefer to merge the PR once accepted. Agree witch @tamasban  - this will saves them possible refactoring of subsequent patches.

utzig:
Why would any refactoring be required? If you need this PR merged you can as well make its commits part of a larger PR later, and can keep rebasing when PRs are merged.

utzig:
@davidvincze @tamasban Since I was asked in other channels, let me clarify that I am not against doing this. My opinion is that this does not serve a purpose standalone, but only once multi image support is properly available. There is no need to refactor at all, if you need to you're doing something wrong. Just make this commit part of a bigger PR where it serves a purpose because the name changes will make more sense once the mechanism for multiple image is there. I would rather spend time reviewing all together than doing in step by step fashion.

tamasban:
@utzig @davidvincze 
Thanks Fabio, fair enough!
Then we will update this PR to include the rest of multiple image update. 
Currently we are struggling with debugging. We  have rebased to encryption feature and the combined test in the simulator are failing: encryption+swap+abort. 
How can we debug the simulator?  How can we find out when has it aborted the mcuboot execution? I would like to reproduce it in another simulator where I can step the code line by line.

davidvincze:
Additional comment for the previous simulator test error:
with encryption (enc-rsa, enc-kw) the perm_with_fails, perm_with_random_fails, revert_with_fails test are FAILED, but without encryption these test cases are PASSED as well.

utzig:
@davidvincze You might have to rebase on recent master (#370 fixed some related issues)

boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -26,13 +26,15 @@
 extern "C" {
 #endif
 
-/** Attempt to boot the contents of slot 0. */
+/** Attempt to boot the contents of primary slot. */
```
d3zd3z: I'll just put one of these here, but there is lots below. "slot 0" and "primary slot" function differently in English, which makes a majority of the uses below ungrammatical. The phrase "slot 0" is definite, and does not use an article. However, "primary slot" does require an article. "Attempt to boot the contents of the primary slot". I would suggest considering terminology that doesn't invoke a grammatical change, or if made, we have a lot more things to change in this commit.

davidvincze: Thank you @d3zd3z for the feedback. I will fix these grammatical faults, however before it I would like to reach a consensus regarding the terminology change.

davidvincze:
**Commits included in the PR**

- _Replace flash partitioning terminology:_
This change replaces the slot 0/1 terminology with primary/secondary slot.

-  _Extend flash layout for multiple images:_
Introduce the BOOT_IMAGE_NUMBER macro and current_image variable to support multiple updateable images and the associated extended flash layout.

- _Enable multi-image boot in overwrite only mode:_
Add the capability to handle multiple firmware images, to update them independently.
Currently this is supported in overwrite-only mode only.

- _Add support for dependency description:_
Add the ability to specify and add dependency TLVs to the firmware image.

- _Add dependency check to multi-image boot:_
Add the capability to check image dependencies in case of multi-image boot
(the dependencies are described with a new type of TLV).

davidvincze:
**Changes in code-base size**

|                                    | **TEXT** **[Byte]** | **DATA** **[Byte]** | **BSS** **[Byte]** |
| ---------------------------------- | ------------------- | ------------------- | ------------------ |
| Upgrade Only<br />Image number = 2 | 33954 (+3.8%)       | 1372 (+0.3%)        | 26648 (+8.8%)      |
| Upgrade Only<br/>Image number = 1  | 33910 (+3.7%)       | 1372 (+0.3%)        | 24496              |
| Upgrade Only<br/>(original)        | 32694               | 1368                | 24496              |
|                                    |                     |                     |                    |
| Swap Images<br/>Image number = 1   | 36594 (+3.6%)       | 1372 (+0.3%)        | 24504 (+0.0%)      |
| Swap Images<br/>(original)         | 35322               | 1368                | 24496              |

boot/boot_serial/test/src/testcases/boot_serial_img_msg.c
```diff
@@ -35,7 +35,7 @@ TEST_CASE(boot_serial_img_msg)
      * 00000018  63 6c 65 6e 1a 00 01 14  |clen....|
      * 00000020  e8 63 6f 66 66 00        |.coff.|
      */
-    static const uint8_t payload[] = {
```
d3zd3z: Although it is fine to fix whitespace errors like this, it should be done in a separate commit.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -156,13 +161,12 @@ boot_magic_off(const struct flash_area *fap)
...
 {
-    switch (fap->fa_id) {
-    case FLASH_AREA_IMAGE_0:
-    case FLASH_AREA_IMAGE_1:
-        return BOOT_STATUS_STATE_COUNT * BOOT_STATUS_MAX_ENTRIES;
-    case FLASH_AREA_IMAGE_SCRATCH:
+    if (fap->fa_id == FLASH_AREA_IMAGE_SCRATCH) {
         return BOOT_STATUS_STATE_COUNT;
-    default:
+    } else if ((fap->fa_id == FLASH_AREA_IMAGE_PRIMARY) ||
```
d3zd3z: There are two changes being made here. One is the terminology change, the other is a stylistic change from a switch to if statements. I would argue this patch should leave this as a switch statement, and feel free to propose a separate patch that changes the switch statement into if statements (in this particular case, the if statement probably generates more code than the switch, just because of the pointer indirection, if you put the fa_id into a temporary, the code would probably be the same).

boot/bootutil/src/bootutil_priv.h
```diff
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright  � 2019 Arm Limited.
```
d3zd3z: Similar comment to the other comments. There is a bit of a conflict here between Apache conventions, and what I believe many companies are going to require. I would be happier if instead of multiple ad-hoc blocks like this, significant contributors would just place an additional copyright line at the top of the file. That is a little harder to figure out how to word when there isn't an existing copyright.

Regardless of Apache origins of this code, we're going to need to come up with a standard way of including copyright markings, or we will be excluding significant contributors (including myself).

boot/bootutil/src/bootutil_priv.h
```diff
@@ -264,20 +286,21 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
...
-        rc = flash_area_to_sectors(flash_area, &num_sectors, state->imgs[0].sectors);
-        state->imgs[0].num_sectors = (size_t)num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_1:
-        rc = flash_area_to_sectors(flash_area, &num_sectors, state->imgs[1].sectors);
-        state->imgs[1].num_sectors = (size_t)num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SCRATCH:
-        rc = flash_area_to_sectors(flash_area, &num_sectors, state->scratch.sectors);
+    if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
```
d3zd3z: Again, the stylistic change from switch to ifs should be made in a separate change (probably a separate pull request).

boot/bootutil/src/bootutil_priv.h
```diff
@@ -309,24 +332,19 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
...
-    case FLASH_AREA_IMAGE_1:
-        num_sectors = BOOT_MAX_IMG_SECTORS;
-        out_sectors = state->imgs[1].sectors;
-        out_num_sectors = &state->imgs[1].num_sectors;
-        break;
-    case FLASH_AREA_IMAGE_SCRATCH:
-        num_sectors = BOOT_MAX_IMG_SECTORS;
+    num_sectors = BOOT_MAX_IMG_SECTORS;
+
+    if (flash_area == FLASH_AREA_IMAGE_PRIMARY) {
```
d3zd3z: Again, here, too. Separate change.

boot/zephyr/flash_map_extended.c
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018 Nordic Semiconductor ASA
  * Copyright (c) 2015 Runtime Inc
+ * Copyright (c) 2019 Arm Limited
```
d3zd3z: This is more in line with how I think we should be doing the copyrights. I would recommend participants update the date ranges when they make modifications, although that is more an issue for the individual contributors to decide. And in this particular instance, it is debatable whether this constitutes a copyrightable work, but having the copyright lines should be a problem.

d3zd3z:
BTW, efforts to trivially rebase this on the tip failed, and I didn't get something that would work with the latest Zephyr. So, I think a rebase is in good order as well.

boot/bootutil/src/bootutil_misc.c
davidvincze: There are five separate commits in this pull request. The terminology change and this "from switch to if" change are in two different commits.

The last change is included in the **Boot: Extend flash layout for multiple images** commit and the main reason for this:
"The FLASH_AREA_IMAGE_* object-like macros are replaced with
function-like ones and therefore some functions have been updated,
because the labels of a switch statement have to be constant
expressions."

boot/boot_serial/test/src/testcases/boot_serial_img_msg.c
davidvincze: I am going to transfer such modifications into a separate commit.

boot/bootutil/src/bootutil_priv.h
davidvincze: Just like in the previous case, the main reason for this:
"The FLASH_AREA_IMAGE_* object-like macros are replaced with
function-like ones and therefore some functions have been updated,
because the labels of a switch statement have to be constant
expressions."

boot/bootutil/src/bootutil_priv.h
davidvincze: Just like in the previous cases, the main reason for this:
"The FLASH_AREA_IMAGE_* object-like macros are replaced with
function-like ones and therefore some functions have been updated,
because the labels of a switch statement have to be constant
expressions."

boot/bootutil/src/bootutil_priv.h
davidvincze: Thank you, could you share your thoughts with the rest of the MCUBoot community please? It would be useful to start a thread for example on Slack in this topic.

Would your initial suggestion be the one line copyright at the top of the file just like in mcuboot_config.h (mcuboot/boot/zephyr/include/mcuboot_config/mcuboot_config.h) ?

d3zd3z:
I really don't like this global variable `current_image` that seems to try an hide that we are working with multiple images. I see why it can make the change look like less, because it is hidden away, but it also means that now most of the code depends on this particular piece of hidden state. Unfortunately, it does kind of fit in with how the code currently works, so maybe no isn't the time to fix this.

boot/bootutil/src/bootutil_priv.h
d3zd3z: I think a one line at the top is fine. I'm not sure what to do with files that don't start out with a copyright. Maybe we need some kind of "For additional copyrights, see NOTICE." Apache doesn't generally put copyrights on individual files, where as many companies will require this.

boot/bootutil/src/bootutil_misc.c
d3zd3z: Ok, this makes sense, I wonder if that should be clearer from the commit text. In fact, I'd be even happier with a separate commit, before the one that adds the multiple images, that just changes the switch statements to if-blocks, with an explanation as to why that will be needed.

davidvincze:
> I really don't like this global variable `current_image` that seems to try an hide that we are working with multiple images. I see why it can make the change look like less, because it is hidden away, but it also means that now most of the code depends on this particular piece of hidden state. Unfortunately, it does kind of fit in with how the code currently works, so maybe no isn't the time to fix this.

Yes, the main purpose of the `current_image`  variable would be to hide from the image update logic the fact that there can be more separate images.
I'm open to other suggestion on how this could be handled, but I would prefer to apply these modifications at a later time.

davidvincze:
@utzig , @nvlsianpu what do you think about the terminology change (slot names) now that the PR comprises this commit along with the multi-image update mechanism?

boot/bootutil/src/bootutil_misc.c
davidvincze: This commit
https://github.com/JuulLabs-OSS/mcuboot/pull/369/commits/1547aed7dbb37b618507aaa956d45e43ec90180aí
is before the one that adds multi-image support. 
It only adds the BOOT_IMAGE_NUMBER macro and current_image variable to MCUBoot and the rest of the change is this "from switch to if" modification. It is the introduction of the current_image variable which creates the need for this and therefore, these together form a coherent commit.

boot/bootutil/src/bootutil_priv.h
davidvincze: For example in this file's license header there is this sentence:  "See the NOTICE file distributed with this work for additional information regarding copyright ownership. ". However, currently it is required of me to add an explicit copyright to these files. I don't know how could we solve this...

boot/bootutil/src/loader.c
```diff
@@ -1512,88 +1558,433 @@ boot_set_image_ok(void)
...
+
+    while(img_cnt < BOOT_IMAGE_NUMBER) {
+        boot_set_active_image(img_cnt);
+
+        if (boot_verify_single_image_dependency() == 0) {
+            /* All dependencies've been satisfied, continue with next image. */
+            img_cnt++;
+        } else {
+            /* Dependency check needs to be restarted. */
+            img_cnt = 0;
```
d3zd3z: The only time I've ever hit this line, it just causes this code to loop forever. I don't think this code is right, and will probably just loop forever if the dependencies haven't been met. I have yet to write tests explicitly for this, but by just having a SUIT manifest, it is easy to get this to loop.

d3zd3z:
@davidvincze, before making any changes to the terminology change, please let me know. I have a rebased tree and have pushed that change to other branches, and we need to make sure to coordinate any changes across them.

d3zd3z:
I've gotten the simulator #435 changes to work with this code (there are a few merge conflicts that aren't difficult to resolve). Many of the test cases pass, but the test "perm_with_fails" fails about half of its test cases:

```
ERROR 2019-03-05T23:03:28Z: bootsim::image: 54 out of 114 failed 47.37%
```

I have not yet written any additional tests to specifically test the funtionality of multi image (making sure dependencies work, adding dependencies to the manifests). It is possible the above failures happen when one image upgrades and the other doesn't, and that is possibly valid because I haven't specified dependencies.

d3zd3z:
Another related issue, we need to make sure that adding this code doesn't make the single-image code any larger. Right now, the dependency checking code is being linked in, even with one image, and this will probably prevent this from working at all on the most memory constrained devices.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -688,3 +696,28 @@ boot_set_confirmed(void)
...
+/**
+ * Check if the version of the image is not older than required.
+ *
+ * @param req         Required minimal image version.
+ * @param ver         Version of the image to be checked.
+ *
+ * @return            0 if the version is sufficient, nonzero otherwise.
+ */
+int
+boot_is_version_sufficient(struct image_version *req,
```
oyvindronningstad: I'm not too familiar with this patch, but I think maybe this logic is wrong. The current logic will e.g. judge 2.0.0.0 to be older than 1.0.0.1.

boot/bootutil/include/bootutil/image.h
```diff
@@ -82,16 +87,24 @@ struct image_version {
     uint32_t iv_build_num;
 };
 
+struct image_dependency {
```
oyvindronningstad: I think we need to solve these problems as well:
 - What if an update introduces breaking changes to an image that other images depend on?
 - What if the incompatible `foo` and `bar` images could both be found under the same `image_id` inside the user's ecosystem?

Possible solutions:
 - Add a Software ID requirement to this struct, that must match an ID in the TLV of the dependency image. This can become complex, especially if updates are allowed to change the ID of an image, but is also the most flexible.
 - Have a maximum version, either:
   - Add an `image_max_version`, or
   - Have an implicit max version, e.g. an image must have the same major version as the requirement.

boot/bootutil/src/loader.c
```diff
@@ -1607,158 +1998,178 @@ boot_swap_if_needed(int *out_swap_type)
...
+        boot_data->imgs[BOOT_SECONDARY_SLOT].sectors =
+                                            secondary_slot_sectors[img_cnt];
+
+        boot_data->scratch.sectors = scratch_sectors;
 
 #ifdef MCUBOOT_ENC_IMAGES
-    /* FIXME: remove this after RAM is cleared by sim */
-    boot_enc_zeroize();
+        /* FIXME: remove this after RAM is cleared by sim */
+        boot_enc_zeroize();
```
oyvindronningstad: Does this need to be done here now, since it's also done in `boot_set_active_image()`?

docs/design.md
```diff
@@ -414,27 +444,115 @@ Procedure:
...
+is a bit more complex but similar to the previously described procedure with
+one image. Every image can be updated independently therefore the flash is
+partitioned further to arrange two slots for each image.
+```
++--------------------+
+| MCUBoot            |
++--------------------+
+        ~~~~~            <- memory might be not contiguous
++--------------------+
+| Image 0            |
```
oyvindronningstad: How much work would it be to allow the primary and secondary slots of an image to be placed apart from one another, e.g. placing all the primary slots together, and all the secondary slots together? I think it will be easier for build systems if the active images don't have to be interspersed with secondary slots.

boot/bootutil/src/loader.c
```diff
@@ -1512,88 +1558,433 @@ boot_set_image_ok(void)
...
+/**
+ * Iterate over all the images and verify whether the image dependencies in the
+ * TLV area are all satisfied and update the related swap type if necessary.
+ */
+static void
+boot_verify_all_image_dependency(void)
+{
+    size_t img_cnt = 0;
+
+    while(img_cnt < BOOT_IMAGE_NUMBER) {
```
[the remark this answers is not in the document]
davidvincze: You are right, I'm going to update the change to include the dependency check only if the BOOT_IMAGE_NUMBER is greater than 1.

boot/bootutil/src/loader.c
davidvincze: There is always at least 1 valid configuration, it is assumed that the initial state is consistent. In the worst case, the system returns to this initial state after dependency check.
The logic in the boot_is_version_sufficient() function was wrong... I've updated it. I guess this is why it stucked in a loop.

davidvincze:
> Another related issue, we need to make sure that adding this code doesn't make the single-image code any larger. Right now, the dependency checking code is being linked in, even with one image, and this will probably prevent this from working at all on the most memory constrained devices.

I'm going to update the code to include the dependency check feature only if the bootloader is handling multiple images (when BOOT_IMAGE_NUMBER is greater than 1.).

boot/bootutil/src/bootutil_misc.c
davidvincze: You're absolutely right... I'm going to update my change. Thank you.

boot/bootutil/include/bootutil/image.h
davidvincze: > What if the incompatible foo and bar images could both be found under the same image_id inside the user's ecosystem?

Basically, the `image_id` here is just an index of the image areas in the flash. It refers to one specific flash area.

davidvincze: > What if an update introduces breaking changes to an image that other images depend on?

Could you clarify this situation, please? What update strategy is used?

boot/bootutil/src/loader.c
davidvincze: No, I've removed it.

boot/bootutil/include/bootutil/image.h
oyvindronningstad: > Could you clarify this situation, please?

Imagine the following: The application (image0) depends on library `foo` (image1). `foo` follows semantic versioning, which means that all versions `n.a.b` are backwards compatible with older versions `n.c.d` as long as the major version `n` remains unchanged. However, if the major version changes, `foo` cannot be guaranteed to be compatible with the application any more. With the current `image_dependency` struct, there is no way to express this fact. So if `foo` is upgradeable without upgrading the app, it might be upgraded to an incompatible version.

oyvindronningstad: > Basically, the image_id here is just an index of the image areas in the flash. It refers to one specific flash area.

Exactly. We need something extra, that helps uniquely identify a particular library.

oyvindronningstad:
cc @hakonfam @sigvartmh @osaether

docs/design.md
tamasban: This can be the done by configuring the build system in the way how you suggested. MCUboot gets the address of the  slots from the platform layer.  So it is up to the user. In the design doc it is just a possible memory layout.

davidvincze: I would say minimal. Nothing depends on the exact location of the image slots in the flash. Every time it executes flash operations it uses the addresses and sizes from the flash map based on the flash area id.

boot/bootutil/include/bootutil/image.h
tamasban: > * What if the incompatible `foo` and `bar` images could both be found under the same `image_id` inside the user's ecosystem?

Image_id reflects to the image area in the flash. It is assumed that the minimum update-able unit is one flash area. If that are contains more than one executable then it is the user responsibility to assure their compatibility. They can be updated only together.

oyvindronningstad: > If that are contains more than one executable

Sorry, I meant if one or the other could be placed in the same image area. For example, if a future update re-purposes the area for another library. Or the user has two classes of devices, which place different libraries in that image area.

docs/design.md
oyvindronningstad: Ok, thanks @tamasban and @davidvincze. Good to hear it's flexible.

boot/bootutil/include/bootutil/image.h
tamasban: > Imagine the following: The application (image0) depends on library `foo` (image1). `foo` follows semantic versioning, which means that all versions `n.a.b` are backwards compatible with older versions `n.c.d` as long as the major version `n` remains unchanged. However, if the major version changes, `foo` cannot be guaranteed to be compatible with the application any more. With the current `image_dependency` struct, there is no way to express this fact. So if `foo` is upgradeable without upgrading the app, it might be upgraded to an incompatible version.

Let say these are compatible
app 1.0.0 - 1.x.x  -  foo 2.0.0 - 2.x.x
app 2.0.0 - 2.x.x  -  foo 3.0.0 - 3-x.x.

In this case user must append a dependency TLV to foo newest (already incompatible) version of foo (3.x.x) to mandate a newer version of the app (at least 2.x.x) . So the user must be careful to attach all the required minimum dependencies to a new image which is not backward compatible. 

Other thing is that user do not use semantic versioning then you might cannot specify the maximum version in  advanced.

d3zd3z: It is interesting to note that, at least currently, SUIT dismisses entirely with versions and any complexity around this in regards to dependencies. The version field is a simple unsigned integer (can be arbitrarily large, though) and is just compared for upgrades. For dependencies, they just specify an exact image that they depend upon.

I'm not sure the bootloader is really the place to try and have meaningful semantic versioning, especially since we don't have enough information in our version structure to represent a proper semantic version.

The code to compare semantic versions is fairly complex, and is certainly not what our current `boot_is_version_sufficient` function is doing (I'll comment on the function itself).

If we do really think there are requirements for richer dependency information, we should probably engage with SUIT so that also gets represented in that manifest.

boot/bootutil/src/bootutil_misc.c
d3zd3z: https://gist.github.com/d3zd3z/2567782363b3bf632eae6ee24b7187df should at least do the comparison in the logical manner, which isn't what semantic versions checks, but does compare them lexicographicaly.

However, I'm not sure that the build num is supposed to be included as part of the comparison, at least as this field was originally intended. Hopefully someone from MyNewt can chime in on the intent.

boot/bootutil/src/loader.c
```diff
@@ -1818,16 +1985,16 @@ split_go(int loader_slot, int split_slot, void **entry)
     if (sectors == NULL) {
         return SPLIT_GO_ERR;
     }
-    boot_data.imgs[loader_slot].sectors = sectors + 0;
-    boot_data.imgs[split_slot].sectors = sectors + BOOT_MAX_IMG_SECTORS;
+    boot_data->imgs[loader_slot].sectors = sectors + 0;
```
hakonfam: Perhaps the `imgs` field in the `boot_loader_state` struct should be changed to `slots` or something? Currently it is quite confusing to be operating with two levels of `image`/`img`. Especially since it actually is slots not images being indexed there.

scripts/imgtool/image.py
```diff
@@ -85,11 +87,12 @@ def get(self):
 class Image():
 
     def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE,
-                 pad_header=False, pad=False, align=1, slot_size=0,
-                 max_sectors=DEFAULT_MAX_SECTORS, overwrite_only=False,
-                 endian="little"):
+                 dependencies={}, pad_header=False, pad=False, align=1,
```
hakonfam: I guess it would be better to use

```
dependencies=None,
...
if dependencies is None:
    dependencies = dict()
```
Not sure if this antipattern applies to constructors as well though. But I guess the dict would be appended to by subsequent calls.

Ref https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

scripts/imgtool/image.py
```diff
@@ -168,11 +173,48 @@ def check(self):
                         len(self.payload), tsize, self.slot_size)
                 raise Exception(msg)
 
-    def create(self, key, enckey):
-        self.add_header(enckey)
+    def create(self, key, enckey, dependencies={}):
+        if dependencies == {}:
```
hakonfam: Again, avoid mutable default arguments.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

boot/bootutil/include/bootutil/image.h
oyvindronningstad: > In this case user must append a dependency TLV to foo newest (already incompatible) version of foo (3.x.x) to mandate a newer version of the app (at least 2.x.x) . So the user must be careful to attach all the required minimum dependencies to a new image which is not backward compatible.

You're right, this does solve the problem. I see a few issues, though.

As a general observation: The "backwards" dependency from foo on the app is not an accurate expression of the relationship between foo and the app, since strictly speaking foo works correctly regardless of the app. Basically, it complicates the dependency tree (cyclic dependencies).

This will cause problems whenever you want to separate responsibility, i.e. when you have multiple FW owners. The maintainer of foo doesn't want to have to sign individual packages for each user of foo. Instead they want to create a single package for a given version, and sign it once.

> Other thing is that user do not use semantic versioning then you might cannot specify the maximum version in advanced.

This is only partially true IMO, just look at semantic versioning, if a new version of foo introduces breaking changes, the major version must be bumped. This is known beforehand by the app.

oyvindronningstad: > It is interesting to note that, at least currently, SUIT dismisses entirely with versions and any complexity around this in regards to dependencies. The version field is a simple unsigned integer (can be arbitrarily large, though) and is just compared for upgrades. For dependencies, they just specify an exact image that they depend upon.

Yes, MCUBoot is ahead of SUIT here.

> I'm not sure the bootloader is really the place to try and have meaningful semantic versioning, 

We need to set a limit somewhere with regards to complexity, I totally agree. I think that this patch, by introducing such rich a version scheme (major+minor+revision+build), should go further, in order to make the feature more complete. Using semver is one way, adding a max_version is another more flexible variant.

> especially since we don't have enough information in our version structure to represent a proper semantic version.

Right, there's no prerelease or build information, but we could have a more complete subset. We do have enough info to support semver with no pre-release versions (hyphens) or build metadata (plus signs)

> The code to compare semantic versions is fairly complex, and is certainly not what our current boot_is_version_sufficient function is doing (I'll comment on the function itself).

Adding the support for the above subset should not be hard, neither should supporting a max version. https://gist.github.com/oyvindronningstad/601ef3e1ef539db579b2316fc567852f

> If we do really think there are requirements for richer dependency information, we should probably engage with SUIT so that also gets represented in that manifest.

Definitely.

boot/bootutil/src/loader.c
```diff
@@ -1551,89 +1558,238 @@ boot_set_image_ok(void)
 #endif /* !MCUBOOT_OVERWRITE_ONLY */
 
 /**
- * Performs an image swap if one is required.
+ * Sets the active image on which the bootloader will continue to work.
+ *
+ * @param image_id                  The index of the image (from 0).
```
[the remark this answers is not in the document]
davidvincze: I rather just tried to follow the "convention". :)

scripts/imgtool/image.py
utzig: Image is only instantiated once, and `create` also only called once, so you could pass it like this, but the main issue seems to be one of design, which is: why are `dependencies` being passed to both the constructor and `create` function? You should decide which one to use...

boot/bootutil/src/loader.c
```diff
@@ -1551,89 +1558,238 @@ boot_set_image_ok(void)
...
+    boot_data = &boot_data_array[image_id];
+
+#ifdef MCUBOOT_ENC_IMAGES
+    /* The keys used for encryption are no longer valid (may belong to
+     * another image).
+     */
+    boot_enc_zeroize();
+#endif
+
+    return;
```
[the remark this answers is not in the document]
davidvincze: If the secondary slot (slot1) is encrypted the keys are stored in an enc_key_data structure which will be marked as "valid" after the keys have been set (the keys won't be reloaded because they are "valid"). When the MCUBoot switches between images the keys must be reloaded.; therefore, it clears the encryption keys and the valid flag too.

boot/bootutil/src/loader.c
```diff
@@ -1551,89 +1558,238 @@ boot_set_image_ok(void)
...
+}
+
+/**
+ * Performs a clean (not aborted) image update.
  *
- * @param out_swap_type         On success, the type of swap performed gets
- *                                  written here.
+ * @param bs                    The current boot status.
```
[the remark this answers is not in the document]
davidvincze: " I rather just tried to follow the "convention". :) "

utzig:
@davidvincze Please rebase this on `master`

scripts/imgtool/image.py
davidvincze: I've updated the script, thank you for the feedback.

boot/bootutil/src/loader.c
davidvincze: I think it really makes sense, however I would appreciate it if it would be part of another following pull request. I'm afraid it would generate even more conflicts in this PR.

hakonfam: Makes sense.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
hakonfam: Hehe, I was way to ambiguous in my comment. I was wondering why the empty return statement was there.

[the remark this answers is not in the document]
davidvincze: I will remove it.

boot/bootutil/src/loader.c
```diff
@@ -1834,11 +1834,6 @@ boot_go(struct boot_rsp *rsp)
 
         boot_data->scratch.sectors = scratch_sectors;
 
-#ifdef MCUBOOT_ENC_IMAGES
-        /* FIXME: remove this after RAM is cleared by sim */
-        boot_enc_zeroize();
```
utzig: squash this commit into 77f5a52a87f3dde96182f5460841174c053f31fd

scripts/imgtool/image.py
```diff
@@ -27,10 +28,10 @@
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 
-IMAGE_MAGIC = 0x96f3b83d
-IMAGE_HEADER_SIZE = 32
-BIN_EXT = "bin"
-INTEL_HEX_EXT = "hex"
+IMAGE_MAGIC =         0x96f3b83d
```
utzig: Please refrain yourself from making stylistic changes, the formatting was already PEP-8 compliant, if in doubt run flake8.

scripts/imgtool/image.py
```diff
@@ -41,16 +42,17 @@
...
-        'ENCRSA2048': 0x30,
-        'ENCKW128': 0x31,
+        'KEYHASH':      0x01,
+        'SHA256':       0x10,
+        'RSA2048':      0x20,
+        'ECDSA224':     0x21,
+        'ECDSA256':     0x22,
+        'ENCRSA2048':   0x30,
+        'ENCKW128':     0x31,
+        'DEPENDENCY':   0x40
```
utzig: PEP-8 non-compliant...

scripts/imgtool/image.py
```diff
@@ -41,16 +42,17 @@
...
 }
 
-TLV_INFO_SIZE = 4
+TLV_INFO_SIZE =  4
```
utzig: PEP-8 non-compliant...

scripts/imgtool/main.py
```diff
@@ -105,6 +107,27 @@ def validate_header_size(ctx, param, value):
             "Minimum value for -H/--header-size is {}".format(min_hdr_size))
     return value
 
+def get_dependencies(ctx, param, value):
+    if value is None:
+        return None
+    else:
```
utzig: could also just be `if value:` and the block below (no return means `return None`)

utzig: or `if value is not None:` in case it could be zero...

scripts/imgtool/image.py
```diff
@@ -207,34 +246,35 @@ def create(self, key, enckey):
             img = bytes(self.payload[self.header_size:])
             self.payload[self.header_size:] = encryptor.update(img) + \
                                               encryptor.finalize()
+        if protected_tlv_size != 0:
+            self.payload += tlv.get()[protected_tlv_size:]
```
utzig: I guess it doesn't make a difference to test the value of `protected_tlv_size`, if it is zero you still get `self.payload += tlv.get()[0:]`, the whole list, or no?

scripts/imgtool/main.py
```diff
@@ -32,9 +34,9 @@ def gen_ecdsa_p256(keyfile, passwd):
 def gen_ecdsa_p224(keyfile, passwd):
     print("TODO: p-224 not yet implemented")
 
-
-valid_langs = ['c', 'rust']
-keygens = {
+dependencies = None
+valid_langs  = ['c', 'rust']
+keygens      = {
```
utzig: PEP-8

scripts/imgtool/image.py
```diff
@@ -168,11 +170,48 @@ def check(self):
...
+                    dependencies["versions"][i].revision,
+                    dependencies["versions"][i].build
+                    )
+                tlv.add('DEPENDENCY', payload)
+            # Full TLV size needs to be calculated in advance, because the
+            # header will be protected as well
+            tlv_header_size     = 4
+            payload_digest_size = 32
+            keyhash_size        = 32
+            cipherkey_size      = 32
```
utzig: PEP-8 ^

scripts/imgtool/image.py
```diff
@@ -168,11 +170,48 @@ def check(self):
...
+            # = 16 Bytes
+            dependencies_num = len(dependencies["images"])
+            protected_tlv_size = (dependencies_num * 16) + TLV_INFO_SIZE
+
+        self.add_header(enckey, protected_tlv_size)
 
         tlv = TLV(self.endian)
 
+        if protected_tlv_size != 0:
+            for i in range(0, dependencies_num):
```
utzig: `range(dependencies_num)` would be the same ;-)

scripts/imgtool/main.py
```diff
@@ -105,6 +107,27 @@ def validate_header_size(ctx, param, value):
...
+        if len(images) != len(raw_versions):
+            raise click.BadParameter(
+                "There's a mismatch between the number of \
+                dependency images and versions in: {}".format(value))
+        for raw_version in raw_versions:
+            try:
+                versions.append(decode_version(raw_version))
+            except ValueError as e:
+                raise click.BadParameter("{}".format(e))
+        if len(images) != 0:
```
utzig: You should throw `click.BadParameter` if `len(images) == 0` before the other tests above; so this test ^ can be removed and you wouldn't die trying to return a non existing `dependencies` variable below.

utzig: Btw, would be interesting to document how the dependencies are to be specified in the help page...

scripts/imgtool/main.py
```diff
@@ -105,6 +107,27 @@ def validate_header_size(ctx, param, value):
...
+            dependencies = dict()
+            dependencies["images"] = images
+            dependencies["versions"] = versions
```
utzig: Instead of using `"images"` and `"versions"` in all the dict accesses please add a const or helper for this...

scripts/imgtool/main.py
```diff
@@ -32,9 +34,9 @@ def gen_ecdsa_p256(keyfile, passwd):
 def gen_ecdsa_p224(keyfile, passwd):
     print("TODO: p-224 not yet implemented")
 
-
-valid_langs = ['c', 'rust']
-keygens = {
+dependencies = None
```
utzig: I don't think this global should be here, what purpose does it serve?

scripts/imgtool/image.py
davidvincze: Fixed.

scripts/imgtool/image.py
davidvincze: I will revert these changes.

scripts/imgtool/image.py
davidvincze: I will revert these changes.

scripts/imgtool/image.py
davidvincze: I will revert these changes.

scripts/imgtool/image.py
davidvincze: Yes, you are right. This test is unnecessary.

scripts/imgtool/main.py
davidvincze: I will revert these changes.

scripts/imgtool/image.py
davidvincze: Fixed.

scripts/imgtool/image.py
davidvincze: Thanks!

scripts/imgtool/main.py
davidvincze:
```
IMAGES = "images"
VERSIONS = "versions"
```
or
```
DEP_IMAGES = "images"
DEP_VERSIONS = "versions"
```
maybe? Or any other suggestion?

utzig: I would probaby use something like:

```
DEP_IMAGES_KEY = "images"
DEP_VERSIONS_KEY = "versions"
```
but all look good, as longs as it's not just text copy/pasted around...

scripts/imgtool/main.py
davidvincze: I will remove it.

scripts/imgtool/main.py
davidvincze: I like the "_KEY" suffix, I'm going to update the script with it.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -64,36 +64,37 @@ struct boot_swap_table {
  * This set of tables maps image trailer contents to swap operation type.
  * When searching for a match, these tables must be iterated sequentially.
  *
- * NOTE: the table order is very important. The settings in Slot 1 always
- * are priority to Slot 0 and should be located earlier in the table.
+ * NOTE: the table order is very important. The settings in secondary slot
+ * always are priority to primary slot and should be located earlier in the
```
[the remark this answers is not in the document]
davidvincze: I will include this fix in another future change.

boot/bootutil/src/loader.c
```diff
@@ -1551,89 +1558,238 @@ boot_set_image_ok(void)
...
+ * Prepare image to be updated if required.
+ *
+ * Prepare image to be updated if required with completing an image swap
+ * operation if one was aborted and/or determining the type of the
+ * swap operation.
+ *
+ * @param bs                    Pointer where the read and possibly updated
+ *                              boot status can be written to.
+ *
+ * @return                      0 on success; nonzero on failure.
```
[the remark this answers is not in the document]
davidvincze: It was a copy-paste mistake, I'd updated the Doxygen comment.

davidvincze:
The "imgtool: Add support for dependency description" (e270cdbb) patch has been moved to a separate [pull request](https://github.com/JuulLabs-OSS/mcuboot/pull/464).

utzig:
@davidvincze Please enable the multi-image test to run on the simulator by changing `.travis.yml` line 26 to:
```
env: MULTI_FEATURES="sig-rsa overwrite-only,sig-ecdsa overwrite-only,multiimage overwrite-only" TEST=sim
```

boot/bootutil/src/loader.c
```diff
@@ -1577,6 +1577,201 @@ boot_set_active_image(size_t image_id)
...
+/**
+ * Iterate over all the images and verify whether the image dependencies in the
+ * TLV area are all satisfied and update the related swap type if necessary.
+ */
+static void
+boot_verify_all_image_dependency(void)
+{
+    size_t img_cnt = 0;
+
+    while(img_cnt < BOOT_IMAGE_NUMBER) {
```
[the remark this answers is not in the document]
utzig: Also is this routine supposed to hang the bootloader on errors?

boot/bootutil/src/loader.c
```diff
@@ -1549,89 +1559,432 @@ boot_set_image_ok(void)
...
+     * slot after the firmware update process.
+     */
+    if (boot_data->swap_type != BOOT_SWAP_TYPE_NONE &&
+        boot_data->swap_type != BOOT_SWAP_TYPE_FAIL) {
+        slot = BOOT_SECONDARY_SLOT;
+    } else {
+        slot = BOOT_PRIMARY_SLOT;
+    }
+
+    rc = boot_verify_all_dependency(slot);
```
utzig: Could just `return boot_verify_all_dependency(slot);` and remove `rc` usage.

boot/bootutil/src/loader.c
```diff
@@ -1549,89 +1559,432 @@ boot_set_image_ok(void)
...
+    dep_slot = (boot_data_array[dep->image_id].swap_type !=
+                BOOT_SWAP_TYPE_NONE) ?
+                BOOT_SECONDARY_SLOT : BOOT_PRIMARY_SLOT;
+    dep_version = boot_img_version(&boot_data_array[dep->image_id],
+                                   dep_slot);
+
+    rc = boot_is_version_sufficient(&dep->image_min_version, dep_version);
+    if (rc != 0) {
+        /* Dependency not satisfied.
+         * Modify the swap type to decrease the version number of the image
```
utzig: Using `swap` here seems a bit confusing since this function is currently only enable for overwrite-only mode.

boot/bootutil/src/loader.c
```diff
@@ -1549,89 +1559,432 @@ boot_set_image_ok(void)
 #endif /* !MCUBOOT_OVERWRITE_ONLY */
 
 /**
- * Performs an image swap if one is required.
+ * Sets the active image on which the bootloader will continue to work.
+ *
+ * @param image_id                  The index of the image (from 0).
+ */
+static void
+boot_set_active_image(size_t image_id)
```
utzig: This function seems to be always called in the following pattern: `1) FOR EACH IMAGE; 2) SET_ACTIVE; 3) DO SOMETHING`. Maybe (1) and (2) could be a macro which in a single-image does nothing...

boot/bootutil/src/loader.c
```diff
@@ -1549,89 +1559,432 @@ boot_set_image_ok(void)
...
+{
+    current_image = image_id;
+    boot_data = &boot_data_array[image_id];
+
+#ifdef MCUBOOT_ENC_IMAGES
+    /* The keys used for encryption are no longer valid (may belong to
+     * another image).
+     */
+    boot_enc_zeroize();
```
utzig: I *think* this would only be required a single time (per image); maybe it could be done once before any swap is operated? Also right now there would be a single key for any number of images, so the only thing that would require reset would possibly be the `valid` state.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
davidvincze: > Also is this routine supposed to hang the bootloader on errors?

Could you elaborate this please?

docs/design.md
```diff
@@ -441,6 +449,85 @@ Procedure:
...
+       swap requested?
+        + Yes: Is the requested image valid (integrity and security check)?
+            + Yes:
+                + Set the previously determined swap type for the current image.
+                + Skip to next image.
+            + No:
+                + Erase invalid image.
+                + Persist failure of swap procedure to image trailers.
+                + Mark the swap type as `Fail`.
+                + Skip to next image.
```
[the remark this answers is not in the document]
davidvincze: I think it should not.  If Image N is valid and it does not depend on Image N-1 (which is invalid) then we can still perform the upgrade of Image N.

boot/bootutil/include/bootutil/image.h
```diff
@@ -82,16 +87,24 @@ struct image_version {
...
 /** Image header.  All fields are in little endian byte order. */
 struct image_header {
     uint32_t ih_magic;
     uint32_t ih_load_addr;
-    uint16_t ih_hdr_size; /* Size of image header (bytes). */
-    uint16_t _pad1;
-    uint32_t ih_img_size; /* Does not include header. */
-    uint32_t ih_flags;    /* IMAGE_F_[...]. */
+    uint16_t ih_hdr_size;           /* Size of image header (bytes). */
+    uint16_t ih_protect_tlv_size;   /* Size of protected TLV area (bytes). */
```
nvlsianpu: in place of _pad1

nvlsianpu: `ih_protect_tlv_size` field should be declared only if (BOOT_IMAGE_NUMBER > 1) is met.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -692,3 +696,43 @@ boot_set_confirmed(void)
...
+    }
+    /* The minor version numbers are equal. */
+    if (ver->iv_revision > req->iv_revision) {
+        return 0;
+    }
+    if (ver->iv_revision < req->iv_revision) {
+        return -1;
+    }
+    /* The revision numbers are equal. */
+    if (ver->iv_build_num < req->iv_build_num) {
```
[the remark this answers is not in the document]
nvlsianpu: I'm agree with that build number has only informative value.

boot/bootutil/include/bootutil/image.h
utzig: I don't think it is a problem to just change the name since `_pad1` was not used before anyway.

nvlsianpu: I see, right

boot/bootutil/src/loader.c
davidvincze: maybe something like this? 
```
ITERATE_THROUGH_IMAGES(img_cnt) {
....
}
```

boot/bootutil/include/bootutil/image.h
```diff
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+/*
+ * Modifications are Copyright  � 2019 Arm Limited.
```
d3zd3z: I think we should figure out how contributors should be inserting copyright statements. Rather than these kind of Ad-hoc declarations, what about the initial copyright lines, similar to the Linaro one in `boot/bootutil/src/caps.c`? The question that remains there is what do we put up there for the files that don't already have a copyright. I would guess that Runtime/Apache's lack of copyright statements will be a minority position for contributors. Perhaps we put in a "Modifications are Copyright" line at the top for those files, but it then gets funny as more and more of the work is done by these various parties, and they are no longer just modifications.

Another option would be to put either an Apache copyright, or a Runtime copyright at the tops of the files, depending on what is appropriate.

tamasban: The best would be to put similar copyright statement in all files in the repo like in `boot/bootutil/src/caps.c`. Then it would be straightforward to extend that list.

Can it be done with those files (retroactively), which currently lack of any copyright?

boot/bootutil/src/loader.c
davidvincze: I've created a boot_enc_mark_keys_invalid() function instead of using the boot_enc_zeroize(). In case of multi image boot it must be called in each main for cycles (in boot_go) where it swaps images or trying to validate the image in the secondary slot.

boot/bootutil/src/loader.c
davidvincze: The reorganization of the code made it unnecessary.

boot/bootutil/src/loader.c
davidvincze: It has been enabled for swapping mode too.

boot/bootutil/src/loader.c
davidvincze: I've updated to code.

boot/bootutil/include/bootutil/image.h
```diff
@@ -82,16 +87,24 @@ struct image_version {
     uint32_t iv_build_num;
 };
 
+struct image_dependency {
+    uint32_t image_id;                      /* Image index (from 0) */
```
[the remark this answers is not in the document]
davidvincze: Now image_id has been declared as uint8_t.

boot/bootutil/src/bootutil_misc.c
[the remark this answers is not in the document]
davidvincze: I've removed the build number check from the function.

boot/bootutil/include/bootutil/image.h
```diff
@@ -67,13 +71,14 @@ struct flash_area;
...
  *   2nd one is the actual signature.
  */
-#define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
-#define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
-#define IMAGE_TLV_RSA2048_PSS       0x20   /* RSA2048 of hash output */
-#define IMAGE_TLV_ECDSA224          0x21   /* ECDSA of hash output */
-#define IMAGE_TLV_ECDSA256          0x22   /* ECDSA of hash output */
-#define IMAGE_TLV_ENC_RSA2048       0x30   /* Key encrypted with RSA-OAEP-2048 */
-#define IMAGE_TLV_ENC_KW128         0x31   /* Key encrypted with AES-KW-128 */
+#define IMAGE_TLV_KEYHASH           0x01  /* hash of the public key */
```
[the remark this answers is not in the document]
davidvincze: Of course not. I've reverted these extra changes.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -116,7 +120,13 @@ struct boot_swap_state {
...
 };
 
-#define BOOT_MAX_IMG_SECTORS       MCUBOOT_MAX_IMG_SECTORS
+#ifdef MCUBOOT_IMAGE_NUMBER
+#define BOOT_IMAGE_NUMBER           MCUBOOT_IMAGE_NUMBER
+#else
+#define BOOT_IMAGE_NUMBER           1
+#endif
+
+#define BOOT_MAX_IMG_SECTORS        MCUBOOT_MAX_IMG_SECTORS
```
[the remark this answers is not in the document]
davidvincze: Done.

boot/zephyr/include/target.h
```diff
@@ -33,16 +35,20 @@
...
+    !defined(FLASH_AREA_IMAGE_0_SIZE) ||                                    \
+    !defined(FLASH_AREA_IMAGE_1_OFFSET) ||                                  \
+    !defined(FLASH_AREA_IMAGE_1_SIZE) ||                                    \
+    !defined(FLASH_AREA_IMAGE_SCRATCH_OFFSET) ||                            \
+    !defined(FLASH_AREA_IMAGE_SCRATCH_SIZE) ||                              \
+    ((MCUBOOT_IMAGE_NUMBER == 2) && !defined(FLASH_AREA_IMAGE_2_OFFSET)) || \
+    ((MCUBOOT_IMAGE_NUMBER == 2) && !defined(FLASH_AREA_IMAGE_2_SIZE)) ||   \
+    ((MCUBOOT_IMAGE_NUMBER == 2) && !defined(FLASH_AREA_IMAGE_3_OFFSET)) || \
+    ((MCUBOOT_IMAGE_NUMBER == 2) && !defined(FLASH_AREA_IMAGE_3_SIZE))
 #error "Target support is incomplete; cannot build mcuboot."
```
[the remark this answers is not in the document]
davidvincze: I agree, I've updated the code.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -4,9 +4,32 @@
...
+                                         DT_FLASH_AREA_IMAGE_0_ID : \
+                                     (current_image == 1) ?         \
+                                         DT_FLASH_AREA_IMAGE_2_ID : \
+                                         255 )
+#define FLASH_AREA_IMAGE_SECONDARY  ((current_image == 0) ?         \
+                                         DT_FLASH_AREA_IMAGE_1_ID : \
+                                     (current_image == 1) ?         \
+                                         DT_FLASH_AREA_IMAGE_3_ID : \
+                                         255 )
+#define FLASH_AREA_IMAGE_SCRATCH         3
```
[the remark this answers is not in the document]
davidvincze: You are right, I've updated the code.

boot/bootutil/src/loader.c
```diff
@@ -42,7 +46,8 @@
 
 MCUBOOT_LOG_MODULE_DECLARE(mcuboot);
 
-static struct boot_loader_state boot_data;
+static struct boot_loader_state boot_data_array[BOOT_IMAGE_NUMBER];
+static struct boot_loader_state *boot_data = &boot_data_array[0];
 uint32_t current_image = 0;
```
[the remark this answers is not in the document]
davidvincze: I've updated the `boot_loader_state` struct accordingly. However, I had to keep the current_image as a separate global variable. Adding the image number parameter to the `FLASH_AREA_IMAGE_(PRIMARY|SECONDARY)` macros would require to spread the currently static boot_data structure among the other source files and to make many functions be aware of the  presence of multiple images. Keeping the current_image as a global makes it easier the "draw a line" and hide the number if images from other functions to keep them simpler.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -692,3 +696,43 @@ boot_set_confirmed(void)
...
+    if (ver->iv_minor < req->iv_minor) {
+        return -1;
+    }
+    /* The minor version numbers are equal. */
+    if (ver->iv_revision > req->iv_revision) {
+        return 0;
+    }
+    if (ver->iv_revision < req->iv_revision) {
+        return -1;
+    }
```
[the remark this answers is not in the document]
davidvincze: I think it would. The left shift operations are fine and I think endianness can not be a problem in this case. Do you think it could be faster?

boot/bootutil/src/bootutil_misc.c
```diff
@@ -291,16 +300,14 @@ boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state)
     const struct flash_area *fap;
     int rc;
 
-    switch (flash_area_id) {
```
nvlsianpu: Previous switch-case consumes 16 B less for for-size optimization using zephyr-sdk-0.10.1

nvlsianpu: I would do similar analize for similar changes.

nvlsianpu:
mcuboot minimal builds against zephyr running master branch
commit 6b40394df90ea29f787a2e8bc4ff7a15bf093909
for target nrf52_pca10040

proj.conf file base content:
```
# Generated by Kconfiglib (https://github.com/ulfalizer/Kconfiglib)
CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=y
CONFIG_BOOT_SIGNATURE_KEY_FILE="root-ec-p256.pem"
# CONFIG_BOOT_VALIDATE_SLOT0 is not set
CONFIG_BOOT_UPGRADE_ONLY=y
# CONFIG_ZEPHYR_TRY_MASS_ERASE is not set
# CONFIG_SYS_POWER_MANAGEMENT is not set
# CONFIG_BOARD_ENABLE_DCDC is not set
CONFIG_SOC_SERIES_NRF52X=y
CONFIG_SOC_NRF52832_QFAA=y
CONFIG_ARM=y
CONFIG_ARM_MPU=n
# CONFIG_MULTITHREADING is not set
CONFIG_MAIN_STACK_SIZE=10240
CONFIG_THREAD_STACK_INFO=y
# CONFIG_TICKLESS_KERNEL is not set
CONFIG_SYSTEM_CLOCK_DISABLE=y
CONFIG_FLASH=y
# CONFIG_PRINTK is not set

CONFIG_CONSOLE=n
CONFIG_SERIAL=n
CONFIG_GPIO=n
CONFIG_DEBUG=n
CONFIG_EARLY_CONSOLE=n
CONFIG_PRINTK=n
```

### davidvincze:terminology_update commit 53f2c793e1cc3fd7c557fd999a73168e577d5035
+ algnment to zephyr running master https://github.com/JuulLabs-OSS/mcuboot/pull/497:

without swap, CONFIG_BOOT_UPGRADE_ONLY=y
```
[117/122]` Linking C executable zephyr/zephyr_prebuilt.elf
          FLASH:       13820 B        48 KB     28.12%
           SRAM:       17040 B        64 KB     26.00%
       IDT_LIST:          40 B         2 KB      1.95%
[122/122] Linking C executable zephyr/zephyr.elf
```

swap enabled, CONFIG_BOOT_UPGRADE_ONLY=n
```
[109/114] Linking C executable zephyr/zephyr_prebuilt.elf
          FLASH:       15796 B        48 KB     32.14%
           SRAM:       17040 B        64 KB     26.00%
       IDT_LIST:          40 B         2 KB      1.95%
[114/114] Linking C executable zephyr/zephyr.elf
```

### mainline master commit f99a4c790a6b4f508938f36e811c80ae1f25c2a3:

without swap, CONFIG_BOOT_UPGRADE_ONLY=y
```
117/122] Linking C executable zephyr/zephyr_prebuilt.elf
          FLASH:       13568 B        48 KB     27.60%
           SRAM:       17040 B        64 KB     26.00%
       IDT_LIST:          40 B         2 KB      1.95%
[122/122] Linking C executable zephyr/zephyr.elf
```

swap enabled, CONFIG_BOOT_UPGRADE_ONLY=n
```
[109/114] Linking C executable zephyr/zephyr_prebuilt.elf
          FLASH:       15252 B        48 KB     31.03%
           SRAM:       17040 B        64 KB     26.00%
       IDT_LIST:          40 B         2 KB      1.95%
[114/114] Linking C executable zephyr/zephyr.elf
```
A config file might be used as follow `west build -d <build_dir> --board nrf52_pca10040 <mcuboot_project_path> -- -DCONF_FILE=/path/to/a/custom_proj.conf`

boot/bootutil/src/bootutil_misc.c
tamasban: The FLASH_AREA_IMAGE_* macros was changed from static #defines to dynamically (run-time) evaluated macros to hide which flash area belongs to a certain image primary and secondary slot. In a switch-case the expression must be constant, due to this reason the switch-case was transformed to the if () statement. I will take a look how can this be optimised for single image case.

nvlsianpu: I see

boot/zephyr/flash_map_extended.c
```diff
@@ -55,9 +55,9 @@ int flash_device_base(uint8_t fd_id, uintptr_t *ret)
  */
 int flash_area_id_from_image_slot(int slot)
 {
-    static const int area_id_tab[] = {FLASH_AREA_IMAGE_PRIMARY,
-                                      FLASH_AREA_IMAGE_SECONDARY,
-                                      FLASH_AREA_IMAGE_SCRATCH};
+    const int area_id_tab[] = {FLASH_AREA_IMAGE_PRIMARY,
```
nvlsianpu: Following change will decrees flash footprint by 24 B.
```suggestion
#if (MCUBOOT_IMAGE_NUMBER == 1)
    static
#endif
    const int area_id_tab[] = {FLASH_AREA_IMAGE_PRIMARY,
```

boot/bootutil/src/loader.c
```diff
@@ -1595,83 +1607,293 @@ boot_set_image_ok(void)
...
+
+    if(!aborted_swap) {
+        if ((BOOT_SWAP_TYPE(&boot_data) == BOOT_SWAP_TYPE_NONE) ||
+            (BOOT_SWAP_TYPE(&boot_data) == BOOT_SWAP_TYPE_REVERT)) {
+            /* Nothing to do */
+            return;
+        }
+    }
+
+    for (uint8_t i = 0; i < current_image; i++) {
```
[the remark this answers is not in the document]
d3zd3z: Although, is there any particular reason for this, other than trying to maintain compatibility with C compilers from the 90s? Declaring on first use means there is no window of uninitialized value. And, especially in this particular case, it limits the scope of the variable.

d3zd3z:
> The decision was to leave those copyright headers for every single file?

Aside from Apache, this is pretty solidly agreed upon by pretty much every other contributor to the project. In fact, it makes things a little clumsy, because there are no Runtime (or Juul) copyright headers, it makes it hard to add a copyright header without making it look like the new contributor is now the original contributor.

I'm sure Apache has their reasons (and hopefully legal counsel), but if we're going to allow contributions, we're going to have to allow copyright headers on a per-file basis. Guidelines like the Linux kernel uses, where the header should be added only for "major" contributors are probably reasonable.

Given the lack of existing copyrights, I think the best we can do is add the somewhat awkward copyrights like the ones in this change.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -259,6 +264,12 @@ boot_img_hdr(struct boot_loader_state *state, size_t slot)
     return &BOOT_IMG(state, slot).hdr;
 }
 
+static inline struct image_version*
+boot_img_version(struct boot_loader_state *state, size_t slot)
```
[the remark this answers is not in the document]
davidvincze: I've removed it.

boot/bootutil/src/image_validate.c
```diff
@@ -78,12 +82,17 @@ bootutil_img_hash(struct image_header *hdr, const struct flash_area *fap,
...
-     * included ATM.
-     */
+    /* Hash is computed over image header and image itself. */
     hdr_size = hdr->ih_hdr_size;
     size = hdr->ih_img_size + hdr_size;
+    /* If dependency TLVs are present then the TLV info header and the
+     * dependency TLVs are also protected and have to be included in the hash
+     * calculation.
+     */
+    if (hdr->ih_protect_tlv_size != 0) {
```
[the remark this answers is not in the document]
davidvincze: Yes, that is correct. I've modified the code accordingly.

boot/zephyr/flash_map_extended.c
davidvincze: Thanks, I've added the conditional directives to the code.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
davidvincze: So, what would be the decision here? :)

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
davidvincze: You can find this part in the current code base at: https://github.com/JuulLabs-OSS/mcuboot/blob/6d079f994aa3ff89a8c8cd85d6f7286487006e91/boot/bootutil/src/loader.c#L1775-L1798 .

Referring to our former conversation:
As far as you could remember (to the previous states of the code), your concern was that it is possible to hang in the loop and never boot an image if any of the upgrades failed. Could you please check the current code regarding to this?

If an image dependency fails, the bootloader refuses the update of that image and in the next cycle it will try to verify the dependencies of the previous version of that image. In worst case the system can return to its previous state (it is assumed that there is always at least one valid state, the initial state - before the bootloader starts - must be consistent).

davidvincze:
> Maybe changing all those #if (BOOT_IMAGE_NUMBER > 1) to a more readable define like #if BOOT_IS_MULTI_IMAGE ?

@utzig had this comment. In my opinion it is pretty clear in its current form and I'm not sure if it is worth to introduce a new `BOOT_IS_MULTI_IMAGE` macro for this purpose. @d3zd3z , @nvlsianpu , @tamasban , what do you think?

nvlsianpu:
^^ For me this is currently clear enough.

boot/bootutil/src/bootutil_misc.c
nvlsianpu: @davidvincze ^^

d3zd3z:
I'm ok with the current `#if (BOOT_IMAGE_NUMBER > 1)` code.

d3zd3z:
I want to make sure I'm understanding the change correctly, and if so, need to point out that the dependency code introduces a security vulnerability. The problem is that the TLVs for the dependencies aren't covered by the signature (the signature is only over the image data). This means it would be easy to tamper with the dependencies, and possibly introduce incompatible versions. How exploitable this is is hard to say, but mismatches between the images could result in exploitable situations.

I still recommend that we merge this as is, but we really need to change how the TLV's are represented. I'll create an issue to cover this in the more general sense.

utzig:
@d3zd3z I think the whole point of adding the protected TLV size field to the header was to be able to protect some amount of TLVs, which in this case happens to be the dependency TLVs. `imgtool` seems to already take those into consideration when hashing/signing and the image validation routine was updated accordingly as well.

tamasban:
@utzig is rigth. This is already in place. `imgtool` is already has this feature (merged to master) and image validation is updated in the this [commit](https://github.com/JuulLabs-OSS/mcuboot/pull/369/commits/1e1e8209ff5c23a49b3c38ab21f77f3f575c7ab1).

d3zd3z:
This change looks good to me. I'll have to analyze the dependency protection a little more closer, but that can be done after this merges.

PR #370, opened 2018-12-14
Fix encrypted support for images with header sizes larger than 256 bytes

utzig:
The visible issue here is that there was an assert checking header size against a buffer that was supposed to run only on encrypted images but was missing the `#ifdef`ery. This cause images with headers larger than 256 bytes to fail, regardless of the encrypted image setting.

This patch fixes the issue by adding correct support for encrypted images, regardless of the header size.

I added a patch to the simulator to run the testing with header size equal to 512 bytes, which I will drop before merging this PR, although I find it valuable to have a test like this, a better way to parametrize this has to be found, and also avoid running every test both on 32 and 512 byte headers.


utzig:
Btw, I still need to test in HW (just to be sure!)

carlescufi:

PR #374, opened 2018-12-18
Update Jekyll due to CVE

utzig:
https://nvd.nist.gov/vuln/detail/CVE-2018-17567


utzig:
@d3zd3z Thanks, I had forgot this one! Turns out `jekyll 3.7.4` was not compatible with our locked version of `github-pages` gem, so I left for bundler to update them all, and now most gems are updated versions in `Gemfile.lock`.

PR #375, opened 2018-12-19
Remove unused hal includes from bootutil

TimoKroeger:
While porting MCUboot to a bare matel chip without OS (NXP LPC824) I found some unused includes.

After this pull request only four headers are required for a new target system:
"sysflash/sysfash.h"
"flash_map_backend/flash_map_backend.h"
"mcuboot_config/mcuboot_config.h"
"os/malloc.h"

The header "os/malloc.h" is only required by the split_go() function.
I my port I also removed split_go() but I’m not sure what it does so I kept it here.

Let’s see if CI passes or if I missed some details.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -23,12 +23,8 @@
 #include <stddef.h>
 
 #include "sysflash/sysflash.h"
-#include "hal/hal_bsp.h"
-#include "hal/hal_flash.h"
-
 #include "flash_map_backend/flash_map_backend.h"
 
-#include "os/os.h"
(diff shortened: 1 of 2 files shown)


nvlsianpu:
Looks like zephyr build is still compatible with this patch.

utzig:
No issues on Mynewt as well...

TimoKroeger:
I can also try to remove the now unneeded files from the zeyphyr build as further cleanup.

What’s your opinion on the removal split_go()? For me it looks like it was copied from mynewt and is not actually used. A seperate issue/pull request would be better to discuss this one though I think.

utzig:
@TimoKroeger Please don't remove split_go.

PR #377, opened 2018-12-21
[DNM] Add imgtool publishing support

utzig:
This adds initial support for publishing imgtool to pypi.org

The main imgtool.py was moved to imgtool package and made into the main file, and a new imgtool.py that calls into the package, was added allowing for the old usage behavior to remain functional.


utzig:
@d3zd3z Yeah, it was mostly a test thing, but I would suggest using the same release numbers as mcuboot uses, and maybe we should sometimes do a point release (as in `1.2.1`, `1.2.2`).

jcarrano:
Sorry for the delay, I was on holidays. This PR works awesome! I tried installing it from PyPI, and also directly from git:

```sh
$ python -m venv imgtool-env
$ . ./imgtool-env/bin/activate
$ pip install 'git+https://github.com/utzig/mcuboot.git@imgtool-package#subdirectory=scripts'
```

And just like that I have the tool.

Regarding the version, if you want to use the same version as mcuboot, perhaps you could extract it directly from `version.yml` in `setup.py`:

```py
import re
import os

version_re = re.compile(r"repo.version\s*:\s*(\S+)\s*")

def get_version():
    with open(os.path.join("..", "version.yml")) as f:
        for line in f:
            vmatch = version_re.fullmatch(line)
            if vmatch:
                return vmatch[1]

        raise RuntimeError("Cannot parse version file")

```

utzig:
@jcarrano Thanks for testing. Unfortunately `version.yml` is metadata for `Mynewt` and does not carry the mcuboot versioning info. Maybe we should also some versioning in .c/.h files before doing the next release!

scripts/setup.py
```diff
@@ -0,0 +1,27 @@
+import setuptools
+
+setuptools.setup(
+    name = "imgtool",
+    version = "0.1.1",
+    author = "The MCUboot commiters",
+    #author_email = "dev-mcuboot@lists.runtime.co",
+    author_email = "utzig@apache.org", #FIXME
```
carlescufi: Why not remove the author email altogether and let people contact us via the GitHub issues?

utzig: Removed, also updated version to `1.2.0` and fixed PEP8 issues.

utzig:
@d3zd3z I'll probably keep running this in-place forever.

PR #381, opened 2019-01-03
zephyr: kconfig: rename MCUBOOT to IS_MCUBOOT

lemrey:
Rename Kconfig option MCUBOOT to IS_MCUBOOT, for clarity.

Used by: https://github.com/zephyrproject-rtos/zephyr/pull/11210


boot/zephyr/Kconfig
```diff
@@ -8,7 +8,7 @@ mainmenu "MCUboot configuration"
 comment "MCUboot-specific configuration options"
 
 # Hidden option to mark a project as MCUboot
-config MCUBOOT
+config IS_MCUBOOT
```
nvlsianpu:
```suggestion
# Hidden option to mark a project as MCUboot
config IMCUBOOT
  select IS_MCUBOOT
```

lemrey: By redefinition, do you mean re-naming? If so I understand, though this one entry has just been introduced so I don't see a big issue.

boot/zephyr/Kconfig
```diff
@@ -13,6 +13,12 @@ config MCUBOOT
 	bool
 	select MPU_ALLOW_FLASH_WRITE if ARM_MPU
 
+# When compiling MCUBoot, the image will be linked to the boot partition.
+# Override .text offset to make sure it is set to zero.
+# This is necessary when other bootloaders set a different default.
```
nvlsianpu: what a other bootloaders?

carlescufi: You could conditionally do this based on `BOARD_HAS_NRF5_BOOTLOADER` perhaps? That way we don't pollute the rest

lemrey: I have address both comments.

boot/zephyr/Kconfig
```diff
@@ -13,6 +13,13 @@ config MCUBOOT
 	bool
 	select MPU_ALLOW_FLASH_WRITE if ARM_MPU
 
+# When compiling MCUBoot, the image will be linked to the boot partition.
+# Override .text offset to make sure it is set to zero.
+# This is necessary when other bootloaders set a different default for
+# application images which are not bootloaders.
```
carlescufi: I wonder if it would be better to have

```
if BOARD_HAS_NRF5_BOOTLOADER
...
```

that way everything is left unaffected when not building for the PCA10059

nvlsianpu: yes, will be safer

lemrey: Please recheck.

PR #386, opened 2019-01-07
imgtool save to format of given output filename

utzig:
Updates imgtool to infer, based on output image filename, which format should be used for output file. Filenames that end with extension `.hex` are saved in Intel HEX, otherwise saves a binary image.


utzig:
This was definitely not super well tested, but I checked that the basics work!

utzig:
Btw, I raised an exception when the input is binary and output is HEX, because the base address is not known. A parameter could be added for providing it, but I doubt this would be a common use case...

carlescufi:
> Btw, I raised an exception when the input is binary and output is HEX, because the base address is not known. A parameter could be added for providing it, but I doubt this would be a common use case...

You could default to base address 0x0 I guess, but this is a very unusual case as you say.

utzig:
> > Btw, I raised an exception when the input is binary and output is HEX, because the base address is not known. A parameter could be added for providing it, but I doubt this would be a common use case...

> You could default to base address 0x0 I guess, but this is a very unusual case as you say.

base address 0x0 will usually be the bootloader itself, so better not to allow it to be overwritten unintentionally.

PR #389, opened 2019-01-11
boot: zephyr: pca10059: disable logging by default

lemrey:
The board cannot output logs on the UART because it is used by MCUBoot's serial recovery for DFU. Remove logging and the RTT console from project defaults.

### DIFF
--- a/boot/zephyr/boards/nrf52840_pca10059.conf
+++ b/boot/zephyr/boards/nrf52840_pca10059.conf
@@ -1,3 +1,8 @@
+# The UART is used for Serial Recovery, so logging requires
+# an RTT console, which is not available out of the box on this board.
+# Disable logging.
+CONFIG_BOOT_HAVE_LOGGING=n
+
 # Serial
 CONFIG_SERIAL=y
 CONFIG_UART_NRFX=y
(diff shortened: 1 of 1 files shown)


nvlsianpu:
@carlescufi LOG increases flash footprint requirement (in compassion to SYS_LOG). This cause that mcuboot witch RTT LOG backend couln't fit into space reserved for it on pca10059. W have two option:

1) by default, the mcuboot project for pca10059 should have logging disabled (change in MCUBOOT)
this is because the board does not have a SEGGER chip and serial recovery is enabled by default so UART can't be used
If we make this change, then the pca10059 "stock" partition table doesn't require changes. it will fit after krzysztof changes (very close to overflow though)

2) increase the default mcuboot partition size for the "debugger" pca10059 partition table (change in ZEPHYR)
this is to accommodate for an RTT console and logging when the debugger is available.

I accept change in patch (I not approved because I wan't to avoid accidental  merge).
If you accept this, pleas merge.

lemrey:
Not so close to overflow in the first sceario, after all :)

carlescufi:
@nvlsianpu thanks, just discussed this briefly with @lemrey and it's fine.

PR #391, opened 2019-01-11
SUIT support

d3zd3z:
This is mostly completed work to add SUIT support to the simulator. This does not add SUIT support to MCUboot itself, so what this mostly does is create a new feature "suit-manifest" that can be enabled to cause test failures.

### DIFF
--- a/boot/bootutil/src/cbor.c
+++ b/boot/bootutil/src/cbor.c
@@ -26,7 +26,7 @@
 
 #include <bootutil/bootutil_log.h>
 
-MCUBOOT_LOG_MODULE_REGISTER(mcuboot);
+MCUBOOT_LOG_MODULE_DECLARE(mcuboot);
 
 #include "cbor.h"
 
--- a/boot/bootutil/src/cbor.c
+++ b/boot/bootutil/src/cbor.c
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2019 Linaro Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>  // TODO: Remove
+#include <stdlib.h>  // TODO: Remove
+
+#include "mcuboot_config/mcuboot_config.h"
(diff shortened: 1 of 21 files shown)


d3zd3z:
Note that github doesn't show the commits in a meaningful order. Probably best to fetch this with `git fetch origin pull/391/head` and look at the changes locally.

d3zd3z:
One big question I have for reviewers is: should I squash the cbor and/or suit commits. There is a bit of development work in here, and these could make sense just having a commit that adds them.

sim/cboria/src/bin/cbor-dump.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    Ok(())
+}
+
+/// Read a CBOR code value, and the associated integer.
+fn get_code<R: Read>(rd: &mut R) -> Result<(u8, usize)> {
+    let b = rd.read_u8()?;
+    let code = b >> 5;
+    let count = match b & 0x1f {
+        count @ 0 ..= 23 => count as usize,
```
utzig: The binding should be `count @ 0 ... 23`, no? What does the `..=` syntax mean?

d3zd3z: `..=` is an inclusive range pattern (as opposed to `0 .. 24` which would be the same range). This is new to Rust 2018.

utzig:
@d3zd3z After reading through some of the cboria commits, it looks like it would be easier to just squash those commits (I just want to review the final version, not the progress of making it!).

d3zd3z:
I've squashed the cbor/suit commits. Now most of the commits are modifications to the simulator to call this code.

boot/bootutil/src/cbor.c
```diff
@@ -0,0 +1,245 @@
...
+    } else if (ext == 24) {
+        want = 1;
+    } else if (ext == 25) {
+        want = 2;
+    } else if (ext == 26) {
+        want = 4;
+    } else {
+        /* If we want uint64 support, add it above here. */
+        return -1;
+    }
```
utzig: *maybe* would gain a few bytes by doing:
```
    ...
    return 0;
} else if (ext > 26) {
    return -1;
}

want = 1 << (ext - 24);
```

d3zd3z: 10-12 bytes. 10 bytes of code difference, possibly causing 12 bytes of padding.

d3zd3z:
I have moved the simulator changes to the front, and pushed to #434. If desired, I can even remove them from this patch series (just a little tweaking with the capabilities flags), perhaps pushing that change through, and these can merge independently.

d3zd3z:
Simulator changes moved into #435. Merge these two changes to be able to test the SUIT changes within the simulator.

utzig:
Depends on #435

d3zd3z:
This has gotten quite out of date, and SUIT has moved to a new version. The pull request will remain, even when closed, but I'm closing this to reduce clutter of the pull request list.

PR #395, opened 2019-01-14
Create ptest utility

d3zd3z:
This is a small utility that facilitates doing the Travis-style parallel tests during development. To use, `cd ptest`, and `cargo run`. It will read the configuration from the `.travis.yml` file.

### DIFF
--- a/ptest/.gitignore
+++ b/ptest/.gitignore
@@ -1,0 +1,2 @@
+target
+.*.swp
--- a/ptest/Cargo.lock
+++ b/ptest/Cargo.lock
@@ -1,0 +1,404 @@
+[[package]]
+name = "aho-corasick"
+version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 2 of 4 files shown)


d3zd3z:
I wonder if I need a limit other than 8. I've got another change that cleans up the status display a bit.

d3zd3z:
I was planning on defaulting to either the number of physical cores, or the number of threads, not sure which.

utzig:
Wouldn't it be better to just squash these commits since it's all part of the first ptest release?

d3zd3z:
Probably :-)

d3zd3z:
All squashed.

d3zd3z:
> Maybe it would be a good idea to remove the commented out code.
Do you mean the printlns in various points? Maybe I should add logging, and make them `debug!()`, which won't normally show up, but can be turned on.

d3zd3z:
I can squash, if we think that is better, but I think review is easier before squashing.

utzig:
Looks good!

PR #404, opened 2019-01-23
Fix trailer erase for some devices

utzig:
Previously it was assumed that the trailer would fit one sector. While this works for most devices, some which happen to have lots of sectors end up using lots of space for their swap status area. The same issue could also happen if sectors are small (less than 2K/4K, for example).

This fixes the issue by checking the trailer size and erasing all sectors that encompass it.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -987,10 +987,14 @@ boot_status_init(const struct flash_area *fap, const struct boot_status *bs)
 
 #ifndef MCUBOOT_OVERWRITE_ONLY
 static int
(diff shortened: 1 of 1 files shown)


utzig:
@d3zd3z This fixes the issue that I was having on STM32L0; it can be added after rc1, because I'd like to do some thorough testing to ensure there are no regressions.

Btw, also have another "minor" issue (for multi-flash systems), which I'll submit a fix once it's well tested!

utzig:
> Btw, also have another "minor" issue (for multi-flash systems), which I'll submit a fix once it's well tested!

Does not need to be fixed, not an issue I think, disregard the comment...

PR #407, opened 2019-01-25
Build Mynewt target configs on CI

utzig:
This extends the CI test runs to build multiple configurations on Mynewt, to avoid untested paths to break the build, like when someone tests some change only on Zephyr, etc


utzig:
testing slack integration - disregard this comment

PR #410, opened 2019-01-28
Do not use an ASSERT to catch the case if no image is loaded

jobrud:
In case if no image is loaded and the `MCUBOOT_VALIDATE_SLOT0` is enabled, the current implementation will stuck at this assertion. This means when assertions will be evaluated that the if-statement after the assertion is dead code.  I think no image loaded shouldn't be a case which will be catched with an assertion.  
What do you think ?

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1734,7 +1734,6 @@ boot_go(struct boot_rsp *rsp)
 
 #ifdef MCUBOOT_VALIDATE_SLOT0
     rc = boot_validate_slot(0, NULL);
-    ASSERT(rc == 0);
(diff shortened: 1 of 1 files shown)


d3zd3z:
The if needs to be there in case assertions are compiled out. The simulator tests pass with or without this change. I think it is a little cleaner without the assertion.

d3zd3z:
Oh, and by the way, thanks for the contribution. It's always nice to see contributions coming from the community, especially with something like MCUboot where there are a lot of users that we don't know about.

jobrud:
For clarification, I don't think that this is a bug, I just think it's better to handle the case if no image is loaded in the same way for development and production environment.

PR #414, opened 2019-01-30
Set pending: don't crash when image slot corrupt

ccollins476ad:
This change affects the "set pending" operation.  That is, the operation that configures mcuboot to temporarily swap images on the next boot.

#### PRIOR TO PR

If the slot 1 trailer contained an invalid 128-bit magic number, an assertion would fail, causing a crash.

#### AFTER PR

If corruption is detected in the slot 1 trailer, the entire image slot is erased, and the "set pending" operation fails with a `BOOT_EBADIMAGE` status.

#### RATIONALE

mcuboot cannot meaningfully recover from data corruption.  The only recourse is to erase the bad data so that future upgrades can be performed.  I was tempted to add a build-time setting to control whether the image slot gets erased when corruption is detected, but I dont think this
freedom justifies the cost of extra config.  A device with a corrupt image slot can no longer be upgraded, so the only reason someone would want to preserve the corrupt data would be for debugging.

#### NOTES

I think mcuboot should detect this type of corruption at startup and erase the bad image slot itself.  This requires some further discussion and can be addressed later.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -615,10 +615,22 @@ boot_set_pending(int permanent)
         flash_area_close(fap);
         return rc;
 
+    case BOOT_MAGIC_BAD:
+        /* The image slot is corrupt.  There is no way to recover, so erase the
+         * slot to allow future upgrades.
+         */
+        rc = flash_area_open(FLASH_AREA_IMAGE_1, &fap);
+        if (rc != 0) {
+            return BOOT_EFLASH;
+        }
+
+        flash_area_erase(fap, 0, fap->fa_size);
+        flash_area_close(fap);
+        return BOOT_EBADIMAGE;
+
     default:
-        /* XXX: Temporary assert. */
         assert(0);
-        return -1;
+        return BOOT_EBADIMAGE;
     }
 }
 


boot/bootutil/src/bootutil_misc.c
```diff
@@ -615,10 +615,16 @@ boot_set_pending(int permanent)
         flash_area_close(fap);
         return rc;
 
+    case BOOT_MAGIC_BAD:
+        /* The image slot is corrupt.  There is no way to recover, so erase the
+         * slot to allow future upgrades.
+         */
+        flash_area_erase(fap, 0, fap->fa_size);
```
utzig: It is weird that you are not calling `flash_area_open` so no idea where `fap` is initialized.

ccollins476ad: Indeed. This is actually a double fail on my part - I meant to test this in the morning before submitting the PR, then forgot to test it!

Thanks, @utzig.  I will fix and test :)

utzig:
@nvlsianpu @d3zd3z Just to make it clear, this affects `mcumgr`/`newtmgr` and it is an error that can happen on Mynewt because we use `newt` + `newtmgr` which never pad binaries, so once data was written to the status area (by any other means), it's impossible to set an image as pending ever again. Users of `imgtool` that employ padding won't ever see this error. I believe this should be safe to merge, but if there are opinions against it let me know. If none is given I will merge by the end of the day.

nvlsianpu:
Zephyr has got its own implementation of boot_set_pending(), so this changes nothing to him.

utzig:
@nvlsianpu Good to know, but maybe not everyone is using that, and also there are people on the slack channel that are not using Mynewt/Zephyr.

d3zd3z:
This doesn't seem to break anything on Zephyr.

However, my concern is that this would be an annoyance to an upgrade that is interrupted. I guess it requires that the upgrade process leave the trailer erased until the end of the process?

Otherwise, looks good.

d3zd3z:
I guess the only question I would ask is should this be conditional? I would argue that fixing this is really the responsibility of the download app, and not the bootloader.

utzig:
@d3zd3z This code is part of the download app, not the bootloader; while the function is defined on the mcuboot repo it's not used by the bootloader itself.

ccollins476ad:
@d3zd3z , could you please clarify your concern:

> However, my concern is that this would be an annoyance to an upgrade that is interrupted. I guess it requires that the upgrade process leave the trailer erased until the end of the process?

?

In Mynewt, at least, the 16-byte magic region is only expected to have one of two values:
1. Erased
2. The magic number itself

If the contents of this region are not one of the above, then something has gone wrong.  In our case, this happened when we rearranged the flash map without erasing the flash.  Consequently, the slot 1 image trailer contained some some leftover garbage.

Does Zephyr write other values to this region during an upgrade?

d3zd3z:
Part of the problem is that I don't know what Zephyr does when doing an upgrade, since Zephyr doesn't yet have anything to do upgrades. That means it is being done by other people. Perhaps we just need to make sure we're clear about what is supposed to happen and that the magic number must be one of these two values (or an indeterminate value, which would indicate we need to release). Thinking more about this, I withdraw my concern, and I think this will be fine.

mbolivar:
> Part of the problem is that I don't know what Zephyr does when doing an upgrade, since Zephyr doesn't yet have anything to do upgrades.

I'm confused about this. There is an upstream sample for doing upgrades: https://github.com/zephyrproject-rtos/zephyr/tree/master/samples/subsys/mgmt/mcumgr/smp_svr

PR #419, opened 2019-02-08
S1000 fix

rgundi:
S1000 got broken due to some changes done in the Zephyr mainline as well as mcuboot. This patchset fixes it.


nvlsianpu:
@rgundi pleas fix CI problems?

PR #425, opened 2019-02-19
zephyr: fix build with zephyr

nvlsianpu:
zephyr macro min() was hanged to MIN()
Kconfig key CDC_ACM_PORT_NAME was changed to CDC_ACM_PORT_NAME_0

zephyr sha of the fixing time is https://github.com/zephyrproject-rtos/zephyr/commit/406dc2cb0e2868392f444677ace416440695829f.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -588,7 +588,11 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
         if (len != *out_off - sizeof(uint16_t)) {
             return 0;
         }
-        len = min(len, *out_off - sizeof(uint16_t));
+
+        if (len > *out_off - sizeof(uint16_t)) {
+            len = *out_off - sizeof(uint16_t);
+        }
+
         out += sizeof(uint16_t);
 #ifdef __ZEPHYR__
         crc = crc16(out, len, CRC_CITT_POLYMINAL, CRC16_INITIAL_CRC, true);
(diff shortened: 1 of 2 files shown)


utzig:
The mentioned commit is unrelated to this change, also please just copy/paste the zephyr commit's URL because GH does correctly link then.

nvlsianpu:
PR which changes min/max to MIN/MAX: https://github.com/zephyrproject-rtos/zephyr/pull/13277 
PR which changes cdc acm port used: https://github.com/zephyrproject-rtos/zephyr/pull/12682/

boot/boot_serial/src/boot_serial.c
```diff
@@ -588,7 +588,7 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
         if (len != *out_off - sizeof(uint16_t)) {
             return 0;
         }
-        len = min(len, *out_off - sizeof(uint16_t));
+        len = MIN(len, *out_off - sizeof(uint16_t));
```
utzig: Please only do this for Zephyr, Mynewt uses `min`.

d3zd3z: We should either define these ourselves (which is less trivial than it seems), or define our own wrapper that uses the appropriate one for the system. Or, my suggestion, since this only occurs once in the code, just change this to:
```c
    if (len > *out_off - sizeof(uint16_t)) {
        len = *out_off - sizeof(uint16_t);
    }
```

nvlsianpu: I did as you suggest.

PR #430, opened 2019-02-22
[WIP] Multi image building

SebastianBoe:
WIP support for multi-image building.

Must be applied simultaneously with https://github.com/zephyrproject-rtos/zephyr/pull/13672

See the Zephyr PR for details.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -134,7 +134,7 @@ if(CONFIG_MCUBOOT_SERIAL)
 
   zephyr_link_libraries_ifdef(
 	CONFIG_TINYCBOR
-	TINYCBOR
+	${IMAGE}TINYCBOR
 	)
 
   zephyr_include_directories_ifdef(
@@ -149,6 +149,14 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
   else()
     set(KEY_FILE ${MCUBOOT_DIR}/${CONFIG_BOOT_SIGNATURE_KEY_FILE})
   endif()
+
+  set_property(
+	GLOBAL
+	PROPERTY
+	KEY_FILE
+	${KEY_FILE}
+	)
+
   set(GENERATED_PUBKEY ${ZEPHYR_BINARY_DIR}/autogen-pubkey.c)
   add_custom_command(
     OUTPUT ${GENERATED_PUBKEY}
@@ -163,3 +171,13 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
     )
   zephyr_library_sources(${GENERATED_PUBKEY})
 endif()
+
+# TODO: Configurable?
+set_property(GLOBAL APPEND PROPERTY
+  HEX_FILES_TO_MERGE
+  ${PROJECT_BINARY_DIR}/zephyr/${KERNEL_HEX_NAME}
+  )
+set_property(GLOBAL APPEND PROPERTY
+  HEX_FILES_TO_MERGE_TARGET
+  ${logical_target_for_zephyr_elf}
+  )
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -13,16 +13,16 @@ set(BOARD qemu_x86)
 # and fits inside, the boot partition. (If the user specified a
 # DTC_OVERLAY_FILE on the CMake command line, we need to append onto
 # the list).
-if(DTC_OVERLAY_FILE)
-  set(DTC_OVERLAY_FILE
-      "${DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/dts.overlay")
+if(${IMAGE}DTC_OVERLAY_FILE)
+  set(${IMAGE}DTC_OVERLAY_FILE
+      "${${IMAGE}DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/dts.overlay")
 else()
-  set(DTC_OVERLAY_FILE ${CMAKE_CURRENT_LIST_DIR}/dts.overlay)
+  set(${IMAGE}DTC_OVERLAY_FILE ${CMAKE_CURRENT_LIST_DIR}/dts.overlay)
 endif()
 
 if (EXISTS ${CMAKE_CURRENT_LIST_DIR}/boards/${BOARD}.overlay)
-  set(DTC_OVERLAY_FILE
-      "${DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/boards/${BOARD}.overlay")
+  set(${IMAGE}DTC_OVERLAY_FILE
+      "${${IMAGE}DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/boards/${BOARD}.overlay")
 endif()
 
 # Enable Zephyr runner options which request mass erase if so
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -60,14 +60,6 @@ config BOOT_SIGNATURE_TYPE_ECDSA_P256
 
 endchoice
 
-config BOOT_SIGNATURE_KEY_FILE
-	string "PEM key file"
-	default ""
-	help
-	  The key file will be parsed by imgtool's getpub command and a .c source
-	  with the public key information will be written in a format expected by
-	  MCUboot.
-
 config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -1,0 +1,39 @@
+if(CONFIG_BOOTLOADER_MCUBOOT)
+  # Build a second bootloader image
+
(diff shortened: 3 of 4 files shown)


utzig:
Some of those changes were already merged as part of 048168aa377e40f5e103a48d634a1c7b7035bedd, please rebase!

SebastianBoe:
Done

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+if(CONFIG_BOOTLOADER_MCUBOOT)
```
mbolivar: Might want to drop a README in this directory or add a comment to this file saying that the `zephyr` directory is for use of mcuboot as a zephyr module (https://docs.zephyrproject.org/latest/application/index.html#modules-external-projects), and that boot/zephyr is where the Zephyr-specific application code is, to avoid confusing newcomers who don't know to look for boot/zephyr yet.

I've had to clear up confusion that CONFIG_BOOTLOADER_MCUBOOT should *not* be enabled when building mcuboot itself, but is rather an option that's enabled by applications using MCUboot; this file muddies that distinction a bit, so a comment would be nice to clear up that what's going on here is kind of complex:

1. Application CMakeLists.txt
2. Application boilerplate.cmake
3. Application zephyr_modules.cmake
4. This file
5. MCUboot CMakeLists.txt (boot/zephyr/CMakeLists.txt)
6. MCUboot boilerplate.cmake
...

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  # Build a second bootloader image
+
+  set(MCUBOOT_BASE ${CMAKE_CURRENT_LIST_DIR}/..)
+
+  zephyr_add_executable(mcuboot require_build)
+
+  if (${require_build})
+    add_subdirectory(${MCUBOOT_BASE}/boot/zephyr ${CMAKE_CURRENT_BINARY_DIR}/mcuboot)
+
+    # TODO: Assert that the bootloader and image use the same key.
```
mbolivar: Tricky: multiple keys are supported simultaneously.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    set_property(GLOBAL APPEND PROPERTY
+      extra_post_build_commands
+      COMMAND
+      ${PYTHON_EXECUTABLE}
+      ${MCUBOOT_BASE}/scripts/imgtool.py
+      sign
+      --key ${MCUBOOT_BASE}/${CONFIG_BOOT_SIGNATURE_KEY_FILE}
+      --header-size ${CONFIG_TEXT_SECTION_OFFSET}
+      --align       ${DT_FLASH_WRITE_BLOCK_SIZE}
+      --version 0.1       # TODO: Configurable?
```
mbolivar: > TODO: Configurable?

I think so. One could set it at CMake time; maybe even let the environment override that. That'd be pretty convenient on macOS and Linux at least.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+      --slot-size 0x32000 # TODO: Configurable?
```
mbolivar: Not configurable; this is from DT (`DT_FLASH_AREA_IMAGE_0_SIZE`).

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  set(MCUBOOT_BASE ${CMAKE_CURRENT_LIST_DIR}/..)
+
+  zephyr_add_executable(mcuboot require_build)
+
+  if (${require_build})
+    add_subdirectory(${MCUBOOT_BASE}/boot/zephyr ${CMAKE_CURRENT_BINARY_DIR}/mcuboot)
+
+    # TODO: Assert that the bootloader and image use the same key.
+
+    set(SIGNED_IMAGE signed.hex)
```
mbolivar: Would you mind zephyr.signed.hex, to match the default from `west sign`?

zephyr/Kconfig
```diff
@@ -0,0 +1,46 @@
+if BOOTLOADER_MCUBOOT
+
+config MCUBOOT_CMAKELISTS_DIR
+	string "Path to the directory of the MCUBoot CMakeLists.txt file"
+	default "$MCUBOOT_BASE/boot/zephyr/"
+
+
+choice
```
mbolivar: These options are boilerplate needed by any child image and should be put into a boilerplate file, like zephyr/subsys/logging/Kconfig.template.log_config, to be included here after setting the name to MCUBOOT or so.

zephyr/Kconfig
```diff
@@ -0,0 +1,46 @@
...
+	prompt "MCUBoot build strategy"
+	default MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
+
+config MCUBOOT_BUILD_STRATEGY_USE_HEX_FILE
+	# Mandatory option when being built through 'zephyr_add_executable'
+	bool "Use hex file instead of building MCUBoot"
+
+if MCUBOOT_BUILD_STRATEGY_USE_HEX_FILE
+
+config MCUBOOT_HEX_FILE
```
mbolivar: I don't think you meant this to be a choice option; did you? Shouldn't it be moved after the `endchoice`?

zephyr/Kconfig
```diff
@@ -0,0 +1,46 @@
...
+config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
+	# Mandatory option when being built through 'zephyr_add_executable'
+	bool "Build from source"
+
+endchoice
+
+endif # BOOTLOADER_MCUBOOT
+
+if MCUBOOT || BOOTLOADER_MCUBOOT
+# TODO: Support sharing Kconfig configuration between images
```
mbolivar: That is a really interesting idea!

boot/zephyr/Kconfig
```diff
@@ -60,14 +60,6 @@ config BOOT_SIGNATURE_TYPE_ECDSA_P256
 
 endchoice
 
-config BOOT_SIGNATURE_KEY_FILE
```
mbolivar: Does this break the build when MCUboot is built as a single image?

boot/zephyr/CMakeLists.txt
```diff
@@ -163,3 +171,13 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
     )
   zephyr_library_sources(${GENERATED_PUBKEY})
 endif()
+
+# TODO: Configurable?
+set_property(GLOBAL APPEND PROPERTY
+  HEX_FILES_TO_MERGE
+  ${PROJECT_BINARY_DIR}/zephyr/${KERNEL_HEX_NAME}
```
mbolivar: Why isn't `zephyr_add_executable()` taking care of these?

boot/zephyr/CMakeLists.txt
```diff
@@ -149,6 +149,14 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
   else()
     set(KEY_FILE ${MCUBOOT_DIR}/${CONFIG_BOOT_SIGNATURE_KEY_FILE})
   endif()
+
+  set_property(
+	GLOBAL
+	PROPERTY
+	KEY_FILE
```
mbolivar: Where is this used?

d3zd3z:
Unless I'm misreading something, the commit is titled "dump: dump\n\ndump". Perhaps a better description is in order.

I'm also wondering if we need a better name for this than "multi image", since we are using multi image to refer to having multiple images that MCUboot knows how to load. I'm not sure what terms are best here. But, "multi-image" from MCUboot's perspective makes sense as MCUboot itself handling multiple images. Maybe "multi build"? I'm not real sure.

SebastianBoe:
> Unless I'm misreading something, the commit is titled "dump: dump\n\ndump". Perhaps a better description is in order.

The commit message has been improved.

SebastianBoe:
> I'm also wondering if we need a better name for this than "multi image", since we are using multi image to refer to having multiple images that MCUboot knows how to load. I'm not sure what terms are best here. But, "multi-image" from MCUboot's perspective makes sense as MCUboot itself handling multiple images. Maybe "multi build"? I'm not real sure.

Since these are two different domains, I'm hoping it will always be obvious from context what we are talking about. It will be difficult to change this terminology as it is in wide use already. To disambiguate when not obvious from context we can try to be more verbose, e.g. multi-image building.

zephyr/CMakeLists.txt
hakonfam: Fixed

zephyr/CMakeLists.txt
hakonfam: Could you elaborate the benefits of not having this as a Kconfig option?

mbolivar: I'm thinking of things like developer workflows where you want to update this on a build-by-build basis every few minutes, without having to go through the overhead of running menuconfig. An environment variable makes that pretty easy -- not sure if it's possible to have a kconfig option fall back to an env var dynamically, but I believe not. OTOH from a reproducible builds perspective I take your point. Tradeoffs here, your call.

utzig:
@SebastianBoe Please rebase on latest master!

SebastianBoe:
Closing, https://github.com/zephyrproject-rtos/zephyr/pull/13672 was not accepted.

PR #435, opened 2019-02-28
Simulator updates for SUIT and multiimage

d3zd3z:
This branch contains the simulator changes to support both SUIT and (eventually) multi-image. These changes should be merged with #391 to be able to test the SUIT changes.

This branch adds support for the SUIT manifest to the simulator, as well as some structural changes to make it easier to test multi-images.


sigvartmh:
Sorry to bump in but what's the point of the `cboria` lib? Is there something missing in https://github.com/BurntSushi/rust-cbor

sigvartmh:
> Sorry to bump in but what's the point of the `cboria` lib? Is there something missing in https://github.com/BurntSushi/rust-cbor

Never mind saw it was unmaintained

sigvartmh:
Still what is the purpose of the cboria lib ?

d3zd3z:
> Still what is the purpose of the cboria lib/writing your own cbor encoder/emitter? Instead of using serde_cbor or any other existing encoder/emitters?

There are no cbor libraries that allow fine enough control over the precise way that the data is encoded, and so none of them would be able to encode a SUIT manifest. The existing Rust cbor libraries are all based around marshaling Rust data structures, not around matching precisely defined CBOR.

d3zd3z:
I have rebased these changes on top of @davidvincze's patch "Replace flash partitioning terminology". Both branches have this same patch, but the rebase after one merges will eliminate the other one. It should now be possible to merge together this simulator change, the multi-image change (or my rebased version of it), and the SUIT implementation.

sigvartmh:
> > Still what is the purpose of the cboria lib/writing your own cbor encoder/emitter? Instead of using serde_cbor or any other existing encoder/emitters?
> 
> There are no cbor libraries that allow fine enough control over the precise way that the data is encoded, and so none of them would be able to encode a SUIT manifest. The existing Rust cbor libraries are all based around marshaling Rust data structures, not around matching precisely defined CBOR.

Makes sense thanks for taking your time answering my question :)

boot/bootutil/src/loader.c
```diff
@@ -1726,33 +1764,34 @@ boot_go(struct boot_rsp *rsp)
...
+         * provide the data for the bootstrap, which previously was at secondary
+         * slot, was updated to primary slot.
          */
-        slot = 0;
+        slot = BOOT_PRIMARY_SLOT;
     }
 
-#ifdef MCUBOOT_VALIDATE_SLOT0
-    rc = boot_validate_slot(0, NULL);
+#ifdef mcuboot_validate_primary_slot
```
davidvincze: I guess it should be MCUBOOT_VALIDATE_PRIMARY_SLOT (upper case).

sigvartmh:
```suggestion
MCUBOOT_VALIDATE_PRIMARY_SLOT
```

@davidvincze you can do
```
   ```suggestion
    <change>
    ```
```
In github now, it's a nice feature for one liner fixes

boot/mynewt/mcuboot_config/syscfg.yml
```diff
@@ -19,8 +19,8 @@
 # Package: boot/mynewt/mcuboot_config
 
 syscfg.defs:
-    BOOTUTIL_VALIDATE_SLOT0:
-        description: 'Validate image at slot 0 on each boot.'
+    BOOTUTIL_VALIDATE_PRIMARY_SLOT:
```
utzig: When you change the name of this `syscfg` you are basically breaking every Mynewt configuration ever created that wants to have validation on slot0.

boot/zephyr/Kconfig
```diff
@@ -71,38 +71,38 @@ config BOOT_SIGNATURE_KEY_FILE
 config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
-config BOOT_VALIDATE_SLOT0
-	bool "Validate image slot 0 on every boot"
+config BOOT_VALIDATE_PRIMARY_SLOT
```
utzig: Although Zephyr defaults to enabled changing this config parameter's name enables it for people who have previously disabled it.

docs/readme-mynewt.md
```diff
@@ -40,10 +40,11 @@ in this new format. That is done by passing the extra parameter `-2` as in:
 # Boot serial functionality with Mynewt
 
 Building with `BOOT_SERIAL: 1` enables some basic management functionality
-like listing images and uploading a new image to `slot0`. The serial bootloader
-requires that `mtu` is set to a value that is less than or equal to `256`.
-This can be done either by editing `~/.newtmgr.cp.json` and setting the `mtu`
-for the connection profile, or specifying you connection string manually as in:
+like listing images and uploading a new image to `primary slot`. The serial
```
utzig: I think it is a bit weird to use this terminology here, because all Mynewt docs use slot0 and slot1, so will be confusing for uses; at least it should be mentioned somewhere that terminology was/will be changed

utzig:
Regarding a4adf9f334799c8494e9034f18108b90c987ea25, I think it should be dropped from this PR and reviewed separately. I don't see anything particularly wrong with it, but I can't really grasp why changing names of stuff is so important to the extent of having to do a whole find/replace that could be just be a (toy example):

```
#define PRIMARY_SLOT 0
#define SLOT0 PRIMARY_SLOT
```

and so on and so forth... I will review the associated PRs, so it might clear it up...

boot/mynewt/mcuboot_config/syscfg.yml
d3zd3z: I agree with this, I think we need to be more careful about changing things that are externally visible, names of features, configuration, etc. I do like the idea of having better terms, but this change is really pervasive, and is making for lots of merge conflicts, etc. I think the feature names should probably be left as SLOT0, just because external entities are already using these names.

boot/zephyr/Kconfig
d3zd3z: I don't think we should change the name of configuration options.

docs/readme-mynewt.md
d3zd3z: This is also one of a handful that are still ungrammatical after the change.

d3zd3z:
> Regarding a4adf9f, I think it should be dropped from this PR and reviewed separately.

Maybe it should be reviewed separately, but it can't exactly be "dropped". I've just been putting it at the beginning of each series of patches. Once one of them gets merged, the rest can be rebased which will drop the patch from that series.

We can review it as a separate change, but whoever works on it needs to realize that we now have 3 active branches depending on it, and changes to it are **painful**. I've basically spent a little over a day just munging these trees to fit with the terminology changes.

d3zd3z:
> commit b3eff76 that adds C code to build.rs should be dropped from this PR and added to PR with C code

I'm going to push back on this one as just not practical to do. The C changes depend on the rest of the simulator changes to create the 'suit' variable. If I moved just the C file changes to the SUIT branch, all it would do is make the simulator not compile on that branch.

d3zd3z:
This should address all of @utzig concerns, except for the one above I don't think is a good idea.

utzig:
Please squash 94b5b4c4780d0ada62047f68addd56820e4d730b and 7d118da3514268bf960fa1da93486a1ee8e98d60 together (and fix SoB)

utzig:
@d3zd3z Any chance you could break the SUIT and multi-image changes apart into 2 PRs?

d3zd3z:
Split this into #458 and #459, so closing this one. Made two new new PRs to avoid having the comments intermingled between the versions.

PR #436, opened 2019-03-01
Fix snprintf warning

SebastianBoe:
Fix https://github.com/zephyrproject-rtos/zephyr/issues/10990.

A warning is triggered when snprintf is used because it is being
declared / configured incorrectly. The MBEDTLS configuration is
defaulting to disabling all of LIBC, and then explicitly enabling the
LIBC functions in use, this has been done for printf with
MBEDTLS_PLATFORM_STD_PRINTF, but not for
MBEDTLS_PLATFORM_STD_SNPRINTF.


SebastianBoe:
This applies the fix proposed by @aurel32, https://github.com/zephyrproject-rtos/zephyr/pull/10991#issuecomment-435034584

carlescufi:
CC @pfalcon @rlubos @jukkar

jukkar:
This looks much clearer solution that the one proposed in https://github.com/zephyrproject-rtos/zephyr/pull/10991

PR #440, opened 2019-03-12
Replace flash partitioning terminology

d3zd3z:
This change replaces the slot 0/1 terminology with primary/secondary
slot and replaces FLASH_AREA_IMAGE_0/1 with
FLASH_AREA_IMAGE_PRIMARY/SECONDARY. This naming convention may be more
understandable, fits better to MCUs with multiple images and it is an
architecture agnostic alternative as well.

Change-Id: I655a585f6ae023852c671ee6635399efe25209c9
Signed-off-by: David Vincze <david.vincze@arm.com>


boot/bootutil/src/bootutil_priv.h
```diff
@@ -129,18 +129,21 @@ struct boot_swap_state {
...
+#define BOOT_SECONDARY_SLOT             1
 
-#define BOOT_FLAG_IMAGE_OK         0
-#define BOOT_FLAG_COPY_DONE        1
+#define BOOT_STATUS_SOURCE_NONE         0
+#define BOOT_STATUS_SOURCE_SCRATCH      1
+#define BOOT_STATUS_SOURCE_PRIMARY_SLOT 2
+
+#define BOOT_FLAG_IMAGE_OK              0
+#define BOOT_FLAG_COPY_DONE             1
```
utzig: To pass Mynewt CI, please add the following at this line:
```
#if defined(MCUBOOT_MYNEWT)                                                     
    #define FLASH_AREA_IMAGE_PRIMARY   FLASH_AREA_IMAGE_0                           
    #define FLASH_AREA_IMAGE_SECONDARY FLASH_AREA_IMAGE_1                           
#endif
```
`newt` auto-generates the `FLASH_AREA_IMAGE_[0|1]` symbols from `bsp.yml` and this adds a compatibility definition.

boot/bootutil/src/encrypted.c
```diff
@@ -305,7 +305,7 @@ boot_enc_load(const struct image_header *hdr, const struct flash_area *fap,
 int
 boot_enc_valid(const struct flash_area *fap)
 {
-    return enc_state[fap->fa_id - FLASH_AREA_IMAGE_0].valid;
```
nvlsianpu: This PR makes me aware of that this mathematics might be not compatible with zephyr flash_area numeration.

boot/bootutil/src/loader.c
```diff
@@ -1726,33 +1765,34 @@ boot_go(struct boot_rsp *rsp)
...
+         * provide the data for the bootstrap, which previously was at the
+         * secondary slot, was updated to the primary slot.
          */
-        slot = 0;
+        slot = BOOT_PRIMARY_SLOT;
     }
 
-#ifdef MCUBOOT_VALIDATE_SLOT0
-    rc = boot_validate_slot(0, NULL);
+#ifdef mcuboot_validate_primary_slot
```
utzig: `MCUBOOT_VALIDATE_PRIMARY_SLOT`

utzig:
@d3zd3z Fix the two issues I mentioned above and it should be OK, tested here, looks good.

boot/zephyr/Kconfig
```diff
@@ -72,37 +72,37 @@ config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
 config BOOT_VALIDATE_SLOT0
```
nvlsianpu: I would change this keyword as well. It produces `#define CONFIG_BOOT_VALIDATE_SLOT0 1` currently.

utzig: `mcuboot_config.h` is checking for this define as is here to avoid breaking compatibility with previous custom configurations, so I don't think it needs to be changed for now.

nvlsianpu: "Previous custom configuration" - do you mena proj.conf file in this repository or end-user configuration?

utzig: If someone made `BOOT_VALIDATE_SLOT0=n` in a custom config

nvlsianpu: ok - can live with this

nvlsianpu: Or maybe we can keep backward compatibility:

```suggestion
config BOOT_VALIDATE_SLOT0
    # Hidden selector for backward compatibility
    # deprecated
    bool
    select BOOT_VALIDATE_PRIMARY_SLOT
    
config BOOT_VALIDATE_PRIMARY_SLOT
```

boot/bootutil/src/encrypted.c
davidvincze: I think it would be much better to pass the slot number instead of the flash area.

utzig: If the values are not sequential for Zephyr, maybe it would be better to change it to something like:
```
int
boot_enc_valid(const struct flash_area *fap)
{
    switch (fap->fa_id) {                                                       
    case FLASH_AREA_IMAGE_0:                                                    
        return enc_state[0].valid;                                              
    case FLASH_AREA_IMAGE_1:                                                    
        return enc_state[1].valid;                                              
    default:                                                                    
        assert(0);
        return 0;                                                               
    }
}
```
But on another PR please!

utzig: Also checking that other `enc_state` accesses do not use the same math.

nvlsianpu: will prepare the patch

boot/zephyr/Kconfig
d3zd3z: Are we happy with it as it is now?

utzig: I am

utzig: But it's a nice suggestion from @nvlsianpu...

PR #441, opened 2019-03-14
image encryption: fix enc_state array indexing for zephyr

nvlsianpu:
enc_state table was indexed with assumption that
image flash area are subsequent and increasing numbers.
It might not be true while building zephyr.

Patch introduce flash_area_id_to_image_slot() implementation for
the zephyr port and uses it to assign proper slot number.
This API is already available in MyNewt.

### DIFF
--- a/boot/bootutil/src/encrypted.c
+++ b/boot/bootutil/src/encrypted.c
@@ -233,7 +233,11 @@ boot_enc_load(const struct image_header *hdr, const struct flash_area *fap,
     uint8_t enckey_type;
     int rc;
 
-    slot = fap->fa_id - FLASH_AREA_IMAGE_PRIMARY;
+    rc = flash_area_id_to_image_slot(fap->fa_id);
+    if (rc < 0) {
+        return rc;
+    }
+    slot = rc;
 
     /* Already loaded... */
     if (enc_state[slot].valid) {
@@ -305,7 +309,16 @@ boot_enc_load(const struct image_header *hdr, const struct flash_area *fap,
 int
 boot_enc_valid(const struct flash_area *fap)
 {
-    return enc_state[fap->fa_id - FLASH_AREA_IMAGE_PRIMARY].valid;
+    int rc;
+
+    rc = flash_area_id_to_image_slot(fap->fa_id);
+    if (rc < 0) {
+        /* can't get proper slot number - skip encryption, */
+        /* postpone the erro for a upper layer */
+        return 0;
+    }
+
+    return enc_state[rc].valid;
 }
 
 void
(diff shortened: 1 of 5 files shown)


boot/bootutil/src/encrypted.c
```diff
@@ -305,7 +309,14 @@ boot_enc_load(const struct image_header *hdr, const struct flash_area *fap,
 int
 boot_enc_valid(const struct flash_area *fap)
 {
-    return enc_state[fap->fa_id - FLASH_AREA_IMAGE_PRIMARY].valid;
+    int rc;
+
+    rc = flash_area_id_to_image_slot(fap->fa_id);
+    if (rc < 0) {
+        return rc;
```
utzig: `boot_enc_valid` is used as bool (maybe it should return bool!), so this would be true?

utzig: I think it would be safer to return `0` here, that would just skip trying encrypted stuff and postpone the error for a later function...

nvlsianpu: You are right -  `boot_enc_valid` is used as it returns boolean. I Will change its type.

boot/bootutil/src/encrypted.c
```diff
@@ -302,10 +306,19 @@ boot_enc_load(const struct image_header *hdr, const struct flash_area *fap,
...
+    int rc;
+
+    rc = flash_area_id_to_image_slot(fap->fa_id);
+    if (rc < 0) {
+        /* can't get proper slot number - skip encryption, */
+        /* postpone the erro for a upper layer */
+        return false;
+    }
+
+    return enc_state[rc].valid;
```
utzig: Since you are at it, maybe you should also make this field `bool`? :-P

boot/bootutil/src/encrypted.c
```diff
@@ -325,7 +339,13 @@ boot_encrypt(const struct flash_area *fap, uint32_t off, uint32_t sz,
...
 
-    enc = &enc_state[fap->fa_id - FLASH_AREA_IMAGE_PRIMARY];
+    rc = flash_area_id_to_image_slot(fap->fa_id);
+    if (rc < 0) {
+        assert(0);
+        return;
+    }
+
+    enc = &enc_state[rc];
     assert(enc->valid == 1);
```
utzig: And if you do the above suggestion, also remove the comparison here!

boot/bootutil/src/encrypted.c
nvlsianpu: it is uint8 - less RAM. I would keep as is.

boot/bootutil/src/encrypted.c
nvlsianpu: will change

PR #446, opened 2019-03-18
dfu: Add support for switching to do USB DFU

takumiando:
Sometimes, we'd like to switch whether to do USB DFU or not
by like button or something else.
This commit supports these cases.

intel_s1000_crb has nothing to do so it returns just 1.

### DIFF
--- a/boot/zephyr/include/usb_dfu/usb_dfu.h
+++ b/boot/zephyr/include/usb_dfu/usb_dfu.h
@@ -1,0 +1,7 @@
+/*
+ * Copyright (c) 2019 Atmark Techno, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+int is_entering_usb_dfu(void);
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -42,6 +42,7 @@ const struct boot_uart_funcs boot_funcs = {
 
 #ifdef CONFIG_BOOT_WAIT_FOR_USB_DFU
 #include <usb/class/usb_dfu.h>
+#include "usb_dfu/usb_dfu.h"
 #endif
 
 MCUBOOT_LOG_MODULE_REGISTER(mcuboot);
@@ -202,9 +203,11 @@ void main(void)
 #endif
 
 #ifdef CONFIG_BOOT_WAIT_FOR_USB_DFU
-    BOOT_LOG_INF("Waiting for USB DFU");
-    wait_for_usb_dfu();
-    BOOT_LOG_INF("USB DFU wait time elapsed");
+    if (is_entering_usb_dfu()) {
+        BOOT_LOG_INF("Waiting for USB DFU");
+        wait_for_usb_dfu();
+        BOOT_LOG_INF("USB DFU wait time elapsed");
+    }
 #endif
 
     rc = boot_go(&rsp);
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -67,6 +67,10 @@ if(EXISTS targets/${BOARD}.h)
   zephyr_library_compile_definitions(MCUBOOT_TARGET_CONFIG="${BOARD}.h")
 endif()
 
+if(EXISTS ${BOOT_DIR}/zephyr/targets/${BOARD}.c)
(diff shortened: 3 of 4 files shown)


utzig:
@rgundi @nvlsianpu Thoughts on this?

boot/zephyr/include/usb_dfu/usb_dfu.h
```diff
@@ -0,0 +1,7 @@
+/*
```
nvlsianpu: this files prototypes a board interface, not really features related strictly to usb_dfu - change name.

boot/zephyr/main.c
```diff
@@ -202,9 +203,11 @@ void main(void)
 #endif
 
 #ifdef CONFIG_BOOT_WAIT_FOR_USB_DFU
-    BOOT_LOG_INF("Waiting for USB DFU");
-    wait_for_usb_dfu();
-    BOOT_LOG_INF("USB DFU wait time elapsed");
+    if (is_entering_usb_dfu()) {
```
nvlsianpu: need default implementation of is_entering_usb_dfu() for any board, or #ifdef'ry

rgundi:
> @rgundi @nvlsianpu Thoughts on this?

I am not quite sure about the purpose of this PR. Currently, there's a Kconfig option which provides a "window of time" where a USB DFU can be performed during early boot. Another possible way to do this is to have a switch/gpio which can be turned ON to ensure the platform waits for USB DFU to occur (may be indefinitely). Of course, this feature would depend upon the platform to have that switch/gpio accessible to the user (something which intel_s1000_crb doesn't have). Is this the feature you want to implement? If yes, it is better to make the changes in Zephyr repository itself. You would have to modify the existing board.c in Zephyr for the relevant platform. You would also have to modify the wait_for_usb_dfu function to make the wait indefinite (may be split the function into 2; the existing outer for loop or a while(1) followed by the crux of the function which polls). Let me know what you think.

takumiando:
Thank you for replying!

I thought this addition like BIOS.
This following sequence is basic:

Power on
↓
Press some keys or switches
↓
select boot device or other menus
(in MCUboot, waiting for USB DFU)
↓
Time out, boot main image

So I think both switching for USB DFU and timeout are compatible.

Of course, some functions depending on boards should be written in board.c in Zephyr for each,
but switching to wait for USB DFU or not should be defined on MCUboot source tree
(and be distributed by boards).
Because "this USB DFU" is a one of the function of MCUboot as application.

How do you think?

boot/zephyr/main.c
takumiando: How about like this?

* common.c
```
__attribute__((weak)) int is_entering_usb_dfu(void)
{
        /* Nothing to do */
        return 1;
}
```

This definition is built by default.

boot/zephyr/include/usb_dfu/usb_dfu.h
takumiando: I have no idea how should I name...
Do you have any good idea?

rgundi:
> Thank you for replying!
> 
> I thought this addition like BIOS.
> This following sequence is basic:
> 
> Power on
> ↓
> Press some keys or switches
> ↓
> select boot device or other menus
> (in MCUboot, waiting for USB DFU)
> ↓
> Time out, boot main image
> 
> So I think both switching for USB DFU and timeout are compatible.
> 
> Of course, some functions depending on boards should be written in board.c in Zephyr for each,
> but switching to wait for USB DFU or not should be defined on MCUboot source tree
> (and be distributed by boards).
> Because "this USB DFU" is a one of the function of MCUboot as application.
> 
> How do you think?

Sorry. I am still not able to figure out the use case. Let me tell you what is currently supported.
1) Let's assume the mcuboot image is created with BOOT_WAIT_FOR_USB_DFU set to 1.
2) When the platform boots up, it waits for USB_DFU_WAIT_DELAY_MS milliseconds to let the user force a USB DFU.
3) If the user fails to force a USB DFU within this time, the platform continues booting and loads the image at slot0. If USB DFU is forced, the platform waits till the DFU is over and then goes ahead with the authentication of the new image at slot1 and the normal procedure ensues.

Can you now please let me know how your new proposal fits in the above scenario?

PR #447, opened 2019-03-18
zephyr bugfix: mcuboot doesn't link into code-partition

nvlsianpu:
Select the code partition as the target for link into.
This patch together with
https://github.com/zephyrproject-rtos/zephyr/pull/14620
fixes the issue which was imposed while zephyr moves
DTC before Kconfig processing in its build process.


nvlsianpu:
cc @SebastianBoe @carlescufi @galak @lemrey @rgundi

PR #450, opened 2019-03-26
Add use and documentation of west manifest to mcuboot

sigvartmh:
This PR will add a west manifest and remove unused symbols used for building cc310 support for nRF52840 in MCUBoot. Now mcuboot can use west to initialize zephyr it can also use revisions in the manifest to lock the support to a specific revision of zephyr.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -66,8 +66,6 @@ set(NRF_DIR "${MCUBOOT_DIR}/ext/nrf")
 if(CONFIG_BOOT_USE_NRF_CC310_BL)
 set(NRFXLIB_DIR ${MCUBOOT_DIR}/../nrfxlib)
 assert_exists(NRFXLIB_DIR)
-# Don't include this if we are using west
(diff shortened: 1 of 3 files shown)


sigvartmh:
@utzig @nvlsianpu

sigvartmh:
Couldn't find a way to put `west.yml` in `boot/zephyr` so I hope having it in the root folder is okei.

nvlsianpu:
@sigvartmh I think we already stated that west manifest should be provided by parent project like fw-nrfconnect-nrf. I something form this PR still needed to be integrated to the mcuboot?

west.yml
```diff
@@ -0,0 +1,35 @@
+west:
```
nvlsianpu: file TBR

sigvartmh: If the maintainers of MCUBoot will allow tight coupling with fw-nrfconnect-nrf I'm fine with that and I'll change this PR

sigvartmh: @nvlsianpu: This is now updated

PR #452, opened 2019-03-28
boot: zephyr: do not override TEXT_SECTION_OFFSET

lemrey:
It is no longer necessary to override TEXT_SECTION_OFFSET when BOARD_HAS_NRF5_BOOTLOADER. The nrf52840_pca10059 board no longer overrides TEXT_SECTION_OFFSET but sets the correct FLASH_LOAD_OFFSET instead, automatically.

Please @nvlsianpu, have a look.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -14,18 +14,6 @@ config MCUBOOT
 	select MPU_ALLOW_FLASH_WRITE if ARM_MPU
 	select USE_CODE_PARTITION if HAS_FLASH_LOAD_OFFSET
 
-if BOARD_HAS_NRF5_BOOTLOADER
-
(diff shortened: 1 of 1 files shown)


nvlsianpu:
@lemrey Is this fixes an issue, or just makes code cleaner?

lemrey:
In short, this is a workaround that is no longer necessary.

It was used when the nRF5 bootloader changed the text offset for applications, and that wouldn't work when compiling MCUboot, since MCUboot doesn't need an offset because it's linked in the `boot_partition`; so we manually set the text offset to zero.

This is no longer necessary after changes to DT / MCUboot overlay in Zephyr.

nvlsianpu:
counterpart of https://github.com/zephyrproject-rtos/zephyr/pull/14978

PR #454, opened 2019-03-29
imgtool: Discard public RSA if set for signing

ChristopherBittner:
If imgtool is used for signing with encryption
it will disallow using public RSA key for
signing.

Signed-off-by: Chris Bittner <chris.bittner@nordicsemi.no>

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -156,9 +156,9 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
     enckey = load_key(encrypt) if encrypt else None
     if enckey:
         if not isinstance(enckey, (keys.RSA2048, keys.RSA2048Public)):
-            raise Exception("Encryption only available with RSA")
-        if key and not isinstance(key, (keys.RSA2048, keys.RSA2048Public)):
-            raise Exception("Encryption with sign only available with RSA")
+            raise Exception("Encryption only available with RSA key")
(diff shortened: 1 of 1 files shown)


ChristopherBittner:
cc @nvlsianpu
details https://github.com/JuulLabs-OSS/mcuboot/issues/455

scripts/imgtool/main.py
```diff
@@ -156,9 +156,9 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
     enckey = load_key(encrypt) if encrypt else None
     if enckey:
         if not isinstance(enckey, (keys.RSA2048, keys.RSA2048Public)):
-            raise Exception("Encryption only available with RSA")
-        if key and not isinstance(key, (keys.RSA2048, keys.RSA2048Public)):
-            raise Exception("Encryption with sign only available with RSA")
+            raise Exception("Encryption only available with RSA key")
+        if key and not isinstance(key, keys.RSA2048):
+            raise Exception("Encryption with sign only available with private RSA signing key")
```
utzig: Your test is correct, but the messagemight induce a user to use the same private key for both signing and encrypting, which is never a secure thing to do. This should make it explicit that encryption requires a public key (works with keypair as well), and signing with private key (or keypair). Maybe removing references to encryption and displaying something like "Signing only available with private RSA" ?

ChristopherBittner: Good point. Will change it.

PR #456, opened 2019-03-31
zephyr: support uart detect and go into recovery mode

lgl88911:
During booting, if finded  "enter" of keypad, go into recovery mode.
This feature is common to all boards of zephyr. which can replace GPIO
detect.
Before go into recovery mode, the serial can output message.

Signed-off-by: Frank Li <lgl88911@163.com>


lgl88911:
@nvlsianpu Could you give some suggest about this pr? Thanks

d3zd3z:
@lgl88911 There are now conflicts with master, which will need to be resolved. Otherwise, I think we should review this change. I'll try to take a look at it.

lgl88911:
@d3zd3z I merge the master to my branch and fix confict, but ci test show:
```
Missing "Signed-off-by: Fabio Utzig <utzig@utzig.org>" in commit fda937a
```
this is not my commit, could you help give some suggest to fix it? Thansk

utzig:
@lgl88911 Do not merge master into your branch, rebase it on master.

lgl88911:
@utzig Thanks I fixed

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -28,11 +28,28 @@ extern "C" {
 
 #ifdef MCUBOOT_HAVE_LOGGING
 
+
```
d3zd3z: There are a few extra blank lines here. One blank should be sufficient.

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -28,11 +28,28 @@ extern "C" {
...
+    }
+#define BOOT_LOG_ERR BOOT_LOG_INF
+#define BOOT_LOG_WRN BOOT_LOG_INF
+#define BOOT_LOG_SIM BOOT_LOG_INF
+
+#else
 #define BOOT_LOG_ERR(...) MCUBOOT_LOG_ERR(__VA_ARGS__)
 #define BOOT_LOG_WRN(...) MCUBOOT_LOG_WRN(__VA_ARGS__)
 #define BOOT_LOG_INF(...) MCUBOOT_LOG_INF(__VA_ARGS__)
-#define BOOT_LOG_DBG(...) MCUBOOT_LOG_DBG(__VA_ARGS__)
```
d3zd3z: Is this a merge error? This seems to be removing the newly added `BOOT_LOG_DBG`.

boot/bootutil/include/bootutil/bootutil_log.h
```diff
@@ -28,11 +28,28 @@ extern "C" {
 
 #ifdef MCUBOOT_HAVE_LOGGING
 
+
+
+#define BOOT_LOG_DBG(...) MCUBOOT_LOG_DBG(__VA_ARGS__)
+
+#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
+#define BOOT_LOG_INF(format, args...)  {\
```
d3zd3z: Instead of adding a BOOT_LOG_INF, which is going to affect every use of BOOT_LOG_INF, perhaps we should add a special print message call just for this uart detect. If logging is enabled, it can just use one of the logging calls, and if not, it can just `console_write` the message. It doesn't look like any of the uses need printf functionality, so this could just output simple strings.

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
     }
 #endif
 
+#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
+
```
d3zd3z: Some lines in this block have trailing whitespace, which should be removed.

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
...
 
+#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
+   
+    char str[2];
+    volatile int newline = 0;
+    volatile int timeout = 10000;
+    int step = 1;
+   
+    BOOT_LOG_INF("Wait:");
+    while(1){
```
d3zd3z: This should be `while (1) {`.

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
...
+        console_read(str, 2, &newline);
+        if(newline == 1){
```
d3zd3z: And this `if (newline == 1) {`, with the space after the `if` and before the brace. There are a few similar below.

boot/zephyr/serial_adapter.c
```diff
@@ -147,8 +147,12 @@ boot_uart_fifo_callback(struct device *dev)
 		if (cur < CONFIG_BOOT_MAX_LINE_INPUT_LEN) {
 			cmd->line[cur++] = byte;
 		}
-
-		if (byte ==  '\n') {
+		#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
+		if(byte == '\r'){
```
d3zd3z: Spacing. Also, do we want to treat '\r' as if it were a '\n', or just ignore it. Most terminals should be sending either just a LF, or a CR/LF pair.

boot/bootutil/include/bootutil/bootutil_log.h
lgl88911: It's good idea, I fix it

boot/zephyr/serial_adapter.c
lgl88911: In nrf52832 serial_adapter read uart always '\r', No matter how I set up the serial tool, I can't get '\n', Maybe It is a issue in uart drv of nrf52832. If you have other hw platform, Could you help double confirm it? Thanks

d3zd3z: I suppose it doesn't hurt anything to allow the '\r' as well, but it does make me wonder how the code was used by others.

lgl88911: In boot_uart_fifo_callback have below code:
```
if (byte ==  '\n'){
	cmd->len = cur;
	sys_slist_append(&lines_queue, &cmd->node);
	cur = 0;
	cmd = NULL;
}
```
It means only meeting '\n' , add cmd to list.  
I hope press **Enter** key to goto recovery, During mcuboot startup,long press Enter key,  boot_uart_fifo_callback   always gets '\r' from serial , It casue no new line to list,  main can't read newline by console_read cause can't go into recovery mode.
So I convert '\r' to '\n' to get newline and go into recovery mode

d3zd3z:
My original comments don't seem to be addressed. There are some lines with trailing whitespace, and the messages for the serial recovery shouldn't go through the BOOT_LOG... macros (which will cause any other use of BOOT_LOG_DEBUG to be compiled in, but have their own macro. Also, please fix the spacing, things should be like `while (1) {`.

boot/zephyr/Kconfig
```diff
@@ -265,12 +264,26 @@ config BOOT_SERIAL_CDC_ACM
 
 endchoice
 
+choice
+	prompt "Detect type"
+	default BOOT_SERIAL_UART_DETECT
```
nvlsianpu:
```suggestion
	default BOOT_SERIAL_GPIO_DETECT
```

prefer previous detection mechanism as the default one.

boot/zephyr/Kconfig
```diff
@@ -294,6 +307,9 @@ config BOOT_SERIAL_DETECT_PIN_VAL
 	  Logic value of the detect pin which triggers serial recovery
 	  mode.
 
+endif #BOOT_SERIAL_GPIO_DETECT
+
+if BOOT_SERIAL_GPIO_DETECT || BOOT_SERIAL_UART_DETECT
```
nvlsianpu: remove this line `if BOOT_SERIAL_GPIO_DETECT || BOOT_SERIAL_UART_DETECT`

boot/zephyr/Kconfig
```diff
@@ -306,6 +322,8 @@ config RECOVERY_UART_DEV_NAME
 	  This option specifies the name of UART device to be used for
 	  serial recovery.
 
+endif #BOOT_SERIAL_GPIO_DETECT || BOOT_SERIAL_UART_DETECT
```
nvlsianpu: remove this line `endif #BOOT_SERIAL_GPIO_DETECT || BOOT_SERIAL_UART_DETECT`

boot/zephyr/boards/uart_detect.conf
```diff
@@ -0,0 +1,9 @@
+CONFIG_MCUBOOT_SERIAL=y
```
nvlsianpu: this is not a board dedicated file, how you want to use it? with -DCONF_FILE?

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
     }
 #endif
 
+#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
```
nvlsianpu: use #elif

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
...
+   
+    char str[2];
+    volatile int newline = 0;
```
nvlsianpu: no reason for make it volatile

boot/zephyr/main.c
```diff
@@ -223,6 +227,36 @@ void main(void)
...
+    volatile int timeout = 10000;
```
nvlsianpu: no reason for make it volatile

boot/zephyr/serial_adapter.c
nvlsianpu: To be analyzed...

boot/bootutil/include/bootutil/bootutil_log.h
nvlsianpu: my NACK for what is done here

lgl88911:
> My original comments don't seem to be addressed. There are some lines with trailing whitespace, and the messages for the serial recovery shouldn't go through the BOOT_LOG... macros (which will cause any other use of BOOT_LOG_DEBUG to be compiled in, but have their own macro. Also, please fix the spacing, things should be like `while (1) {`.

So sorry for this, I modify these content in other branch, but push add_uart_detect, it's my mistake, I sync it and push again, Please review again, Thanks so much

boot/zephyr/boards/uart_detect.conf
lgl88911: Yes, Below is a sample:
cmake -DBOARD=nrf52_moderate -DBOARD_ROOT=/home/frank/work/project/nrf52_moderate -DCONF_FILE="prj.conf boards/uart_detect.conf" ..

boot/zephyr/serial_adapter.c
```diff
@@ -148,7 +148,13 @@ boot_uart_fifo_callback(struct device *dev)
...
 		}
 
-		if (byte ==  '\n') {
+#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
+		if (byte == '\r') {
+			byte = '\n';
+		}
+#endif
+
+		if (byte ==  '\n'){
```
d3zd3z: Missed a space between "){".

lgl88911: Thanks,  I fixed . Is there a tool or script to detect this format?

d3zd3z: It might be possible to adapt checkpatch.pl from the Linux kernel, but it probably also detects a lot of things we do want. It might also be possible to configure indent or equivalent to check for this.

lgl88911:
@nvlsianpu 
I modify below, Do you think it OK? Thanks
from
```
#define BOOT_LOG_DBG(...) MCUBOOT_LOG_DBG(__VA_ARGS__)

#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
#define BOOT_LOG_INF(format, args...)  {\
```
to
```
#ifdef CONFIG_BOOT_SERIAL_UART_DETECT
#define BOOT_UART_DETECT_LOG_INF(str)   console_write(str, strlen(str))

#define BOOT_LOG_ERR(...) IGNORE(__VA_ARGS__)
#define BOOT_LOG_WRN(...) IGNORE(__VA_ARGS__)
#define BOOT_LOG_INF(...) IGNORE(__VA_ARGS__)
#define BOOT_LOG_DBG(...) IGNORE(__VA_ARGS__)
#define BOOT_LOG_SIM(...) IGNORE(__VA_ARGS__)
```

heinwessels:
Any notice on the last state of this PR? Would it be possible for me to pick it up and finish it? We might find this useful.

d3zd3z:
@nvlsianpu have your concerns been addressed?

nvlsianpu:
@lgl88911 I believe that recently similar feature was adopted, see `BOOT_SERIAL_WAIT_FOR_DFU`

nvlsianpu (commented):
^^ see my concern of having similar feature

heinwessels:
> @lgl88911 I believe that recently similar feature was adopted, see `BOOT_SERIAL_WAIT_FOR_DFU`

This does not help our use-case, since we do not have a DFU connection, only UART. Maybe the current functionality can be expanded to be able to wait for a UART connection as well.

nvlsianpu:
Option I mentioned allows MCUboot to wait for a prescribed duration to see if DFU is invoked by receiving a mcumgr comand.
So, within the timeout the serial recovery can be entered by receiving any mcumgr command. Mcumgr command are received over UART in most cases.
> This does not help our use-case, since we do not have a DFU connection, only UART. 

I don't get why. Aren't you process with mcumgr commands anyway?

lgl88911:
@nvlsianpu Yes,  `BOOT_SERIAL_WAIT_FOR_DFU` can do the same. It is a relatively simple way to monitor the DFU data packets from the serial port and enter the recovery  mode when starting up. 
But Waiting for the DFU header is more prone to loss than waiting for a single character, which means that the boot needs to wait a little longer for the timeout, which may prolong the normal boot time. This is just my inference, and no actual test comparison is made.

nvlsianpu:
@lgl88911 I understand your case now.
I wonder whether would be possible to make BOOT_SERIAL_WAIT_FOR_DFU implementation optionally sensitive for a single character.

lgl88911:
> @lgl88911 I understand your case now. I wonder whether would be possible to make BOOT_SERIAL_WAIT_FOR_DFU implementation optionally sensitive for a single character, so the codebase will be shared with your feature?

Do you mean to check both specific single character and DFU header？

nvlsianpu:
> Do you mean to check both specific single character and DFU header

I mean check specific single character for one option, or DFU request for the second on. Selected at build-time.

nvlsianpu:
^^looks like proper place.

PR #458, opened 2019-04-04
Add simulator support for multi-image

d3zd3z:
Support testing the multi-image code in the simulator. This is based on #435 but with only the multi-image code present.


d3zd3z:
This can  be tested by merging with #317. In order to test, you will also need to run:

```bash
cargo test --features multiimage,validate-primary-slot,overwrite-only,sig-rsa
```

A patch to add this to `.travis.yml` will make sense once both of these prs can be combined.

boot/bootutil/src/caps.c
```diff
@@ -47,3 +47,12 @@ uint32_t bootutil_get_caps(void)
 
         return res;
 }
+
+uint32_t bootutil_get_num_images(void)
+{
+#if defined(MCUBOOT_IMAGE_NUMBER)
+	return MCUBOOT_IMAGE_NUMBER;
```
utzig: tab -> spaces

sim/mcuboot-sys/src/area.rs
```diff
@@ -176,11 +176,8 @@ pub enum FlashId {
     Image0 = 1,
     Image1 = 2,
     ImageScratch = 3,
-    Nffs = 4,
-    Core = 5,
-    RebootLog = 6,
-    Image2 = 7,
-    Image3 = 8,
+    Image2 = 4,
+    Image3 = 5,
```
utzig: If you had to do this, I assume the prev commit (a15e8f8ca297f4aab787b7b9374fd7f87d2606dc) so why not just squash there? Also would not it be better to make it:
```
    Image0 = 1,
    Image1 = 2,
    ImageScratch = 3,
    Image2 = 4,
    Image3 = 5,
    Nffs = 6,
    Core = 7,
    RebootLog = 8,
```

sim/src/image.rs
```diff
@@ -87,11 +90,11 @@ impl ImagesBuilder {
             dev_id: slot1_dev_id,
         };
 
-        ImagesBuilder {
+        Some(ImagesBuilder {
```
utzig: Does it ever return `None` though?

sim/src/image.rs
```diff
@@ -104,22 +110,28 @@ impl ImagesBuilder {
...
             total_count: None,
         }
     }
 
     /// Construct an `Images` for normal testing.
     pub fn make_image(self) -> Images {
         let mut images = self.make_no_upgrade_image();
-        mark_upgrade(&mut images.flash, &images.slots[1]);
+        for image in &images.images {
+            mark_upgrade(&mut images.flash, &image.slots[1]);
```
utzig: This is OK, but semantics needs to be defined later, do we have to mark every upgrade slot?

sim/src/image.rs
```diff
@@ -852,6 +873,50 @@ impl Images {
 
         (flash, resets)
     }
+
+    /// Verify the image in the given flash device, the specified slot
+    /// against the expected image.
+    fn verify_images(&self, flash: &SimMultiFlash, slot: usize, against: usize) -> bool {
```
utzig: Maybe would be clearer to add some enumerations like `slotPrimary, slotSecundary, againstPrimary, againSecondary` and so on, since we'll be working not only with `0` and `1` but also `2`, `3`... (yes, while they are still `0` and `1` for that particular slot set)

sim/src/image.rs
```diff
@@ -852,6 +873,50 @@ impl Images {
...
+                             }) {
+                return false;
+            }
+        }
+        true
+    }
+
+    /// Verify that the trailers of the images have the specified
+    /// values.
+    fn verify_trailers(&self, flash: &SimMultiFlash, slot: usize,
```
utzig: It might be possible that checking both images together in those tests might break some of the testing that expects errors, etc. This applies here and for `verify_images` above, so it's good to be aware. For mono images this should remain working as before, so looks good!

sim/src/image.rs
d3zd3z: Yes, there are several earlier returns of None.

sim/src/image.rs
d3zd3z: I'd like to write some additional tests that are able to upgrade one slot without the other. There also need to be tests that verify that dependencies are checked properly. But, currently, it always upgrades both slots.

sim/src/image.rs
utzig: Ok, I see you added them with another commit (a15e8f8ca2)...

sim/src/image.rs
d3zd3z: Will they ever be other than zero and 1? At least the `slot` argument is always used as an array index. The against is against the 'primaries' and 'upgrades' fields (which reminds me that I need to rename those to 'primary' and 'upgrade' since they only refer to one thing.

It actually seems clearer to me to have these just be 0 and 1, although I can see how an enum makes it clearer. The 'slot' argument would end up being cast back to a usize anyway, though. Thoughts?

sim/src/image.rs
d3zd3z: These loops essentially are an 'and' condition. It at least seems to work for all of the current tests with multiple images. It will get an error if either of the images is incorrect.

boot/bootutil/src/caps.c
d3zd3z: Given that the file already has a mix of tabs and spaces, maybe I should just have a separate patch after this that fixes it all to spaces.

sim/mcuboot-sys/src/area.rs
d3zd3z: Is there any reason to have the extra areas that we don't use?

boot/bootutil/src/caps.c
utzig: Or just change it all together now

d3zd3z: I guess the other approach is to start with a patch that fixes the existing tabs, and then make this patch correct.

sim/mcuboot-sys/src/area.rs
utzig: Whatever about the extra areas, does it work like this?
```
    Image0 = 1,
    Image1 = 2,
    ImageScratch = 3,
    Nffs = 4,
    Core = 5,
    RebootLog = 6,
    Image2 = 7,
    Image3 = 8,
```
If it does not, you are fixing a broken previous commit, so squash...

d3zd3z: Ok, yeah, I'll squash it.

PR #459, opened 2019-04-04
Add SUIT draft 3 support to the simulator

d3zd3z:
This builds on earlier simulator changes in #458 to add support for SUIT draft 3. As this draft is now deprecated, which will not merge as is, but be replaced by a version that supports the new manifest. This replaces the work in #435.


d3zd3z:
Closing as this is also deprecated, and future work is not scheduled.

PR #466, opened 2019-04-12
Remove bootutil unit test for Mynewt

sjanc:
mcuboot tests are done under sim and there is no need to maintain
separate tests just for Mynewt. Especially that those no longer
compile nor are maintained.

Also Mynewt code will rather depend on release and not master so
there is no need to run those on Mynewt anyway.


sjanc:
@utzig @ccollins476ad

utzig:
Ain't you removing a bit too much? (e.g `boot/bootutil/include/bootutil/bootutil.h`)

utzig:
I think you should only remove stuff under `boot/bootutil/test/` tree, no?

utzig:
LGTM, any thoughts @ccollins476ad?

sjanc:
those tests are non functional and not used anyway... can we just remove them?

sjanc:
I mean, if someone ever needs them can restore and fix..  it is all in git history

PR #467, opened 2019-04-12
zephyr: fix logging

mbolivar:
The removal of CONFIG_BOOT_HAVE_LOGGING in 15aa6ef ("zephyr: remove BOOT_HAVE_LOGGING Kconfig option") missed some uses. In particular, the one in mcuboot_config.h that converts Kconfig options to MCUboot's platform-agnostic configuration macros.

This broke logging even when the relevant Kconfig options were on.

Fix things up by using CONFIG_LOG instead.

Tested logging works again on nrf52840_pca10056. Build tested nrf51_pca10056 to make sure it still fits. The nrf52840_pca10059 build overflows flash by 0.16%, but that was a problem before 15aa6ef.

@nvlsianpu I think we should fix the nrf52840_pca10059 build break in a separate patch if it's OK with you. This PR is just to fix the logging regression.

### DIFF
--- a/boot/zephyr/boards/nrf51_pca10028.conf
+++ b/boot/zephyr/boards/nrf51_pca10028.conf
@@ -1,5 +1,6 @@
-# Due the small boot partition can't use logging and debug optimalization
-# out-off-the-box. For using these features need to increase boot partition via
-# zephyr DTS.
-CONFIG_BOOT_HAVE_LOGGING=n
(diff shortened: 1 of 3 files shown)


mbolivar:
cc @mnkp

nvlsianpu:
thanks for the fix!

nvlsianpu:
cc @carlescufi

PR #469, opened 2019-04-16
boot: bootutil: Change conditional to valid C

d3zd3z:
The change

    boot: bootutil: Simplify check for crypto backends to reduce complexity

    Adding multiple crypto backends will grow quadraticly. This change will
    ensure that the growth will be linear and generate less complexity.

unfortunately is not legal C.  It can be fixed by eliminating the
intermediate macro, so that the `defined` keywords are within the `#if`.

### DIFF
--- a/boot/bootutil/include/bootutil/sha256.h
+++ b/boot/bootutil/include/bootutil/sha256.h
@@ -31,10 +31,9 @@
 
 #include "mcuboot_config/mcuboot_config.h"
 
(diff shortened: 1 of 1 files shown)


utzig:
@d3zd3z I think you can merge this without waiting for the CI, since it does not fail on the CI anyway (only for newer gcc versions)

mbolivar:
My bad! I originally suggested the change in that form. Thanks for the fix.

PR #471, opened 2019-04-18
Add Mynewt specific release information

utzig:


utzig:
@ccollins476ad Thanks!

PR #472, opened 2019-04-24
[SIM] Relax trailer verification for multi-image

utzig:
The simulator does trailer verification to ensure the status is what would be expected when operating a swap. When doing a multi-image swap upgrade one of the images must be in a known status but others might not be, depending on which image number is being upgraded at the moment. This patch relaxes the verification to assure at least one is in a known status.


utzig:
@d3zd3z Please do not merge, I think I will change this slightly...

utzig:
This won't be the correct solution...

PR #473, opened 2019-04-25
boot: serial recovery: Add pca10090 default detect pin

andvib:
Add button 1 as default serial detect pin for nrf9160_pca10090

Signed-off-by: Andreas Vibeto <andreas.vibeto@nordicsemi.no>


andvib:
cc @nvlsianpu

PR #475, opened 2019-05-07
Add watchdog feeding macro

utzig:
When HW / OS provides an always enabled watchdog, this macro can optionally be implemented to avoid resets which are expected to occur under normal conditions when swapping very large images or
running on slower devices.

### DIFF
--- a/samples/mcuboot_config/mcuboot_config.template.h
+++ b/samples/mcuboot_config/mcuboot_config.template.h
@@ -116,4 +116,17 @@
  * "assert" is used. */
 /* #define MCUBOOT_HAVE_ASSERT_H */
 
+/*
+ * Watchdog feeding
+ */
+
+/* This macro might be implemented if the OS / HW watchdog is enabled while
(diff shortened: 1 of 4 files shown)


samples/mcuboot_config/mcuboot_config.template.h
```diff
@@ -116,4 +116,17 @@
  * "assert" is used. */
 /* #define MCUBOOT_HAVE_ASSERT_H */
 
+/*
+ * Watchdog tickling
```
d3zd3z: The conventional term for this is "kicking" the watchdog. The typical more polite term is "petting" the watchdog. Tickling makes for an interesting metaphor.

nvlsianpu: We use term `feed` in Nordic nrfx watchdog driver.

ccollins476ad: For what it's worth, I've heard "tickle" used before.

PR #476, opened 2019-05-08
Add support for RSA-3072 signatures

utzig:
Tested on HW with Mynewt/Zephyr; imgtool changes should be done, simulator support as well. 

PS: This was not tested with encryption.


utzig:
@tamasban ^

boot/bootutil/src/image_validate.c
```diff
@@ -182,7 +186,7 @@ bootutil_img_validate(struct image_header *hdr, const struct flash_area *fap,
     int key_id = -1;
 #endif
     struct image_tlv tlv;
-    uint8_t buf[256];
+    uint8_t buf[384];  /* XXX RSA3072 has the largest key! */
```
nvlsianpu: shouldn't the size be compile time optional? RSA3072 might be not enabled.

utzig: Could be, it was previously sized to the largest possible key; could also be smaller than 256 for EC.

boot/bootutil/src/image_rsa.c
```diff
@@ -21,7 +21,7 @@
 
 #include "mcuboot_config/mcuboot_config.h"
 
-#ifdef MCUBOOT_SIGN_RSA
+#if defined(MCUBOOT_SIGN_RSA) || defined(MCUBOOT_SIGN_RSA3072)
```
mbolivar: Looks like support for this is being added to bootutil before it's being added to mcuboot_config.h. Why? That wouldn't be the right order.

I also think there should be an `MCUBOOT_SIGN_RSA_LEN` (or `...MAX_LEN`) instead of a separate MCUBOOT_SIGN_RSA3072 option. It won't be long before 4096, after all (actually, I think that proving this design out to RSA 4096 as well would be a good idea to see how it will scale over the next few decades).

That would also allow you to set the size of `buf` below more accurately on smaller targets, conserving memory more carefully.

boot/bootutil/src/image_validate.c
utzig: This should be OK now.

boot/bootutil/src/image_rsa.c
utzig: I changed the order of the commits and now Zephyr/Mynewt config is added before the `bootutil` changes but I did not break for `mcuboot_config` specific commits. I also made the size of the key more parameterizable instead of having to create new ifdefery now to add RSA-4096 or some other size does not required `bootutil` changes (just OS specific configs).

boot/zephyr/Kconfig
```diff
@@ -66,6 +66,17 @@ config BOOT_SIGNATURE_TYPE_RSA
 	select BOOT_USE_MBEDTLS
 	select MBEDTLS
 
+if BOOT_SIGNATURE_TYPE_RSA
+choice
```
mbolivar: You might want to consider using a `range` here (between the 2048 and 3072 values), and just documenting that only those two values are valid. Not sure though; would like to know what people think.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -19,7 +19,14 @@
 #ifndef __BOOTSIM__
 
 #ifdef CONFIG_BOOT_SIGNATURE_TYPE_RSA
-#define MCUBOOT_SIGN_RSA
+#  define MCUBOOT_SIGN_RSA
+#  if CONFIG_BOOT_SIGNATURE_TYPE_RSA_2048
+#    define MCUBOOT_SIGN_RSA_LEN 2048
```
mbolivar: Using a range would allow you to avoid these extra ifdefs.

boot/zephyr/prj.conf
```diff
@@ -14,11 +14,16 @@ CONFIG_BOOT_BOOTSTRAP=n
 CONFIG_BOOT_SIGNATURE_TYPE_RSA=y
 CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=n
 
+### RSA key size
+CONFIG_BOOT_SIGNATURE_TYPE_RSA_2048=y
```
mbolivar: And would let you do something like `CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN` here.

scripts/imgtool/keys/__init__.py
```diff
@@ -53,13 +53,13 @@ def load(path, passwd=None):
                 backend=default_backend())
 
     if isinstance(pk, RSAPrivateKey):
-        if pk.key_size != 2048:
+        if pk.key_size not in [2048, 3072]:
```
mbolivar: It looks like there's a variable for this (`RSA_KEY_SIZES`); why not use it in this file?

boot/zephyr/Kconfig
utzig: The problem is that `imgtool` is still only able to generate RSA 2048 and 3072 at the moment, and I am not sure we should allow for free choice of number of bits (even between a range).

mbolivar: Totally fair -- I'm thinking that the value could be a `_LEN` here, which could e.g. be checked in the header file (`#if CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN != 2048 && .... != 3072 / #error / #endif`).

Would make the zephyr config easier to change later -- just one option, which is the number of bits, but without sacrificing the restriction that only some are valid.

utzig: Ok, updated to use a `range`.

scripts/imgtool/keys/__init__.py
utzig: Fixed

boot/zephyr/Kconfig
mbolivar: Thanks!

PR #479, opened 2019-05-10
Add ed25519 signature support

utzig:
This adds `ed25519` support for imgtool, so that it can generate keys, export public keys, and sign images; `bootutil` supports validation of ed25519 under Zephyr (where I am testing).


utzig:
Removing "[WIP]" because this is now feature complete. Mynewt support was tested with https://github.com/apache/mynewt-newt/pull/299; simulator support has been added; a few issues and cleanups might remain but it is already completely functional.

d3zd3z:
First commit, we should probably be explicit in the documentation that we are implementing "prehash" Ed25519, aka SHA256-Ed25519, since we are feeding the Ed25519 algorithm a hash of the input. Ed25519 is a little unusual in that it can sign arbitrarily long data (there is a Hash internally), but we are using it in prehash mode, where we always feed it the output of a SHA256 hash.

Security-wise, this is no weaker than the other signing algorithms we use, although it does prevent one of Ed25519's mitigations against a certain hash prefix attack. I don't foresee this being an issue with the way that signatures work in MCUboot.

ext/fiat/README.mcuboot
```diff
@@ -0,0 +1,13 @@
+This code is based on the boringssl version of fiat-crypto (https://github.com/mit-plv/fiat-crypto)
+
+The version bundled here had most if its non-used code removed, like signing,
+X25519 curve operations not required by verification, etc.
+
+Under boringssl tree, fiat can be found in third_party/fiat. At the time it was
+copied, the last commit was:
+
+commit f109f2087349712d3ac717d15fab48e130618110
```
d3zd3z: I think this information should be placed in the commit text that adds the fiat code. It doesn't necessarily hurt anything to put it in this `README.mcuboot` file, since it is a good record of the initial version. But it would be good for any commit that brings in changes from the upstream version of this tree have at a minimum the URL and sha hash of the tree version that was brought in.

Another option is to use the `git subtree` command, whose explicit purpose is to bring code from another tree, and keep track of the versions as brought in. The end result wouldn't be different, but it has a specific formatting of the parent URL and hash that allow the tool to later be used to bring patches into the current tree.

ext/fiat/pkg.yml
```diff
@@ -0,0 +1,24 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
```
d3zd3z: Extraneous space

utzig:
@d3zd3z Your suggestions where applied, and `verify` support was added to `imgtool`.

Issue #480, opened 2019-05-16
Watchdog reset during bootloader stage breaks image swap


MariuszSkamra:
mcuboot version: 1.2.0/v1.3.0-68-g272702a + 0001-MCUBOOT-Config-file-changes.patch
Zephyr version: 1.13.0/zephyr-v1.14.0-834-ga2fa8e9c1c + 0001-ZEPHYR-Add-watchdog-to-smp_svr-sample.patch
Board: nrf52840-Preview-DK
Zephyr SDK: 0.9.5 (for 1.13.0)/0.10.0
mcumgr 0.0.0-dev

I enclose binaries, so that one can build yourself or use mine.
In the patch I've added Watchdog to Zephy's smp_srv sample.

Steps to reproduce:

1. Build boot, application and sign it.
```
$ ./scripts/imgtool.py sign --key root-ec-p256.pem --slot-size 0x69000 --header-size 0x200 --align 8 --version 1.0.0 ../zephyrproject/zephyr/samples/subsys/mgmt/mcumgr/smp_svr/build/zephyr/zephyr.hex zephyr_signed.hex
```
```
$ mergehex -m boot/zephyr/build/zephyr/zephyr.hex zephyr_signed.hex -o zephyr_boot_signed.hex
```

2. Chiperase and flash:

```
$ nrfjprog --program zephyr_boot_signed.hex --chiperase --reset
Parsing hex file.
Erasing user available code and UICR flash areas.
Applying system reset.
Checking that the area to write is not protected.
Programing device.
Applying system reset.
Run.
```

3. Verify
```
$ mcumgr --conntype ble --connstring 'peer_name=Zephyr' image list
Images:
 slot=0
    version: 1.0.0
    bootable: true
    flags: active confirmed
    hash: 36699217adc9ff5c08a66cd0845a187830977528ccab345480fac6fa54d9c67f
Split status: N/A (0)
```
```
$ nrfjprog --memrd 0x74e58 --n 424
0x00074E58: FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFF   |................|
0x00074E68: FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFF   |................|
0x00074E78: FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFF   |................|
0x00074E88: FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFF   |................|
[...]

PR #482, opened 2019-05-17
Fix interrupted revert

utzig:
This extends the test+revert case to also force interruptions on revert. Previously only the test stage was interrupted, and this oversight on testing has masked an issue where the revert swapping is not resilient to failures, causing an extra swap after it has finished, ending up with the test image as confirmed in the primary slot and the old image in the secondary slot.

It might be related to #480, although the issue that is being tackled here does not end in a revert forever loop.


utzig:
Merged with #485

PR #485, opened 2019-05-23
Fix double swap on interrupted revert

### LINKED ISSUE #480: Watchdog reset during bootloader stage breaks image swap

ccollins476ad:
This fixes #480.

When mcuboot rewrites image trailers during a swap, some information is lost.  If a reset occurs before the swap completes, mcuboot may not be able to determine which swap type to resume upon startup.  Specifically, if a "revert" swap gets interupted, mcuboot will perform an extraneous swap on the subsequent boot.  See https://github.com/JuulLabs-OSS/mcuboot/issues/480 for details.

**NOTE:** A different PR (#482) adds unit tests which test this fix.

This PR adds an additional field to the image trailer: `swap-type`.  The new trailer structure is illustrated below:

```
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    ~                                                               ~
    ~    Swap status (BOOT_MAX_IMG_SECTORS * min-write-size * 3)    ~
    ~                                                               ~
    |                 Encryption key 0 (16 octets) [*]              |
    |                                                               |
    |                 Encryption key 1 (16 octets) [*]              |
    |                                                               |
    |                      Swap size (4 octets)                     |
    |   Swap type   |           0xff padding (7 octets)             |
    |   Copy done   |           0xff padding (7 octets)             |
    |   Image OK    |           0xff padding (7 octets)             |
    |                       MAGIC (16 octets)                       |
    |                                                               |

[*]: Only present if the encryption option is enabled (`MCUBOOT_ENC_IMAGES`).
```

The `swap-type` field contains one of the `BOOT_SWAP_TYPE_[...]` constants.  Every time a trailer is written, this field is written along with it.  When resuming an interrupted swap, mcuboot uses this field alone to determine the type of swap being resumed. For new swap operations (non-resume case), this field is not read at all; instead, mcuboot consults the `boot_swap_tables` array to determine the swap operation to perform (as it did prior to this PR).

Some additional changes were necessary to make all the simulated unit tests pass:

* Before initiating a new swap operation, always write the image trailer to the scratch area.  This step allows mcuboot to persist the `swap-type` field somewhere before erasing the trailer in the primary slot.  If a reset occurs immediately after the erase, mcuboot recovers by using the trailer in the scratch area.

* Related to the above: if the scratch area is being used to hold status bytes (because there are no spare sectors in the primary slot), erase the scratch area immediately after the trailer gets written to the primary slot.  This eliminates ambiguity regarding the location of the current trailer in case a reset occurs shortly afterwards.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -119,6 +119,32 @@ boot_flag_decode(uint8_t flag)
     return BOOT_FLAG_SET;
 }
 
+/**
+ * Determines if a status source table is satisfied by the specified magic
+ * code.
+ *
+ * @param tbl_val               A magic field from a status source table.
+ * @param val                   The magic value in a trailer, encoded as a
+ *                                  BOOT_MAGIC_[...].
+ *
+ * @return                      1 if the two values are compatible;
+ *                              0 otherwise.
+ */
+int
+boot_magic_compatible_check(uint8_t tbl_val, uint8_t val)
+{
+    switch (tbl_val) {
+    case BOOT_MAGIC_ANY:
+        return 1;
+
+    case BOOT_MAGIC_NOTGOOD:
+        return val != BOOT_MAGIC_GOOD;
+
+    default:
+        return tbl_val == val;
+    }
+}
+
 uint32_t
 boot_trailer_sz(uint8_t min_write_sz)
 {
@@ -136,7 +162,6 @@ boot_trailer_sz(uint8_t min_write_sz)
 static uint32_t
 boot_magic_off(const struct flash_area *fap)
 {
-    assert(offsetof(struct image_trailer, magic) == 16);
     return fap->fa_size - BOOT_MAGIC_SZ;
 }
 
@@ -168,17 +193,21 @@ boot_status_off(const struct flash_area *fap)
     return fap->fa_size - off_from_end;
 }
 
+uint32_t
+boot_swap_type_off(const struct flash_area *fap)
+{
+    return fap->fa_size - BOOT_MAGIC_SZ - BOOT_MAX_ALIGN * 3;
+}
+
 static uint32_t
 boot_copy_done_off(const struct flash_area *fap)
 {
-    assert(offsetof(struct image_trailer, copy_done) == 0);
     return fap->fa_size - BOOT_MAGIC_SZ - BOOT_MAX_ALIGN * 2;
 }
 
 static uint32_t
 boot_image_ok_off(const struct flash_area *fap)
 {
-    assert(offsetof(struct image_trailer, image_ok) == 8);
     return fap->fa_size - BOOT_MAGIC_SZ - BOOT_MAX_ALIGN;
 }
 
@@ -214,6 +243,16 @@ boot_read_swap_state(const struct flash_area *fap,
         state->magic = BOOT_MAGIC_UNSET;
     } else {
         state->magic = boot_magic_decode(magic);
+    }
+
+    off = boot_swap_type_off(fap);
+    rc = flash_area_read_is_empty(fap, off, &state->swap_type,
+            sizeof state->swap_type);
+    if (rc < 0) {
+        return BOOT_EFLASH;
+    }
+    if (rc == 1 || state->swap_type > BOOT_SWAP_TYPE_REVERT) {
+        state->swap_type = BOOT_SWAP_TYPE_NONE;
     }
 
     off = boot_copy_done_off(fap);
@@ -405,34 +444,19 @@ boot_write_magic(const struct flash_area *fap)
 }
 
 static int
-boot_write_flag(int flag, const struct flash_area *fap)
+boot_write_trailer_byte(const struct flash_area *fap, uint32_t off,
+                        uint8_t val)
 {
-    uint32_t off;
-    int rc;
     uint8_t buf[BOOT_MAX_ALIGN];
     uint8_t align;
     uint8_t erased_val;
-
-    switch (flag) {
-    case BOOT_FLAG_COPY_DONE:
-        off = boot_copy_done_off(fap);
-        BOOT_LOG_DBG("writing copy_done; fa_id=%d off=0x%x (0x%x)",
-                     fap->fa_id, off, fap->fa_off + off);
-        break;
-    case BOOT_FLAG_IMAGE_OK:
-        off = boot_image_ok_off(fap);
-        BOOT_LOG_DBG("writing image_ok; fa_id=%d off=0x%x (0x%x)",
-                     fap->fa_id, off, fap->fa_off + off);
-        break;
-    default:
-        return BOOT_EBADARGS;
-    }
+    int rc;
 
     align = flash_area_align(fap);
     assert(align <= BOOT_MAX_ALIGN);
     erased_val = flash_area_erased_val(fap);
     memset(buf, erased_val, BOOT_MAX_ALIGN);
-    buf[0] = BOOT_FLAG_SET;
+    buf[0] = val;
 
     rc = flash_area_write(fap, off, buf, align);
     if (rc != 0) {
@@ -445,13 +469,39 @@ boot_write_flag(int flag, const struct flash_area *fap)
 int
 boot_write_copy_done(const struct flash_area *fap)
 {
-    return boot_write_flag(BOOT_FLAG_COPY_DONE, fap);
+    uint32_t off;
+
+    off = boot_copy_done_off(fap);
+    BOOT_LOG_DBG("writing copy_done; fa_id=%d off=0x%x (0x%x)",
+                 fap->fa_id, off, fap->fa_off + off);
+    return boot_write_trailer_byte(fap, off, BOOT_FLAG_SET);
 }
 
 int
 boot_write_image_ok(const struct flash_area *fap)
 {
-    return boot_write_flag(BOOT_FLAG_IMAGE_OK, fap);
+    uint32_t off;
+
+    off = boot_image_ok_off(fap);
+    BOOT_LOG_DBG("writing image_ok; fa_id=%d off=0x%x (0x%x)",
+                 fap->fa_id, off, fap->fa_off + off);
+    return boot_write_trailer_byte(fap, off, BOOT_FLAG_SET);
+}
+
+/**
+ * Writes the specified value to the `swap-type` field of an image trailer.
+ * This value is persisted so that the boot loader knows what swap operation to
+ * resume in case of an unexpected reset.
+ */
+int
+boot_write_swap_type(const struct flash_area *fap, uint8_t swap_type)
+{
+    uint32_t off;
+
+    off = boot_swap_type_off(fap);
+    BOOT_LOG_DBG("writing swap_type; fa_id=%d off=0x%x (0x%x), swap_type=0x%x",
+                 fap->fa_id, off, fap->fa_off + off, swap_type);
+    return boot_write_trailer_byte(fap, off, swap_type);
 }
 
 int
@@ -524,10 +574,10 @@ boot_swap_type(void)
     for (i = 0; i < BOOT_SWAP_TABLES_COUNT; i++) {
         table = boot_swap_tables + i;
 
-        if ((table->magic_primary_slot == BOOT_MAGIC_ANY     ||
-                table->magic_primary_slot == primary_slot.magic) &&
-            (table->magic_secondary_slot == BOOT_MAGIC_ANY   ||
-                table->magic_secondary_slot == secondary_slot.magic) &&
+        if (boot_magic_compatible_check(table->magic_primary_slot,
+                                        primary_slot.magic) &&
+            boot_magic_compatible_check(table->magic_secondary_slot,
+                                        secondary_slot.magic) &&
             (table->image_ok_primary_slot == BOOT_FLAG_ANY   ||
                 table->image_ok_primary_slot == primary_slot.image_ok) &&
             (table->image_ok_secondary_slot == BOOT_FLAG_ANY ||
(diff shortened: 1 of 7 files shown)


MariuszSkamra:
I confirm this fix. I've tested this PR with my app from https://github.com/JuulLabs-OSS/mcuboot/issues/480.

ccollins476ad:
Note: I force pushed a commit with documentation and comment changes.  I did not change any code with this force push. (I messed up the trailer structure representation in my first commit).

utzig:
@ccollins476ad Please cherry-pick the commit from PR #482

ccollins476ad:
Thanks for the reminder, @utzig.  I have pulled in your commits from #482.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -119,6 +119,32 @@ boot_flag_decode(uint8_t flag)
...
+ *
+ * @param tbl_val               A magic field from a status source table.
+ * @param val                   The magic value in a trailer, encoded as a
+ *                                  BOOT_MAGIC_[...].
+ *
+ * @return                      1 if the two values are compatible;
+ *                              0 otherwise.
+ */
+int
+boot_magic_eq(uint8_t tbl_val, uint8_t val)
```
nvlsianpu: Function name is not descriptive. Maybe `boot_magic_compatible _check`

utzig:
@ccollins476ad From your commit message:

> Prior to this change, the scratch image trailer had a different format from a slot image trailer.  Specifically, the `copy_done` field was excluded from the scratch trailer.

More importantly than `copy_done` was that in the scratch area we only ever store a single sector's state, so for it's usage calculation there was: `BOOT_STATUS_STATE_COUNT * min_write_sz`, while for slotX it was: `BOOT_STATUS_MAX_ENTRIES * BOOT_STATUS_STATE_COUNT * min_write_sz`. Now `BOOT_STATUS_MAX_ENTRIES` is user configurable and might end up being a big value; not to say there's anything wrong with the change, but I thought this correction was requried! :-P

ccollins476ad:
@utzig, that is a good point, and something that I completely missed.  At the very least, I will update the commit message.

Do you think this could cause a problem?  I think the scratch area is guaranteed to be large enough to accommodate this change, but I could use a second opinion!

boot/bootutil/src/bootutil_misc.c
ccollins476ad: Thanks, I have renamed it as suggested.  To be honest, I did start out with a longer and more descriptive name, but I had a feeling some of these identifiers were getting too long (the switch from 0/1 to primary/secondary definitely contributes to this!).

utzig:
@ccollins476ad I don't think it will cause issues; so I was thinking about this and it seems to me that if you cannot fit the data of the last sector with metadata on scratch, than it won't fit on the last sector as well, so in the end this seems like a good move into simplifying the code. Also metadata on scratch is only ever used it there's data to be swapped in any slot's last sectors, thus it's a bit of a corner case that would be better off having a specific test added on the simulator. Your changes LGTM, I just need to finish reviewing one commit (the big one!).

boot/bootutil/src/bootutil_misc.c
```diff
@@ -616,6 +642,15 @@ boot_set_pending(int permanent)
             rc = boot_write_image_ok(fap);
         }
 
+        if (rc == 0) {
```
nvlsianpu: In BOOT_MAGIC_GOOD case swap_type should be checked and checked in case operation was interrupted here.

utzig: I don't think anyone would be able to schedule a new swap operation if we were resuming an interruption; what exact issue do you have mind?

utzig:
@ccollins476ad Please fixup the last commit.

ccollins476ad:
Thanks, @utzig.  I have fixed up that commit, and reworded the "scratch trailer structure" commit as discussed above.

boot/bootutil/src/bootutil_misc.c
ccollins476ad: I also don't think we need to worry about the resume case here.  `boot_set_pending()` doesn't get called from the boot loader, it gets called from the application.  If the application is running, there can't be a swap in progress.

This is a little weird, though.  If the application tries to schedule a swap operation, and one is already scheduled, we just return 0.  It seems like we ought to check if the two swap operations have the same type.  If they are different, we should return an error.  I think this should go in a separate PR, though.

nvlsianpu: What I meant was a power-down had occurred just before writing the swap_type. Shouldn't an application be able to recognize such a undone request unambiguously, and be able to renew boot_set_pending() after that? For sure it is possible to check that the application had already collected proper upgrade image.

ccollins476ad: @nvlsianpu I think I see what you mean.

Application schedules a "test" swap.  This requires two flash writes:
1. Write `magic` to secondary trailer.
2. Write `swap_type` to secondary trailer.

If the device reboots after step 1 (before step 2), the trailer will be incomplete.

However, I don't think this is a problem.  Since this is a new swap (not a resume), mcuboot won't even look at the `swap_type` field at startup.  It should execute the "test" swap as expected.

If mcuboot then reboots in the middle of this swap, then we have a potential issue.  Mcuboot will recognize that a swap was interrupted and needs to be resumed, but it won't be able to determine the swap type.  I think this just puts us back where we were before this bug fix ("revert" will fail; all other swap types happen to work by chance).  Since you cannot specify a "revert" swap type with the `boot_set_pending()` function, I don't think there is a problem here.

I do think there is a general issue of the device resetting with a partially-written trailer.  This issue has always existed, though, so it is probably best left for a separate PR.

Please let me know if I am still missing something.

Issue #486, opened 2019-05-27
imgtool supports verifying signed firmware binaries


robbawebba:
I recently asked on the mcuboot slack channel if it's possible to verify if a firmware binary is signed properly, whether it be with imgtool or even openssl. It turns out this feature is not implemented but might be useful in the future.

The main benefit of this feature that I could think of, besides checking if the binary is signed, would be to determine which private key was used to sign a given firmware image. In development I use the default key provided by this repository, but I use a separate key for deployment in production. I've run into issues in the past where there was a mismatch in signing keys, so this feature would be very helpful for debugging this issue if it occurs, as well as validating a signed image before deploying it.

I imagine this feature would be another command built into imgtool.py. The user would have to provide a path to the signed firmware image and the path to the private key for verification. I'm not very familiar with the way images are signed, so I would appreciate anyone's input on this feature.

utzig:
Support for verification was added with #487. Still lacks support for encrypted images but that should arrive eventually or another issue can be created for it.

robbawebba:
@utzig Thank you for implementing this feature! My team and I really appreciate it :100:

PR #487, opened 2019-05-27
Add new verify command

utzig:
`imgtool verify -k <some-key.(pub|sec)> `

Allow imgtool to validate that an image has a valid sha256sum and that it was signed by the supplied key.

This implements #486

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -19,6 +19,7 @@ Image signing and management.
 """
 
 from . import version as versmod
+from enum import Enum
 from intelhex import IntelHex
 import hashlib
 import struct
@@ -27,6 +28,7 @@ from cryptography.hazmat.primitives.asymmetric import padding
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
+from cryptography.exceptions import InvalidSignature
 
 IMAGE_MAGIC = 0x96f3b83d
 IMAGE_HEADER_SIZE = 32
@@ -55,6 +57,7 @@ TLV_VALUES = {
         'DEPENDENCY': 0x40
 }
 
+TLV_SIZE = 4
 TLV_INFO_SIZE = 4
 TLV_INFO_MAGIC = 0x6907
 
@@ -68,6 +71,13 @@ STRUCT_ENDIAN_DICT = {
         'little': '<',
         'big':    '>'
 }
+
+VerifyResult = Enum('VerifyResult',
+                    """
+                    OK INVALID_MAGIC INVALID_TLV_INFO_MAGIC INVALID_HASH
+                    INVALID_SIGNATURE
+                    """)
+
 
 class TLV():
     def __init__(self, endian):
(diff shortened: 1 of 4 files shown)


scripts/imgtool/image.py
```diff
@@ -307,3 +309,48 @@ def pad_to(self, size):
...
+                tlv_sig = b[off:off+tlv_len]
+                payload = b[:header_size+img_size]
+                try:
+                    key.verify(tlv_sig, payload)
+                    print("signature validated!")
+                    return
+                except InvalidSignature:
+                    raise Exception("Invalid signature for image")
+            tlv_off += TLV_SIZE + tlv_len
+        raise Exception("Signature not found")
```
d3zd3z: I wonder if we should be pickier about what is in the TLVs? Also, does this work if there are multiple signatures (say with different signers)? It could check that at least one key is valid instead of failing on the first key.

utzig: > It could check that at least one key is valid instead of failing on the first key.

I'll fix this.

utzig: Updated to not fail on the first key that didn't match and just printing the exception on the main call, instead of throwing the exception on the user. Also updated commit message to mention that encrypted messages are not yet supported.

scripts/imgtool/image.py
```diff
@@ -307,3 +309,49 @@ def pad_to(self, size):
...
         self.payload += pbytes
+
+    @staticmethod
+    def verify(imgfile, key):
+        with open(imgfile, "rb") as f:
+            b = f.read()
+
+        magic, _, header_size, _, img_size = struct.unpack('IIHHI', b[:16])
+        if magic != IMAGE_MAGIC:
+            raise Exception("Invalid image magic; is this a MCUBoot image?")
```
mbolivar: s/a MCUBoot/an MCUboot/ (change "a" to "an" and "MCUBoot" to "MCUboot")

scripts/imgtool/image.py
```diff
@@ -307,3 +309,49 @@ def pad_to(self, size):
...
+
+        tlv_info = b[header_size+img_size:header_size+img_size+TLV_INFO_SIZE]
+        magic, tlv_tot = struct.unpack('HH', tlv_info)
+        if magic != TLV_INFO_MAGIC:
+            raise Exception("Invalid TLV magic; is this a MCUBoot image?")
```
mbolivar: s/a MCUBoot/an MCUboot/

scripts/imgtool/image.py
```diff
@@ -307,3 +309,49 @@ def pad_to(self, size):
...
+        tlv_off = header_size + img_size
+        tlv_end = tlv_off + tlv_tot
+        tlv_off += TLV_INFO_SIZE  # skip tlv info
+        while tlv_off < tlv_end:
+            tlv = b[tlv_off:tlv_off+TLV_SIZE]
+            tlv_type, _, tlv_len = struct.unpack('BBH', tlv)
+            if tlv_type == TLV_VALUES["SHA256"]:
+                off = tlv_off + TLV_SIZE
+                if digest == b[off:off+tlv_len]:
+                    print("sha256sum matched!")
```
mbolivar: Did you mean to leave this in? It looks like it might be debug code.

scripts/imgtool/image.py
```diff
@@ -307,3 +309,49 @@ def pad_to(self, size):
...
+                        return
+                else:
+                    raise Exception("Invalid hash for image")
+            elif key is not None and tlv_type == TLV_VALUES[key.sig_tlv()]:
+                off = tlv_off + TLV_SIZE
+                tlv_sig = b[off:off+tlv_len]
+                payload = b[:header_size+img_size]
+                try:
+                    key.verify(tlv_sig, payload)
+                    print("signature validated!")
```
mbolivar: Same here, did you mean this?

utzig:
@mbolivar Thanks, I updated the code with the suggested separation of responsibilities between the "library" and the calling app.

PR #491, opened 2019-06-14
Update mbedTLS sha256 usage to avoid deprecation

utzig:
mbedTLS made sha256 functions that do not return errors deprecated. This updates to use the new functions avoiding the extra functions calls, and breakage when the deprecated calls are effectively removed.

### DIFF
--- a/boot/bootutil/include/bootutil/sha256.h
+++ b/boot/bootutil/include/bootutil/sha256.h
@@ -61,20 +61,20 @@ typedef mbedtls_sha256_context bootutil_sha256_context;
 static inline void bootutil_sha256_init(bootutil_sha256_context *ctx)
 {
     mbedtls_sha256_init(ctx);
-    mbedtls_sha256_starts(ctx, 0);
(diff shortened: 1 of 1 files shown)


boot/bootutil/include/bootutil/sha256.h
```diff
@@ -61,20 +61,20 @@ typedef mbedtls_sha256_context bootutil_sha256_context;
 static inline void bootutil_sha256_init(bootutil_sha256_context *ctx)
 {
     mbedtls_sha256_init(ctx);
-    mbedtls_sha256_starts(ctx, 0);
+    (void)mbedtls_sha256_starts_ret(ctx, 0);
```
d3zd3z: I think this is fine for this commit, but I'm guessing these are made to return errors for some reason. Perhaps if there is a failure with a hardware crypto module. I wonder if we should consider handling the errors (not that there is much we can do).

utzig: I did not want to apply changes up the stack; there are not really many errors that could happen for us; also tinycrypt does not return errors.

PR #492, opened 2019-06-18
requirements.txt: bump cryptography Py package version

nvlsianpu:
Since Ed25519 curve primitives are available since cryptography package
version 2.6 need to bump its version.


nvlsianpu:
cc @hakonfam

PR #500, opened 2019-07-04
Fix bug that prevents split images from working.

alvarop:
There is a bug in split_go that never opens the loader image flash
area, but opens the app/split image flash area twice. This prevents
split_image_check from passing and the app from ever loading.

This should take care of https://github.com/JuulLabs-OSS/mcuboot/issues/499


utzig:
@alvarop Looks good, but you need to sign-off you commit!

alvarop:
Added sign-off :D

PR #512, opened 2019-07-29
image: use explitic padding on dependency version

d3zd3z:
This struct currently relies on padding being done by the C compiler.
Although this is generally done the same across targets, this isn't
required, and it is better to make any padding explicit.  This also
makes it clear that there is extra space in this structure that can be
used for another purpose.

The python imgtool is also fixed to match.  It was using a 32-bit int,
which only worked in the little endian case.


d3zd3z:
@davidvincze, @tamasban, is it difficult for you to test this? I don't believe it should change any behavior.

davidvincze:
Thank you for the fix. I've tested it. LGTM

PR #516, opened 2019-07-31
[RFC] Multi image refactoring

utzig:
This currently brings down an optimized build for Mynewt by 350 bytes, but still 200 bytes to get to 1.3!

Not ready for review just yet.


utzig:
@d3zd3z With https://github.com/JuulLabs-OSS/mcuboot/pull/516/commits/5d55be18e0e33a8114281a2690a73e9e75c215df I removed the global locking on the simulator, if you want to do some double checking. Right now all of `bootutil`, but encrypted images, is already thread safe on the simulator.

utzig:
@d3zd3z @nvlsianpu I'm adding RFC; some feedback might be interesting...

utzig:
I moved the Mynewt specific watchdog change to #520 and squashed in a few changes to fix build errors on Mynewt. I removed `[WIP]` because it will be easier to add other changes into new PRs...

boot/bootutil/src/bootutil_misc.c
```diff
@@ -170,14 +171,20 @@ boot_magic_off(const struct flash_area *fap)
 }
 
 int
-boot_status_entries(const struct flash_area *fap)
+boot_status_entries(int image_index, const struct flash_area *fap)
 {
+#if (BOOT_IMAGE_NUMBER == 1)
+    (void)image_index;
```
d3zd3z: Would it be useful to put the `(void)image_index` into `FLASH_AREA_IMAGE_PRIMARY` and `SECONDARY` when there is only a single image. After all, it is the macro that isn't using its argument. This would avoid having to add this block to a bunch of functions.

d3zd3z: Although, this could fairly easily be fixed in another commit. It might be nice to figure if there are other of these "unused" blocks that can be eliminated this way, and to figure the best way to "use" a macro argument.

boot/bootutil/src/loader.c
```diff
@@ -659,31 +750,46 @@ boot_write_status(struct boot_status *bs)
  * Validate image hash/signature in a slot.
  */
 static int
-boot_image_check(struct image_header *hdr, const struct flash_area *fap,
-        struct boot_status *bs)
+boot_image_check(struct boot_loader_state *state, struct image_header *hdr,
+                 const struct flash_area *fap, struct boot_status *bs)
 {
-    static uint8_t tmpbuf[BOOT_TMPBUF_SZ];
+#if !defined(__BOOTSIM__)
```
d3zd3z: Perhaps have a single ifdef on `__BOOTSIM__` and define something like `SIMSTATIC` (or maybe `TARGET_STATIC`) as either `static` or empty, and then just use that symbol in the places where there is an ifdef, especially since there are a bunch of these.

boot/bootutil/src/bootutil_misc.c
utzig: Nice one, I changed the definition to:

```
#define FLASH_AREA_IMAGE_PRIMARY(x)    (((x) == 0) ?                \           
                                         DT_FLASH_AREA_IMAGE_0_ID : \           
                                         DT_FLASH_AREA_IMAGE_0_ID)
```
And removed all `(void)image_index`...

boot/bootutil/src/loader.c
utzig: I have gone with `TARGET_STATIC`

boot/bootutil/src/bootutil_misc.c
d3zd3z: As long as we don't build with too stupid of a compiler, this should be fine.

sim/mcuboot-sys/src/api.rs
```diff
@@ -7,97 +7,202 @@ use log::{Level, log_enabled, warn};
...
+    let ctx = map.get_mut(&thread_id).unwrap();
+    ctx.flash_map.remove(&dev_id);
 }
 
 // This isn't meant to call directly, but by a wrapper.
 
+#[no_mangle]
+pub extern fn sim_get_flash_areas() -> *const CAreaDesc {
+    let map = THREAD_CTX.lock().unwrap();
+    let thread_id = thread::current().id();
```
d3zd3z: Rust std has thread local variables (https://doc.rust-lang.org/std/macro.thread_local.html), I wonder if this would be cleaner to just use those?

d3zd3z: I'm also curious if making these thread local values will remove the need to derive Send for the wrapper types.

sim/mcuboot-sys/src/api.rs
```diff
@@ -7,97 +7,202 @@ use log::{Level, log_enabled, warn};
...
     };
 }
 
 // Set the flash device to be used by the simulation.  The pointer is unsafely stashed away.
 pub unsafe fn set_flash(dev_id: u8, dev: &mut dyn Flash) {
-    let mut flash_params = FLASH_PARAMS.lock().unwrap();
-    flash_params.insert(dev_id, FlashParams {
+    let mut map = THREAD_CTX.lock().unwrap();
+    let thread_id = thread::current().id();
+    if !map.contains_key(&thread_id) {
```
d3zd3z: (aside from this not being needed with thread local variables), using map.entry(...) should make this a bit cleaner.

sim/mcuboot-sys/src/api.rs
utzig: I updated it to use `thread_local` there's no need to derive `Send` anymore.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
 use log::{Level, log_enabled, warn};
 use std::{
     collections::HashMap,
     mem,
-    ops::Deref,
     slice,
-    sync::Mutex,
 };
+use crate::area::CAreaDesc;
+use std::cell::RefCell;
```
d3zd3z: These should be in the `use std::{ ...` block above.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
 use libc;
 use log::{Level, log_enabled, warn};
 use std::{
     collections::HashMap,
     mem,
-    ops::Deref,
     slice,
-    sync::Mutex,
 };
+use crate::area::CAreaDesc;
```
d3zd3z: Although the simflash one seems to be out of order, lets at least try to keep these in alphabetical order.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
-    flash.remove(&dev_id);
+    THREAD_CTX.with(|ctx| {
+        ctx.borrow_mut().flash_map.remove(&dev_id);
+    });
 }
 
 // This isn't meant to call directly, but by a wrapper.
 
+#[no_mangle]
+pub extern fn sim_get_flash_areas() -> *const CAreaDesc {
```
d3zd3z: You should just be able to have
```rust
pub extern fn sim_get_flash_areas() -> *const CAreaDesc {
    THREAD_CTX.with(|ctx| {
        ctx.borrow().flash_areas.ptr
    })
}
```

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
+
+#[no_mangle]
+pub extern fn sim_reset_flash_areas() {
+    THREAD_CTX.with(|ctx| {
+        ctx.borrow_mut().flash_areas.ptr = ptr::null();
+    });
+}
+
+#[no_mangle]
+pub extern fn sim_get_context() -> *const CSimContext {
```
d3zd3z: Likewise, similar to `sim_get_flash_areas` above.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
+    });
+    rc
 }
 
 #[no_mangle]
 pub extern fn sim_flash_align(id: u8) -> u8 {
-    let flash_params = FLASH_PARAMS.lock().unwrap();
-    let params = flash_params.get(&id).unwrap();
-    params.align
+    let mut align: u8 = 0;
```
d3zd3z: Again here, remove the `align` variable, the assign part, and the semicolon in the `THREAD_CTX.with` block, and it should just return the value.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
...
+    });
+    align
 }
 
 #[no_mangle]
 pub extern fn sim_flash_erased_val(id: u8) -> u8 {
-    let flash_params = FLASH_PARAMS.lock().unwrap();
-    let params = flash_params.get(&id).unwrap();
-    params.erased_val
+    let mut erased_val: u8 = 0;
```
d3zd3z: Here, too.

sim/mcuboot-sys/src/api.rs
```diff
@@ -1,104 +1,198 @@
 //! HAL api for MyNewt applications
 
 use simflash::{Result, Flash, FlashPtr};
-use lazy_static::lazy_static;
```
d3zd3z: This does remove all uses of lazy_static, so it is good to remove that from the Cargo.toml file as well.

sim/mcuboot-sys/src/api.rs
utzig: I reordered all of them and added `std` stuff to the block.

sim/mcuboot-sys/src/api.rs
utzig: done

sim/mcuboot-sys/src/api.rs
utzig: done

sim/mcuboot-sys/src/api.rs
utzig: done

sim/mcuboot-sys/src/api.rs
utzig: done

sim/mcuboot-sys/src/api.rs
utzig: done

nvlsianpu:
Great patch. For zephyr dual bank minimal size decreases:
- overwrite only mode: 13820 B -> 13500 B, it is less than was before multi-mage patch (13568 B)
- swap mode: 15796 B -> 13376 B, Just a few dozen more than before multi-mage patch  (13252 B)

Reference was results described here: https://github.com/JuulLabs-OSS/mcuboot/pull/369#issuecomment-507695141

PR #517, opened 2019-08-05
Sim tests for dependencies

d3zd3z:
This adds some changes that support testing dependencies within the simulator. Currently, there is a simple test case for incorrect dependencies, but it is disabled because the mcuboot code hangs in this configuration.

### DIFF
--- a/sim/src/depends.rs
+++ b/sim/src/depends.rs
@@ -1,0 +1,133 @@
+//! Support and tests related to the dependency management for multi-image
+//! support.
+
+use crate::image::ImageVersion;
+
+pub trait Depender {
+    /// Generate a version for this particular image.  The slot indicates
+    /// which slot this is being put in.
+    fn my_version(&self, offset: usize, slot: usize) -> ImageVersion;
+
+    /// Return dependencies for this image/slot combination.
+    fn my_deps(&self, offset: usize, slot: usize) -> Vec<ImageVersion>;
+
+    /// Return the image ID of the other version.
+    fn other_id(&self) -> u8;
+}
+
+/// A boring image is used when we aren't testing dependencies.  There will
+/// be meaningful version numbers.  The size field is the image number we
+/// are.
+pub struct BoringDep(pub usize);
+
+impl Depender for BoringDep {
+    fn my_version(&self, _offset: usize, slot: usize) -> ImageVersion {
+        ImageVersion::new_synthetic(self.0 as u8, slot as u8, 0)
+    }
+
+    fn my_deps(&self, _offset: usize, _slot: usize) -> Vec<ImageVersion> {
+        vec![]
+    }
+
+    fn other_id(&self) -> u8 {
+        0
+    }
+}
+
+/// An individual test of the dependency mechanism describes one of the
+/// possibilities for the dependency information for each image, and what
+/// the expected outcome is.
+#[derive(Clone, Debug)]
+pub struct DepTest {
+    /// What kinds of dependency should be installed in the image.
+    pub depends: [DepType; 2],
+
+    /// What is the expected outcome of the upgrade.
+    pub upgrades: [UpgradeInfo; 2],
+}
+
+/// Describes the various types of dependency information that can be
+/// provided.
+#[derive(Clone, Debug)]
+pub enum DepType {
+    /// Do not include dependency information
+    Nothing,
+    /// Provide dependency information that matches the other image.
+    Correct,
+    /// Provide dependency information describing something newer than the
+    /// other image.
+    Newer,
+}
+
+/// Describes what our expectation is for an upgrade.
+#[derive(Clone, Debug)]
+pub enum UpgradeInfo {
+    /// The current version should be held.
+    Held,
+    /// The image should be upgraded
+    Upgraded,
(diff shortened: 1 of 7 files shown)


d3zd3z:
I removed the DNM. There is no reason to not merge this. It is the code under test that is broken, not the test.

sim/src/tlv.rs
```diff
@@ -205,8 +263,39 @@ impl ManifestGen for TlvGen {
...
+        // efficiency view.  If this is shown to be a performance issue
+        // with the tests, the protected data could be appended to the
+        // payload, and then removed after the signature is done.  For now,
+        // just make a signed payload.
+        let mut sig_payload = self.payload.clone();
+        if self.protect_size > 0 {
+            assert_eq!(self.protect_size as usize, result.len());
+            sig_payload.extend_from_slice(&result);
+        }
+        let sig_payload = sig_payload;
```
utzig: Is this done to remove mutability?

sim/src/tlv.rs
```diff
@@ -289,7 +291,7 @@ impl ManifestGen for TlvGen {
         // just make a signed payload.
         let mut sig_payload = self.payload.clone();
         if self.protect_size > 0 {
-            assert_eq!(self.protect_size as usize, result.len());
+            assert_eq!(self.protect_size as usize + 4, result.len());
```
utzig: Would have been nice squashing this change into the previous commit...

sim/src/depends.rs
```diff
@@ -0,0 +1,140 @@
+//! Support and tests related to the dependency management for multi-image
+//! support.
+
+use crate::image::ImageVersion;
+
+pub trait Depender {
+    /// Generate a version for this particular image.  The slot indiates
```
utzig: type: "indicates"

sim/src/depends.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// A boring image is used when we aren't testing dependencies.  There will
+/// be meaningful version numbers.  The size field is the image number we
+/// are.
+pub struct BoringDep(pub usize);
+
+impl Depender for BoringDep {
+    fn my_version(&self, offset: usize, slot: usize) -> ImageVersion {
+        ImageVersion {
+            major: self.0 as u8 * 20 + slot as u8,
```
utzig: I am not sure why you are multiplying by `20` (instead of any other value) here and in many other places below, but maybe this could be a constant?

sim/src/image.rs
```diff
@@ -139,11 +146,17 @@ impl ImagesBuilder {
...
-            let upgrades = install_image(&mut flash, &slots[1], 46928, false);
+        let num_images = self.slots.len();
+        let images = self.slots.into_iter().enumerate().map(|(image_num, slots)| {
+            let dep: Box<dyn Depender> = if num_images > 1 {
+                Box::new(PairDep::new(num_images, image_num, deps))
+            } else {
+                Box::new(BoringDep(image_num))
+            };
+            let primaries = install_image(&mut flash, &slots[0], 42784, &*dep, false);
+            let upgrades = install_image(&mut flash, &slots[1], 46928, &*dep, false);
```
utzig: TODO: maybe in the future it would be interesting to update this to install different images to avoid having hash collisions on the multiple images.

sim/src/image.rs
```diff
@@ -139,11 +146,17 @@ impl ImagesBuilder {
     }
 
     /// Construct an `Images` that doesn't expect an upgrade to happen.
-    pub fn make_no_upgrade_image(self) -> Images {
+    pub fn make_no_upgrade_image(self, deps: &DepTest) -> Images {
         let mut flash = self.flash;
-        let images = self.slots.into_iter().map(|slots| {
-            let primaries = install_image(&mut flash, &slots[0], 42784, false);
-            let upgrades = install_image(&mut flash, &slots[1], 46928, false);
+        let num_images = self.slots.len();
```
utzig: `let num_images = self.num_images();` ?

sim/src/tlv.rs
d3zd3z: Yes.

sim/src/depends.rs
d3zd3z: I just wanted the version numbers to be distinct and clear between the two images. I'm not sure what I'd even call the constant, but I can name it something.

d3zd3z: Even better, I'll make a special constructor for ImageVersion that makes the arguments clearer as well.

sim/src/image.rs
d3zd3z: They should be different. The image number is used in the header, as well as in the textual information at the start of the image. Maybe change the seed to splat? But, the hashes should definitely be different.

sim/src/image.rs
d3zd3z: Good catch. I added the `num_images` method after writing this, and didn't notice it. One tricky thing is that the call has to be moved above the `let mut flash = self.flash`, since that moves a value out of `self` making a method call invalid.

d3zd3z:
This should pick all of the nits, except for the TODO, which is commented on above.

PR #525, opened 2019-08-09
zephyr: add minimal footprint configuration for nrf52

nvlsianpu:
Added Kconfig project configuration whit shows the
minimal flash footprint on nRF2832 SoC targets.


boot/zephyr/boards/nrf52_minimal_footprint.conf
```diff
@@ -0,0 +1,29 @@
+# Minima MCUBoot flash footprint configuration
```
utzig: typo: Minimal

boot/zephyr/boards/nrf52_minimal_footprint.conf
```diff
@@ -0,0 +1,29 @@
...
+# for nRF52832 SoC targets
+
+# Generated by Kconfiglib (https://github.com/ulfalizer/Kconfiglib)
+CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=y
+CONFIG_BOOT_SIGNATURE_KEY_FILE="root-ec-p256.pem"
+# CONFIG_BOOT_VALIDATE_SLOT0 is not set
```
utzig: Although this might not be minimal I would strongly encourage people to have it set! :-P

nvlsianpu:
This is configuration I used as base for different revision comparison recently.

boot/zephyr/boards/nrf52_minimal_footprint.conf
nvlsianpu: It is the minimal, not the recommend one. Will consider addition of a comment on implication.

nvlsianpu: added comment.

boot/zephyr/boards/nrf52_minimal_footprint.conf
```diff
@@ -0,0 +1,39 @@
+# Minimal MCUBoot flash footprint configuration
+# for nRF52832 SoC targets
+# This is not recomendet configuration because of security and reliability
+# reasons.
+
```
carlescufi: You are missing quite a few from:
https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/basic/minimal/common.conf
and
https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/basic/minimal/no-timers.conf

nvlsianpu: #CONFIG_SYS_CLOCK_EXISTS=y is required for reboot routine.

I added the rest form minimal - it has no impact on footprint as configuration was by default the same or an API was no used.

carlescufi: > #CONFIG_SYS_CLOCK_EXISTS=y is required for reboot routine.

Really? That's a shame, are you sure we cannot reboot without a timer? Seems counter intuitive to me.

> I added the rest form minimal - it has no impact on footprint as configuration was by default the same or an API was no used.

It has no impact on the footprint? That's strange, there were a bunch there that I know reduced the code size.

nvlsianpu: looks like only zephyr API which requires CONFIG_SYS_CLOCK_EXISTS is `sys_clock_disable()` which is called before jump to the application. Can do this optional - but probably the call is for something here,

nvlsianpu: https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/zephyr/main.c#L77

carlescufi: Can you then change that code to:

```
#ifdef(CONFIG_SYS_CLOCK_EXISTS)
sys_clock_disable();
#endif
```

nvlsianpu: Yes - actually I was about asking whether this is good idea.

PR #527, opened 2019-08-13
[RFC] Add AES T-Tables implementation for Tinycrypt

utzig:
This adds an optional implementation of t-tables for AES in Tinycrypt, which can be enabled to speed up the execution of encryption/decryption.

The t-table implementaion is based on the MbedTLS implementation, using single forward and reverse tables; no table generation code was added, so this uses extra 2KB of flash (256 * sizeof(uint32_t) * 2).

PS: I have not tested in HW yet; this might need more clean ups; have to check how to contribute this upstream, since the project seems dead.

### DIFF
--- a/ext/tinycrypt/lib/source/aes_decrypt.c
+++ b/ext/tinycrypt/lib/source/aes_decrypt.c
@@ -59,10 +59,151 @@ static const uint8_t inv_sbox[256] = {
 	0x55, 0x21, 0x0c, 0x7d
 };
 
+#ifdef TC_USE_AES_TTABLES
+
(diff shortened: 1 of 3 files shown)


utzig:
Btw, I did a bit of testing of MbedTLS AES options to gather some numbers; the results can be seen here: https://github.com/apache/mynewt-core/pull/1958

Adding the table generation would be interesting, because a single set of tables (sbox, inverse sbox, forward and reverse t-tables) that are RAM generated brings the best trade off in terms of speed, flash and RAM usage.

PR #528, opened 2019-08-13
sim: Add debugging ptable to image dumps

d3zd3z:
Add a `debug_dump()` method to `Images` to allow the images to be
written to a file.  The dependency test will call this if the
environment variable MCUBOOT_DEBUG_DUMP is set.

In order to make these debug dumps more useful, add a simple partition
table to the beginning of the image (where MCUboot would reside on
target).  This has a simple header, and then entries for each partition,
using the partition ids used within the simulator.  This allows the
image to be more easily used by external tools.

As an example, `scripts/mcubin.bt` is a binary template for the [010
Editor](https://www.sweetscape.com/010editor/), allowing it to decode
and show the details of images from MCUboot.

### DIFF
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -1,9 +1,19 @@
-use log::{info, warn, error};
+use byteorder::{
+    LittleEndian, WriteBytesExt,
+};
+use log::{
+    Level::Info,
+    error,
+    info,
+    log_enabled,
+    warn,
+};
 use rand::{
     distributions::{IndependentSample, Range},
     Rng, SeedableRng, XorShiftRng,
 };
 use std::{
+    collections::HashSet,
     io::{Cursor, Write},
     mem,
     slice,
@@ -163,6 +173,7 @@ impl ImagesBuilder {
                 primaries: primaries,
                 upgrades: upgrades,
             }}).collect();
+        install_ptable(&mut flash, &self.areadesc);
         Images {
             flash: flash,
             areadesc: self.areadesc,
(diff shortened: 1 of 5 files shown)


sim/src/image.rs
```diff
@@ -1337,6 +1356,55 @@ fn verify_trailer(flash: &SimMultiFlash, slot: &SlotInfo,
...
+        // avoids messing up images already written.
+        if areadesc.iter_areas().any(|area| {
+            area.device_id == id &&
+                area.off == 0 &&
+                area.flash_id != FlashId::BootLoader
+        }) {
+            let special: Vec<FlashId> = areadesc.iter_areas()
+                .filter(|area| area.device_id == id && area.off == 0)
+                .map(|area| area.flash_id)
+                .collect();
```
utzig: Wouldn't it be possible to avoid iterating twice, by just collecting the `flash_ids` and checking if there is any?

utzig:
Would be nicer if there was a similar tool that was free!

sim/src/image.rs
d3zd3z: I could be, but I'm not sure which would be better. I probably should pub the `let special ...` in an if statement based on whether info logging is even enabled, since right now we compute the whole thing, and then immediately discard it in the normal case of there not being a message printed.

The first `any` iteration will stop as soon as it finds something matching the criteria. Even in the case of printing, it is probably faster to just do that instead of allocation a Vec to hold the result.

I'll push a version with the conditional.

d3zd3z:
I wish there was a free tool as well, but I'm not aware of anything. It's kind of just an example, and useful if anyone has that tool.

PR #532, opened 2019-08-20
zephyr: enable MULTITHREADING by default except on nRF

mbolivar:
The main point of this PR is to enable CONFIG_MULTITHREADING unless we know it's safe to do so. Currently, the only platform where we know for sure it is safe is on nRF SoCs.

Note this doesn't touch nrf52840_pca10059's final .config setting for this option -- defaults are overridden by .conf fragments.


nvlsianpu:
I changed the title to more descriptive as it helps to redact the release note.

PR #535, opened 2019-08-26
scripts: allow user to set load_addr

hakonfam:
Allow the user to set the load_addr field of the header.
This could be useful in multi image situations to help
deduce the image number of an update without having to
look at the swap info.

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -182,6 +182,8 @@ class BasedIntParamType(click.ParamType):
 
 @click.argument('outfile')
 @click.argument('infile')
+@click.option('-L', '--load-addr', type=BasedIntParamType(), required=False,
+              help='Load address for image when it is in its primary slot.')
 @click.option('-E', '--encrypt', metavar='filename',
               help='Encrypt image using the provided public key')
 @click.option('-e', '--endian', type=click.Choice(['little', 'big']),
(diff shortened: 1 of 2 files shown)


utzig:
If I remember correctly we discussed way back that this field was reserved to put the address (in RAM) where an image is supposed to be loaded when not XIP. I don't think it is a good idea to use it for some other purpose, just because at the moment nobody seems to be using this "feature". At least we should discuss and agree on it first...

hakonfam:
Thanks for the info. IMO there is nothing stopping non-XIP users from using it for that purpose while weird multi-image with one secondary slot and two primary slots-users use it for a different purpose :)

Laczen:
I think it is a good idea to extend the idea of load address to include XIP systems. For non-XIP it is used to determine where the image should be placed in RAM (without the header) but it also is the address to jump to. For XIP it could only have the meaning of the address to jump to (so pointing to the address just after the header). This would also allow the verification if a image is compiled for the correct slot before moving it to this slot.

d3zd3z:
> For XIP it could only have the meaning of the address to jump to (so pointing to the address just after the header).
Arm already defines a "header" at the beginning of the image that gives the address to jump to, as well as the initial stack pointer. I don't see a real compelling reason to do things differently. Even for non-XIP, it doesn't determine the address to jump to, that comes from the beginning of the image itself (which follows the header).

hakonfam:
This is true @d3zd3z. Although I don't think I will be able to extract the same information as what was found in `ih_load_addr`, I will be able to find the information we require. So feel free to close this PR if no one else wants this.

scripts/imgtool/main.py
```diff
@@ -205,16 +205,18 @@ def convert(self, value, param, ctx):
 @click.option('--align', type=click.Choice(['1', '2', '4', '8']),
               required=True)
 @click.option('-k', '--key', metavar='filename')
+@click.option('-L', '--load-addr', type=BasedIntParamType(), required=False,
```
utzig: Could you move this option up? `click` sorts them in reverse (bottom to top) and I would rather see the `required` ones first in the help page.

utzig:
Looks good apart from the comment... Btw, I don't like using some already defined/documented field for a different purpose, but since this just writes it to the image, users are free to do whatever they want with it! :-P

nvlsianpu:
@hakonfam looks good to me, pleas apply utzig suggestion.

PR #542, opened 2019-09-03
Fix533

d3zd3z:
Not intended to be merged, but a few additional patches on top of #533 for Fabio to pull in.


utzig:
Commits added to #533

PR #544, opened 2019-09-10
Fix encrypted images with dependencies

utzig:
This fixes an issue where dependencies were being decrypted before hashing when encrypted images are used. When using dependencies, some "protected" TLVs are added to the image. Those TLVs take part into the image hash calculation, but are not encrypted and must be sent plain to the hash update routine.


utzig:
This fixes tests that employ multi-image and encryption in the simulator, eg, `cargo test --features "sig-rsa enc-rsa multiimage`. No such tests are currently enabled to run on Travis-CI.

PR #545, opened 2019-09-10
Use better naming for swap copy/erase functions

### LINKED ISSUE #541: Swap status documentation issue
The documentation states that in the trailer the first 3 records are the swap status of the last sector, and the next 3 records are the status of last-1 sector. And so on. But I don't think that is true. The first 3 records are the status of the first `boot_swap_sectors` operation. But a single `boot_swap_sectors` operation can swap multiple sectors as long as the scratch partition is larger than 1 sector. `boot_copy_sector` should be named `boot_copy_sectors`, because it can copy multiple sectors. `boot_write_status` only writes status once following a call to `boot_copy_sector`.

so if the images you are swapping each 10 sectors large. and the scratch area is 2 sectors . Then 5 swap operations will take place (15 copy operations), and status will be written 3 times for each swap, or 15 total times. 
In the Swap status in the trailer,
The first 3 records will thus represent the swap status for sectors 8 and 9.
The next 3 records will represent swap status for sectors 7 and 6.
..
The last 3 records will represent swap status for sectors 1 and 0.

I am almost positive this is true. I am watching the memory get filled in in my debugger. Its possible I misunderstood the documentation. If that is the case then maybe it needs to be clarified.

Assuming I am correct,
I have yet to test whether this is an actual bug or just an error in documentation. I posted about a related issue that I think is actually a bug https://github.com/JuulLabs-OSS/mcuboot/issues/427 . has anyone actually tested that an interrupted swap can be recovered on a subsequent boot. I would hope so. That seems like core functionality of MCUboot. It is on my agenda for next week, as I slog through developing unit tests of my customized version of MCUboot.

utzig:
Since a copy/erase during a swap can result in some number of sectors erased and copied, which is not necessarily equal to 1, update design document and bootloader function naming to reflect this fact, by using the postfix "_region" instead of "_sector" in those cases.

This fixes #541

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2262,18 +2262,17 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
                  * magic, so also run validation on the primary slot to be
                  * sure it's not OK.
                  */
-                 if (boot_check_header_erased(state, BOOT_PRIMARY_SLOT) == 0 ||
-                     boot_validate_slot(state, BOOT_PRIMARY_SLOT, bs) != 0) {
-                     if (boot_img_hdr(state,
-                          BOOT_SECONDARY_SLOT)->ih_magic == IMAGE_MAGIC &&
-                         boot_validate_slot(state, BOOT_SECONDARY_SLOT, bs) == 0)
-                     {
+                if (boot_check_header_erased(state, BOOT_PRIMARY_SLOT) == 0 ||
+                        boot_validate_slot(state, BOOT_PRIMARY_SLOT, bs) != 0) {
+                    if (boot_img_hdr(state,
+                            BOOT_SECONDARY_SLOT)->ih_magic == IMAGE_MAGIC &&
+                            boot_validate_slot(state, BOOT_SECONDARY_SLOT, bs) == 0) {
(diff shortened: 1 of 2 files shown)


boot/bootutil/src/loader.c
```diff
@@ -2262,12 +2262,11 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
                  * magic, so also run validation on the primary slot to be
                  * sure it's not OK.
                  */
-                 if (boot_check_header_erased(state, BOOT_PRIMARY_SLOT) == 0 ||
-                     boot_validate_slot(state, BOOT_PRIMARY_SLOT, bs) != 0) {
-                     if (boot_img_hdr(state,
-                          BOOT_SECONDARY_SLOT)->ih_magic == IMAGE_MAGIC &&
-                         boot_validate_slot(state, BOOT_SECONDARY_SLOT, bs) == 0)
-                     {
+                if (boot_check_header_erased(state, BOOT_PRIMARY_SLOT) == 0 ||
```
utzig: Oops, this was just an indentation issue.

d3zd3z: It does look better, but should probably be a separate commit.

utzig: It was separated.

PR #547, opened 2019-09-11
bootutil_misc: fix printf types in log macros

mbolivar:
These were caught by GCC printf-like macro attribute warnings.

(The relevant types are all uint32_t, but %lx wants a long unsigned.)


ccollins476ad:
This may fix warnings on some architectures, but it will cause some on others, won't it?  I guess the correct format specifier to use for `uint32_t` is the `PRIu32` macro, but I have always had mixed results with this when cross compiling.

As ugly as it is, I think the best solution is to just cast the the argument to the appropriate type (`unsigned long` in this case).

utzig:
This was changed yesterday (04eec77c659a1a256f392b9d048c55ea3e78d826) because of similar issues on another platform. I agree with @ccollins476ad on the cast suggestion.

PR #551, opened 2019-09-11
New protected TLV format

### LINKED ISSUE #539: Protected area prevents appending TLV entries
Before adding the TLV protected area, it was possible to extend the TLV section (for example, to add additional signatures). This was a requested/required feature.

Now that the TLV protected area has been added, the TLV header, containing the length of the TLV is covered within image signatures. This means that the TLV length cannot be changed, without invalidating existing signatures (it would still be possible to append a new signature, but the old one would then become invalid).

One possible fix would be to skip the TLV header would computing the hash for the signature. This makes the checking code a bit more complex, however.

Related to #538.

image

utzig:
When using protected TLVs (for dependencies) the image validation was run over header + payload + tlv_info + protected TLVs. As a side-effect it was impossible to update the TLVs without rehashing (+ resigning) an image. This updates the validation routine to skip the tlv_info, so that hashing is run over header + payload + protected TLVs.

Fixes #539.

This is WIP, needs rebase after #549 is merged; `imgtool` and documentation changes in progress.

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -41,6 +41,7 @@ struct flash_area;
 #define IMAGE_MAGIC_V1              0x96f3b83c
 #define IMAGE_MAGIC_NONE            0xffffffff
 #define IMAGE_TLV_INFO_MAGIC        0x6907
+#define IMAGE_TLV_PROT_INFO_MAGIC   0x6908
 
 #define IMAGE_HEADER_SIZE           32
 
@@ -150,7 +151,7 @@ struct image_tlv_iter {
     const struct flash_area *fap;
     uint8_t type;
     bool prot;
-    uint16_t prot_len;
+    uint16_t prot_end;
     uint32_t tlv_off;
     uint32_t tlv_end;
 };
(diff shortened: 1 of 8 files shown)


utzig:
@tamasban @davidvincze This might interest you.

utzig:
Now everything was updated to work with the new protected TLVs; I have not tested the updated `imgtool` in HW yet but the hexdumps look OK!

utzig:
@d3zd3z @nvlsianpu @tamasban @davidvincze I tested a bunch of combinations of options on single image HW and now everything looks good. I'll try to do some testing of dependencies in multi image but if someone would be able to help, I'd be glad. Removing WIP.

d3zd3z:
The only suggestion I'd make is maybe being a little more consistent with the git subjects, e.g. starting the ones on the simulator with "sim: ", which I've been trying to be more consistent with.

utzig:
> The only suggestion I'd make is maybe being a little more consistent with the git subjects, e.g. starting the ones on the simulator with "sim: ", which I've been trying to be more consistent with.

Yes, that looks pretty good, I just updated the subject line of all commits (but design doc).

PR #552, opened 2019-09-12
[RFC] New protected TLV format proposal

utzig:
Updates the protected TLV specification to have their own TLV info struct, to ease hashing without having to skip parts of the image, while adding the ability to append new non-protected TLVs.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -72,6 +72,9 @@ struct image_header {
     uint32_t _pad1;
 };
 
+#define IMAGE_TLV_INFO_MAGIC        0x6907
+#define IMAGE_TLV_PROT_INFO_MAGIC   0x6908
+
 /** Image TLV header.  All fields in little endian. */
 struct image_tlv_info {
     uint16_t it_magic;
@@ -111,9 +114,10 @@ Optional type-length-value records (TLVs) containing image metadata are placed
 after the end of the image.
 
 The `ih_protect_tlv_size` field indicates the length of the protected TLV area.
-If dependency TLVs are present then the TLV info header and the dependency TLVs
-are also protected and have to be included in the hash calculation. Otherwise
-the hash is only calculated over the image header and the image itself. In this
+If protected TLVs are present then a TLV info header with magic equal to
+`IMAGE_TLV_PROT_INFO_MAGIC` must be present and the protected TLVs (plus the
+info header itself) have to be included in the hash calculation. Otherwise the
+hash is only calculated over the image header and the image itself. In this
 case the value of the `ih_protect_tlv_size` field is 0.
 
 The `ih_hdr_size` field indicates the length of the header, and therefore the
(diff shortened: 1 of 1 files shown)


d3zd3z:
So, my concern with not having the header describe the length of the protected header is that it then seems like it might be easier to just remove the protected header. Not specifying what the length of the trailer is makes us susceptible to a length-extension attack (well, maybe, but it makes me leery). A flag that indicates that a protected trailer must be present is a bit better. Having the length also makes it a little easier to find the unprotected trailer.

utzig:
> So, my concern with not having the header describe the length of the protected header is that it then seems like it might be easier to just remove the protected header.

Yes, that is a good point; it would still require the image creator to release a correctly signed version with the same header information, and no protected trailers, and then later release an update with the protected trailers added for it to be possible for someone to just remove them, no?

Anyway I think it's fine to leave the protected length in the header (and I don't need to update the proposal!).

docs/design.md
```diff
@@ -874,22 +881,15 @@ producing signed images, see: [signed_images](signed_images.md).
...
 
-If the TLV area contains dependency TLV entries, then these are required to be
+If the TLV area contains protected TLV entries, by beginning with a `struct
+image_tlv_info` with a magic value of `IMAGE_TLV_PROT_INFO_MAGIC` then the
+data of those TLVs with size given by `ih_protect_tlv_size` must also be
 integrity and authenticity protected. In this case the SHA256 has to be
-calculated over not just the image header and the image but also the TLV info
-header and the dependency TLVs.
+calculated over not just the image header and the image but also the TLV
+info header and the dependency TLVs.
```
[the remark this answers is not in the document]
utzig: The text was update to clarify the issue.

docs/design.md
```diff
@@ -898,9 +903,11 @@ A +---------------------+
...
   | TLV area            |
   | +-----------------+ |
-  | | TLV area header | | <- struct image_tlv_info
+  | | TLV area header | | <- struct image_tlv_info (optional)
   | +-----------------+ |
-  | | Dependency      | | <- Dependency entry (struct image_tlv)
+  | | Protected TLVs  | | <- Protected TLVs (struct image_tlv)
+  | +-----------------+ |
+  | | TLV area header | | <- struct image_tlv_info
 B | +-----------------+ |
```
tamasban: Should not be B in line 909? The unprotected TLV header should not be included.

docs/design.md
```diff
@@ -898,9 +903,11 @@ A +---------------------+
   +---------------------+
   | TLV area            |
   | +-----------------+ |
-  | | TLV area header | | <- struct image_tlv_info
+  | | TLV area header | | <- struct image_tlv_info (optional)
```
tamasban: might put also the protected TLV area header magic here

docs/design.md
```diff
@@ -898,9 +903,11 @@ A +---------------------+
...
   | +-----------------+ |
-  | | Dependency      | | <- Dependency entry (struct image_tlv)
+  | | Protected TLVs  | | <- Protected TLVs (struct image_tlv)
+  | +-----------------+ |
+  | | TLV area header | | <- struct image_tlv_info
```
tamasban: might put also the unprotected TLV area header magic here

docs/design.md
utzig: Yes, thanks for point this out! Fixed.

docs/design.md
utzig: Done.

docs/design.md
utzig: Done.

utzig:
Thanks for the reviews, I added this commit to #551, and will add reviewers when implementation is done.

PR #558, opened 2019-10-01
zephyr/serial_recovery: fix serial device binding

nvlsianpu:
zephyr changes the way in which it provide name of
serial device used by serial recover.
see: https://github.com/zephyrproject-rtos/zephyr/pull/18752

This path introduce build time configuration of device name
which is set to exact the same device which was set before.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -248,6 +248,18 @@ config BOOT_SERIAL_DETECT_PIN_VAL
 	  Logic value of the detect pin which triggers serial recovery
 	  mode.
 
+# Workaround for not being able to have commas in macro arguments
+DT_CHOSEN_Z_CONSOLE := zephyr,console
+
+config RECOVERY_UART_DEV_NAME
+	string "UART Device Name for Recovery UART"
+	default "$(dt_chosen_label,$(DT_CHOSEN_Z_CONSOLE))" if HAS_DTS
+	default "UART_0"
+	depends on BOOT_SERIAL_UART
+	help
+	  This option specifies the name of UART device to be used for
+	  serial recovery.
+
 endif # MCUBOOT_SERIAL
 
 endmenu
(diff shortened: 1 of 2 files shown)


nvlsianpu:
cc @galak

boot/zephyr/Kconfig
```diff
@@ -248,6 +248,18 @@ config BOOT_SERIAL_DETECT_PIN_VAL
 	  Logic value of the detect pin which triggers serial recovery
 	  mode.
 
+# Workaround for not being able to have commas in macro arguments
+DT_CHOSEN_Z_CONSOLE := zephyr,console
+
+config RECOVERY_UART_DEV_NAME
+	string "UART Device Name for Recovery UART"
+	default "$(dt_chosen_label,$(DT_CHOSEN_Z_CONSOLE))" if HAS_DTS
```
mbolivar: Is there any reason not to use DT_CHOSEN_Z_CONSOLE directly? Are there any platforms we're interested in here that have CONFIG_HAS_DTS=n? If so, I think this patch seems like making the best of a bad situation, but otherwise I don't understand why we have to have this kconfig option.

nvlsianpu: I might `CONFIG_HAS_DTS=n` PLATFORM, that reasonable.
Yeah I introduce something which allow to get the same DTS configuration to mcuboot. Not possible without kconfig i believe. Do you think I should just declare stand-alone config property here, UART_) by default?

galak: > Is there any reason not to use DT_CHOSEN_Z_CONSOLE directly? Are there any platforms we're interested in here that have CONFIG_HAS_DTS=n? If so, I think this patch seems like making the best of a bad situation, but otherwise I don't understand why we have to have this kconfig option.

not sure I follow, what do you mean by use it directly?

nvlsianpu: @galak I believe directly mean "get string constructed by syntax: `"$(dt_chosen_label,$(DT_CHOSEN_Z_CONSOLE))"` above into C code, without kconfig as the intermediary.

nvlsianpu:
@mbolivar I'm going to merge this as release is coming, and this is bugfix.

mbolivar:
> @mbolivar I'm going to merge this as release is coming, and this is bugfix.

Sounds good.

PR #560, opened 2019-10-01
Eliminate a few `extern const` items

d3zd3z:
Values of type `extern const` generate possibly an extra load (depending on the value), but also make it possible to introduce VLA (variable length arrays) into the code. These changes eliminate two of these consts that really should just be #defines. Instead of the simulator accessing them directly, provide an accessor function to make the constant value available.

### DIFF
--- a/sim/mcuboot-sys/csupport/run.c
+++ b/sim/mcuboot-sys/csupport/run.c
@@ -454,3 +454,8 @@ uint32_t boot_max_align(void)
 {
     return BOOT_MAX_ALIGN;
 }
+
+uint32_t boot_magic_sz(void)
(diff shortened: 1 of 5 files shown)


sim/mcuboot-sys/csupport/run.c
```diff
@@ -449,3 +449,13 @@ void sim_assert(int x, const char *assertion, const char *file, unsigned int lin
         }
     }
 }
+
+uint32_t get_boot_max_align(void)
```
utzig: For our typical nomenclature, it would be better if those functions didn't have the `get_` prefix.

PR #565, opened 2019-10-11
imgtool: allow address adjustment in output hex

utzig:
Add a new flag `-x` (or `--hex_addr`) which adjusts the memory address where this file has to be written to. This is useful when generating upgrade images that will go to the secondary slot, in cases where the user is not using mcumgr or some other delivery mechanism, and has to manually adjust the addresses using `objcopy`.

PS: this uses internals of the intelhex python module (protected variables) because no existing mechanism to do the adjustment seems to exist.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -164,20 +164,26 @@ class Image():
 
         self.check()
 
-    def save(self, path):
+    def save(self, path, hex_addr=None):
         """Save an image from a given file"""
-        if self.pad:
-            self.pad_to(self.slot_size)
-
         ext = os.path.splitext(path)[1][1:].lower()
         if ext == INTEL_HEX_EXT:
             # input was in binary format, but HEX needs to know the base addr
-            if self.base_addr is None:
-                raise Exception("Input file does not provide a base address")
+            if self.base_addr is None and hex_addr is None:
+                raise Exception("No address exists in input file neither was "
+                                "it provided by user")
             h = IntelHex()
+            if hex_addr is not None:
+                self.base_addr = hex_addr
             h.frombytes(bytes=self.payload, offset=self.base_addr)
+            if self.pad:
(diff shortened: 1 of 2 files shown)


scripts/imgtool/image.py
```diff
@@ -176,6 +176,12 @@ def save(self, path):
                 raise Exception("Input file does not provide a base address")
             h = IntelHex()
             h.frombytes(bytes=self.payload, offset=self.base_addr)
+            if hex_addr is not None:
```
d3zd3z: Is it possible to do this with h.segments(), h.gets(), and using h.puts() on a new IntelHex object?

utzig: Oh, I completely missed those methods! Fixed, thanks.

utzig:
@d3zd3z So on further analysis, it seems that there is no need to create a new IntelHex object, so I just adjusted the address while loading the payload hex. Also I didn't know this concept of segments existed (shame on me!) so I just fixed the padding code to simply add the magic at the end, which now generates final images that are slightly larger than the original firmware instead of slot sized!

nvlsianpu:
Interesting.
cc @hakonfam

scripts/imgtool/image.py
```diff
@@ -164,20 +164,25 @@ def load(self, path):
...
         """Save an image from a given file"""
-        if self.pad:
-            self.pad_to(self.slot_size)
-
         ext = os.path.splitext(path)[1][1:].lower()
         if ext == INTEL_HEX_EXT:
             # input was in binary format, but HEX needs to know the base addr
-            if self.base_addr is None:
+            if self.base_addr is None and hex_addr is None:
                 raise Exception("Input file does not provide a base address")
```
nvlsianpu: neither Input file nor parameter does provide a base address (please correct syntax as I'm not english native)

utzig: Should have a better message now.

PR #567, opened 2019-10-11
Move around mbed TLS

d3zd3z:
Move `ext/mbedtls` to `ext/mbedtls-asn1` to make it clearer that this is just the ASN1 parser taken out of mbed TLS. Then move the mbedtls submodule out of the sim code into this ext directory.

This should merge post 1.4.0.


sim/mcuboot-sys/build.rs
```diff
@@ -143,13 +143,13 @@ fn main() {
         }
```
utzig: The build failed on the CI because the line above was not updated.

utzig:
This is a good move; while you are on it, you could try updating the submodule from `2.14.1` to `2.16.3` (or whatever the latest stable is when ready for merge) !

d3zd3z:
Fixed the original problem, and included an update of mbedtls to 2.16.3.

PR #574, opened 2019-10-18
mynewt: allow overriding the startup routine

utzig:
Add a new Mynewt option that allows for overriding the default `hal_system_start` routine; this allows for startup customizations required by some BSPs.

There's also a commit with some example custom code that will be dropped before merge!

### DIFF
--- a/boot/mynewt/src/main.c
+++ b/boot/mynewt/src/main.c
@@ -46,6 +46,10 @@
 #include "bootutil/bootutil.h"
 #include "bootutil/bootutil_log.h"
 
+#if MYNEWT_VAL(BOOT_CUSTOM_START)
+void boot_custom_start(uintptr_t flash_base, struct boot_rsp *rsp);
+#endif
+
 #if defined(MCUBOOT_SERIAL)
 #define BOOT_SERIAL_REPORT_DUR  \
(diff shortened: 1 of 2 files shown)


boot/mynewt/src/main.c
```diff
@@ -46,6 +46,12 @@
 #include "bootutil/bootutil.h"
 #include "bootutil/bootutil_log.h"
 
+#if MYNEWT_VAL(BOOT_CUSTOM_START)
+static struct boot_rsp rsp;
+static uintptr_t flash_base;
+void boot_custom_start(uintptr_t flash_base, struct boot_rsp *rsp);
+#endif
```
agross-korg: i don't understand why we need these.  The function call below can just use the flash_base and rsp we have already.

boot/mynewt/src/main.c
```diff
@@ -207,8 +213,10 @@ int flash_device_base(uint8_t fd_id, uintptr_t *ret)
 int
 main(void)
 {
+#if !MYNEWT_VAL(BOOT_CUSTOM_START)
```
agross-korg: let's just use rsp and flash_base here in the custom call as is.

utzig: Fixed

agross-korg:
lgtm

agross-korg:
go ahead and drop the sample.  it's not needed now.

PR #577, opened 2019-10-22
Some minor Zephyr cleanups

d3zd3z:
~~The first patch fixes a problem with how BOARD is defined. Despite the hope of the commit text, CMake doesn't use set(...) as a default, but instead sets the value for the current directory. Although the rest of Zephyr will indeed build correctly, the board-specific overlay files won't ever be selected, since that CMakeLists.txt will always look for just an x86_qemu overlay.~~
Remove old support for DTC_OVERLAY_FILE, which is now done for us in the Zephyr biolerplate CMake file. What we had would never work (it would use a BOARD value that would get replaced by something else in the Zephyr build).

The other path just removes a legacy check for the Kinetis board that hasn't applied to Zephyr for almost 2 years.


boot/zephyr/CMakeLists.txt
```diff
@@ -7,8 +7,12 @@
 cmake_minimum_required(VERSION 3.8.2)
 
 # Board-specific CONF_FILES should get merged into the build as well.
-# Default to qemu_x86 if no board has been specified.
-set(BOARD qemu_x86)
+# If the board is not set by the build system, fail, otherwise the use
+# of BOARD below will mismatch with how this value gets set by the
+# Zephyr build system.
+if(NOT DEFINED BOARD)
+  assert(0 "Must set a BOARD to build on. Please configure a board with -DBOARD=boardname")
```
SebastianBoe: This implies that setting the board through the environment variable BOARD is not supported, but this should also be supported.

SebastianBoe: Not having a default BOARD is sane. But we can't assert like this because there are more ways to set BOARD than -DBOARD=.

SebastianBoe: Doing this would set BOARD correctly:

```
if(NOT BOARD)
        set(BOARD $ENV{BOARD})
endif()

```

One could assert that BOARD is set after this snippet, or one could safely omit the assert and let the assert in boilerplate.cmake enforce that BOARD is set.

mbolivar:
> let the assert in boilerplate.cmake enforce that BOARD is set.

Why not just do that? That's what that logic is for. And anyway, west build will have complained a long time ago, so this must just be for power users who are calling CMake directly.

boot/zephyr/targets/soc_family_kinetis.h
```diff
@@ -1,8 +0,0 @@
-/*
- * Copyright (c) 2017 Linaro
- * SPDX-License-Identifier: Apache-2.0
```
mbolivar: You also need to update `boot/zephyr/include/target.h`.

d3zd3z:
The problem is that the Zephyr CMakeLists.txt file is using the BOARD definition before including the Zephyr project boilerplate. There is 30 or so lines of cmake code that figure out where to get a value for BOARD. I'm open for ideas, but all three solutions I've seen (the if(NOT DEFINED BOARD) that was there before Sebastian's change, the version with Sebastian's change that just sets it, and my current patch which fails if it isn't set a certain way) are wrong in some way. The current version is wrong in the sense that the board overlay is _always_ wrong.

d3zd3z:
> First patch I'm not sure is necessary; there are already a couple of layers in west and the Zephyr build system itself that detect this condition and report an error.

The problem is that the current version always does the wrong thing. It never finds the overlay file, because it always compares it against x86_qemu instead of the real target that boilerplate.cmake decides to use.

boot/zephyr/CMakeLists.txt
d3zd3z: This still doesn't do the right thing. The only solutions I see are 1. replicate the 30 or so lines in `boilerplate.cmake` that figure out what BOARD should be set to, or 2. Figure out how to move the DTC_OVERLAY setting to after `boilerplate.cmake` is included.

mbolivar:
> The problem is that the Zephyr CMakeLists.txt file is using the BOARD definition before including the Zephyr project boilerplate.

Can you please use the precise paths of the files you are talking about? It makes it difficult to understand what you mean otherwise, and there are overnight latencies with the Europeans here :).

I think you mean:

- "Zephyr CMakeLists.txt" == zephyr/CMakeLists.txt
- "Zephyr project boilerplate" == zephyr/cmake/app/boilerplate.cmake

But if so, I don't understand what you're saying, and that doesn't seem correct to me. It goes:

#. Application CMakeLists.txt includes boilerplate.cmake
#. boilerplate.cmake has BOARD detection logic starting around line 120
#. boilerplate.cmake calls `add_subdirectory(${ZEPHYR_BASE} ...)` around line 534, pulling in zephyr/CMakeLists.txt

So board detection happens before zephyr/CMakeLists.txt can do anything at all.

Can you please provide some steps to reproduce the problem you're trying to solve?

d3zd3z:
I'm referring to mcuboot:`boot/zephyr/CMakeLists.txt`. The easiest way to reproduce the problem is to create a `boot/zephyr/boards/target.overlay` file for a given target, and then build MCUboot for that target. Currently, the DTS_OVERLAY_FILE detection code will not pick up that file, because BOARD is being set to x86_qemu while that code is being run.

boot/zephyr/CMakeLists.txt
SebastianBoe: Can you describe a situation (reproducable steps) where this does not do the right thing?

We have been using this trick to replicate the lines in boilerplate.cmake to great effect so far.

SebastianBoe: Never mind, assuming Marti's approach (https://github.com/JuulLabs-OSS/mcuboot/pull/577#issuecomment-545640142) works, this point is moot.

d3zd3z:
Is there a mechanism that is supposed to replace the DTC_OVERLAY_FILE mechanism? I'm trying something that does need an overlay, for a specific board. I'd be happy to remove that code from Zephyr, especially since it clearly isn't ever working anyway.

d3zd3z:
It looks like the boilerplate will use an overlay file named `boards/${BOARD}.overlay`, which should be including the board files as we have them. I'll make the patch to just remove all of this code.

d3zd3z:
Replaced my patch with Martí's, and removed the reference to the file I deleted in the Kinetis patch.

boot/zephyr/CMakeLists.txt
```diff
@@ -4,31 +4,7 @@
...
-  set(DTC_OVERLAY_FILE ${CMAKE_CURRENT_LIST_DIR}/dts.overlay)
-endif()
-
-if (EXISTS ${CMAKE_CURRENT_LIST_DIR}/boards/${BOARD}.overlay)
-  set(DTC_OVERLAY_FILE
-    "${DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/boards/${BOARD}.overlay"
-    CACHE STRING "" FORCE
-    )
-endif()
+cmake_minimum_required(VERSION 3.31.1)
```
mbolivar: Looks like I made a typo here -- 3.13.1 is the real version

d3zd3z:
Fixed the typo. That's what I get for applying patches by hand.

PR #578, opened 2019-10-23
imgtool: fix header padding value check

utzig:
Commit 9117fded476768b458253902a7d9c87e3663802a added support for supplying an `--erased-val`, which defaults to 0xff. When padding a header, the checks were still validating against 0, which is the value used for padding by the Zephyr build system. Now when the user passes `--pad-header` this check is skipped (we already know the contents).

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -197,7 +197,7 @@ class Image():
         """Perform some sanity checking of the image."""
         # If there is a header requested, make sure that the image
         # starts with all zeros.
-        if self.header_size > 0:
+        if self.header_size > 0 and not self.pad_header:
             if any(v != 0 for v in self.payload[0:self.header_size]):
                 raise Exception("Padding requested, but image does not start with zeros")
         if self.slot_size > 0:


d3zd3z:
This will cause Zephyr builds to fail. The zephyr build always pads with zero. Perhaps we could check against either the erased value or zero? Are there build systems that have image padding that isn't zero?

utzig:
Yes, but why would anyone ever `--pad-header` building with Zephyr, if the build system already does it for you? I always used this option to create images that came from a Mynewt build.

d3zd3z:
The check is for when the build system already pads the header. This does fail zephyr builds, just trying building one of the `samples/zephyr` targets, such as test-good-rsa.

d3zd3z:
It isn't failing with `--pad-header` its that this changes what is checked for without it.

utzig:
@d3zd3z Ah, makes sense now. Since Zephyr does the padding with zeroes I will have to check for erased_val only when the padding was requested by the user!

scripts/imgtool/image.py
```diff
@@ -198,8 +198,9 @@ def check(self):
         # If there is a header requested, make sure that the image
         # starts with all zeros.
         if self.header_size > 0:
-            if any(v != 0 for v in self.payload[0:self.header_size]):
-                raise Exception("Padding requested, but image does not start with zeros")
+            val = self.erased_val if self.pad_header else 0
```
d3zd3z: Is there any reason to even check this if pad_header is added, since we just added that header earlier?

utzig: Yeah, no, fixed.

utzig:
@d3zd3z Could you review again?

PR #579, opened 2019-10-23
ECIES-P256

utzig:
Add support for encrypted images using ECIES (using secp256r1 aka ecdsa-256).


utzig:
Removed WIP because this is mostly functional now (with possibly a few `FIXME`/`TODO` to check). I will be adding Zephyr configuration soon. I also found a bug with encrypted passwords, that is probably `sim` related, but I don't think it is relevant for this PR, so temporarily this uses a global password like it was already for RSA-OAEP e AES-128-KW. `newt` support will be added, but does not block this PR.

boot/zephyr/Kconfig
```diff
@@ -135,7 +135,15 @@ config BOOT_BOOTSTRAP
 	  If unsure, leave at the default value.
 
 config BOOT_ENCRYPT_RSA
-	bool "Support for encrypted upgrade images"
+	bool "Support for encrypted upgrade images using RSA"
+	default n
+	help
+	  If y, images in the secondary slot can be encrypted and are decrypted
```
mbolivar: Random: might be better to go into a bit more detail about the RSA vs ECIES algorithms used in the help here instead of copy/pasting the same text.

utzig: The copy/paste remains but I also added a bit of explanation of why on option is different from the other.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -62,6 +62,11 @@
 #define MCUBOOT_ENCRYPT_RSA
 #endif
 
+#ifdef CONFIG_BOOT_ENCRYPT_ECIES
```
d3zd3z: There is a bit of a potential of confusion with this name. Unlike RSA-OEP, ECIES was designed to encrypt (and protect) arbitrary data. When used that way, it offers additional security properties over RSA-OEP. When used as a keywrap, it doesn't offer these properties. However, the name "ENCRYPT_ECIES" kind of suggests we are encrypting the data itself using ECIES.

Perhaps we should call this ENCRYPT_ECIES_KW (or keywrap). Or just use the EC256 that we use elsewhere, which doesn't really mean anything by itself. "ENCRYPT_RSA" doesn't really mean anything by itself, so is less likely to cause confusion.

docs/encrypted_images.md
```diff
@@ -69,13 +69,36 @@ but randomizing a 16-byte block with a TRNG should make it highly
 improbable that duplicates ever happen.
 
 To distribute this AES-CTR-128 key, new TLVs were defined. The key can be
-encrypted using either RSA-OAEP or AES-KW-128. Also in the future support
-for EICES (using EC) can be added.
+encrypted using either RSA-OAEP, AES-KW-128 or ECIES.
```
d3zd3z: This use of ECIES is less ambiguous, because we are at least saying we are using it to wrap the key.  I wonder if something like ECIES-KW might be less confusing, though.

docs/encrypted_images.md
```diff
@@ -69,13 +69,36 @@ but randomizing a 16-byte block with a TRNG should make it highly
...
-AES-KW-128 a new TLV with value `0x31` is added to the image. The contents
-of both TLVs are the results of applying the given operations over the
-AES-CTR-128 key.
+AES-KW-128 a new TLV with value `0x31` is added to the image, and for ECIES
+a new TLV with value `0x32` is added. The contents of those TLVs are the
+results of applying the given operations over the AES-CTR-128 key.
+
+## ECIES encryption
+
+ECIES follows a well defined process to generate an encryption key. There are
```
d3zd3z: I think we shouldn't just call this "ECIES encryption" but perhaps "ECIES keywrap encryption", and then add some text to say that we use this to wrap the key similar to how the other algorithms do it. I'd like to keep things open to someday perhaps be able to use ECIES directly to encrypt the images.

boot/zephyr/CMakeLists.txt
```diff
@@ -95,8 +95,8 @@ zephyr_library_sources(
 
 if(NOT DEFINED CONFIG_FLASH_PAGE_LAYOUT)
   zephyr_library_sources(
-	flash_map_legacy.c
-	)
+    flash_map_legacy.c
```
d3zd3z: Please pull out the tab/space patch into its own patch.

sim/mcuboot-sys/build.rs
```diff
@@ -167,11 +169,41 @@ fn main() {
         }
     }
 
+    if enc_ec256 {
```
d3zd3z: This conflicts with #571, however it's fine to leave it, since that is still marked as WIP, and this is close to ready.

boot/zephyr/Kconfig
mbolivar: Looks great, that's a clear distinction with pointers to more info. Thanks!

boot/zephyr/include/mcuboot_config/mcuboot_config.h
utzig: I did not do exactly that, but I think it is properly described in the encrypted images doc now.

docs/encrypted_images.md
utzig: I updated all references to be either `ECIES_P256` or `EC256`. It does not feel appropriate to add `-KW` because that terminology does not exists in the papers. But the docs are somewhat clear of what we are doing.

boot/zephyr/CMakeLists.txt
utzig: Done.

PR #580, opened 2019-10-23
stats: Get rid of static qualifier for STATS_NAME_START

manish-sudo:


utzig:
@manish-sudo You need to sign-off your commit.

PR #583, opened 2019-10-30
imgtool: ed25519: add method to return raw bytes

nkaje:
Add a method to return the 32 bytes Public Key in
Raw Encoding and Public Raw format.
The existing `get_public_bytes` returns DER format which is 44 bytes long.

This enables the clients to use the API instead of
doing a
```
    string_stream = (
        key._get_public()
        .public_bytes(
            encoding=serialization.Encoding.Raw, format=serialization.PublicFormat.Raw
        )
        .hex()
    )
```

Signed-off-by: Naveen Kaje <naveen.kaje@juul.com>

### DIFF
--- a/scripts/imgtool/keys/ed25519.py
+++ b/scripts/imgtool/keys/ed25519.py
@@ -32,6 +32,11 @@ class Ed25519Public(KeyClass):
                 encoding=serialization.Encoding.DER,
                 format=serialization.PublicFormat.SubjectPublicKeyInfo)
 
+    def get_public_raw_bytes(self):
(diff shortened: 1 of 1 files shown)


utzig:
I don't think the use case is enough for adding this very specific API. At least you would have to define it in the base class, add some tests, and implement this for all key classes. For your use case why can't you just use `get_public_bytes()[-32:]` instead?

nkaje:
Yes, get_public_bytes[-32:] is sufficient. Thanks for the tip! Closing this.

Issue #586, opened 2019-11-06
NRF Cryptocell is not actually being disabled.


solsbarry:
https://github.com/JuulLabs-OSS/mcuboot/blob/8554ed5be658bf710b8e79cbac74e162c3b27338/ext/nrf/cc310_glue.h#L39

Is there a reason the NRF Cryptocell is not actually being disabled?

Currently
```
static inline void nrf_cc310_disable(void)
{
    NRF_CRYPTOCELL->ENABLE=1;
}
```
why not
```
static inline void nrf_cc310_disable(void)
{
    NRF_CRYPTOCELL->ENABLE=0;
}
```
**From the NRF52840 Manual V1.1**

![image](https://user-images.githubusercontent.com/8915651/68334241-33b18880-00a8-11ea-8e05-743d31302f61.png)

sigvartmh:
Copypasta bug, will fix it. There is no actual reason it is not disabled. Thanks for spotting it :)

solsbarry:
Thanks all, that was super fast.

PR #587, opened 2019-11-06
Bugfix: nRF Cryptocell not being disabled

sigvartmh:
Copy paste bug in cc310 glue layer for nRF where disable does not set
the enable register to 0. Thanks to @solsbarry for pointing this out.

Fixes: #586

Signed-off-by: sigvartmh <sigvart.m@gmail.com>


utzig:
Fixes #586

PR #591, opened 2019-11-22
sim: add randomly generated image encryption keys

utzig:
Removes the hard-coded image encryption keys, and updates with keys generated randomly before use. This tests the correct behavior of how ephemeral keys should be used when generating new images.


utzig:
This will fail, I will add another commit with the `bootutil` fix.

PR #593, opened 2019-11-26
boot: zephyr: Add check of RESETREAS for nRF SoC

jhn-nordic:
Only execute serial recovery if valid pin condition is detected
during power up reset or pin reset on nRF devices with RESETREAS
register. This is to prevent going into serial recovery when
waking up from power off mode with the same pin conditions that
is configured to trigger serial recovery.

Signed-off-by: Jon Helge Nistad <jon.helge.nistad@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -44,6 +44,26 @@ const struct boot_uart_funcs boot_funcs = {
 #include <usb/class/usb_dfu.h>
 #endif
 
+#ifdef CONFIG_SOC_FAMILY_NRF
+#include <hal/nrf_power.h>
+
+static inline bool boot_skip_serial_recovery()
+{
+#if NRF_POWER_HAS_RESETREAS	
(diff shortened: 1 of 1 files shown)


nvlsianpu:
> This is to prevent going into serial recovery when
waking up from power off mode with the same pin conditions that
is configured to trigger serial recovery.

How to trigger serial recovery for such a case?

nvlsianpu:
@jhn-nordic ^^

jhn-nordic:
> > This is to prevent going into serial recovery when
> > waking up from power off mode with the same pin conditions that
> > is configured to trigger serial recovery.
> 
> How to trigger serial recovery for such a case?

Power on reset or pin reset. 
(rr == 0 || (rr & NRF_POWER_RESETREAS_RESETPIN_MASK))

nvlsianpu:
tested on regression - OK.

PR #596, opened 2019-11-28
Fix Insufficient image encryption feature

### LINKED ISSUE #489: Insufficient image encryption feature
I want to encrypt my image to prevent potential attackers to analyse it. As far as I understood the encryption feature works as follows:

1. The image will be encrypted after build using a randomly generated key.
2. The randomly generated key will be encrypted with a fixed key defined by mcuboot (defined in boot/zephyr/keys.c).
3. The encrypted randomly generated key will be attached to the FW image.

This will lead to following security leak:

1. An attacker can read/sniff the encrypted image.
2. The attacker uses the openly accessible key from keys.c to decrypt the randomly generated key contained in the image.
3. The attacker uses the decrypted randomly generated key to decrypt the complete image.

Maybe I'm wrong here, but if this is true the encryption feature doesn't increase security at all and is completely useless.

Hope you can give me feedback to this issue.

gon1332:
Fixes #489 

This patch is targeting and is tested on RSA encryption scheme.

I haven't checked if the docs need any update. This is to come, if you are happy with the changes.


scripts/imgtool/keys/general.py
```diff
@@ -19,17 +19,46 @@ def _public_emit(self, header, trailer, indent, file=sys.stdout, len_format=None
         if len_format is not None:
             print(len_format.format(len(encoded)), file=file)
 
-    def emit_c(self, file=sys.stdout):
+    def _private_emit(self, header, trailer, indent, file=sys.stdout, len_format=None):
+        print(AUTOGEN_MESSAGE, file=file)
+        print(header, end='', file=file)
+        encoded = self.get_private_bytes()
```
utzig: This looks a bit exaggerated in the copypasta; this line above seems to be the only one changed from `_public_emit`. Why not call it `_emit` or `_emit_from_bytes` and pass in the byte buffer?

scripts/imgtool/keys/general.py
```diff
@@ -19,17 +19,46 @@ def _public_emit(self, header, trailer, indent, file=sys.stdout, len_format=None
...
+
+    def emit_c_private(self, file=sys.stdout):
+        self._private_emit(
+                header="const unsigned char enc_priv_key[] = {",
+                trailer="};",
+                indent="    ",
+                len_format="const unsigned int enc_priv_key_len = {};",
+                file=file)
+
+    def emit_rust_private(self, file=sys.stdout):
```
utzig: This won't be used by the simulator, so it should be removed.

scripts/imgtool/keys/general.py
gon1332: Done

scripts/imgtool/keys/general.py
gon1332: Done

boot/zephyr/prj.conf
```diff
@@ -6,7 +6,7 @@ CONFIG_SYS_POWER_MANAGEMENT=n
 CONFIG_MAIN_STACK_SIZE=10240
 CONFIG_MBEDTLS_CFG_FILE="mcuboot-mbedtls-cfg.h"
 
-CONFIG_BOOT_ENCRYPT_RSA=n
+CONFIG_BOOT_ENCRYPT_RSA=y
```
d3zd3z: Why are we changing the default for the encryption feature? Shouldn't that still be up to the user?

boot/zephyr/CMakeLists.txt
```diff
@@ -213,5 +213,19 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
...
+    COMMAND
+    ${PYTHON_EXECUTABLE}
+    ${MCUBOOT_DIR}/scripts/imgtool.py
+    getpriv
+    -k
+    ${KEY_FILE}
+    > ${GENERATED_PRIVKEY}
+    DEPENDS ${KEY_FILE}
+    )
+  zephyr_library_sources(${GENERATED_PRIVKEY})
```
d3zd3z: As I understand, we're changing the behavior from every build using the same hardcoded key, to generating a new one that varies for each build directory this is run in. If I were to clean and rebuild, I would get another key. Aren't we going to need a little better key management to coordinate these keys to whoever is making the build?

boot/zephyr/prj.conf
gon1332: You're right. It's a testing leftover of my part.

boot/zephyr/CMakeLists.txt
gon1332: The current patch hands out the responsibility of the key generation (when and how often) to the user. This could be documented. However, if we were to generate a new private key on each build, we should add an extra command here doing this job.

d3zd3z: Except that the user won't likely be aware of it. The first time they build, they will get a new key. If they clean their tree and rebuild, they will get a new key. @utzig any thoughts on this?

utzig:
Please drop 3b75545aef99af58acb6a1962395f55fa2529f47

gon1332:
> Please drop [3b75545](https://github.com/JuulLabs-OSS/mcuboot/commit/3b75545aef99af58acb6a1962395f55fa2529f47)

Done.

boot/zephyr/CMakeLists.txt
```diff
@@ -213,5 +213,19 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
     DEPENDS ${KEY_FILE}
     )
   zephyr_library_sources(${GENERATED_PUBKEY})
+
+  set(GENERATED_PRIVKEY ${ZEPHYR_BINARY_DIR}/autogen-privkey.c)
```
d3zd3z: I'll be honest, I don't want this feature. It isn't going to do anything for me, other than make development slightly less convenient, when, after I make clean, I have a new keypair being used.

At a minimum, can we make this a configurable option? The default could be to use the hard-coded key, and with an option, it will generate one for that build. Really, to do secure key management, the private key would need to come from an HSM, which is something on our longer-term task list.

d3zd3z:
I believe this will address #1094.

boot/zephyr/CMakeLists.txt
d3zd3z: Thinking more about this, this would address #1094, which causes us to get periodic security vulnerability reports, because of the in-repo keys.

d3zd3z (commented):
Doing this for Zephyr is a good start. It probably needs to be done for all platforms and the simulator, in order to be able to remove the keys from the repo, and satisfy #1094.

PR #602, opened 2019-12-03
Add html anchors to design and imgtool documents

utzig:


utzig:
Yes, the anchors are generated but there is not easy way to link them, unless one browses the page source.

utzig:
> I'm ok with the change itself, but I think it would be useful to have a bit of an explanation in the commit text as to why this is being done.

Done.

PR #603, opened 2019-12-03
zephyr: Fix build problem with missing overlay

d3zd3z:
The commit:

commit 0e3fa72df4849eb42c99efb34b8617046c2e33cb
Author: Martí Bolívar <marti.bolivar@nordicsemi.no>
Date:   Tue Oct 22 14:39:33 2019 -0600

    zephyr: Fix board logic

Mistakenly removed the dts.overlay file, despite it still being
referenced.  What I believe it intended to remove was the ifdef that is
no longer needed, but we still reference this overlay, and still need it
in order to put mcuboot into the boot partition of flash.

### DIFF
--- a/boot/zephyr/dts.overlay
+++ b/boot/zephyr/dts.overlay
@@ -1,0 +1,8 @@
+/*
+ * MCUboot should always be built in the boot partition.
+ */
+/ {
+	chosen {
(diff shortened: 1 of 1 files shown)


d3zd3z:
This is insufficient. @SuperPrower is going to make a correct PR for this.

PR #604, opened 2019-12-03
zephyr: restore device tree overlay

### LINKED ISSUE #601: zephyr: mcuboot is not being linked to the boot partition
0e3fa72df4849eb42c99efb34b8617046c2e33cb removed device tree overlay that is required to flash mcuboot to correct partition and adding a CMake variable that pointed to this file; however, without this, resulting image is not linked to the beginning of the flash:

```
...
(gdb) file ./mcuboot/boot/zephyr/build/zephyr/zephyr.elf
A program is being debugged already.
Are you sure you want to change the file? (y or n) y
Load new symbol table from "../mcuboot/boot/zephyr/build/zephyr/zephyr.elf"? (y or n) y
Reading symbols from ../mcuboot/boot/zephyr/build/zephyr/zephyr.elf...
(gdb) load
Loading section text, size 0xdc lma 0xc000 <- incorrect!
...
```

As far as I understand this is happens because ec134775d05e455d70f55339c06252115d62e638 (or one of the later patches on that file) made it such that Zephyr [will now link](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_USE_CODE_PARTITION.html)  application into `zephyr,code-partition`, which is  `&slot0_partition` for big portion of ARM boards. Thus, an overlay that will specify correct code partition is needed (which was deleted in commit mentioned at the beginning).

If my assumption is correct, I have a PR ready to fix this, but first I would like to hear opinion from more experienced people.

SuperPrower:
0e3fa72df4849eb42c99efb34b8617046c2e33cb removed device tree overlay that configures Zephyr to link mcuboot at the beginning of the flash, and CMake variable that pointed to it (required for Zephyr). This resulted in mcuboot not being linked at some offset, rendering it unable to boot. Restore overlay and required code.

Without it, Zephyr will use `zephyr,code-partition` configuration from board device tree that usually points to the `slot0_partition`. This behaviour is due to the [CONFIG_USE_CODE_PARTITION](https://docs.zephyrproject.org/latest/reference/kconfig/CONFIG_USE_CODE_PARTITION.html) being used in MCUboot's configuration for Zephyr.

Fixes #601.

Signed-off-by: Sergey Koziakov <dya.eshshmai@gmail.com>

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -6,6 +6,19 @@
 
 cmake_minimum_required(VERSION 3.13.1)
 
+# Add a common dts overlay necessary to ensure mcuboot is linked into,
+# and fits inside, the boot partition. (If the user specified a
+# DTC_OVERLAY_FILE on the CMake command line, we need to append onto
+# the list).
+if(DTC_OVERLAY_FILE)
+  set(DTC_OVERLAY_FILE
+    "${DTC_OVERLAY_FILE} ${CMAKE_CURRENT_LIST_DIR}/dts.overlay"
+    CACHE STRING "" FORCE
+    )
+else()
+  set(DTC_OVERLAY_FILE ${CMAKE_CURRENT_LIST_DIR}/dts.overlay)
+endif()
+
 # Enable Zephyr runner options which request mass erase if so
 # configured.
 #
--- a/boot/zephyr/dts.overlay
+++ b/boot/zephyr/dts.overlay
@@ -1,0 +1,5 @@
+/ {
+	chosen {
+		zephyr,code-partition = &boot_partition;
+	};
+};


mbolivar:
> Do not restore adding "boards/${BOARD}.overlay" to the list since Zephyr
> detects this file automatically now

I believe this is incorrect. It only detects it automatically if DTC_OVERLAY_FILE is not already set. See this comment for a fuller description: https://github.com/JuulLabs-OSS/mcuboot/pull/577#pullrequestreview-323964150

mbolivar:
This is exactly what I was talking about when I say I wish the Zephyr behavior were additive instead of mutually exclusive.

SuperPrower:
@mbolivar , yes, you are [correct](https://docs.zephyrproject.org/latest/application/index.html#devicetree-overlays). I am unfamiliar with changes of Zephyr functionality and assumed that mentioned part was added because previously Zephyr didn't automatically scan for the board-specific overlay. 
I can edit the commit message to make it more correct if you believe this is required.

mbolivar:
I more meant that the BOARD.overlay behavior is desirable.

d3zd3z:
We do need to do something, though. Right now, Zephyr doesn't build correctly, because there is no overlay, and it goes in the wrong partition? We do need a general overlay file that is used for all MCUboot targets, in order to set the default partition.

mbolivar:
Then I think there's no choice but to implement the desired "additive" logic by hand before including the boilerplate file.

d3zd3z:
@mbolivar Are you ok with this approach for now?

mbolivar:
> @mbolivar Are you ok with this approach for now?

No objections from me on any of this, just chiming in with information I thought might be relevant

PR #608, opened 2019-12-05
Make TLVs 16 bits

d3zd3z:
Two changes, one for the code, and one for the simulator to make TLVs 16 bits. The format is compatible with the existing TLVs, so it isn't necessary to change imgtool until we want to support a TLV value larger than 8-bits (presumably that can happen with the user-defined TLV support).

### DIFF
--- a/sim/src/tlv.rs
+++ b/sim/src/tlv.rs
@@ -268,8 +268,7 @@ impl ManifestGen for TlvGen {
             protected_tlv.write_u16::<LittleEndian>(self.protect_size()).unwrap();
             for dep in &self.dependencies {
                 protected_tlv.write_u16::<LittleEndian>(TlvKinds::DEPENDENCY as u16).unwrap();
-                protected_tlv.push(12);
-                protected_tlv.push(0);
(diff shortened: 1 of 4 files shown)


sim/src/tlv.rs
```diff
@@ -267,8 +267,7 @@ impl ManifestGen for TlvGen {
             protected_tlv.push(0x69);
             protected_tlv.write_u16::<LittleEndian>(self.protect_size()).unwrap();
             for dep in &self.dependencies {
-                protected_tlv.push(TlvKinds::DEPENDENCY as u8);
-                protected_tlv.push(0);
+                protected_tlv.write_u16::<LittleEndian>(TlvKinds::DEPENDENCY as u16).unwrap();
                 protected_tlv.push(12);
                 protected_tlv.push(0);
```
utzig: Those ordered byte pushes read much better; I would suggest you to also change the `len` pushes to use it.

d3zd3z: I seem to have missed all of the ones that are just hardcoded values. I did change the lengths that come from a value. I'll make a commit that changes the rest as well.

PR #611, opened 2019-12-07
ext/mbedtls-asn1: Add support for mbedtls `sha512.c`

rajivr:
`sha512.c` is needed by `curve22519.c`. By having `sha512.c` support in `ext/mbedtls-asn1`, we can do away with the requirement of having to import mbedtls into our tree just for `sha512.c` support.

As per PR #567, the imported mbedtls files are from mbedtls LTS version 2.16.3.

### DIFF
--- a/ext/mbedtls-asn1/include/mbedtls/config.h
+++ b/ext/mbedtls-asn1/include/mbedtls/config.h
@@ -33,6 +33,7 @@
 #define MBEDTLS_MD_C
 #define MBEDTLS_OID_C
 #define MBEDTLS_PEM_PARSE_C
+#define MBEDTLS_SHA512_C
 
 /**
  * \def MBEDTLS_PLATFORM_C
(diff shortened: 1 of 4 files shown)


utzig:
> we can do away with the requirement of having to import mbedtls into our tree just for sha512.c support.

Are you using tinycrypt for the sha256? This does not include any changes to build scripts, which means sha512.c will only stay there without use (for most users); or are you planning on also adding the build changes?

rajivr:
> Are you using tinycrypt for the sha256?

Yes, I think so. I am still working on the port and [here](https://gist.github.com/rajivr/98fb8b2b369f3bde596c449b3c567cc8) is how our `mcuboot_config.h` looks.

> This does not include any changes to build scripts, which means sha512.c will only stay there without use (for most users); or are you planning on also adding the build changes?

I was not planning on making any changes to the build scripts. However, if there is interest upstream in adding `enc_ec256` + `sig_ed25519` configuration to `build.rs` without mbedtls, I can submit another PR enabling that configuration.

We are using Rust as well so I am somewhat familiar with `build.rs`, but not the rest of the simulator code

rajivr:
Closing in favor of #627

PR #618, opened 2019-12-12
zephyr: switch to recent defines and include headers

nvlsianpu:
In zephyr code-base some generic include patches
and defines had been deprecated since a while.
This condition was maintained until zephyr v2.1.0 inclusive.

Recently these deprecated items were removed completely from
the zephyr master branch. Zephyr current SHA of this patch is
e124c1cd34938f02bbe879533c6bd29b5ff53707

This patch switch mcuboot to using most recent items.


utzig:
Btw, the CI failed because there are some files that are used by both Zephyr and the sim so have to adjust those includes...

boot/zephyr/include/flash_map_backend/flash_map_backend.h
```diff
@@ -8,7 +8,7 @@
 #ifndef __FLASH_MAP_BACKEND_H__
 #define __FLASH_MAP_BACKEND_H__
 
-#include <flash_map.h> // the zephyr flash_map
+#include <storage/flash_map.h> // the zephyr flash_map
```
utzig: Moving `sim/mcuboot-sys/csupport/flash_map.h` -> `sim/mcuboot-sys/csupport/storage/flash_map.h` should probably fix this breakage. Otherwise you can use `defined(__BOOTSIM__)` here...

nvlsianpu: I moved the file as prefer not to increase if-def-ry

PR #619, opened 2019-12-12
zephyr: Integrate with Zephyr's build and test systems

nvlsianpu:
Added sample.yaml which is required for automatic building
mcuboot using zephyr saniticheck.

Can be run by zephyr sanitycheck locally using
--testcase-root=/path/to/mcuboot/boot/zephyr
to make it detect the sample.yaml added.

Added here as we want to keep the zephyr fork only pure mirror repo.


boot/zephyr/sample.yaml
```diff
@@ -0,0 +1,14 @@
+sample:
+  description: mcuboot test build sample
+  name: mcuboot
+common:
+    tags: introduction
+    harness: console
+    harness_config:
+      type: one_line
+      regex:
+        - "Hello World! (.*)"
```
mbolivar: I think this part is from the hello_world sample and is not needed. The rest of the sample.yaml looks good to me. I will leave it up to you MCUboot developers if you want to have the header changes.

nvlsianpu: of course - this is just a sandbox for now, will be fixed
Actually I would add at last test with configuration for serial recovery as well.

carlescufi:
CC @nashif

boot/zephyr/sample.yaml
```diff
@@ -0,0 +1,15 @@
+sample:
+  description: mcuboot test build sample
+  name: mcuboot
+
+tests:
+  sample.bootloader.mcuboot:
+    tags: bootloader_mcuboot
+    platform_whitelist:  nrf52840_pca10056
```
carlescufi: Should we build it for another platform too? I know mcuboot supports many other Zephyr boards. Perhaps start with this one and then add more later?

nvlsianpu: That my thought as well.

utzig:
What is Zephyr sanitycheck?

carlescufi:
> What is Zephyr sanitycheck?
It's Zephyr's tool to build and run samples and tests during CI:

https://docs.zephyrproject.org/latest/guides/test/ztest.html

boot/zephyr/sample.yaml
d3zd3z: The frdm_k64f should always work, so maybe that one as well?

utzig: I agree, `nrf52840_pca10056` and `frdm_k64f` are the two boards I use for all my Zephyr testing.

nvlsianpu: Added frdm_k64f. Are you using it to test serial recovery?

nvlsianpu:
@carlescufi  module.yaml added

boot/zephyr/sample.yaml
nvlsianpu: ^^ if yes - please give me proj.conf for it.

utzig: I don't think I've used `boot_serial` with the `frdm_k64f`, but I'm gonna test it later and can send you the `proj.conf`; but it should not block this PR.

carlescufi: It doesn't work, so not enabling it right now. Have disabled it now.

carlescufi:
@mbolivar would you mind giving this a +1?

boot/zephyr/sample.yaml
```diff
@@ -0,0 +1,15 @@
+sample:
+  description: mcuboot test build sample
+  name: mcuboot
+
+tests:
+  sample.bootloader.mcuboot:
+    tags: bootloader_mcuboot
+    platform_whitelist: nrf52840_pca10056 frdm_k64f
```
mbolivar: The whitelist doesn't contain pca10059. Is that ok? Will it get built for this board with CDC ACM recovery enabled?

carlescufi: It will get built with CDC ACM recovery, yes, I tested it locally.

carlescufi:
This is no longer an RFC, it's ready for merging.

carlescufi:
@utzig @d3zd3z this is ready for merging. Any objections?

boot/zephyr/sample.yaml
nvlsianpu: BTW - build for pca10059 without CDC ACM would be almost identical as for pca10056 (only different flash layout).

PR #620, opened 2019-12-12
Encrypted image improvements

utzig:
This applies a few improvements to a commit previously included in PR #596:

* Move functions to dump a private key to the private key classes
* Remove language option; always dumps in C format
* Add option to generate a minimal dump. This will remove extra parameters that are present in keys generated with the `keygen` command. For P256 this will remove the public point, which is already ignored by the parsing function. The resulting key dump shrinks from 138 to 70 bytes. For RSA it will remove the primes P/Q which can be calculated at runtime, and DP/DQ/QP which are only used with CRT and can also be calculated at runtime. This reduces the size of a key dump from around 1190 bytes to somewhere close to 530 bytes. A patch to the RSA parsing routine will be added in another commit.

DISCLAIMER: very "hacky" DER updating here!


utzig:
@d3zd3z Yes, that is correct but it requires extra code to generate the values that are removed from the ASN1, so there's some removal of size because of the shrank key, but added size in extra code, so I have to do some experiments to see if this is valuable. And lso it's very slow, but I did not "bench". The main thing I don't like about this is that hacky way of updating the ASN1, because I have only skimmed at existing ASN1 python libraries, but nothing looks really great. The EC ASN1 change is totally straight-forward and should be a default actually, once I can do ASN1 updating without hacky code.

d3zd3z:
I would think if the code to re-compute the CRT values is larger than the space saved, there isn't any benefit. And, as you say, it takes a while to compute them (although it is still probably faster than ECDSA).

utzig:
> And, as you say, it takes a while to compute them (although it is still probably faster than ECDSA).

This is not related to signing so no ECDSA here. But I got a few numbers:

| description | parse privkey + decrypt TLV time | mcuboot size (bytes)
|----------------|---------------------------------------------|---------------------
| RSA with CRT enabled (default) + full key | 0.909s | 50136
| RSA with CRT enabled (default) + minimized key | 1.012s |  49152
| RSA with CRT disabled + minimized key | 1.266s | 48284
| ECIES-P256 | 0.235s | 40704

I updated the RSA minimization code to always include P/Q because it fails randomly on Zephyr when those are not available; it works on the sim but I did not debug to find the reason; also it is really slow to deduce the primes in the sim!

I tried a few of the python ASN1 libraries to work with the DER but none can decode the input correctly or make it much easier when re-constructing so staying with the manual parsing. The exceptions should be enough to at least protect from invalid parsing and `minimal` stays with `False` for default.

utzig:
@d3zd3z Yeah, good point, if someone runs passing in `--minimal` it won't work because the parser still will be expecting PQ/DQ/QP to exist, so I will swap those commits.

d3zd3z:
Although it is less of an issue, since it does take an extra argument to cause the failure.

PR #621, opened 2019-12-12
docs: Write up some docs about ecdsa padding

d3zd3z:
Start with some documentation on ECDSA signatures, and the problems with
the current padding approach.  Present a plan to support correctly
formatted ECDSA signatures, and how to handle the transition both in the
C code, as well as the tooling that signs images.

### DIFF
--- a/docs/ecdsa.md
+++ b/docs/ecdsa.md
@@ -25,8 +25,8 @@ There are a couple of ways to fix this:
       generated.  These tools are all written in higher-level
       languages and this change should not be difficult.
 
-      However, this will also break compatibility with older version,
-      significantly in that images generated with newer tools will not
+      However, this will also break compatibility with older versions,
+      specifically in that images generated with newer tools will not
       work with older versions of MCUboot.
 
 This document proposes a multi-stage approach, to give a transition
@@ -40,7 +40,7 @@ period.
   - MCUboot will be modified to allow unpadded signatures right away.
     The existing EC256 implementations will still work (with or
     without padding), and the existing EC implementation will begin
-    accepting all signatures.
+    accepting padded and unpadded signatures.
 
   - An mbed TLS implementation of EC256 can be added, but will require
     the `--valid-ecdsa` signature to be able to boot all generated
@@ -48,7 +48,7 @@ period.
     padding, and be considered invalid).
 
 After one or more MCUboot release cycles, and announcements over
-relevant channels, the arguments to mcuboot will change:
+relevant channels, the arguments to `imgtool.py` will change:
 
   - `--valid-ecdsa` will still be accepted, but have no effect.
 
@@ -76,11 +76,11 @@ where both `r` and `s` are 256-bit numbers.  Because these are
 unsigned numbers that are being encoded in ASN.1 as signed values, if
 the high bit of the number is set, the DER encoded representation will
 require 33 bytes instead of 32.  This means that the length of the
-signature will vary by a couple of bytes, depending on whether on of
+signature will vary by a couple of bytes, depending on whether one of
 both of these numbers has the high bit set.
 
 Originally, MCUboot added padding to the entire signature, and just
-removed 0 bytes from the data block.  This would be fine 255/256
+removed any trailing 0 bytes from the data block.  This would be fine 255/256
 times, when the last byte of the signature was non-zero, but if the
 signature ended in a zero, it would remove too many bytes, and the
 signature would be considered invalid.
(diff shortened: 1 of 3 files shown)


d3zd3z:
Added a change to fix the length check, since that doesn't cause any incompatibility.

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+      signatures.  EC code would only reliably work in the new
+      combination).
+
+  2.  Remove the padding entirely.  Depending on which tool, this will
+      require some rethinking of how TLV generation is implemented so
+      that the length does not need to be known until the signature is
+      generated.  These tools are all written in higher-level
+      languages and this change should not be difficult.
+
+      However, this will also break compatibility with older version,
```
mbolivar:
```suggestion
      However, this will also break compatibility with older versions
```

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+    ECDSA-Sig-Value ::= SEQUENCE {
+      r  INTEGER,
+      s  INTEGER
+    }
+
+where both `r` and `s` are 256-bit numbers.  Because these are
+unsigned numbers that are being encoded in ASN.1 as signed values, if
+the high bit of the number is set, the DER encoded representation will
+require 33 bytes instead of 32.  This means that the length of the
+signature will vary by a couple of bytes, depending on whether on of
```
mbolivar:
```suggestion
signature will vary by a couple of bytes, depending on whether one or
```

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+both of these numbers has the high bit set.
+
+Originally, MCUboot added padding to the entire signature, and just
+removed 0 bytes from the data block.  This would be fine 255/256
```
mbolivar:
```suggestion
removed any trailing 0 bytes from the data block.  This would be fine 255/256
```

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+
+  - First, add a `--valid-ecdsa` argument to the sign command in
+    `imgtool.py`.  Without this, the images will be padded with the
+    existing scheme, and with the argument, the ecdsa will be encoded
+    without any padding.
+
+  - MCUboot will be modified to allow unpadded signatures right away.
+    The existing EC256 implementations will still work (with or
+    without padding), and the existing EC implementation will begin
+    accepting all signatures.
```
mbolivar:
```suggestion
    accepting padded and unpadded signatures.
```

Is that correct? I'm not sure. The existing wording makes it sound like it will always accept any signature, even if it's invalid, which I'm guessing is not what you meant.

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+
+  - An mbed TLS implementation of EC256 can be added, but will require
+    the `--valid-ecdsa` signature to be able to boot all generated
+    images (without the argument 3 of out 4 images generated will have
+    padding, and be considered invalid).
+
+After one or more MCUboot release cycles, and announcements over
+relevant channels, the arguments to mcuboot will change:
```
mbolivar:
```suggestion
relevant channels, the arguments to imgtool will change:
```

docs/ecdsa.md
```diff
@@ -0,0 +1,89 @@
...
+
+  - `--valid-ecdsa` will still be accepted, but have no effect.
+
+  - `--invalid-ecdsa` will now bring back the old padding behavior.
```
mbolivar: Just to be clear, you're going to implement support for both options at the same time, right?

That way, people can add `--invalid-ecdsa` to their scripts today if they are signing images for an MCUboot that can't handle unpadded signatures, and not have to make changes to their script when the default is flipped to 'valid'.

utzig:
Not a review, but maybe we should think about deprecating secp224r1; I have not used it in quite a while and the simulator does not test it. Would not surprise me if it is currently broken.

docs/ecdsa.md
d3zd3z: That does make sense to do it that way. I'll have to look at how click handles this, it might be best to have `--valid-ecdsa` and `--no-valid-ecdsa`, or I can just make them two separate arguments, and figure out the value from there (and error if they ask for both valid and invalid).

d3zd3z: I'm also going to change these to `--pad-sig` and `--no-pad-sig`.

d3zd3z:
I've pushed the doc changes as two separate commits (everything got rebased because I messed up, though), to make it easier to review the new changes. I'll squash the minor doc edits into the main once people are happy with this.

PR #625, opened 2019-12-19
Encrypted image upgrade security fixes

utzig:
This adds security fixes to the use of encrypted images. It was tested in the simulator although I did not left it enabled by default.


sim/mcuboot-sys/src/c.rs
```diff
@@ -40,6 +40,10 @@ pub fn boot_trailer_sz(align: u8) -> u32 {
     unsafe { raw::boot_trailer_sz(align) }
 }
 
+pub fn boot_status_sz(align: u8) -> u32 {
```
d3zd3z: `align` should now be a `u32`. The C code picked up the change.

sim/mcuboot-sys/src/c.rs
```diff
@@ -82,6 +86,7 @@ mod raw {
         pub fn invoke_boot_go(sim_ctx: *mut CSimContext, areadesc: *const CAreaDesc) -> libc::c_int;
 
         pub fn boot_trailer_sz(min_write_sz: u8) -> u32;
+        pub fn boot_status_sz(min_write_sz: u8) -> u32;
```
d3zd3z: Likewise here, `min_write_sz` should also be a `u32`. Hopefully this is consistent with the C code.

sim/src/image.rs
```diff
@@ -660,16 +660,8 @@ impl Images {
...
     fn status_sz(&self, align: usize) -> usize {
-        let bias = if Caps::EncRsa.present() || Caps::EncKw.present() ||
-                Caps::EncEc256.present() {
-            32
-        } else {
-            0
-        };
-
-        self.trailer_sz(align) - (16 + 32 + bias)
+        c::boot_status_sz(align as u8) as usize
```
d3zd3z: Again, it should be ` u32`.

utzig:
Now this is also tested in Mynewt and Zephyr.

Issue #224, opened 2018-02-05
Add support for optional anti-rollback protection


aditihilbert:
Some systems require anti-rollback protection to be in place. The easiest way for us to do this is to prevent upgrades to versions that are older than the currently running version. 

In order to do this, we need to define what the ordering means with the version fields, and then add an option to enforce it.

tamasban:
I would like to open a discussion about the implementation of rollback protection in MCUBoot:
 - Should it be mandatory or just optional?
 - How to implement it?
 - Where to store the rollback counter (header or TLV section)?
 - How does SUIT manifest handle it? 
 - Security concerns with image swapping

Aim of anti-rollback protection:
 - Prevent to downgrade the device to an older version of  SW, which contains a known security vulnerability.
 - Manifest data of firmware images must contain a security counter, which is increased if a potential security vulnerability is fixed in the new version.
 - Boot loader must ensure that only such a firmware is accepted for update which related security counter is the same or higher than the security counter of the current active image on the device. 

Some potential implementation of rollback protection:
 -  Always use the latest version. Compare the full version numbers which are extracted from image header and do not accept any downgrade.
 -  Check only a part of the image version. i.e: if (major+minor) part of version number must be the same or higher to accept an image for update. Beyond these the rest of the image version does not matter. Downgrading image up to certain image version is possible.
- Introduce a new security counter which can be independent from image main version number. This counter must be increased only in the case if a potential security vulnerability is fixed in the new version of the SW. After this version will be deployed then it is not allowed to downgrade to older ones.

What to use for security counter:
 - Security counter must be in stored in NVM memory (on-chip eFlash, OTP, trusted subsytem).  
 - Version number in image header can be a good candidate because it is in the integrity protected part of the image and active image is stored in eFlash.
 -  Introduce a new rollback/security counter. This counter should be placed in the integrity protected part of the image. This could go in the header but there is not too much empty field in the header. 

Extend the the integrity protection to some part of the TLV section:
 - Currently the TLV section is not integrity protected, not covered by the hash & signature.
 - It would be beneficial to cover some part of the TLV section with the signature. In this case new TLV-s could be introduced easily which are requires integrity protection and the header should not but overloaded. There is an implementation of this in [multi image boot PR](https://github.com/JuulLabs-OSS/mcuboot/pull/369/commits/ee2d6046b1c3857d109c7cf6b8194b04acafd937). In that case the dependency TLV entries are stored in the protected part of the TLV section.

[...]

PR #628, opened 2020-01-03
add option for rollback protection

hakonfam:
Depends on 'MCUBOOT_OVERWRITE_ONLY' option since swap info is not protected
by signature

Proposal to issue #224 

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -510,6 +510,42 @@ boot_check_header_erased(struct boot_loader_state *state, int slot)
     return 0;
 }
 
+#if (BOOT_IMAGE_NUMBER > 1) || \
+    (defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION))
+/**
+ * Check if the version of the image is not older than required.
+ *
+ * @param req         Required minimal image version.
+ * @param ver         Version of the image to be checked.
+ *
+ * @return            0 if the version is sufficient, nonzero otherwise.
+ */
+static int
+boot_is_version_sufficient(struct image_version *req,
+                           struct image_version *ver)
+{
+    if (ver->iv_major > req->iv_major) {
+        return 0;
+    }
+    if (ver->iv_major < req->iv_major) {
+        return BOOT_EBADVERSION;
+    }
+    /* The major version numbers are equal. */
+    if (ver->iv_minor > req->iv_minor) {
+        return 0;
+    }
+    if (ver->iv_minor < req->iv_minor) {
+        return BOOT_EBADVERSION;
+    }
+    /* The minor version numbers are equal. */
+    if (ver->iv_revision < req->iv_revision) {
+        return BOOT_EBADVERSION;
+    }
+
+    return 0;
+}
+#endif
+
 /*
  * Check that there is a valid image in a slot
  *
@@ -541,6 +577,24 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         goto out;
     }
 
+#if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
+    if (slot != BOOT_PRIMARY_SLOT) {
+        /* Check if version of secondary slot is sufficient */
+        rc = boot_is_version_sufficient(
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
+        if (rc != 0 && boot_check_header_erased(state, BOOT_PRIMARY_SLOT)) {
+            BOOT_LOG_ERR("insufficient version in secondary slot");
+            flash_area_erase(fap, 0, fap->fa_size);
+            /* Image in the secondary slot does not satisfy version requirement.
+             * Erase the image and continue booting from the primary slot.
+             */
+            rc = 1;
+            goto out;
+        }
+    }
+#endif
+
     if (!boot_is_header_valid(hdr, fap) || boot_image_check(state, hdr, fap, bs)) {
         if (slot != BOOT_PRIMARY_SLOT) {
             flash_area_erase(fap, 0, fap->fa_size);
@@ -552,7 +606,7 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         BOOT_LOG_ERR("Image in the %s slot is not valid!",
                      (slot == BOOT_PRIMARY_SLOT) ? "primary" : "secondary");
 #endif
-        rc = -1;
+        rc = 1;
         goto out;
     }
 
@@ -966,39 +1020,6 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
 #endif
 
 #if (BOOT_IMAGE_NUMBER > 1)
-/**
- * Check if the version of the image is not older than required.
- *
- * @param req         Required minimal image version.
- * @param ver         Version of the image to be checked.
- *
- * @return            0 if the version is sufficient, nonzero otherwise.
- */
-static int
-boot_is_version_sufficient(struct image_version *req,
-                           struct image_version *ver)
-{
-    if (ver->iv_major > req->iv_major) {
-        return 0;
-    }
-    if (ver->iv_major < req->iv_major) {
-        return BOOT_EBADVERSION;
-    }
-    /* The major version numbers are equal. */
-    if (ver->iv_minor > req->iv_minor) {
-        return 0;
-    }
-    if (ver->iv_minor < req->iv_minor) {
-        return BOOT_EBADVERSION;
-    }
-    /* The minor version numbers are equal. */
-    if (ver->iv_revision < req->iv_revision) {
-        return BOOT_EBADVERSION;
-    }
-
-    return 0;
-}
-
 /**
  * Check the image dependency whether it is satisfied and modify
  * the swap type if necessary.
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -346,4 +346,14 @@ config UPDATEABLE_IMAGE_NUMBER
 	help
 	  Enables support of multi image update.
 
+config MCUBOOT_DOWNGRADE_PREVENTION
+	bool "Downgrade prevention"
+	depends on BOOT_UPGRADE_ONLY
+	help
+	  Prevent downgrades by enforcing incrementing version numbers.
+	  When this option is set, any upgrade must have greater major version
+	  or greater minor version with equal major version. This mechanism
+	  only protects against some attacks against version downgrades (for
+	  example, a JTAG could be used to write an older version).
+
 source "$ZEPHYR_BASE/Kconfig.zephyr"
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -89,6 +89,10 @@
 #define MCUBOOT_IMAGE_NUMBER    1
 #endif
 
+#ifdef CONFIG_MCUBOOT_DOWNGRADE_PREVENTION
+#define MCUBOOT_DOWNGRADE_PREVENTION 1
+#endif
+
 /*
  * Enabling this option uses newer flash map APIs. This saves RAM and
  * avoids deprecated API usage.
--- a/docs/design.md
+++ b/docs/design.md
@@ -942,3 +942,11 @@ state after dependency check.
 
 For more information on adding dependency entries to an image,
 see: [imgtool](imgtool.md).
+
+## [Downgrade Prevention](#downgrade-prevention)
+
+Downgrade prevention is a feature which enforces that the new image must have a
+higher version number than the image it is replacing. This feature is enabled
+with the `MCUBOOT_DOWNGRADE_PREVENTION` option. Downgrade prevention is only
+available when the overwrite-based image update strategy is used
+(i.e. `MCUBOOT_OVERWRITE_ONLY` is set).
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ matrix:
       env: MULTI_FEATURES="sig-ecdsa enc-kw validate-primary-slot" TEST=sim
     - os: linux
       env: MULTI_FEATURES="sig-rsa validate-primary-slot overwrite-only large-write,sig-ecdsa enc-ec256 validate-primary-slot" TEST=sim
+    - os: linux
+      env: MULTI_FEATURES="sig-rsa validate-primary-slot overwrite-only downgrade-prevention" TEST=sim
 
     - os: linux
       language: go
--- a/boot/bootutil/include/bootutil/caps.h
+++ b/boot/bootutil/include/bootutil/caps.h
@@ -44,6 +44,7 @@ uint32_t bootutil_get_caps(void);
 #define BOOTUTIL_CAP_ED25519                (1<<9)
 #define BOOTUTIL_CAP_ENC_EC256              (1<<10)
 #define BOOTUTIL_CAP_SWAP_USING_MOVE        (1<<11)
+#define BOOTUTIL_CAP_DOWNGRADE_PREVENTION   (1<<12)
 
 /*
  * Query the number of images this bootloader is configured for.  This
--- a/boot/bootutil/src/caps.c
+++ b/boot/bootutil/src/caps.c
@@ -57,6 +57,9 @@ uint32_t bootutil_get_caps(void)
 #if defined(MCUBOOT_VALIDATE_PRIMARY_SLOT)
     res |= BOOTUTIL_CAP_VALIDATE_PRIMARY_SLOT;
 #endif
+#if defined(MCUBOOT_DOWNGRADE_PREVENTION)
+    res |= BOOTUTIL_CAP_DOWNGRADE_PREVENTION;
+#endif
 
     return res;
 }
--- a/sim/Cargo.toml
+++ b/sim/Cargo.toml
@@ -20,6 +20,7 @@ enc-ec256 = ["mcuboot-sys/enc-ec256"]
 bootstrap = ["mcuboot-sys/bootstrap"]
 multiimage = ["mcuboot-sys/multiimage"]
 large-write = []
+downgrade-prevention = ["mcuboot-sys/downgrade-prevention"]
 
 [dependencies]
 byteorder = "1.3"
--- a/sim/mcuboot-sys/Cargo.toml
+++ b/sim/mcuboot-sys/Cargo.toml
@@ -47,6 +47,9 @@ bootstrap = []
 # Support multiple images (currently 2 instead of 1).
 multiimage = []
 
+# Check (in software) against version downgrades.
+downgrade-prevention = []
+
 [build-dependencies]
 cc = "1.0.25"
 
--- a/sim/mcuboot-sys/build.rs
+++ b/sim/mcuboot-sys/build.rs
@@ -22,6 +22,7 @@ fn main() {
     let enc_ec256 = env::var("CARGO_FEATURE_ENC_EC256").is_ok();
     let bootstrap = env::var("CARGO_FEATURE_BOOTSTRAP").is_ok();
     let multiimage = env::var("CARGO_FEATURE_MULTIIMAGE").is_ok();
+    let downgrade_prevention = env::var("CARGO_FEATURE_DOWNGRADE_PREVENTION").is_ok();
 
     let mut conf = cc::Build::new();
     conf.define("__BOOTSIM__", None);
@@ -31,12 +32,20 @@ fn main() {
     conf.define("MCUBOOT_MAX_IMG_SECTORS", Some("128"));
     conf.define("MCUBOOT_IMAGE_NUMBER", Some(if multiimage { "2" } else { "1" }));
 
+    if downgrade_prevention && !overwrite_only {
+        panic!("Downgrade prevention requires overwrite only");
+    }
+
     if bootstrap {
         conf.define("MCUBOOT_BOOTSTRAP", None);
     }
 
     if validate_primary_slot {
         conf.define("MCUBOOT_VALIDATE_PRIMARY_SLOT", None);
+    }
+
+    if downgrade_prevention {
+        conf.define("MCUBOOT_DOWNGRADE_PREVENTION", None);
     }
 
     // Currently no more than one sig type can be used simultaneously.
--- a/sim/src/caps.rs
+++ b/sim/src/caps.rs
@@ -16,6 +16,7 @@ pub enum Caps {
     Ed25519              = (1 << 9),
     EncEc256             = (1 << 10),
     SwapUsingMove        = (1 << 11),
+    DowngradePrevention  = (1 << 12),
 }
 
 impl Caps {
--- a/sim/src/depends.rs
+++ b/sim/src/depends.rs
@@ -18,11 +18,28 @@ pub trait Depender {
 /// A boring image is used when we aren't testing dependencies.  There will
 /// be meaningful version numbers.  The size field is the image number we
 /// are.
-pub struct BoringDep(pub usize);
+pub struct BoringDep {
+    number: usize,
+    test: DepTest,
+}
+
+impl BoringDep {
+    pub fn new(number: usize, test: &DepTest) -> BoringDep {
+        BoringDep {
+            number: number,
+            test: test.clone(),
+        }
+    }
+}
 
 impl Depender for BoringDep {
     fn my_version(&self, _offset: usize, slot: usize) -> ImageVersion {
-        ImageVersion::new_synthetic(self.0 as u8, slot as u8, 0)
+        let slot = if self.test.downgrade {
+            1 - slot
+        } else {
+            slot
+        };
+        ImageVersion::new_synthetic(self.number as u8, slot as u8, 0)
     }
 
     fn my_deps(&self, _offset: usize, _slot: usize) -> Vec<ImageVersion> {
@@ -44,6 +61,10 @@ pub struct DepTest {
 
     /// What is the expected outcome of the upgrade.
     pub upgrades: [UpgradeInfo; 2],
+
+    /// Should this be considered a downgrade (cause the version number to
+    /// decrease).
+    pub downgrade: bool,
 }
 
 /// Describes the various types of dependency information that can be
@@ -77,6 +98,15 @@ pub enum UpgradeInfo {
 pub static NO_DEPS: DepTest = DepTest {
     depends: [DepType::Nothing, DepType::Nothing],
     upgrades: [UpgradeInfo::Upgraded, UpgradeInfo::Upgraded],
+    downgrade: false,
+};
+
+/// A "test" with no dependency information, and the images marked as a
+/// downgrade.
+pub static REV_DEPS: DepTest = DepTest {
+    depends: [DepType::Nothing, DepType::Nothing],
+    upgrades: [UpgradeInfo::Held, UpgradeInfo::Held],
+    downgrade: true,
 };
 
 /// A PairDep describes the dependencies between two pairs.
@@ -102,6 +132,11 @@ impl PairDep {
 
 impl Depender for PairDep {
     fn my_version(&self, _offset: usize, slot: usize) -> ImageVersion {
+        let slot = if self.test.downgrade {
+            1 - slot
+        } else {
+            slot
+        };
         ImageVersion::new_synthetic(self.number as u8, slot as u8, 0)
     }
 
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -39,6 +39,7 @@ use crate::depends::{
     Depender,
     DepTest,
     DepType,
+    NO_DEPS,
     PairDep,
     UpgradeInfo,
 };
@@ -171,7 +172,7 @@ impl ImagesBuilder {
             let dep: Box<dyn Depender> = if num_images > 1 {
                 Box::new(PairDep::new(num_images, image_num, deps))
             } else {
-                Box::new(BoringDep(image_num))
+                Box::new(BoringDep::new(image_num, deps))
             };
             let primaries = install_image(&mut flash, &slots[0], 42784, &*dep, false);
             let upgrades = match deps.depends[image_num] {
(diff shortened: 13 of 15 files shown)


hakonfam:
Please provide feedback on the solution in general, and also what work is required to get something like this merged (docs, simulator tests etc).

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -89,6 +89,10 @@
 #define MCUBOOT_IMAGE_NUMBER    1
 #endif
 
+#ifdef CONFIG_MCUBOOT_VERIFY_VERSION
+#define MCUBOOT_VERIFY_VERSION 1
```
oyvindronningstad: `#define MCUBOOT_DOWNGRAD_PREVENTION`?

Laczen:
@hakonfam, nice that you are trying to add this feature. I have only one concern: What will stop a malicious user from updating the image version of an older image?

hakonfam:
> @hakonfam, nice that you are trying to add this feature. I have only one concern: What will stop a malicious user from updating the image version of an older image?

I assume that the signature spans the version in the metadata as well. If this is the case, then malicious user can not alter the version  (the signature check will fail). If this is **not** the case, then this must be fixed ASAP.

Laczen:
> @hakonfam, nice that you are trying to add this feature. I have only one concern: What will stop a malicious user from updating the image version of an older image?
> 
> I assume that the signature spans the version in the metadata as well. If this is the case, then malicious user can not alter the version  (the signature check will fail). If this is not the case, then this must be fixed ASAP.

You are correct, the header is included in the hash calculation.

boot/bootutil/src/loader.c
```diff
@@ -965,7 +965,6 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
 }
 #endif
 
-#if (BOOT_IMAGE_NUMBER > 1)
```
utzig: To pass the simulator you need to fix the non used function warning which happens in most build configurations. Enclose this with something like:

```
#if (BOOT_IMAGE_NUMBER > 1) || (MCUBOOT_OVERWRITE_ONLY && MCUBOOT_DOWNGRADE_PREVENTION)
```

hakonfam: Should be fixed

utzig:
Upgrades in overwrite-only cannot be rolled back, unless re-uploading a new image with the previous number counts as a roll-back, so this is more properly a downgrade protection. It looks OK to me (I have not tested yet).

hakonfam:
> Upgrades in overwrite-only cannot be rolled back, unless re-uploading a new image with the previous number counts as a roll-back, so this is more properly a downgrade protection. It looks OK to me (I have not tested yet).

Note that I have not tested this sufficiently, I just want to hear the opinion of the community before I go on making the final soliution, as these kind of things historically are subject to a lot of discussions :)

Laczen:
Maybe something to add to the discussion: Should downgrade protection be disabled for unconfirmed images ? And something more: should images in overwrite_only be considered as confirmed, as there is no rollback option?

hakonfam:
> Should downgrade protection be disabled for unconfirmed images ? 
I don't see any reason for this. 

Any higher level protocol can itself check the current version by reading the mcuboot metadata before accepting an upgrade to be stored in the secondary slot. 

Also, what would be gained from disabling it? To disable it for unconfirmed images would require special handling, and unless there is a gain from disabling it, it would be better to have the same handling for both confirmed and unconfirmed images.

> should images in overwrite_only be considered as confirmed, as there is no rollback option?

I don't see how this relates to the issue of downgrade prevention.

Laczen:
> > Should downgrade protection be disabled for unconfirmed images ?
> > I don't see any reason for this.
> 
> Any higher level protocol can itself check the current version by reading the mcuboot metadata before accepting an upgrade to be stored in the secondary slot.
> 
> Also, what would be gained from disabling it? To disable it for unconfirmed images would require special handling, and unless there is a gain from disabling it, it would be better to have the same handling for both confirmed and unconfirmed images.
> 

When an image is not confirmed it is some kind of test image (regardless of the version). What if the test image fails and you have no image with a bigger version ? You would get stuck, so maybe special handling is required.

> > should images in overwrite_only be considered as confirmed, as there is no rollback option?
> 
> I don't see how this relates to the issue of downgrade prevention.

It is not related (yet).

boot/zephyr/Kconfig
```diff
@@ -346,4 +346,12 @@ config UPDATEABLE_IMAGE_NUMBER
 	help
 	  Enables support of multi image update.
 
+config MCUBOOT_DOWNGRADE_PREVENTION
+	bool "Downgrade prevention"
+	depends on BOOT_UPGRADE_ONLY
```
d3zd3z: Can you explain further why this depends on upgrade only? We can check before starting an upgrade. There are plenty of other issues with a malicious user manipulating swap, but that would be caught at the end, and even in that case, trying to prevent an upgrade is at least better than not doing it at all. Any thoughts on how this could be used with swap?

To get this in, we should have some simulation tests for this functionality.

We should also think about naming here, to make sure it doesn't end up too confusing. There is another feature that will be coming from TFM that adds support for a hardware anti-rollback counter. That will protect against even an attacker that can replace the primary image. But, not all devices have this counter, so we definitely want this software version.

hakonfam: I assume the following (please let me know if any of these are incorrect):
1. The only reason to perform a swap instead of an overwrite is to allow rollback.
2. Whether or not the rollback is performed depends solely on data found in the image trailers, assuming that the image hashes are correct.
3. The image trailers are not protected from modification in any way. (*)
4. Given 1, 2 and 3 it is clear that the application booted by MCUboot can trigger a swap operation by modifying the contents of the image trailers.
5. Given 4, it is not possible for MCUBoot to distinguish between if the image trailers has been written by MCUBoot or a malicious application.

Because of 1, MCUBoot with downgrade prevention has to allow a "REVERT" operation to be performed with a lower version number (otherwise there is no point in supporting swapping).
Because of 4 and 5 it is possible to break the downgrade prevention by modifying the image trailers.

(*)
If primary and secondary is open, simply write 0xff to image OK to bypass downgrade prevention
If primary is locked and secondary is open during first boot -> downgrade prevention can be bypassed by downloading an old image to secondary, and not writing IMAGE-OK, which will trigger the old image to be swapped in.
The solution to this problem is to lock the secondary slot during TEST boot, and then lock primary slot in subsequent boots.
Another solution to this problem is (as you point out) monotonic version counters (anti rollback counter).

Regarding naming confusion, these two features provide the same functionality, so one would typically choose one of them.
Since the term "rollback" is used in MCUBoot to describe the two step upgrade process, I prefer to call this feature "downgrade prevention" in this context.

In the future we could have an option to select between the two variants (SW or HW based).

nvlsianpu: Thanks for elaboration. Looks like downgrade prevention witch swap make sense only if we can trust to image trailers. Other actors here are application images. May for some reason/some case the image be trusted (as it has signature verified)?

hakonfam: > Looks like downgrade prevention witch swap make sense only if we can trust to image trailers.

Note that downgrade prevention with swapping is possible if we implement the scheme proposed above, where you:
1. During TEST boot of newly swapped image, lock the secondary slot.
2. During second boot after image has self-verified (IMAGE-OK), lock the primary slot

>  Other actors here are application images. May for some reason/some case the image be trusted (as it has signature verified)?

No, you can not trust the image. 

Firstly there is no guarantee that the application has any level of flash-write protection. So bugs in the application could allow modifications in run time, and these changes would not be discovered until after reboot, which is too late.

Secondly, some attacks do not modify the firmware, such as [return oriented programming](https://en.wikipedia.org/wiki/Return-oriented_programming)

Laczen:
@hakonfam, thanks for the elaboration. Can this be summarized as:

To get a real downgrade protection we:
a. Cannot support a swap mode (your reasoning is completely logical).
b. Even in overwrite mode test images cannot be replaced by older ok images.

These seem to be very limiting on mcuboot features.

hakonfam:
> b. Even in overwrite mode test images cannot be replaced by older ok images.

AFAICT the concept of "test image"  is only valid when performing swaps. If you don't do a swap, you  simply overwrite the existing application. Hence without swap, there is no "test image" as the update operation can not be reverted in any way.

Note that adding flash protection support of the primary and secondary slot as described in a different comment can give downgrade prevention with swapping.

Laczen:
> > b. Even in overwrite mode test images cannot be replaced by older ok images.
> 
> AFAICT the concept of "test image" is only valid when performing swaps. If you don't do a swap, you simply overwrite the existing application. Hence without swap, there is no "test image" as the update operation can not be reverted in any way.
>
OK, so no support to reload an older version in any way.

> Note that adding flash protection support of the primary and secondary slot as described in a different comment can give downgrade prevention with swapping.
>
I don't think this kind of flash protection would be easy/safe to implement.

hakonfam:
> I don't think this kind of flash protection would be easy/safe to implement.

I would say that the job of enforcing the flash protection actually is easy (at least on the nordic HW, and I assume that other vendors have similar features). 

When setting the flash protection through the "flash protection" peripheral, these configurations can only be un-set with a reboot. This allows MCUBoot to effectively lock down any part of flash which is not already configured, and know that the application can not mess up the configuration.

This blocks the application from modifying its own flash partition, but this sounds logical to me, as file systems etc. are typically in a different partition.

However ensuring that the protection enforced by mcuboot is correct in the different states (first non-confirmed boot vs second confirmed boot) would require some work I agree.

Laczen:
@hakonfam, I am thinking about a different scheme for another bootloader, maybe this can trigger ideas: what if the version check is added as a dependency (the image in the image-1 is only valid for versions xxx until yyy in image-0), then there is no more need to include the header in the hash and the header version value can be used as the placeholder for the anti-rollback counter. The header is updated in the upload slot before the swapping is done.

boot/zephyr/Kconfig
nvlsianpu: > Note that adding flash protection support of the primary and secondary slot as described in a different comment can give downgrade prevention with swapping.

That's good. Such protection mechanism sounds like a another patch objective.

nvlsianpu:
:heavy_check_mark: I'm fine with this proposal.

Laczen:
After studying dependencies in mcuboot the above proposal would not work, mcuboot dependencies only allow to specify a minimum required version.

But maybe it is possible to add this also to the swap method: what if the version check is not done as long as the copy_done flag is "set", in that case it is known that the old image is still the original one.

hakonfam:
> But maybe it is possible to add this also to the swap method: what if the version check is not done as long as the copy_done flag is "set", in that case it is known that the old image is still the original one.

TL;DR: It won't work since the application can modify the `copy_done` flag.

Remember that the application has to be able to modify the secondary slot (at least after first boot as specified in an earlier comment). Hence, the application can also modify the `copy_done` flag (and any other image trailer data). This implies that MCUBoot can not perform unsafe actions based on the information, such as skipping version check.

boot/bootutil/src/loader.c
```diff
@@ -965,7 +965,8 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
...
+#if (BOOT_IMAGE_NUMBER > 1) || \
+    (MCUBOOT_OVERWRITE_ONLY && MCUBOOT_DOWNGRADE_PREVENTION)
```
d3zd3z: This should be:
```
#if (BOOT_IMAGE_NUMBER > 1) || \
    (defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION))
```
as these are based on whether or not they are defined, not what value they are defined to.

boot/bootutil/src/loader.c
```diff
@@ -1583,6 +1586,33 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
         /* Set the previously determined swap type */
         bs.swap_type = BOOT_SWAP_TYPE(state);
 
+#if MCUBOOT_OVERWRITE_ONLY && MCUBOOT_DOWNGRADE_PREVENTION
```
d3zd3z: Likewise here `#if defined(...) && defined(...)`.

boot/zephyr/Kconfig
d3zd3z: This catches the misconfiguration in Zephyr, but as I just discovered, the code seems perfectly happy to build without overwrite only, but with downgrade protection enabled. I think this configuration should cause a compilation error, rather than just failing silently.

Laczen:
> > But maybe it is possible to add this also to the swap method: what if the version check is not done as long as the copy_done flag is "set", in that case it is known that the old image is still the original one.
> 
> TL;DR: It won't work since the application can modify the `copy_done` flag.
> 
> Remember that the application has to be able to modify the secondary slot (at least after first boot as specified in an earlier comment). Hence, the application can also modify the `copy_done` flag (and any other image trailer data). This implies that MCUBoot can not perform unsafe actions based on the information, such as skipping version check.

Agreed.

Maybe there is another way to protect against a stuck situation for test (unconfirmed) images: what if unconfirmed images are limited to only upgrade images that allow reverting based on image versions.

d3zd3z:
> Maybe there is another way to protect against a stuck situation for test (unconfirmed) images: what if unconfirmed images are limited to only upgrade images that allow reverting based on image versions.

I think most people that use test images will want to use it on every image, which kind of conflicts with the rollback protection. A good compromise would be to somehow only allow rollback to the image that was previously there, but agreeing with the above, I don't see all that good of a way to allow that. If we had a device private key, information could be written that would only be accessible to the bootloader, but this isn't common in a lot of devices.

I think a similar situation will happen with the hardware rollback counter. When that is incremented, there will still be boundaries where it will be impossible to revert to.

Two ways to handle this. If the image boots and can do upgrades, but just isn't "usable", we could rollback by deploying a new image based on the old code, but with a newer version. But, I don't think most boot failures are going to be workable with that.

One question I have, how common are reverts? The only circumstance I can see is where an image is deployed, but due to hardware variants, doesn't boot on some subset of devices. If it didn't boot at all, it should get discovered in testing, and signatures and such will protect against corrupt images, or partial updates.

d3zd3z:
I have written some simulator code to test this code. I still need to track down a write of unerased data that seems to happen when this feature is enabled.

boot/bootutil/src/loader.c
```diff
@@ -1583,6 +1586,33 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
+                const struct flash_area *fap;
+
+                rc = flash_area_open(
+                        FLASH_AREA_IMAGE_SECONDARY(BOOT_CURR_IMG(state)),
+                        &fap);
+                if (rc != 0) {
+                        BOOT_LOG_ERR("Unable to open flash area");
+                        goto out;
+                }
+                rc = boot_write_swap_info(fap,
```
d3zd3z: This isn't going to reliably work. In fact, this causes an error in the sim trying to write to data that isn't erased. I think we will just need to detect this each time we boot, and avoid doing the upgrade. Trying to write to the status like this won't work.

boot/bootutil/src/loader.c
```diff
@@ -1583,6 +1586,33 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 
+#if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
+        /* Check if version of secondary is sufficient */
+        rc = boot_is_version_sufficient(
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
+        if (rc != 0) {
+                const struct flash_area *fap;
+
+                rc = flash_area_open(
```
d3zd3z: This open is leaked (but see below, it won't really work anyway).

boot/bootutil/src/loader.c
hakonfam: I removed the entire "update the boot state" logic. AFAICT it is only used when resuming an interrupted swap operation (and this only works without swapping), so it is not needed.

boot/bootutil/src/loader.c
hakonfam: I removed the entire "update the boot state" logic. AFAICT it is only used when resuming an interrupted swap operation (and this only works without swapping), so it is not needed.

hakonfam:
I added a section in design.md. Not sure if this is really needed, input is welcome.

Laczen:
> > Maybe there is another way to protect against a stuck situation for test (unconfirmed) images: what if unconfirmed images are limited to only upgrade images that allow reverting based on image versions.
> 
> I think most people that use test images will want to use it on every image, which kind of conflicts with the rollback protection. A good compromise would be to somehow only allow rollback to the image that was previously there, but agreeing with the above, I don't see all that good of a way to allow that. If we had a device private key, information could be written that would only be accessible to the bootloader, but this isn't common in a lot of devices.
> 
> I think a similar situation will happen with the hardware rollback counter. When that is incremented, there will still be boundaries where it will be impossible to revert to.
> 
> Two ways to handle this. If the image boots and can do upgrades, but just isn't "usable", we could rollback by deploying a new image based on the old code, but with a newer version. But, I don't think most boot failures are going to be workable with that.
> 
> One question I have, how common are reverts? The only circumstance I can see is where an image is deployed, but due to hardware variants, doesn't boot on some subset of devices. If it didn't boot at all, it should get discovered in testing, and signatures and such will protect against corrupt images, or partial updates.

@d3zd3z, IIRC the origin of mcuboot is in devices that are deployed massively (>100k) and cover multiple hardware revisions, the test images would be deployed to a limited set of these devices (e.g. 10k) and based upon the result of this test upgrades it would be decided to roll out new versions. As these devices would all be remote the revert is an essential property. @utzig might have more information about this. Limiting test images to only those that could be reverted seems like a good fit with this scenario and downgrade protection.

d3zd3z:
I have a suspicion that many users of MCUboot have missed the overall design of swap, and just enable it everywhere, because somehow it feels better. I wonder if it would be helpful to have a document describing a typical scenario for deployment and updates.

boot/bootutil/src/loader.c
```diff
@@ -1583,6 +1586,16 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
+#if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
+        /* Check if version of secondary is sufficient */
+        rc = boot_is_version_sufficient(
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
+        if (rc != 0) {
+                BOOT_LOG_WRN("insufficient version");
+                goto out;
+        }
+#endif
```
davidvincze: I'm not sure if this version check is happening in the right stage (after image validation and swap type determination), and also it may requires further actions depending on the swap type.
In it's current form if we have valid images in both the primary and secondary slots, but the image in the secondary slot has an older version number the `boot_is_version_sufficient` will return with an error an the code will jump to the `out` flag. In this case the device won't boot any image.
I would suggest the move this check into the `boot_validated_swap_type` or `boot_validate_slot` functions (it should be checked against the secondary slot's image only). In this case if the version number comparison fails it would simply set the swap type to `NONE` or `FAIL`, erase the contect if the secondary slot and then boot the previously used image from the  primary slot (without performing an upgrade). Basically I would just make this check part of the image validation process.

hakonfam: I moved the check now. It looks a bit bloated, but I want to perform the version check before doing the signature check, and it was not easy to cram it in to the existing if checks without making them look weird. Feedback and proposals is, as always, welcome.

boot/bootutil/src/loader.c
```diff
@@ -965,7 +983,8 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
 }
 #endif
 
-#if (BOOT_IMAGE_NUMBER > 1)
+#if (BOOT_IMAGE_NUMBER > 1) || \
+    (defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION))
```
d3zd3z: The definition will need to be moved earlier in the file. Currently, this causes a compile fail in the simulator (and probably at least warnings on any target). It needs to be moved earlier in the file that its use above.

docs/design.md
```diff
@@ -942,3 +942,11 @@ state after dependency check.
 
 For more information on adding dependency entries to an image,
 see: [imgtool](imgtool.md).
+
+## [Downgrade Prevention](#downgrade-prevention)
+
+Downgrade prevention is a feature which enforce that the new image must have a
```
d3zd3z: "enforce" -> "enforces".

boot/bootutil/src/loader.c
```diff
@@ -541,6 +541,24 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
+        rc = boot_is_version_sufficient(
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
+        if (rc != 0) {
+            BOOT_LOG_ERR("insufficient version in secondary slot");
+            flash_area_erase(fap, 0, fap->fa_size);
+            /* Image in the secondary slot does not satisfy version requirement.
+             * Erase the image and continue booting from the primary slot.
+             */
+            rc = -1;
```
d3zd3z: Shouldn't this be returning '1' to indicate that there is no image found? I'm not sure it affects the behavior much, but at least that is how I interpret the description in the header.

boot/bootutil/src/loader.c
hakonfam: Fixed

docs/design.md
hakonfam: fixed

boot/bootutil/src/loader.c
hakonfam: Well, there is an image found, the same way in the case below, its just that its invalid. It is "bootable" however, its just that it doesn't satisfy all the criteria. IMO it should be -1 to be consistent with the signature check fail, but I can change it if you want to.

d3zd3z: I'm not sure it affects any behavior. But, there really isn't an image at that point, since it just got erased.

hakonfam: Makes sense, changing to 1, should I change the other case as well, so that the behavior is consistent? The image in secondary is also erased in that case.

hakonfam: Set both to 1. -1 is still used for the error case.

d3zd3z:
I will try to push my sim changes to test this. There is one test from that that fails, which we will need to figure out before we can merge this.

utzig:
@hakonfam Could you cherry-pick 207a961e920be52cdaa233f4cd37b4468146266b and 276e07888e0070ea4cf18a448037c2bb9c2679bc from #638 and add here? (Btw, you'll probably have to add your sign-off in those commits as well); then we can close #638.

utzig:
@hakonfam You can't remove the original signed-off-by line, you just add yours, so it will have two signed-off-by lines! ;-)

hakonfam:
> @hakonfam You can't remove the original signed-off-by line, you just add yours, so it will have two signed-off-by lines! ;-)

That was my plan, but there were no signed-off lines as far as I could tell. I will add some anyway :)

boot/bootutil/src/loader.c
```diff
@@ -541,6 +577,24 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         goto out;
     }
 
+#if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
+    if (slot != BOOT_PRIMARY_SLOT) {
+        /* Check if version of secondary slot is sufficient */
+        rc = boot_is_version_sufficient(
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
+        if (rc != 0) {
```
utzig: So this fails in the simulator. The reason is that when an upgrade is started, the primary slot is fully erased first, and if a reset happens at the "right" time it won't have an image which will end up in `boot_is_version_sufficient` failing.

The simple solution is to add a check for the erased head like this:

```
if (rc != 0 && boot_check_header_erased(state, BOOT_PRIMARY_SLOT)) {
```

This should fix the simulator (it fails on my machine so I assume it will fail on the CI as well!)

hakonfam: Added your patch, thanks

Laczen: @utzig, @hakonfam, and this is an opening to remove the downgrade protection !

hakonfam: Could you explain in detail?

Laczen: If you stop the upgrade just after the erase has started, you write an older image to the upgrade slot and this older image will be accepted.

utzig: > If you stop the upgrade just after the erase has started, you write an older image to the upgrade slot and this older image will be accepted.

That is unfortunately correct in a sense. I think this only an issue if someone is using an external SPI flash. It is an issue as well if you have JTAG/SWD enabled, but in this particular case it would be easier to just rebuild/rewrite the bootloader itself. Since you have no image running after the deletion of the primary slot's image, the only way to upload something would be through physical access, do you agree or do you have other attack vectors?

PS: `boot_serial` could also be used to upload a new older image, if enabled, but that would be the case with slot0 erased or not...

d3zd3z: There are devices that would allow write protection of the bootloader, that could allow an SWD update of the upgrade slot, but not a replacement of the bootloader.

d3zd3z:
Do you mind changing the "Attempt to test the anti-downgrade code" subject to "sim: Test software rollback protection".

boot/bootutil/src/loader.c
d3zd3z: > The simple solution is to add a check for the erased head

Isn't there still a window where the primary image is only partially erased? Should we just do the check if the header is actually valid?

As far as the window that would allow a downgrade, yes it exists, and there are other ways that the software downgrade protection only catches most cases instead of all. This is why, for example, the PSA spec requires a hardware downgrade prevention mechanism. However, what we are doing here is still much better than not doing anything at all.

Laczen: > However, what we are doing here is still much better than not doing anything at all.

Allow me to disagree on that, what is done might be a nuisance for the normal user, but a joke for users with bad intentions.

d3zd3z: It is unclear to me how this could be exploited (for devices where all flash is internal) via an attack that couldn't also just replace the primary image with an older one? The application won't boot until the upgrade is completed, so this would have to be done via something like jtag.

hakonfam:
I added a comment in the help text of the option describing that for this feature to be trustworthy, the primary slot must be flash protected.

utzig:
> I added a comment in the help text of the option describing that for this feature to be trustworthy, the primary slot must be flash protected.

It is not enough to have the primary slot flash protected; if the secondary slot can be messed with, this feature can be defeated.

hakonfam:
> It is not enough to have the primary slot flash protected; if the secondary slot can be messed with, this feature can be defeated.

How? The signature check will fail if it is tampered with.

utzig:
> > It is not enough to have the primary slot flash protected; if the secondary slot can be messed with, this feature can be defeated.
> 
> How? The signature check will fail if it is tampered with.

Install a newer image in the secondary slot; start an upgrade; allow the upgrade procedure to erase flash in the primary slot; reset after finishing the erase; install an older version into the secondary slot.

hakonfam:
> Install a newer image in the secondary slot; start an upgrade; allow the upgrade procedure to erase flash in the primary slot; reset after finishing the erase; install an older version into the secondary slot.

Understood. So you really can't provide any guarantee if an attacker is able to 
- Restart the device and write to secondary slot
OR
- Write to primary slot

Is that correct?

utzig:
> Understood. So you really can't provide any guarantee if an attacker is able to
> 
> * Restart the device and write to secondary slot
>   OR
> * Write to primary slot
> 
> Is that correct?

Kinda...

> Write to primary slot

If someone can write to the primary slot, they only need to get a properly signed image, whatever version, and write it; and delete the one in the secondary slot if necessary.

> Restart the device and write to secondary slot

It would be possible to make this secure if the primary slot can be protected, by writing version information as metadata, similarly to what swap does with its trailer; but much simpler. It would complicate the current approach for `ovewrite-only` though (obviously!).

hakonfam:
Then I propose that we clearly state in the helptext of this configuration that this feature is only intended for stopping unintentional downgrades (i.e. user error). And that when there has been a security related bugfix, the "owner" actually needs to push out two updates with the same fw only with increasing version numbers, to get rid of the variant with flawed security. 

What do you guys think?

d3zd3z:
> And that when there has been a security related bugfix, the "owner" actually needs to push out two updates with the same fw only with increasing version numbers, to get rid of the variant with flawed security.

I don't understand how doing this solves any security problem. The old versions are still all present. But, this vulnerability for downgrade I still don't think is that significant. At least it isn't any more significant than other attacks that could be made by the same attacker. I don't really see anyone applying this attack, since they could just as easily replace the primary image with an earlier one.

But, it isn't just to prevent unintentional downgrades, and will still prevent downgrades done by a user without a JTAG/SWD, for example by spoofing a network server that makes an older image appear to be an upgrade.

d3zd3z:
@tamasban I'm not sure if you're aware of this change, but this will probably go in fairly soon, and we can then integrate the SW implementation with what you are working on for HW rollback.

boot/zephyr/Kconfig
```diff
@@ -346,4 +346,15 @@ config UPDATEABLE_IMAGE_NUMBER
...
+	help
+	  Prevent downgrades by enforcing incrementing version numbers.
+	  When this option is set, any upgrade must have greater major version
+	  or greater minor version with equal major version. For this mechanism
+	  to be trustworthy, the application must enable flash protection of
+	  the primary partition. Otherwise, the downgrade prevention can be
+	  circumvented by erasing the header of the primary slot.
```
utzig: I am OK with merging this PR, but this comment does not seem correct regarding security implications. Do you plan to fix it @hakonfam?

d3zd3z:
What about something like this for the Kconfig text:

> Prevent downgrades by enforcing incrementing version numbers.
> When this option is set, any upgrade must have greater major version
> or greater minor version with equal major version. This mechanism
> only protects against some attacks against version downgrades (for
> example, a JTAG could be used to write an older version).

boot/zephyr/Kconfig
utzig: btw, https://github.com/JuulLabs-OSS/mcuboot/pull/628#issuecomment-577825787 is a good suggestion!

hakonfam:
Thanks @d3zd3z , I added your proposal.

PR #629, opened 2020-01-04
ci: add tool to check for Cargo features

utzig:
When running tests from .travis.yml, the passed in features are first checked locally for support in the current simulator. The list of supported features was manually maintained, allowing newly implemented features to be skipped, also skipping the related test (without warnings). This adds a new tool that parses and prints the list of features directly from the given Cargo.toml.


ci/get_features.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
d3zd3z: This is added as `python` in the first commit, and then fixed to `python3` in the last commit. Better to just make it correct when the script is added.

ci/get_features.py
```diff
@@ -0,0 +1,40 @@
...
+
+parser = argparse.ArgumentParser(description='Print features from a Cargo.toml.')
+parser.add_argument('infile', help='Input file to parse')
+
+args = parser.parse_args()
+if not os.path.isfile(args.infile):
+    print("File not found")
+    exit(1)
+
+config = configparser.ConfigParser()
```
d3zd3z: There is no requirement that Cargo.toml be a valid configparser file. I think this would be better done with a real toml parser, or we will possibly see strange errors if we use some of the flexibility of toml that isn't in configparser.

ci/get_features.py
utzig: Oops, completely issed this, will fix.

ci/get_features.py
utzig: Yeah, I did not want to add a new external dependency, but it's ok, I will update the code.

ci/get_features.py
utzig: Ah, squashed into the wrong commit! :-P

ci/get_features.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
```
d3zd3z: Now, this gets put back as `python`.

ci/sim_install.sh
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+pip install --user toml
```
d3zd3z: I wonder if it'd be useful to add this to the (or a) requirements file?

ci/get_features.py
utzig: I will drop this once if fixup the last commit, I was just gonna test if it works on Python2.

ci/sim_install.sh
utzig: Can do.

ci/get_features.py
d3zd3z: I guess it might be interesting to check if it works on python 2, but there isn't really much reason to do so.

utzig:
@d3zd3z Should be all fixed up now.

PR #630, opened 2020-01-06
Extra swap without scratch check; better sim coverage; remove Zephyr [EXPERIMENTAL] option

utzig:
*From commit messages*:

* boot: Add free space check for swap without scratch

Add a missing test which ensures that there is enough free sectors to perform an upgrade when using the move strategy; this basically checks that the sectors used by the trailer don't overlap the last sector required for a move up operation.

* sim: enable saving encrypted TLV for ECIES

This changes the simulator to save the encryption TLV itself instead of the unecrypted AES-128 key when doing the ECIES encryption test, to add proper test coverage of this configuration option.

* boot: zephyr: remove [EXPERIMENTAL] from config option

Swap without scratch has been tested enough and no obvious bugs (or "show stopper) bugs seem to exist; so remove [EXPERIMENTAL] and make it "stable".


utzig:
This PR is self-contained but ideally https://github.com/zephyrproject-rtos/zephyr/pull/21721 should be merged together.

PR #632, opened 2020-01-09
Compute sizes instead of magic numbers

d3zd3z:
Remove a bunch of magic numbers from the TLV code, and just compute the size directly after encoding the data.  The size has to be back-patched after the TLV is built. The protected TLV size is a little harder, since it needs to be known before the image can be hashed. This is still done dynamically instead of with a variable that we adjust.

### DIFF
--- a/sim/src/tlv.rs
+++ b/sim/src/tlv.rs
@@ -98,8 +98,6 @@ pub trait ManifestGen {
 pub struct TlvGen {
     flags: u32,
     kinds: Vec<TlvKinds>,
-    /// The total size of the payload.
-    size: u16,
     payload: Vec<u8>,
     dependencies: Vec<Dependency>,
     enc_key: Vec<u8>,
@@ -121,7 +119,6 @@ impl TlvGen {
     pub fn new_hash_only() -> TlvGen {
         TlvGen {
             kinds: vec![TlvKinds::SHA256],
-            size: 4 + 32,
             ..Default::default()
         }
     }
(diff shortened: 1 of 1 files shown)


sim/src/tlv.rs
```diff
@@ -564,6 +545,11 @@ impl ManifestGen for TlvGen {
             result.extend_from_slice(&buf);
         }
 
+        // Patch the size back into the TLV header.
+        let size = result.len() as u16;
```
utzig: I believe this should discount the preprended protected TLVs, eg:

```
let size = (result.len() - protected_tlv.len()) as u16;
```
but since it passed the CI I might be wrong, or we might be missing some coverage.

sim/src/tlv.rs
```diff
@@ -282,10 +279,11 @@ impl ManifestGen for TlvGen {
     fn make_tlv(self: Box<Self>) -> Vec<u8> {
         let mut protected_tlv: Vec<u8> = vec![];
 
-        if self.protect_size > 0 {
+        if !self.dependencies.is_empty() {
```
utzig: Maybe using `if self.protect_size() > 0` to avoid depending on `dependencies` (no pun intented!) too much?

d3zd3z: Yeah, that's probably fine. I think we'll want to abstract it a little more when we add the next protected TLV thing.

sim/src/tlv.rs
d3zd3z: At least the size of the dependencies was included before these patches, although I think it was missing the size of the protected header. I think the correct answer, though, is to make the size `result.len() - size_pos - 2`, or something a little cleaner. So, it wasn't correct before, it is now incorrect in a different way, and apparently not checked.

d3zd3z:
I pushed to the main repo accidentally. I'll be sure to delete after merging.

d3zd3z:
I've squashed the fixes, and will merge this once it is finished CI.

PR #635, opened 2020-01-15
dts: Include devicetree.h instead of generated_dts_board.h

ulfalizer:
Needed for https://github.com/zephyrproject-rtos/zephyr/pull/20757, to
avoid a warning-turned-error.


ulfalizer:
Replaces https://github.com/zephyrproject-rtos/mcuboot/pull/2.

utzig:
@ulfalizer You need to fix your sign-off-by to use your @gmail.com account or rewrite the commit using your @nordicsemi account.

ulfalizer:
> @ulfalizer You need to fix your sign-off-by to use your @gmail.com account or rewrite the commit using your @nordicsemi account.

Fixed now. Always forget when switching repos...

utzig:
> > @ulfalizer You need to fix your sign-off-by to use your @gmail.com account or rewrite the commit using your @nordicsemi account.
> 
> Fixed now. Always forget when switching repos...

It is still broken the same way!

ulfalizer:
Weird. Wonder if CI didn't retrigger.

I rebased it now as well.

utzig:
@ulfalizer The `author` is @nordic... the `committer` is still @gmail. Just signoff with your gmail as well...

ulfalizer:
Ah... hopefully fixed now.

Wonder if there's a nicer way than

    GIT_COMMITTER_NAME="Ulf Magnusson" GIT_COMMITTER_EMAIL="Ulf.Magnusson@nordicsemi.no" git commit --amend --author="Ulf Magnusson <Ulf.Magnusson@nordicsemi.no>"

to hack it for a single commit. I'm a lousy sysadmin...

utzig:
@ulfalizer Oops, you also need to rename `mcuboot-sys/csupport/generated_dts_board.h`.

nvlsianpu:
Fabio was a few second faster XD

ulfalizer:
> @ulfalizer Oops, you also need to rename `mcuboot-sys/csupport/generated_dts_board.h`.

Fixed now.

ulfalizer:
Oh wait... should rename the include guards too for consistency.

d3zd3z:
What if we conditionalize this ifdef based on `KERNEL_VERSION_MAJOR` and `KERNEL_VERSION_MINOR`?

carlescufi:
> What dependencies does this place on specific versions of Zephyr? I wonder if we should use the old name for a "while", and after a stable release or two, rename the include within MCUboot?

@d3zd3z we now tightly couple MCUBoot inside the Zephyr "distribution".  We have [our own copy](https://github.com/zephyrproject-rtos/mcuboot) of MUCBoot that is referenced in the [west manifest](https://github.com/zephyrproject-rtos/zephyr/blob/master/west.yml#L76)
Users are encouraged to use that copy, which has no downstream changes.

> Or am I misunderstanding where Zephyr is in this process? The old name should continue to work for a while, and that will allow us to support older version of Zephyr for a while.

The old name will continue to work but with warnings. The reason we are posting this PR is to avoid having downstream changes in [Zephyr's MCUBoot copy ](https://github.com/zephyrproject-rtos/mcuboot). We want to upstream the change first, then update the copy. 

If we don't make this change upstream in MCUboot users that use MCUboot:master and zephyr:master will start seeing warnings. In the past we've always assumed that the master branches on both projects always work together, but not going back in time. We've done similar chagnes in the past, and now users are anyway encouraged to use the MCUBoot that is "bundled" with Zephyr and thus locked at a particular revision.

> Otherwise, is there a define from Zephyr that would let us conditionalize on the old name of the header when using older LTS versions of Zephyr?

No, there is not.

carlescufi:
> What if we conditionalize this ifdef based on `KERNEL_VERSION_MAJOR` and `KERNEL_VERSION_MINOR`?

That won't work because this was introduced during the 2.1.99 cycle. See my comment above about versioning and now Zephyr using a fixed SHA for MCUBoot.

d3zd3z:
Seems like we should try to use the version macros in the future, when that is possible. Does this mean that the next mcuboot release won't work with the latest Zephyr LTS? I don't think I'd consider that acceptable.

utzig:
@carlescufi @nvlsianpu @d3zd3z it would probably be better to merge this after next release, which is supposed to happen soon. What do you think? (And we still have to wait the Zephyr PR to be approved/merged as well!)

nvlsianpu:
@utzig 
Patch consists only renaming file changes. I see no risk in having it into release candidate. From zephyr perspective should be merged ASAP after zephyrproject-rtos/zephyr#20757 will be merged.

PR #636, opened 2020-01-15
Fix imgtool checking of oversized app

### LINKED ISSUE #226: Detect oversized app image
Right now, if an application overlaps the image trailer, it just causes failures. Ideally, we should detect this, but there are difficulties since this information depends on the flash alignment, which isn't readily available in the host tools.

First question - can we detect this?

utzig:
This partly fixes #226 (the image generation side).

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -156,13 +156,16 @@ class Image():
     def load(self, path):
         """Load an image from a given file"""
         ext = os.path.splitext(path)[1][1:].lower()
-        if ext == INTEL_HEX_EXT:
-            ih = IntelHex(path)
-            self.payload = ih.tobinarray()
-            self.base_addr = ih.minaddr()
-        else:
-            with open(path, 'rb') as f:
-                self.payload = f.read()
+        try:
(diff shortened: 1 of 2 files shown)


utzig:
@d3zd3z I will update `docs/imgtool.py` accordingly after comments...

scripts/imgtool/image.py
```diff
@@ -360,26 +373,37 @@ def add_header(self, enckey, protected_tlv_size):
...
-                raise Exception("Invalid alignment: {}".format(write_size))
+                raise click.BadParameter("Invalid alignment: {}".format(
+                    write_size))
             m = DEFAULT_MAX_SECTORS if max_sectors is None else max_sectors
             trailer = m * 3 * write_size  # status area
             if enckey is not None:
-                trailer += 16 * 2  # encryption keys
+                if save_enctlv:
+                    try:
+                        keylen = len(enckey.get_private_bytes(minimal_enctlv))
```
utzig: This actually needs to be rounded to `align` size to avoid the bootloader possibly copying some non-aligned size (if image data overlaps the last sector containing the trailer).

PR #640, opened 2020-01-22
cmake: use zephyr base in relative path

hakonfam:
To be more robust in the placement of MCUBoot directory,
use the ZEPHYR_BASE env variable to locate
nrfxlib directory.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -60,7 +60,7 @@ assert_exists(MBEDTLS_ASN1_DIR)
 set(NRF_DIR "${MCUBOOT_DIR}/ext/nrf")
 
 if(CONFIG_BOOT_USE_NRF_CC310_BL)
-set(NRFXLIB_DIR ${MCUBOOT_DIR}/../nrfxlib)
+set(NRFXLIB_DIR $ENV{ZEPHYR_BASE}/../nrfxlib)
 assert_exists(NRFXLIB_DIR)
 # Don't include this if we are using west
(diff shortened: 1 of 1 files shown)


boot/zephyr/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ assert_exists(MBEDTLS_ASN1_DIR)
 set(NRF_DIR "${MCUBOOT_DIR}/ext/nrf")
 
 if(CONFIG_BOOT_USE_NRF_CC310_BL)
-set(NRFXLIB_DIR ${MCUBOOT_DIR}/../nrfxlib)
+set(NRFXLIB_DIR $ENV{ZEPHYR_BASE}/../nrfxlib)
```
utzig: I think this is OK, but it will break the build config for people that currently have `nrfxlib` cloned below the MCUBoot dir, if Zephyr is located somewhere else, correct? (e.g. my personal setup does not have MCUBoot and Zephyr cloned under the same hierarchy!)

hakonfam: Correct, this follows the scheme enforced by using fw-nrfconnect-nrf's west script. Not sure what approach we use for situations like this. @carlescufi - any comment?

utzig: To be clear, I am not against the change, I'm just pointing out that I will have to move `nrfxlib` around, so other users might have to as well. If `nrfxlib` is auto-pulled by `west update` then even better (still I can remove my copy cloned under MCUBoot!).

Issue #597, opened 2019-12-02
Implement HW anti-rollback counter


d3zd3z:
This is separate from #224, which is a software anti-rollback feature request.  This issue tracks bringing the TF-M anti-rollback support into mainline MCUboot.

hakonfam:
Should this be closed?

utzig:
Fixed by #642

PR #642, opened 2020-01-22
introduce HW rollback protection

davidvincze:
- Add new security counter to protected TLV area
- Add new command line option for imgtool to specify the value of this counter
- Modify the simulator to prevent the tests from failing (temporary solution)
- Make the security counter verification part of the image validation process
- Introduce security counter interface and also provide a SW only weak implementation (currently does absolutely nothing)

Related GitHub issue: #597 

Please provide feedback and ideas how it could be combined with a SW only solution (if we need it at all). Currently the verification of a new image's security counter value is mandatory.
Please see my additional comments below.

### DIFF
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -18,10 +18,11 @@
  */
 
 /*
- * Modifications are Copyright (c) 2019 Arm Limited.
+ * Modifications are Copyright (c) 2019-2020 Arm Limited.
  */
 
 #include <stddef.h>
+#include <stdint.h>
 #include <inttypes.h>
 #include <string.h>
 
@@ -30,6 +31,7 @@
 #include "bootutil/image.h"
 #include "bootutil/sha256.h"
 #include "bootutil/sign_key.h"
+#include "bootutil/security_cnt.h"
 
 #include "mcuboot_config/mcuboot_config.h"
 
@@ -200,6 +202,68 @@ bootutil_find_key(uint8_t *keyhash, uint8_t keyhash_len)
 }
 #endif
 
+#ifdef MCUBOOT_HW_ROLLBACK_PROT
+/**
+ * Reads the value of an image's security counter.
+ *
+ * @param hdr           Pointer to the image header structure.
+ * @param fap           Pointer to a description structure of the image's
+ *                      flash area.
+ * @param security_cnt  Pointer to store the security counter value.
+ *
+ * @return              0 on success; nonzero on failure.
+ */
+int32_t
+bootutil_get_img_security_cnt(struct image_header *hdr,
+                              const struct flash_area *fap,
+                              uint32_t *img_security_cnt)
+{
+    struct image_tlv_iter it;
+    uint32_t off;
+    uint16_t len;
+    int32_t rc;
+
+    if ((hdr == NULL) ||
+        (fap == NULL) ||
+        (img_security_cnt == NULL)) {
+        /* Invalid parameter. */
+        return BOOT_EBADARGS;
+    }
+
+    /* The security counter TLV is in the protected part of the TLV area. */
+    if (hdr->ih_protect_tlv_size == 0) {
+        return BOOT_EBADIMAGE;
+    }
+
+    rc = bootutil_tlv_iter_begin(&it, hdr, fap, IMAGE_TLV_SEC_CNT, true);
+    if (rc) {
+        return rc;
+    }
+
+    /* Traverse through the protected TLV area to find
+     * the security counter TLV.
+     */
+
+    rc = bootutil_tlv_iter_next(&it, &off, &len, NULL);
+    if (rc != 0) {
+        /* Security counter TLV has not been found. */
+        return -1;
+    }
+
+    if (len != sizeof(*img_security_cnt)) {
+        /* Security counter is not valid. */
+        return BOOT_EBADIMAGE;
+    }
+
+    rc = flash_area_read(fap, off, img_security_cnt, len);
+    if (rc != 0) {
+        return BOOT_EFLASH;
+    }
+
+    return 0;
+}
+#endif /* MCUBOOT_HW_ROLLBACK_PROT */
+
 /*
  * Verify the integrity of the image.
  * Return non-zero if image could not be validated/does not validate.
@@ -222,6 +286,11 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
     uint8_t buf[SIG_BUF_SIZE];
     uint8_t hash[32];
     int rc;
+#ifdef MCUBOOT_HW_ROLLBACK_PROT
+    uint32_t security_cnt = UINT32_MAX;
+    uint32_t img_security_cnt = 0;
+    int32_t security_counter_valid = 0;
+#endif
 
     rc = bootutil_img_hash(enc_state, image_index, hdr, fap, tmp_buf,
             tmp_buf_sz, hash, seed, seed_len);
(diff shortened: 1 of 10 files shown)


davidvincze:
This initial solution is coming from TF-M's mcuboot fork:
- https://developer.trustedfirmware.org/w/tf_m/design/trusted_boot/rollback_protection/
(https://git.trustedfirmware.org/trusted-firmware-m.git/tree/docs/design_documents/secure_boot_rollback_protection.rst)
- https://git.trustedfirmware.org/trusted-firmware-m.git/tree/bl2/ext/mcuboot/include/security_cnt.h

d3zd3z:
I guess I'm not so sure about just making this mandatory, since that is going to make images made before this incompatible. I'm ok introducing this as an optional feature, though. More review coming.

Note, that a change was merged that added copyrights to the sim code, so that will have to be fixed up when you rebase.

boot/bootutil/src/image_validate.c
```diff
@@ -200,6 +202,77 @@ bootutil_find_key(uint8_t *keyhash, uint8_t keyhash_len)
...
+
+    rc = bootutil_tlv_iter_begin(&it, hdr, fap, IMAGE_TLV_SEC_CNT, true);
+    if (rc) {
+        return rc;
+    }
+
+    /* Traverse through the protected TLV area to find
+     * the security counter TLV.
+     */
+    while (true) {
```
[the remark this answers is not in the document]
davidvincze: I agree, fixed.

scripts/imgtool/main.py
```diff
@@ -154,6 +154,23 @@ def validate_version(ctx, param, value):
...
+        if value == '':
+            return value
+        else:
+            try:
+                if value[:2].lower() == '0x':
+                    return int(value[2:], 16)
+                elif value[:1] == '0':
+                    return int(value, 8)
+                else:
+                    return int(value, 10)
```
utzig: This conversion code is the same as in the `convert` function inside `BasedIntParamType`. You can create a new global function `convert_based_int(value)` that does the conversion (or throws an exception) and call it here and inside `BasedIntParam:convert`.

docs/design.md
```diff
@@ -946,7 +947,30 @@ see: [imgtool](imgtool.md).
...
+
+During the software based downgrade prevention the image version numbers are
+compared. This feature is enabled with the `MCUBOOT_DOWNGRADE_PREVENTION`
+option. In this case downgrade prevention is only available when the
+overwrite-based image update strategy is used (i.e. `MCUBOOT_OVERWRITE_ONLY`
+is set).
+
+### [HW Based Downgrade Prevention](#hw-downgrade-prevention)
+
+Each signed image can contain a security counter in its protected TLV area.
```
oyvindronningstad: The software based downgrade prevention could also use this counter, to make things more consistent. Just compare it to the counter in the other slot, instead of the HW counter.

oyvindronningstad:
Do you have any example implementations of the boot_nv_security_counter_* API functions?

scripts/imgtool/main.py
d3zd3z: Please, just use `return int(value, 0)`, this functionality is already builtin to Python.

boot/zephyr/Kconfig
```diff
@@ -360,8 +361,12 @@ config UPDATEABLE_IMAGE_NUMBER
 	help
 	  Enables support of multi image update.
 
+choice
```
d3zd3z: If this is a choice, we need an option to choose neither of these. There are existing uses of MCUboot where rolling back is a required feature, and that still needs to be possible.

scripts/imgtool/main.py
```diff
@@ -154,6 +154,23 @@ def validate_version(ctx, param, value):
         raise click.BadParameter("{}".format(e))
 
 
+def validate_security_counter(ctx, param, value):
```
d3zd3z: I wonder if we should use something like `-s auto` instead of just always auto generating the security counter. If the feature is not enabled, this just wastes space in the TLV.

scripts/imgtool/image.py
```diff
@@ -246,14 +258,22 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
+        if self.security_counter is not None:
+            # TLV info header + security counter TLV
+            # Size of the security counter TLV: header ('HH') + payload ('I')
+            #                                   = 4 + 4 = 8 Bytes
+            protected_tlv_size = TLV_INFO_SIZE + TLV_SIZE + 4
+
+        if dependencies is not None:
+            if protected_tlv_size == 0:
+                # TLV info header
+                protected_tlv_size = TLV_INFO_SIZE
```
utzig: I believe you could remove this `if` block from here and add one that increments after all protected TLVs are handled, so in the future new code that is added does not need to keep copy/pasting this.

scripts/imgtool/image.py
```diff
@@ -246,14 +258,22 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
+            # Size of a Dependency TLV = Header ('HH') + Payload('IBBHI')
+            # = 4 + 12 = 16 Bytes
             dependencies_num = len(dependencies[DEP_IMAGES_KEY])
-            protected_tlv_size = (dependencies_num * 16) + TLV_INFO_SIZE
+            protected_tlv_size += (dependencies_num * 16)
```
utzig: Following my previous comments, you could add

```
if protected_tlv_size != 0:
￼                # add TLV info header
￼                protected_tlv_size += TLV_INFO_SIZE
```

here; at least I think this should work!

scripts/imgtool/image.py
```diff
@@ -246,14 +258,22 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
-        else:
-            # Size of a Dependency TLV = Header ('BBH') + Payload('IBBHI')
-            # = 16 Bytes
+        protected_tlv_size = 0
+
+        if self.security_counter is not None:
+            # TLV info header + security counter TLV
+            # Size of the security counter TLV: header ('HH') + payload ('I')
+            #                                   = 4 + 4 = 8 Bytes
+            protected_tlv_size = TLV_INFO_SIZE + TLV_SIZE + 4
```
utzig: `TLV_INFO_SIZE` should be removed if changed as suggested in other comments...

scripts/imgtool/main.py
utzig: > If the feature is not enabled, this just wastes space in the TLV.

If you don't pass `-s` no TLV is added.

boot/zephyr/Kconfig
davidvincze: I marked this choice as `optional` , so it can be completely switched off.
FYI: HW rollback protection allows for rolling back the firmware to some extent
(the security counter doesn't need to be increased with each sw release).

scripts/imgtool/main.py
davidvincze: Yes. To auto generate the security counter from the image version an empty string had to be passed, but I changed it to `auto` (really a better solution).

scripts/imgtool/image.py
```diff
@@ -61,7 +61,8 @@
         'ENCRSA2048': 0x30,
         'ENCKW128': 0x31,
         'ENCEC256': 0x32,
-        'DEPENDENCY': 0x40
+        'DEPENDENCY': 0x40,
+        'SEC_CNT': 0x50
```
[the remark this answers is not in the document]
davidvincze: Done.

scripts/imgtool/image.py
davidvincze: Yes indeed. I've already modified it locally. I have another patch that adds a new TLV to the protected area and I encountered with this problem.:)

scripts/imgtool/image.py
davidvincze: Done.

scripts/imgtool/image.py
davidvincze: Done.

docs/design.md
davidvincze: Yes, it was the original idea before the sw downgrade prevention was introduced.
However, I decided to leave this sw based solution as it is, because it's more simple and doesn't
require an extra TLV.

davidvincze:
> Out of curiosity, do you have any example implementations of the boot_nv_security_counter_* API functions?

There is one in the TF-M repository:
https://git.trustedfirmware.org/trusted-firmware-m.git/tree/bl2/src/security_cnt.c
https://git.trustedfirmware.org/trusted-firmware-m.git/tree/platform/ext/common/template/nv_counters.c

Note that this implementation is for testing purposes only and as such it has
limitations (e.g. it is not resistant to asynchronous power failures).

More information about the design: https://ci.trustedfirmware.org/job/tf-m-build-test-nightly/lastSuccessfulBuild/artifact/build-docs/tf-m_documents/install/doc/user_guide/html/docs/design_documents/secure_boot_rollback_protection.html

utzig:
@d3zd3z Just to be sure you know that @davidvincze would be able to merge this himself, right? :-P

Issue #631, opened 2020-01-08
Unable to use mcuboot with zephyr and nrf52840_pca10059 (dongle)


RIJU55:
Hi, I am trying the procedure for using **mcuboot with nrf52840_pca10059 (dongle)** as given in the documentation ( [https://docs.zephyrproject.org/latest/boards/arm/nrf52840_pca10059/doc/index.html?highlight=nrf52840_pca10059]), I have achieved up to **step 4 of flashing the zip using** `nrfutil dfu usb-serial -pkg mcuboot.zip -p /dev/ttyACM0` but next when I do the subsequent steps **Boot into MCUboot serial recovery mode** by plugging the board in with the SW1 button pressed down I **do not get** the mentioned `/dev/serial/by-id/usb-ZEPHYR_MCUBOOT_0.01-if00 `as I am using Ubuntu 19.10 even my `/dev/ttyACM0` is **missing** and hence I am unable to perform the steps which ensue

utzig:
You do not have any file(s) under `/dev/serial/by-id/`? Also check that the board was detect with `dmesg`.

RIJU55:
no I do not have any file(s) under `/dev/serial/by-id/` will try` dmesg` again with mcuboot and get back to you, when I ran the zephyr/samples/subsys/usb/cdc_acm with nordic open bootloader I got the following on dmesg
```
[ 1749.112399] cdc_acm 3-3:1.0: ttyACM0: USB ACM device
[ 1758.442850] usb 3-3: USB disconnect, device number 6
[ 1758.443211] cdc_acm 3-3:1.0: failed to set dtr/rts
```

pjpetree:
I too am seeing the same thing.  I have used macOS, Linux and Windows hosts and no USB serial port ever shows up.

Any ideas?

pjpetree:
Found the issue.  It seems that there really should be a call to `usb_enable(NULL)` in main in order to properly initialize the USB serial port on the nrf52840_pca10059.

I have a fix, should I go ahead and submit a PR?

RIJU55:
Is that the solution if I just add `usb_enable(NULL)` it should run or do you have any further changes as well

pjpetree:
With just that change, the rest of the instructions work perfectly.

nvlsianpu:
fixed by #651

PR #643, opened 2020-01-23
boot: zephyr: Call usb_enable() when CONFIG_USB enabled

pjpetree:
USB CDC ACM wasn't being enabled on nRF52840 PCA10059 (Dongle) when booting
while holding the power button.  Adding an explicit call to usb_enable(NULL)
in main() allows the USB device to properly enumerate.

This should address issue #631 

Signed-off-by: P.J. Petree <pjpetree@gmail.com>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -198,6 +198,11 @@ void main(void)
     }
 #endif
 
+#ifdef CONFIG_USB
+    rc = usb_enable(NULL);
+    __ASSERT(rc == 0, "Error initializing USB.\n");
+#endif
+
 #ifdef CONFIG_MCUBOOT_SERIAL
 
     struct device *detect_port;


boot/zephyr/main.c
```diff
@@ -198,6 +198,11 @@ void main(void)
     }
 #endif
 
+#ifdef CONFIG_USB
+    rc = usb_enable(NULL);
+    __ASSERT(rc == 0, "Error initializing USB.\n");
+#endif
```
utzig: Seems that `usb_enable(NULL)` now needs be explicitly called. I would suggest you add this code in:

https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/zephyr/serial_adapter.c#L196

You probably don't need to test for `CONFIG_USB` there, and it will only be enabled if required.

nvlsianpu: Good recommendation.

emob-nordic: Yes, USB enable must be called explicitly. This was introduced [here](https://github.com/zephyrproject-rtos/zephyr/pull/20375).  You can find a note about that in [2.2 release notes](https://github.com/zephyrproject-rtos/zephyr/blob/master/doc/releases/release-notes-2.2.rst).
I recommend to call
```
usb_enable(NULL);
```
after 
```
device_get_binding()
```

nvlsianpu:
``` C
uart_dev = device_get_binding(CONFIG_USB_CDC_ACM_DEVICE_NAME "_0");
if (uart_dev) {
	rc = usb_enable(NULL);
	if (rc) {
		return -1;
	}
}
```

nvlsianpu: @emob-nordic thanks for your help.

pjpetree: Wouldn't this also be required in the `wait_for_usb_dfu()` case as well?  Moving to `serial_adapter.c` wouldn't solve that case ...

nvlsianpu:
superseded by #651

@pjpetree Thanks for pointing the bug and solution, just need to have fix now.

PR #644, opened 2020-01-24
mynewt: remove version.yml and references

utzig:
With the merge of https://github.com/apache/mynewt-newt/pull/365 `newt` does not use `version.yml` files anymore, so remove the file and references to updating it in the release process.

Testing a MCUBoot release now depends on `newt` `master` or `> 1.7.0` (to be released).


ccollins476ad:
It might make sense to keep `version.yml` around for a little while for backwards compatibility.  At least until the next official newt release.

I don't think there are any serious problems with removing this file, even for older versions of newt.  However, old newt will print a warning about the missing file!

d3zd3z:
I'll plan on merging this after the post 1.8.0 mynewt release.

utzig:
Mynewt 1.8.0 was released on April 9, 2020.

PR #645, opened 2020-01-24
zephyr: use move swap instead of scratch

hakonfam:


utzig:
I think this is not going to work with many STM32 devices, and if the secondary partition in the nrf52x boards is configured to be located in the SPI flash, among other possible targets, because the implementation is currently dependent on all sectors of primary/secondary slot being the same size. Wouldn't it be better to add overlays for some boards?

hakonfam:
@nvlsianpu @mbolivar  should we keep this patch downstream then?

utzig:
> @nvlsianpu @mbolivar should we keep this patch downstream then?

Ain't it possible to add an overlay for nrf boads only? Also, if someone changes the secondary slot to be located in the SPI flash it won't fail silently, it should refuse to upgrade with a clear message stating the slot incompatibility, so at least then the person can switch the upgrade method! :-P

hakonfam:
I updated so that the default in Kconfig is 'y' for nRF devices only.

When writing firmware images one will use a special library for this in zephyr context, and this library does not support other flash devices (e.g. SPI), so only when this support is added do you have to ensure that this option is not set.

utzig:
When I run my usual build inside `boot/zephyr/build` for the `nrf52840_pca10056` I still get swap-move disabled. It is disabled in `prj.conf`, doesn't that always override the defaults in `Kconfig` or do I need to build some other way?

hakonfam:
> When I run my usual build inside `boot/zephyr/build` for the `nrf52840_pca10056` I still get swap-move disabled. It is disabled in `prj.conf`, doesn't that always override the defaults in `Kconfig` or do I need to build some other way?

Fixed, I'm not on my regular PC, so not able to do proper checks at the moment :D

utzig:
> Fixed, I'm not on my regular PC, so not able to do proper checks at the moment :D

No problem, it worked! I have approved but I'll leave the final approval and merge to @nvlsianpu!

nvlsianpu:
> When I run my usual build inside boot/zephyr/build for the nrf52840_pca10056 I still get swap-move disabled. It is disabled in prj.conf, doesn't that always override the defaults in Kconfig or do I need to build some other way?

Yes, values from config_files from have priority on default from Kconfig.
It is also possible to divide the config_file into more files - the general one and per board one (actually this is done in mcuboot, but without care about config duplicates, but that NoP, until duplicates are in line). For zephyr CI also can use config-override file, but i doubt whether this works for regular project, moreover override project configuration would be a bad development practice.

nvlsianpu:
@hakonfam zephyr fork was synchronized.

PR #647, opened 2020-01-27
Fix out-of-bounds buffer access detected by Coverity

utzig:
This fixes an issue where a memset of length 0 was being called passing an out-of-bounds index in a buffer.


d3zd3z:
So, I believe this is an issue with Coverity understanding two edge-case aspects of C, but not understanding how they interact.

1. It is valid to take the address of one past the end of an array. This is considered a valid pointer.
2. `memset` can have a length of zero, as long as the pointer is valid.

Adding this `if` would only serve to add additional code for the check, which would be redundant with memset already working with a zero pointer.

Therefore, this should just be marked as "false positive" in Coverity, and this PR closed.

PR #648, opened 2020-01-30
boot: zephyr: move to non-deprecated GPIO flags

pabigot:
GPIO_DIR_IN has been replaced by GPIO_INPUT, GPIO_PUD_PULL_UP by
GPIO_PULL_UP, and gpio_pin_read() by gpio_pin_get_raw().  Update the
code to use the preferred API if it available.  This avoids
deprecation warnings in the build.

Signed-off-by: Peter Bigot <peter.bigot@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -201,18 +201,32 @@ void main(void)
 #ifdef CONFIG_MCUBOOT_SERIAL
 
     struct device *detect_port;
-    u32_t detect_value;
+    u32_t detect_value = !CONFIG_BOOT_SERIAL_DETECT_PIN_VAL;
 
     detect_port = device_get_binding(CONFIG_BOOT_SERIAL_DETECT_PORT);
     __ASSERT(detect_port, "Error: Bad port for boot serial detection.\n");
 
+    /* The default presence value is 0 which would normally be
+     * active-low, but historically the raw value was checked so we'll
+     * use the raw interface.
+     */
     rc = gpio_pin_configure(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN,
-                            GPIO_DIR_IN | GPIO_PUD_PULL_UP);
+#ifdef GPIO_INPUT
+                            GPIO_INPUT | GPIO_PULL_UP
+#else
+                            GPIO_DIR_IN | GPIO_PUD_PULL_UP
+#endif
+	    );
     __ASSERT(rc == 0, "Error of boot detect pin initialization.\n");
 
+#ifdef GPIO_INPUT
+    rc = gpio_pin_get_raw(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN);
+    detect_value = rc;
+#else
     rc = gpio_pin_read(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN,
                        &detect_value);
-    __ASSERT(rc == 0, "Error of the reading the detect pin.\n");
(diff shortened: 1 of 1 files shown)


nvlsianpu:
> Update the code to use the preferred API if it available. 

As I understand API is available since certain zephyr SHA, right?

utzig:
> > Update the code to use the preferred API if it available.
> 
> As I understand API is available since certain zephyr SHA, right?

I've found some PRs adding this eg: https://github.com/zephyrproject-rtos/zephyr/pull/22279

But since #ifdefery is used here it should work on Zephyr now and after the PRs are merged.

pabigot:
> > Update the code to use the preferred API if it available.
> 
> As I understand API is available since certain zephyr SHA, right?

The new API has been in development since April of last year in a branch.  We're finally at the point where the branch is expected to be merged, and part of the preparation is making use of the old API produce build errors that are enabled in Zephyr's CI.

The patch has been tested both with the current API and the new one, and it's intended to work with both.  Though just in case some last minute change occurs, it would be reasonable to wait until next week for further testing to complete before committing it.

nvlsianpu:
I rather prefer to have master to master compatibility than keeping backward compatibility, especially as we have mcuboot reference in the zephy repo (it exist for ensure compatibility).

Want to no introduce much ifdefry.

We used to keep this since the zephyr port beginning.

nvlsianpu:
Is it needed for zephyr 2.2?

pabigot:
> Is it needed for zephyr 2.2?

It is, as long as topic-gpio is going into 2.2, which is still likely.  Zephyr CI fails because of the use of the deprecated API.

I don't think Zephyr's decision to change the GPIO API should break use of updated mcuboot with older Zephyr releases, so the ifdef checking is necessary for backwards compatibility.

utzig:
> I don't think Zephyr's decision to change the GPIO API should break use of updated mcuboot with older Zephyr releases, so the ifdef checking is necessary for backwards compatibility.

Also the #ifdefery can be removed after the APIs have been there for some time.

nvlsianpu:
> Also the #ifdefery can be removed after the APIs have been there for some time.
Agree, I recommend to have then a comment for that near #ifdefry.

@pabigot Should I merge this in connection to merge of zephyr PR I ques?

pabigot:
I'm not quite sure what all is going on here and who expects to do what, but:
* I will update this to add a comment stating that the conditions can be removed once mcuboot decides not to support for Zephyr before 2.2.
* I will test that in Zephyr on both master and topic-gpio.
* When I report back here that it's good to go, this can be merged in this repository.
* Then @nvlsianpu  can update the Zephyr fork to mirror this repository.
* Then I will update zephyrproject-rtos/zephyr#22347 to reference the commit in the zephyr fork.
* Then zephyrproject-rtos/zephyr#22347 can be merged
* Then the topic branch can be rebased on zephyr master

I will update here when I believe it's time to merge this to this repository.

nvlsianpu:
^^ @pabigot  OK, thanks for precise process description. OK, I will follow that.

pabigot:
This patch works against both master and topic-gpio in Zephyr.  @nvlsianpu

nvlsianpu:
zephyr fork synchronized, please upgrade the reference in your convenience.

nvlsianpu:
@pabigot ^^ https://github.com/zephyrproject-rtos/zephyr/pull/22369

PR #650, opened 2020-02-03
Update license headers more appropriate for non-Apache project

d3zd3z:
I'd like to start a discussion as to how we can more clearly format copyright and licenses in our source files.  This is modeled after what many modern projects are moving to, explicit copyrights, and an SPDX line that identifies the license.

The problem with the existing format is that there are no copyright statements from any party. This means some additional parties will have to add awkward lines like: "Modifications are copyright...".

### DIFF
--- a/boot/bootutil/include/bootutil/bench.h
+++ b/boot/bootutil/include/bootutil/bench.h
@@ -1,4 +1,6 @@
 /*
+ * SPDX-License-Identifier: Apache-2.0
+ *
  * Copyright (c) 2019 Linaro Limited
  *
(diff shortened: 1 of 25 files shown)


davidvincze:
May I ask what is the status of this PR?

d3zd3z:
> May I ask what is the status of this PR?

I haven't heard from anyone. Maybe the answer is to just approve and merge it, and fix any complaints that come up later. I'll need to rebase this, and push and update. Feel free to approve if you think the Arm aspects are ok.

PR #659, opened 2020-02-11
Revert "dts: Include devicetree.h instead of generated_dts_board.h"

d3zd3z:
This reverts commit e96b6873bd8a64185796ac47b941f56dd29b6e3f.

Renaming this will break the ability to compile against the latest
release of Zephyr (v2.1.0 at the time).  This patch can be reapplied
after one or more additional Zephyr releases.

We need to be careful to not break compatible with the latest release version of Zephyr. Zephyr has a 2-release policy for breaking changes, so we need to wait for a full release cycle of Zephyr before making this change.


d3zd3z:
@ulfalizer I think we need to hold off on this change until another Zephyr release cycle. Ideally, Zephyr would go through one release without a warning, but until then, we kind of need to just put up with the warning.

utzig:
Hadn't we agreed that only the tip of `master` needs to be supported, while the responsibility for supporting other releases is deferred to https://github.com/zephyrproject-rtos/mcuboot ?

nvlsianpu:
Zephyr build system generates generates version.h header which contains usefully versions description e.g.:
``` C
#ifndef _KERNEL_VERSION_H_
#define _KERNEL_VERSION_H_

/*  values come from cmake/version.cmake */

#define ZEPHYR_VERSION_CODE 131427
#define ZEPHYR_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))

#define KERNELVERSION          0x2016300
#define KERNEL_VERSION_NUMBER  0x20163
#define KERNEL_VERSION_MAJOR   2
#define KERNEL_VERSION_MINOR   1
#define KERNEL_PATCHLEVEL      99
#define KERNEL_VERSION_STRING  "2.1.99"

#endif /* _KERNEL_VERSION_H_ */
```

utzig:
Otherwise we could start a `branch-1.5`, revert this commit there (but not on `master`) and support the latest "stable" (or what Zephyr calls it) release with our latest release and Zephyr `master` as well if possible. For mcuboot `master` would be better to just support Zephyr `master`. What do you think?

nvlsianpu:
@ulfalizer  1.5 should be able to be compiled against zephyr master, so coming zephyr 2.2 as well.

carlescufi:
> Hadn't we agreed that only the tip of `master` needs to be supported, while the responsibility for supporting other releases is deferred to https://github.com/zephyrproject-rtos/mcuboot ?

@d3zd3z 
I agree that the tip of master is the only one to be supported @utzig. For anything else there's Zephyr's fork.

utzig:
Have we agreed on something? Can this be closed?

d3zd3z:
This change doesn't break anything with regards to master. It still compiles, it just generates warnings. It'd be nice if Zephyr wasn't so aggressive about warnings (ideal one release cycle would go through without a warning, then one with, then the old functionality would be removed).

But, yeah, I forgot about the Zephyr fork of MCUboot which can be used to have working versions against releases. I do think a given release of mcuboot should work with a given release of Zephyr, though, so hopefully we can work with 2.2.

nvlsianpu:
mcuboot 1.5.0 will be compatible with zephyr 2.2 (on 99.9999%) 
There are examples of backward compatibility assurance in `to a zephyr changes adaptation` PRs like e.g. https://github.com/JuulLabs-OSS/mcuboot/pull/648

PR #660, opened 2020-02-11
Change CBOR decoding code in boot_serial

oyvindronningstad:
Change from using TinyCBOR for decoding to using code generated from a CDDL description.

This is a pilot application of cddl_gen.py (https://github.com/oyvindronningstad/cddl_gen) which generates C code for decoding CBOR.

The result is a ~1.1 k reduction in code size for MCUboot with serial recovery.

The change is split into 2 commits. The first makes the necessary changes to existing code and adds the CDDL description and a shell script for getting the generated code. The second adds the generated files as well as the script used for generation and a README.

Encoding is still using TinyCBOR.


utzig:
I only skimmed over this, but it looks very interesting! IMO it would be better to not include `cddl_gen.py` in the PR, and when auto-generating the .c files you could add a comment header stating that it was generated by a tool and link to the official repo. Also `boot_serial.c` still includes `cbor.h`, is that supposed to be there?

oyvindronningstad:
> I only skimmed over this, but it looks very interesting! IMO it would be better to not include `cddl_gen.py` in the PR, and when auto-generating the .c files you could add a comment header stating that it was generated by a tool and link to the official repo.

Sounds good! I'll add the header and remove the script from the PR.

> Also `boot_serial.c` still includes `cbor.h`, is that supposed to be there?

Yes, since TinyCBOR is still used for encoding. At least I assume it's still needed, I'll double check.

nvlsianpu:
cc @de-nordic

boot/boot_serial/src/serial_recovery_cbor_readme.txt
```diff
@@ -0,0 +1,16 @@
+The serial_recovery_cbor.c|h files have been generated with cddl_gen.py from the
+description in serial_recovery.cddl.
+cddl_gen.py has been copied from the "cddl_gen" repo, commit
```
de-nordic: Please refer here to the original used cddl_gen repo (https://github.com/oyvindronningstad/cddl_gen) as there is a chance that cddl_gen may get forked a lot and this may cause confusion.

boot/boot_serial/src/boot_serial.c
```diff
@@ -257,102 +252,38 @@ bs_upload(char *buf, int len)
...
+                img_blen = slen;
+                break;
+            case _Member_len:
+                data_len = upload._Upload_members[i]._Member._Member_len._Member_len;
+                break;
+            case _Member_off:
+                off = upload._Upload_members[i]._Member._Member_off._Member_off;
+                break;
+            case _Member_sha:
+                // upload._Upload_members[i]._Member._Member_sha._Member_sha
```
de-nordic: This has been left commented out intentionally?

boot/boot_serial/src/boot_serial.c
```diff
@@ -257,102 +252,38 @@ bs_upload(char *buf, int len)
...
-    cbor_parser_init(&reader.r, 0, &parser, &root_value);
-#endif
-    if (!cbor_value_is_container(&root_value)) {
+    Upload_t upload;
+    if (!cbor_decode_Upload((const uint8_t *)buf, len, &upload)) {
         goto out_invalid_data;
     }
-    if (cbor_value_enter_container(&root_value, &value)) {
+
+    if (upload._Upload_members_count > 5) {
```
de-nordic: Instead of literal 5, can you use some named define? Why 5?

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+	case 24: return 1;
+	case 25: return 2;
+	case 26: return 4;
+	case 27: return 8;
+	default: return 0;
+	}
+}
+
+/** Extract the major type, i.e. the first 3 bits of the header byte. */
+#define MAJOR_TYPE(header_byte) ((header_byte) >> 5) & 0x7
```
de-nordic: Missing outer parentheses around macro expansion.

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+
+/** Extract the additional info, i.e. the last 5 bits of the header byte. */
+#define ADDITIONAL(header_byte) (header_byte) & 0x1F;
```
de-nordic: Excess ; at the end of macro definition and missing outer parentheses around macro expression.

utzig:
@oyvindronningstad Tested on Mynewt as well; this brings a simple bootloader with no signature verification from 18144 to 16892 bytes, good improvement! How much work would it be to also do encoding and get rid of `tinycbor`? :-P

oyvindronningstad:
> @oyvindronningstad Tested on Mynewt as well; this brings a simple bootloader with no signature verification from 18144 to 16892 bytes, good improvement! How much work would it be to also do encoding and get rid of `tinycbor`? :-P

It's mostly a matter of writing a cbor_encode.c|h library following the same API as cbor_decode.c|h. I expect the changes to the python script to be minor in comparison.

boot/boot_serial/src/boot_serial.c
oyvindronningstad: It was because we have 5 different keys in the CBOR map ("image", "sha" etc.). I removed the checks altogether because it's not strictly needed, and will cause incompatibility with newer formats with additional keys.

oyvindronningstad:
Thanks, @utzig and @de-nordic, your comments should all be addressed now.

boot/boot_serial/src/boot_serial.c
de-nordic: Thx

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+	}
+	if (!size_decode(p_state,
+			&p_str_result->len, (size_t *)p_min_len,
+			(size_t *)p_max_len)) {
+		FAIL();
+	}
+	p_str_result->value = p_state->p_payload;
+	return true;
+}
+
```
de-nordic: Missing newline.

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+
+bool primx_decode(cbor_decode_state_t * p_state,
+		uint8_t *p_result, void *p_min_result, void *p_max_result)
+{
+	uint8_t major_type = MAJOR_TYPE(*p_state->p_payload);
+	uint32_t val;
+
+
```
de-nordic: Excess newline.

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+			p_state->elem_count = elem_count_bak;
+			if (i < min_decode) {
+				FAIL();
+			} else {
+				cbor_decode_print("Found %d elements.\n", i);
+			}
+			return true;
+		}
+	}
+	cbor_decode_print("Found %d elements.\n", max_decode);
```
de-nordic: I think 'z' is correct letter for printing size_t and ssize_t types.

boot/boot_serial/src/serial_recovery_cbor.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <string.h>
+#include "cbor_decode.h"
+
+
+typedef struct {
+	union {
```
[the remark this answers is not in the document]
oyvindronningstad: Very good question. I implemented it now. It's a lot cleaner, and code size is a bit smaller as well :)

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,410 @@
...
+	CBOR_MAJOR_TYPE_MAP  = 5, ///! Map
+	CBOR_MAJOR_TYPE_TAG  = 6, ///! Semantic Tag
+	CBOR_MAJOR_TYPE_PRIM = 7, ///! Primitive Type
+} cbor_major_type_t;
+
+/** Return value length from additional value.
+ */
+static uint32_t additional_len(uint8_t additional)
+{
+	switch (additional) {
```
de-nordic: May I suggest some code, that shaves a few bytes, here?

boot/boot_serial/src/cbor_decode.c
```diff
@@ -0,0 +1,412 @@
...
+#include <string.h>
+#include "cbor_decode.h"
+
+/** Enumeration representing the major types available in CBOR.
+ *
+ * The major type is represented in the 3 first bits of the header byte.
+ */
+typedef enum
+{
+	CBOR_MAJOR_TYPE_PINT = 0, ///! Positive Integer
```
nvlsianpu: Nip: I believe mcuboot convention are C like comments. `/**< xyx */`. Use elsewhere as well.

utzig:
@oyvindronningstad Maybe it would be a good idea to actually make your repo a submodule here, so that reviews of those sources are not mixed with mcuboot source code reviews, what do you think?

oyvindronningstad:
> @oyvindronningstad Maybe it would be a good idea to actually make your repo a submodule here, so that reviews of those sources are not mixed with mcuboot source code reviews, what do you think?

Sure, no problem. Should the submodule be inside the boot_serial/src folder, or somewhere else?

oyvindronningstad:
> > @oyvindronningstad Maybe it would be a good idea to actually make your repo a submodule here, so that reviews of those sources are not mixed with mcuboot source code reviews, what do you think?
> 
> Sure, no problem. Should the submodule be inside the boot_serial/src folder, or somewhere else?

@utzig Done now, take a look.

nvlsianpu:
> @oyvindronningstad Maybe it would be a good idea to actually make your repo a submodule here, so that reviews of those sources are not mixed with mcuboot source code reviews, what do you think?

I see one small problem - zephyr build system doesn't support sub modules. It uses `west` tool instead. Will be possible to add https://github.com/oyvindronningstad/cddl_gen reference to the zephyr manifest (https://github.com/zephyrproject-rtos/zephyr/blob/master/west.yml) and use it for this patch in case of zephyr build?

utzig:
> > Sure, no problem. Should the submodule be inside the boot_serial/src folder, or somewhere else?
> 
> @utzig Done now, take a look.

Putting the submodule under `src` in a Mynewt package (like `boot_serial` that has a `pkg.yml`) won't work well with Mynewt because it tries to build everything under that tree; it did fail because your tests are Zephyr only but even if they were Mynewt compatible it would still be a bad idea. Maybe using `boot/boot_serial/ext`? Or maybe we should put all submodules for everything under `/ext`? I haven't thought much about what the best solution would be...

oyvindronningstad:
@nvlsianpu @utzig As a first measure I moved the submodule to ext, and copied the cbor_decode.c|h files into boot_serial/src so tests will run. We can see if there's a better way still, maybe involving west.yml etc.

nvlsianpu:
In my opinion need to make your repo a sub-module for building with Mynewt and zephyr-rtos (in zephyr-rtos/zephyr for sure) west reference to build with zephyr. We are using this approach for mbedtls since a while here.

oyvindronningstad:
> In my opinion need to make your repo a sub-module for building with Mynewt and zephyr-rtos (in zephyr-rtos/zephyr for sure) west reference to build with zephyr. We are using this approach for mbedtls since a while here.

This currently works without checking the submodule out, since the relevant files are copied. The submodule is only there for reference and ease of updating later. I propose we merge this and add cddl_gen to zephyr's west manifest later if there's a compelling reason.

nvlsianpu:
@oyvindronningstad OK, agree on that. Will wait with zephyr's west manifest then for now.
Can you resolve the merge conflict?

utzig:
> In my opinion need to make your repo a sub-module for building with Mynewt and zephyr-rtos (in zephyr-rtos/zephyr for sure) west reference to build with zephyr. We are using this approach for mbedtls since a while here.

Our mbedTLS submodule was only used by the simulator until recently, and now also by the cypress port. Zephyr and Mynewt provide its own library.

nvlsianpu:
@utzig I would merge this as is?

utzig:
> @utzig I would merge this as is. ?

Please don't, this is not OK, I didn't take the time to comment much but I don't like the structure of how it was integrated, I would suggest another approach which follows in the lines of our tinycrypt integration. Just wait a bit...

d3zd3z:
> Zephyr and Mynewt provide their own library.

And we have had a few times when at least Zephyr updated mbed TLS, and there was an API change. I think in this case, we ended up making our own copy of these files.

nvlsianpu:
> And we have had a few times when at least Zephyr updated mbed TLS, and there was an API change. I think in this case, we ended up making our own copy of these files.

I was talking with Øyvind of channel, It seams this would be best solution until his CBOR parsing solution will be used wider.

d3zd3z:
I think we're ready with this, but now there are merge conflicts, and it will need to be rebased.

oyvindronningstad:
> I think we're ready with this, but now there are merge conflicts, and it will need to be rebased.

@d3zd3z Done

PR #661, opened 2020-02-11
sim: update dependencies

utzig:
Updates a few sim dependencies, reducing the amount of crates required for a build; also gets rid of cases of more than one version required for some crates. Results in depending on 76 crates instead of 89 before the PR.


utzig:
> It seems like every time I do a version update, it requires code changes to go with it.

Unfortunately some crates I did not upgrade because of that, eg, `rand`.

PR #663, opened 2020-02-17
loader: Fixed warning about uninitialized rc variable

erik-johnson-nimbelink:
Might get past the loop without going in at least once.

Signed-off-by: Erik Johnson <erik.johnson@nimbelink.com>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1145,7 +1145,7 @@ done:
 static int
 boot_verify_dependencies(struct boot_loader_state *state)
 {
-    int rc;
+    int rc = -1;
     uint8_t slot;
 
     BOOT_CURR_IMG(state) = 0;


utzig:
> Might get past the loop without going in at least once.

This function is only enabled `#if (BOOT_IMAGE_NUMBER > 1)` which means:

```
BOOT_CURR_IMG(state) = 0;
while (BOOT_CURR_IMG(state) < BOOT_IMAGE_NUMBER) {
```
will run *at least* once! This is a `gcc` issue, which version are you using?

erik-johnson-nimbelink:
It's not a GCC "issue", it's that GCC is being extra careful to not assume someone else won't happen to edit the passed-in state object between the set and the check -- which is 100% a valid thing for GCC to do. If you do something slightly different to assure GCC someone won't mess with the state, the warning goes away:

static int
boot_verify_dependencies(struct boot_loader_state *state)
{
    int rc;
    uint8_t slot;

    uint8_t currentImage = BOOT_CURR_IMG(state);

    while (currentImage < BOOT_IMAGE_NUMBER) {
        if (BOOT_SWAP_TYPE(state) != BOOT_SWAP_TYPE_NONE &&
            BOOT_SWAP_TYPE(state) != BOOT_SWAP_TYPE_FAIL) {
            slot = BOOT_SECONDARY_SLOT;
        } else {
            slot = BOOT_PRIMARY_SLOT;
        }

        rc = boot_verify_slot_dependencies(state, slot);
        if (rc == 0) {
            /* All dependencies've been satisfied, continue with next image. */
            currentImage++;
        } else if (rc == BOOT_EBADVERSION) {
            /* Cannot upgrade due to non-met dependencies, so disable all
             * image upgrades.
             */
            for (int idx = 0; idx < BOOT_IMAGE_NUMBER; idx++) {
                currentImage = idx;
            }
            break;
        } else {
            /* Other error happened, images are inconsistent */
            break;
        }
    }

    BOOT_CURR_IMG(state) = currentImage;
    return rc;
}

utzig:
What `gcc` version are you using?

erik-johnson-nimbelink:
gcc-arm-none-eabi-9-2019-q4-major

d3zd3z:
This kind of warning tends to come and go as subtle changes are made to the optimizer within gcc. I personally have no problem initializing this value. If anything, it makes it clearer what the empty case would mean, and we'd have to decide what value that should have.

Issue #529, opened 2019-08-14
MCUBoot not cleaning up properly before booting Zephyr?


endian-benjamin:
Cross post from: https://github.com/zephyrproject-rtos/zephyr/issues/18231

If you enable UART or RTT logging in MCUBoot, it seems these won't be uninitialized after booting to Zephyr, at least on nRF52.

1. Build a sample for nRF52 that implements proper power management (suspending all UARTs etc) and flash it "as usual" i.e. at offset `0x00` with bootloader disabled
2. Observe that power consumption is expectedly low when UARTs are disabled, about 4 uA hopefully
3. Build/flash mcuboot with UART enabled and redo step 1, flashing at `slot-0`
4. Observe that power consumption of `slot-0` application behaves as if UART never gets suspended.

Expected behavior:
- `slot-0` application behaves the same wrt power consumption regardless if you're bootloading or not.

Impact:
- If one does not rely upon UART inside bootloader, simply disabling it in the mcuboot `prj.conf` seems like a fine workaround. But I figured it was worth reporting anyway in case this is an indication that far worse bugs could be possible.

nvlsianpu:
Thanks for report it here!

utzig:
@mbolivar You were looking into something to handle this, correct?

mbolivar:
> @mbolivar You were looking into something to handle this, correct?

We briefly discussed this at an MCUboot meeting and decided Zephyr needs better power management for disabling devices. I'm not actively working on that but it would be the next step.

endian-benjamin:
So basically the issue is that we're not uniniting because Zephyr lacks "uninit" APIs?

mbolivar:
> So basically the issue is that we're not uniniting because Zephyr lacks "uninit" APIs?

Yep. I think @pabigot is starting to work on device power management but I am not sure.

pabigot:
I am, yes.  I expect it to take a while.  One might say that Zephyr's got some architectural weaknesses in areas that are critical to power conservation.

endian-benjamin:
@pabigot I am well aware :P I'll pm you on slack, since power management is dear to my heart.

endian-benjamin:
@pabigot Situation unchanged?

pabigot:
Yes.

nvlsianpu:
I think its time to add cleanup functionality to mcuboot, at last temporary.

pabigot:
[...]

PR #665, opened 2020-02-18
zephyr: allow recovery over usb cdc-acm with logs enabled

nvlsianpu:
Modyfied serial_adapter so log are alowed when using USB
CDC ACM serial port emulation.

As log needs to work asynchronously in its own thread,
mcuboot need to give some time for execution this thread using
k_sleep().

Added configuration for nrf52840_pca10056 which shows how
to enable looging along with USB - among other, thread log
processing is required.

build command (form zephyr-project root directory)
west build -d build/mcuboot/nrf52840_pca10056 -b nrf52840_pca10056
./bootloader/mcuboot/boot/zephyr/
 -- -DDTC_OVERLAY_FILE=./boards/nrf52840_pca10056_big.overlay
-DOVERLAY_CONFIG=./boards/nrf52840_pca10056_usb_cdc_log.conf

### DIFF
--- a/boot/zephyr/boards/nrf52840_pca10056_big.overlay
+++ b/boot/zephyr/boards/nrf52840_pca10056_big.overlay
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2020 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/delete-node/ &boot_partition;
+/delete-node/ &slot0_partition;
+/delete-node/ &slot1_partition;
+
+&flash0 {
+	partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			boot_partition: partition@0 {
+				label = "mcuboot";
+				reg = <0x000000000 0x00010000>;
+			};
+			slot0_partition: partition@10000 {
+				label = "image-0";
+				reg = <0x000010000 0x000074000>;
+			};
+			slot1_partition: partition@75000 {
+				label = "image-1";
+				reg = <0x00084000 0x000074000>;
+			};
+	};
+};
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -51,11 +51,10 @@ const struct boot_uart_funcs boot_funcs = {
 #else
 #include <logging/log_ctrl.h>
 
-#define BOOT_LOG_STACK_SIZE 768
 #define BOOT_LOG_PROCESSING_INTERVAL 30 /* [ms] */
 
 /* log are processing in custom routine */
-K_THREAD_STACK_DEFINE(boot_log_stack, BOOT_LOG_STACK_SIZE);
+K_THREAD_STACK_DEFINE(boot_log_stack, CONFIG_MCUBOOT_LOG_THREAD_STACK_SIZE);
 struct k_thread boot_log_thread;
 volatile bool boot_log_stop = false;
 K_SEM_DEFINE(boot_log_sem, 1, 1);
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -44,6 +44,28 @@ const struct boot_uart_funcs boot_funcs = {
 #include <usb/class/usb_dfu.h>
 #endif
 
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE)
+#ifdef CONFIG_LOG_PROCESS_THREAD
+#warning "The log internal thread for log processing can't transfer the log"\
+         "well for MCUBoot."
+#else
+#include <logging/log_ctrl.h>
+
+#define BOOT_LOG_STACK_SIZE 768
+#define BOOT_LOG_PROCESSING_INTERVAL 30 /* [ms] */
+
+/* log are processing in custom routine */
+K_THREAD_STACK_DEFINE(boot_log_stack, BOOT_LOG_STACK_SIZE);
+struct k_thread boot_log_thread;
+
+/* log processing need to be initalized by the application */
+#define ZEPHYR_BOOT_LOG_START() zephyr_boot_log_start()
+#endif /* CONFIG_LOG_PROCESS_THREAD */
+#else
+/* synchronous log mode doesn't need to be initalized by the application */
+#define ZEPHYR_BOOT_LOG_START() do { } while (false)
+#endif /* defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) */
+
 #ifdef CONFIG_SOC_FAMILY_NRF
 #include <hal/nrf_power.h>
 
@@ -175,6 +197,45 @@ static void do_boot(struct boot_rsp *rsp)
 }
 #endif
 
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
+    !defined(CONFIG_LOG_PROCESS_THREAD)
+/* The log internal thread for log processing can't transfer log well as has too
+ * low priority.
+ * Dedicated thread for log processing below uses highest application
+ * priority. This allows to transmit all logs without adding k_sleep/k_yield
+ * anywhere else int the code.
+ */
+
+/* most simple log processing theread */
+void boot_log_thread_func(void *dummy1, void *dummy2, void *dummy3)
+{
+    (void)dummy1;
+    (void)dummy2;
+    (void)dummy3;
+
+     log_init();
+
+     while (1) {
+             if (log_process(false) == false) {
+                    k_sleep(BOOT_LOG_PROCESSING_INTERVAL);
+             }
+     }
+}
+
+void zephyr_boot_log_start(void)
+{
+        /* start logging thread */
+        k_thread_create(&boot_log_thread, boot_log_stack,
+                K_THREAD_STACK_SIZEOF(boot_log_stack),
+                boot_log_thread_func, NULL, NULL, NULL,
+                K_HIGHEST_APPLICATION_THREAD_PRIO, 0,
+                BOOT_LOG_PROCESSING_INTERVAL);
+
+        k_thread_name_set(&boot_log_thread, "logging");
+}
+#endif/* defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
+        !defined(CONFIG_LOG_PROCESS_THREAD) */
+
 void main(void)
 {
     struct boot_rsp rsp;
@@ -183,6 +244,8 @@ void main(void)
     BOOT_LOG_INF("Starting bootloader");
 
     os_heap_init();
+
+    ZEPHYR_BOOT_LOG_START();
 
 #if (!defined(CONFIG_XTENSA) && defined(DT_FLASH_DEV_NAME))
     if (!flash_device_get_binding(DT_FLASH_DEV_NAME)) {
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -57,13 +57,17 @@ const struct boot_uart_funcs boot_funcs = {
 /* log are processing in custom routine */
 K_THREAD_STACK_DEFINE(boot_log_stack, BOOT_LOG_STACK_SIZE);
 struct k_thread boot_log_thread;
+volatile bool boot_log_stop = false;
+K_SEM_DEFINE(boot_log_sem, 1, 1);
 
 /* log processing need to be initalized by the application */
 #define ZEPHYR_BOOT_LOG_START() zephyr_boot_log_start()
+#define ZEPHYR_BOOT_LOG_STOP() zephyr_boot_log_stop()
 #endif /* CONFIG_LOG_PROCESS_THREAD */
 #else
 /* synchronous log mode doesn't need to be initalized by the application */
 #define ZEPHYR_BOOT_LOG_START() do { } while (false)
+#define ZEPHYR_BOOT_LOG_STOP() do { } while (false)
 #endif /* defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) */
 
 #ifdef CONFIG_SOC_FAMILY_NRF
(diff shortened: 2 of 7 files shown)


nvlsianpu:
DNM as I still need to add either k-sleep() or k_yield() in few more places for improve log output, but this shows that log works along with usb cdc acm recovery.

@nordic-krch pleas look at what I have done.

nvlsianpu:
Notice that https://github.com/zephyrproject-rtos/zephyr/pull/22798 is required for robust USB CDC ACM  communication.

boot/boot_serial/src/boot_serial.c
```diff
@@ -666,6 +666,9 @@ boot_serial_start(const struct boot_uart_funcs *f)
 
     off = 0;
     while (1) {
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE)
+        k_sleep(3); /* give time to process the log */
```
nvlsianpu: should be zephyr only patch

boot/zephyr/boards/nrf52840_pca10056_big.overlay
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/delete-node/ &boot_partition;
+/delete-node/ &slot0_partition;
+/delete-node/ &slot1_partition;
+/delete-node/ &scratch_partition;
+
+&flash0 {
+partitions {
```
mbolivar: Should be indented, right?

boot/zephyr/sample.yaml
```diff
@@ -13,3 +13,8 @@ tests:
   sample.bootloader.mcuboot.usb_cdc_acm_recovery:
     tags: bootloader_mcuboot
     platform_whitelist:  nrf52840_pca10059
+  sample.bootloader.mcuboot.usb_cdc_acm_recovery_log:
+    extra_args: OVERLAY_CONFIG=./boards/nrf52840_pca10056_usb_cdc_log.conf
+      DTC_OVERLAY=./boards/nrf52840_pca10056_big.overlay
+    platform_whitelist:  nrf52840_pca10056
+    tags: bootloader_mcuboot
```
mbolivar: Missing newline at end of file?

boot/zephyr/boards/nrf52840_pca10056_usb_cdc_log.conf
```diff
@@ -0,0 +1,37 @@
+# The UART is used for Serial Recovery, so logging requires
```
mbolivar: Is there any way to do this conditionally with if statements in the Kconfig file instead of using a board-specific file? This makes it not portable to other boards.

nvlsianpu: ok, will try to do that

nvlsianpu:
I Added dedicated thread for log processing of the highest application
priority. This allows to transmit all logs without adding k_sleep
anywhere else int the code.

Introduced boot log thread is simpler than the default log threat
which decreases flash footprint by a few dozen bytes.

@mbolivar what you think about such a solution.

nvlsianpu:
@hakonfam you opinion on this will be valuable.

boot/boot_serial/src/boot_serial.c
hakonfam: Magic value `3` should have a define with some comment on rationale. Otherwise it is extremely hard for future developers to understand the implications of it, and how to maintain it on other HW (if needed)

boot/zephyr/boards/nrf52840_pca10056_big.overlay
```diff
@@ -0,0 +1,35 @@
...
+		};
+		slot0_partition: partition@10000 {
+			label = "image-0";
+			reg = <0x000010000 0x000065000>;
+		};
+		slot1_partition: partition@75000 {
+			label = "image-1";
+			reg = <0x00075000 0x000065000>;
+		};
+		scratch_partition: partition@da000 {
```
hakonfam: Do we need this partition if we default to using the non-scratch-based but moved-based swap operation?

boot/boot_serial/src/boot_serial.c
hakonfam: Lol, I was doing per commit review, and I see now that this is indeed removed all together :)

boot/zephyr/main.c
```diff
@@ -175,6 +183,40 @@ static void do_boot(struct boot_rsp *rsp)
 }
 #endif
 
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
+    !defined(CONFIG_LOG_PROCESS_THREAD)
+#include <logging/log_ctrl.h>
+
+/* most simple log processing theread */
+void boot_log_therad_func(void *dummy1, void *dummy2, void *dummy3)
```
hakonfam:
```suggestion
void boot_log_thread_func(void *dummy1, void *dummy2, void *dummy3)
```

boot/zephyr/main.c
```diff
@@ -175,6 +183,40 @@ static void do_boot(struct boot_rsp *rsp)
...
+void zephyr_boot_log_start(void)
+{
+		/* start logging thread */
+		k_thread_create(&boot_log_thread, boot_log_stack,
+				K_THREAD_STACK_SIZEOF(boot_log_stack),
+				boot_log_therad_func, NULL, NULL, NULL,
+				K_HIGHEST_APPLICATION_THREAD_PRIO, 0, 30);
+		k_thread_name_set(&boot_log_thread, "logging");
+}
+#endif
```
hakonfam:
```suggestion
#endif /* defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
    !defined(CONFIG_LOG_PROCESS_THREAD) */
```

hakonfam: Not sure if this is really needed, but some comment on what the condition for the corresponding `if` would be welcome IMO

boot/zephyr/main.c
```diff
@@ -175,6 +183,40 @@ static void do_boot(struct boot_rsp *rsp)
...
+     log_init();
+
+     while (1) {
+             if (log_process(false) == false) {
+                    k_sleep(30);
+             }
+     }
+}
+
+K_THREAD_STACK_DEFINE(boot_log_stack, 768);
```
hakonfam: Aren't stack sizes of threads typically user configurable? In this case I kinda agree that it can be hardcoded, because its logging, and will never be in released code (AFAIK), you could consider if you want to add some kconfig option for this though.

boot/zephyr/main.c
```diff
@@ -175,6 +183,40 @@ static void do_boot(struct boot_rsp *rsp)
...
+{
+    (void)dummy1;
+    (void)dummy2;
+    (void)dummy3;
+
+     log_init();
+
+     while (1) {
+             if (log_process(false) == false) {
+                    k_sleep(30);
```
hakonfam: Magic value `30` should have some define with a comment behind its rationale.

boot/zephyr/main.c
```diff
@@ -175,6 +183,40 @@ static void do_boot(struct boot_rsp *rsp)
 }
 #endif
 
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
+    !defined(CONFIG_LOG_PROCESS_THREAD)
+#include <logging/log_ctrl.h>
```
mbolivar: Why is it necessary to do this twice instead of just once, outside of `CONFIG_SOC_FAMILY_NRF`?

nvlsianpu:
> Do you need to shut down this thread before jumping to the application? CC @nordic-krch

No it doesn't need that.

nvlsianpu:
@mbolivar @hakonfam DNM removed, passed zephyr sanitycheck locally, yours comments were taken into account.

boot/zephyr/main.c
```diff
@@ -44,6 +44,28 @@ const struct boot_uart_funcs boot_funcs = {
...
 #endif
 
+#if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE)
+#ifdef CONFIG_LOG_PROCESS_THREAD
+#warning "The log internal thread for log processing can't transfer the log"\
+         "well for MCUBoot."
+#else
+#include <logging/log_ctrl.h>
+
+#define BOOT_LOG_STACK_SIZE 768
```
mbolivar: I think @hakonfam 's comment about making this Kconfigurable (and the log processing interval too) is not addressed.

mbolivar:
Thanks! Looking much more generic now.



> > Do you need to shut down this thread before jumping to the application? CC @nordic-krch
> 
> No it doesn't need that.

I have some doubts here. Can you explain your reasoning?

nordic-krch:
> Do you need to shut down this thread before jumping to the application?

I assume that bootloader simply jumps to the application, right? If so then i'm a bit afraid about what happens with system clock which is running (RTC) and will expire after jumping to the application. For application kernel that will be confusing and may lead to serious, hard to detect issues.

I don't know mcuboot details but imo jumping to the application from bootloader should always happen after reset (with clean state). There are some resources started (usb, clock, rtc, mpu?, etc.) and application is not expecting that.

nvlsianpu:
> Do you need to shut down this thread before jumping to the application? CC @nordic-krch

@mbolivar @nordic-krch I didn't observed any problem which causes any system failure.
What I'm aware is #529 issue, what you are afraid is the same problem.

I can kill the log processing thread - but this still not guaranties that RTOS release/clean all HW clocks and timers resources (I beet it doesn't).

@nordic-krch 
I was thinking about chain-load the application via reboot before jumping to the application might create Time of check, time of use issue. I would consider that execution flow on separate feature-discussion. Do you want to start the discussion via mcboot's feature issue creation?

@utzig How similar problem is solved for Mynewt

nordic-krch:
> I was thinking about chain-load the application via reboot before jumping to the application might create Time of check, time of use issue.

It's not about validation of the application. That should be performed before jump to application or omitted only on user responsibility, e.g. after special reset like wakeup from deep sleep. From my experience with the bootloader it nrf5 sdk, there should be no direct jump to the application once bootloader performed any actions other than the one needed to start application. That is image update or recovery. It's usually quite difficult to get back to clean state and reset is the safest and simplest way.

> I didn't observed any problem

That can be very random. Especially with active timeout.

nvlsianpu:
@mbolivar @nordic-krch In the most recent commit I added termination of the log processing thread.

boot/zephyr/main.c
```diff
@@ -175,6 +200,58 @@ static void do_boot(struct boot_rsp *rsp)
...
+
+        k_thread_name_set(&boot_log_thread, "logging");
+}
+
+void zephyr_boot_log_stop(void)
+{
+    boot_log_stop = true;
+
+    /* busy wait until log procesing thread expired*/
+    while (boot_log_stop) {
```
mbolivar: Are you sure this is enough? Might be worth using k_kthread_abort() afterwards:

https://docs.zephyrproject.org/latest/reference/kernel/threads/index.html#_CPPv414k_thread_abort7k_tid_t

(I just am being paranoid about cleaning up according to Zephyr API promises, rather than what happens to work today.)

boot/zephyr/main.c
mbolivar: > I think @hakonfam 's comment about making this Kconfigurable (and the log processing interval too) is not addressed.

Do you intend to address this one?

boot/zephyr/main.c
nvlsianpu: quote [thread termination behavior description](
https://docs.zephyrproject.org/latest/reference/kernel/threads/index.html#thread-termination):
> Once a thread is started it typically executes forever. However, a thread may synchronously end its execution by returning from its entry point function. This is known as termination.

In the next paragraph (about aborting):
> it is typically preferable to signal a thread to terminate itself gracefully, rather than aborting it.

So thread ends execution and that all needed. I might call abort but according to this, it is not  required.

boot/zephyr/main.c
nvlsianpu: addressed - I followed log internal thread stack size from zephyr codebase

boot/zephyr/main.c
mbolivar: Right, but the current implementation looks racy. How do you know the thread is done, instead of just having set the variable and gotten scheduled out? Are all the threads cooperative?

mbolivar: Looking at https://github.com/zephyrproject-rtos/zephyr/issues/21500, it seems like the common pattern right now is to use a semaphore given at the end of the exiting thread, since there is no `k_thread_join()` in Zephyr.

 From that issue:

> There's no good way to block until another thread exits.
> 
> Our test cases currently manage this in a few ways:
>
>    1. Create an equal priority co-op thread, or higher priority thread, and then yield() the CPU, or sleep for a while. Assuming the child thread never blocks and then exits, control returns after yield() completes...if we are in uniprocessor. In SMP the child could spawn on another CPU in which case after the yield() returns the child is probably still running. The workaround is the 1cpu variant of test case declarations, which spawn a busy-looping co-op thread on all other CPUs.
>    2. Synchronize by having the parent take a semaphore given by the child at the end of its entry function. Potentially race-prone if the child hasn't completely exited and the parent tries to re-use the thread object, resulting in a crash, but still safer than #1.

So I think on Nordic SoCs, the current implementation, which uses a higher priority thread and a flag global, is probably safe. But just for the sake of keeping to the common pattern, how about using approach 2. for now (give a semaphore at log thread exit), and add a comment to switch to using `k_thread_join()` as soon as that is available?

nvlsianpu: ^^ updated.
I could even use the same semaphore (but numeric one) to request thread termination - but I fell this as a overkill, so kept trigger via volatile variable,

boot/zephyr/main.c
```diff
@@ -216,7 +299,7 @@ void main(void)
 #else
                             GPIO_DIR_IN | GPIO_PUD_PULL_UP
 #endif
-	    );
+           );
```
mbolivar: Unnecessary whitespace change

boot/zephyr/main.c
```diff
@@ -175,6 +200,62 @@ static void do_boot(struct boot_rsp *rsp)
...
+                BOOT_LOG_PROCESSING_INTERVAL);
+
+        k_thread_name_set(&boot_log_thread, "logging");
+}
+
+void zephyr_boot_log_stop(void)
+{
+    boot_log_stop = true;
+
+    /* busy wait until log procesing thread expired
```
mbolivar: Technically this is not busy waiting, because calling k_sem_take should put the thread to sleep until the semaphore is available. 

```suggestion
    /* wait until log procesing thread expired
```

boot/zephyr/main.c
nvlsianpu: I replaced Tab indentation. spaces indentation is mcuboot CS contention,

PR #674, opened 2020-02-26
Share boot data with runtime software

davidvincze:
This patch enables measured boot by implementing an interface for mcuboot to share various attributes/measurement of the authenticated images (boot data) with the runtime software (sharing arbitrary data is also possible). As a result it also creates the prerequisites for an attestation service
(runtime) by preserving this data from the boot stage.

The imgtool was updated to optionally add a new type of TLV to the image manifest with all the attributes and measurements that are known at build time. Before adding this data to the TLV area it is encoded in CBOR format. Later mcuboot will read and update this binary data and add it to a shared memory area between the bootloader and runtime software.

The list of shared measurements is based on the recommendations of Arm's Platform Security Architecture (PSA) (https://developer.arm.com/architectures/security-architectures/platform-security-architecture) because another purpose of this patch is to support compliance with it. By encoding part of the boot data at build time the increase in code size can be minimized and the CBOR encoded data can be used directly later in the creation of an initial attestation token which is also required by PSA.

### DIFF
--- a/boot/bootutil/include/bootutil/boot_record.h
+++ b/boot/bootutil/include/bootutil/boot_record.h
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2018-2020 Arm Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __BOOT_RECORD_H__
+#define __BOOT_RECORD_H__
+
+#include <stdint.h>
+#include "bootutil/image.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Add an image's all boot status information to the shared memory area
+ * between the bootloader and runtime SW.
+ *
+ * @param[in]  sw_module  Identifier of the SW component.
+ * @param[in]  hdr        Pointer to the image header stored in RAM.
+ * @param[in]  fap        Pointer to the flash area where image is stored.
+ *
+ * @return                0 on success; nonzero on failure.
+ */
+int boot_save_boot_status(uint8_t sw_module,
+                          const struct image_header *hdr,
+                          const struct flash_area *fap);
+
+/**
+ * Add application specific data to the shared memory area between the
+ * bootloader and runtime SW.
+ *
+ * @param[in]  hdr        Pointer to the image header stored in RAM.
+ * @param[in]  fap        Pointer to the flash area where image is stored.
+ *
+ * @return                0 on success; nonzero on failure.
+ */
+int boot_save_shared_data(const struct image_header *hdr,
+                          const struct flash_area *fap);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __BOOT_RECORD_H__ */
--- a/boot/bootutil/include/bootutil/boot_status.h
+++ b/boot/bootutil/include/bootutil/boot_status.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020 Arm Limited
+ * Copyright (c) 2020 Linaro Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,8 +39,9 @@ extern "C" {
  *    header structure: struct shared_data_tlv_entry and the data. In the entry
  *    header is a type field (tly_type) which identify the consumer of the
  *    entry in the runtime SW and specify the subtype of that data item. There
- *    is a size field (tlv_len) which covers the size of the entry header and
- *    the data. After this structure comes the actual data.
+ *    is a size field (tlv_len) which covers the size of the the data. After
+ *    this structure comes the actual data.
+ *
  *  - Arbitrary number and size of data entry can be in the shared memory area.
  *
  * This table gives of overview about the tlv_type field in the entry header.
@@ -68,9 +70,10 @@ extern "C" {
 #define MINOR_MASK 0xFFF   /* 12 bit */
 
 #define SET_TLV_TYPE(major, minor) \
-        ((((major) & MAJOR_MASK) << MAJOR_POS) | ((minor) & MINOR_MASK))
-#define GET_MAJOR(tlv_type) ((tlv_type) >> MAJOR_POS)
-#define GET_MINOR(tlv_type) ((tlv_type) &  MINOR_MASK)
+        (((uint16_t)((major) & MAJOR_MASK) << MAJOR_POS) \
+        | ((minor) & MINOR_MASK))
+#define GET_MAJOR(tlv_type) ((uint16_t)(tlv_type) >> MAJOR_POS)
+#define GET_MINOR(tlv_type) ((tlv_type) & MINOR_MASK)
 
 /* Magic value which marks the beginning of shared data area in memory */
 #define SHARED_DATA_TLV_INFO_MAGIC    0x2016
@@ -98,9 +101,10 @@ extern "C" {
 #define CLAIM_MASK 0x3F            /* 6 bit */
 #define MEASUREMENT_CLAIM_POS 3    /* 3 bit */
 
-#define GET_IAS_MODULE(tlv_type) (GET_MINOR(tlv_type) >> MODULE_POS)
-#define GET_IAS_CLAIM(tlv_type)  (GET_MINOR(tlv_type)  & CLAIM_MASK)
-#define SET_IAS_MINOR(sw_module, claim) (((sw_module) << 6) | (claim))
+#define GET_IAS_MODULE(tlv_type) ((uint16_t)GET_MINOR(tlv_type) >> MODULE_POS)
+#define GET_IAS_CLAIM(tlv_type)  (GET_MINOR(tlv_type) & CLAIM_MASK)
+#define SET_IAS_MINOR(sw_module, claim) \
+        (((uint16_t)(sw_module) << MODULE_POS) | (claim))
 
 /**
  * Shared data TLV header.  All fields in little endian.
@@ -127,7 +131,7 @@ struct shared_data_tlv_header {
  */
 struct shared_data_tlv_entry {
     uint16_t tlv_type;
-    uint16_t tlv_len; /* size of single TLV entry (including this header). */
+    uint16_t tlv_len; /* TLV data length (not including this header). */
 };
 
 #define SHARED_DATA_ENTRY_HEADER_SIZE sizeof(struct shared_data_tlv_entry)
--- a/boot/bootutil/include/bootutil/boot_status.h
+++ b/boot/bootutil/include/bootutil/boot_status.h
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2018-2020 Arm Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __BOOT_STATUS_H__
+#define __BOOT_STATUS_H__
+
+#include <stdint.h>
+#include <stddef.h>
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * The shared data between boot loader and runtime SW is TLV encoded. The
+ * shared data is stored in a well known location in memory and this is a
+ * contract between boot loader and runtime SW.
+ *
+ * The structure of shared data must be the following:
+ *  - At the beginning there must be a header: struct shared_data_tlv_header
+ *    This contains a magic number and a size field which covers the entire
+ *    size of the shared data area including this header.
+ *  - After the header there come the entries which are composed from an entry
+ *    header structure: struct shared_data_tlv_entry and the data. In the entry
+ *    header is a type field (tly_type) which identify the consumer of the
+ *    entry in the runtime SW and specify the subtype of that data item. There
+ *    is a size field (tlv_len) which covers the size of the entry header and
+ *    the data. After this structure comes the actual data.
+ *  - Arbitrary number and size of data entry can be in the shared memory area.
+ *
+ * This table gives of overview about the tlv_type field in the entry header.
+ * The tlv_type always composed from a major and minor number. Major number
+ * identifies the addressee in runtime SW, who should process the data entry.
+ * Minor number used to encode more info about the data entry. The actual
+ * definition of minor number could change per major number.
+ *
+ * In case of boot status data, which can be processed by an attestation
+ * service the minor number is split further to two part: sw_module and claim.
+ * The sw_module identifies the SW component in the system which the data item
+ * belongs to and the claim part identifies the exact type of the data.
+ *
+ * |---------------------------------------|
+ * |            tlv_type (16)              |
+ * |---------------------------------------|
+ * | tlv_major(4)|      tlv_minor(12)      |
+ * |---------------------------------------|
+ * | MAJOR_IAS   | sw_module(6) | claim(6) |
+ * |---------------------------------------|
+ */
+
+/* General macros to handle TLV type */
+#define MAJOR_MASK 0xF     /* 4  bit */
+#define MAJOR_POS  12      /* 12 bit */
+#define MINOR_MASK 0xFFF   /* 12 bit */
+
+#define SET_TLV_TYPE(major, minor) \
+        ((((major) & MAJOR_MASK) << MAJOR_POS) | ((minor) & MINOR_MASK))
+#define GET_MAJOR(tlv_type) ((tlv_type) >> MAJOR_POS)
+#define GET_MINOR(tlv_type) ((tlv_type) &  MINOR_MASK)
+
+/* Magic value which marks the beginning of shared data area in memory */
+#define SHARED_DATA_TLV_INFO_MAGIC    0x2016
+
+/* Initial attestation specific macros */
+
+/**
+ * Major numbers (4 bit) to identify the
+ * consumer of shared data in runtime SW.
+ */
+#define TLV_MAJOR_IAS      0x1
+
+/* Initial attestation: Claim per SW components / SW modules */
+/* Bits: 0-2 */
+#define SW_VERSION       0x00
+#define SW_SIGNER_ID     0x01
+/* Reserved              0x02 */
+#define SW_TYPE          0x03
+/* Bits: 3-5 */
+#define SW_MEASURE_VALUE 0x08
+#define SW_MEASURE_TYPE  0x09
+#define SW_BOOT_RECORD   0x3F
+
+#define MODULE_POS 6               /* 6 bit */
+#define CLAIM_MASK 0x3F            /* 6 bit */
+#define MEASUREMENT_CLAIM_POS 3    /* 3 bit */
+
+#define GET_IAS_MODULE(tlv_type) (GET_MINOR(tlv_type) >> MODULE_POS)
+#define GET_IAS_CLAIM(tlv_type)  (GET_MINOR(tlv_type)  & CLAIM_MASK)
+#define SET_IAS_MINOR(sw_module, claim) (((sw_module) << 6) | (claim))
+
+/**
+ * Shared data TLV header.  All fields in little endian.
+ *
+ *    -----------------------------------
+ *    | tlv_magic(16) | tlv_tot_len(16) |
+ *    -----------------------------------
+ */
+struct shared_data_tlv_header {
+    uint16_t tlv_magic;
+    uint16_t tlv_tot_len; /* size of whole TLV area (including this header) */
+};
+
+#define SHARED_DATA_HEADER_SIZE sizeof(struct shared_data_tlv_header)
+
+/**
+ * Shared data TLV entry header format. All fields in little endian.
+ *
+ *    -------------------------------
+ *    | tlv_type(16) |  tlv_len(16) |
+ *    -------------------------------
+ *    |         Raw data            |
+ *    -------------------------------
+ */
+struct shared_data_tlv_entry {
+    uint16_t tlv_type;
+    uint16_t tlv_len; /* size of single TLV entry (including this header). */
+};
+
+#define SHARED_DATA_ENTRY_HEADER_SIZE sizeof(struct shared_data_tlv_entry)
+#define SHARED_DATA_ENTRY_SIZE(size) (size + SHARED_DATA_ENTRY_HEADER_SIZE)
+
+/* Structure to store the boot data for the runtime SW. */
+struct shared_boot_data {
+    struct shared_data_tlv_header header;
+    uint8_t data[];
+};
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __BOOT_STATUS_H__ */
--- a/boot/bootutil/src/boot_record.c
+++ b/boot/bootutil/src/boot_record.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020 Arm Limited
+ * Copyright (c) 2020 Linaro Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -87,21 +88,23 @@ boot_add_data_to_shared_area(uint8_t        major_type,
     /* Iterates over the TLV section looks for the same entry if found then
      * returns with error: SHARED_MEMORY_OVERWRITE
      */
-    for (; offset < tlv_end; offset += tlv_entry.tlv_len) {
+    while (offset < tlv_end) {
         /* Create local copy to avoid unaligned access */
         memcpy(&tlv_entry, (const void *)offset, SHARED_DATA_ENTRY_HEADER_SIZE);
         if (GET_MAJOR(tlv_entry.tlv_type) == major_type &&
             GET_MINOR(tlv_entry.tlv_type) == minor_type) {
             return SHARED_MEMORY_OVERWRITE;
         }
+
+        offset += SHARED_DATA_ENTRY_SIZE(tlv_entry.tlv_len);
     }
 
     /* Add TLV entry */
     tlv_entry.tlv_type = SET_TLV_TYPE(major_type, minor_type);
-    tlv_entry.tlv_len  = SHARED_DATA_ENTRY_SIZE(size);
+    tlv_entry.tlv_len  = size;
 
     if (!boot_u16_safe_add(&boot_data_size, boot_data->header.tlv_tot_len,
-                           tlv_entry.tlv_len)) {
+                           SHARED_DATA_ENTRY_SIZE(size))) {
         return SHARED_MEMORY_GEN_ERROR;
     }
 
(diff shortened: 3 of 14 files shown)


utzig:
@davidvincze The Mynewt build fails because `MCUBOOT_SHARED_DATA_BASE` and `MCUBOOT_SHARED_DATA_SIZE` are not defined, I don't see where they are defined, what am I missing?

davidvincze:
> @davidvincze The Mynewt build fails because `MCUBOOT_SHARED_DATA_BASE` and `MCUBOOT_SHARED_DATA_SIZE` are not defined, I don't see where they are defined, what am I missing?

I didn't surround the code in boot_record.c with an #ifdef and the mynewt test was failing due to the missing macro definitions even if this feature was disabled. These definitions are platform dependent and should come from the platform code. I'll update the documentation accordingly.

scripts/imgtool/boot_record.py
```diff
@@ -0,0 +1,75 @@
...
+import cbor
+
+
+# SW component IDs
+SW_COMPONENT_RANGE = 0
+SW_COMPONENT_TYPE = SW_COMPONENT_RANGE + 1
+MEASUREMENT_VALUE = SW_COMPONENT_RANGE + 2
+SW_COMPONENT_VERSION = SW_COMPONENT_RANGE + 4
+SIGNER_ID = SW_COMPONENT_RANGE + 5
+MEASUREMENT_DESCRIPTION = SW_COMPONENT_RANGE + 6
```
utzig: Why not use an `Enum`?

scripts/imgtool/boot_record.py
```diff
@@ -0,0 +1,75 @@
...
+                             sw_measurement_value, sw_signer_id):
+
+    # List of SW component claims (key ID + value)
+    key_value_list = [
+        SW_COMPONENT_TYPE, sw_type,
+        SW_COMPONENT_VERSION, sw_version,
+        SIGNER_ID, sw_signer_id,
+        MEASUREMENT_DESCRIPTION, sw_measurement_type,
+        MEASUREMENT_VALUE, sw_measurement_value
+    ]
```
utzig: This could actually be a `dict`. If it is a `dict` there is no need to check it is even in length, and there should be no need for `claim_number` because `cbor.dumps` can encode it to `bytes` and no hack (`+ 0xA0`) is required.

scripts/imgtool/image.py
```diff
@@ -19,6 +19,7 @@
 """
 
 from . import version as versmod
+from . import boot_record as br
```
utzig: Since only one function is being used maybe `from .boot_record import create_sw_component_data` would be better?

scripts/imgtool/boot_record.py
```diff
@@ -0,0 +1,75 @@
...
+
+def create_sw_component_data(sw_type, sw_version, sw_measurement_type,
+                             sw_measurement_value, sw_signer_id):
+
+    # List of SW component claims (key ID + value)
+    key_value_list = [
+        SW_COMPONENT_TYPE, sw_type,
+        SW_COMPONENT_VERSION, sw_version,
+        SIGNER_ID, sw_signer_id,
+        MEASUREMENT_DESCRIPTION, sw_measurement_type,
```
utzig: Would not it be better if the names matched? (either `MEASUREMENT_TYPE` or `sw_measurement_description`)

scripts/imgtool/image.py
```diff
@@ -255,16 +259,50 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
         self.enckey = enckey
 
+        # Calculate the hash of the public key
+        if key is not None:
+            pub = key.get_public_bytes()
+            sha = hashlib.sha256()
+            sha.update(pub)
+            pubbytes = sha.digest()
+        else:
+            pubbytes = bytes(SHA256_DIGEST_SIZE)
```
utzig: Could get the digest directly from `hashlib` with `bytes(hashlib.sha256().digest_size)`

scripts/imgtool/image.py
```diff
@@ -255,16 +259,50 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
         if self.security_counter is not None:
             # Size of the security counter TLV: header ('HH') + payload ('I')
             #                                   = 4 + 4 = 8 Bytes
             protected_tlv_size += TLV_SIZE + 4
 
+        if sw_type is not None:
+            if len(sw_type) > MAX_SW_TYPE_LENGTH:
+                msg = "'{}' is too long ({} characters) for sw_type. Its " \
+                    "maximum allowed length is 12 characters.".format(
+                    sw_type, len(sw_type))
```
utzig: `flake8` tells me this line should have on extra indentation

scripts/imgtool/image.py
```diff
@@ -255,16 +259,50 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
+                raise click.UsageError(msg)
+
+            image_version = (str(self.version.major) + '.'
+                          + str(self.version.minor) + '.'
+                          + str(self.version.revision))
```
utzig: `flake8` tells me the `+` in those 2 lines should align one space after the `(` in the line above.

scripts/imgtool/image.py
```diff
@@ -255,16 +259,50 @@ def ecies_p256_hkdf(self, enckey, plainkey):
...
+
+            # The image hash is computed over the image header, the image itself
+            # and the protected TLV area. However, the boot record TLV (which is
+            # part of the protected area) should contain this hash before it is
+            # even calculated. For this reason the script fills this field with
+            # zeros and the bootloader will insert the right value later.
+            digest = bytes(SHA256_DIGEST_SIZE)
```
utzig: `bytes(hashlib.sha256().digest_size)` ?

davidvincze: Done (thank you).

scripts/imgtool/image.py
davidvincze: Done.

scripts/imgtool/image.py
davidvincze: I believe it's easier, this way I can entirely abandon the module name.

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,228 @@
...
+     * the protected TLV area (which should already include the image hash as
+     * part of the boot record TLV). For this reason this field has been
+     * filled with zeros during the image signing process.
+     */
+    offset = record_len - sizeof(image_hash);
+    /* Avoid buffer overflow. */
+    if ((offset + sizeof(image_hash)) > sizeof(buf)) {
+        return BOOT_STATUS_ERROR;
+    }
+    memcpy(buf + offset, image_hash, sizeof(image_hash));
```
utzig: Not sure why you are reading the sha256 TLV, appending it here, and then seem to be passing the length of the buffer without the added hash to `boot_add_data_to_shared_area`, which itself doesn't seem to be using the hash at all; what am I missing?

utzig:
> The list of shared measurements is based on the recommendations of Arm's Platform Security Architecture (PSA) (https://developer.arm.com/architectures/security-architectures/platform-security-architecture) because another purpose of this patch is to support compliance with it. By encoding part of the boot data at build time the increase in code size can be minimized and the CBOR encoded data can be used directly later in the creation of an initial attestation token which is also required by PSA.

Could you please provide direct links to the relevant documents?

scripts/imgtool/boot_record.py
davidvincze: I'm not sure if it's worth it. For example the `SIGNER_ID` as an enum would look like: `SwCompFields.SIGNER_ID.value`.

utzig: > I'm not sure if it's worth it. For example the `SIGNER_ID` as an enum would look like: `SwCompFields.SIGNER_ID.value`.

A better naming would look like this: `SwComponent.SIGNER_ID`

utzig: Btw, why is `SW_COMPONENT_VERSION = SW_COMPONENT_RANGE + 4` using `+ 4` instead of `+ 3`?

scripts/imgtool/boot_record.py
davidvincze: Currently I can use a `list` only, it must be ensured that the `MEASUREMENT_VALUE` is the last item:
>"A dictionary would be the best suited data structure to store these key-value pairs (claims), however dictionaries are not sorted, but for example the lists do keep to order of items which we care about now. An ordered dictionary could be used instead, but it would be converted to a dict before the encoding and this conversion may not keep the order of the items."

utzig: `dict` is ordered since 3.7: https://mail.python.org/pipermail/python-dev/2017-December/151283.html

davidvincze: I think the SW component claims without the `MEASUREMENT_VALUE` could be encoded directly as a `dict` and that all 0's byte array (representing the measurement value) could be appended to it later but the resulting binary data still needs to be modified (because it's length has changed).

utzig: I feel a bit lost, let me just PoC something...

davidvincze:
> Did a "preliminary" review, I'll have to read some of the PSA material to understand some of the concepts...

Let me know if I can help you navigating through the documents. ;)

scripts/imgtool/boot_record.py
utzig:
```python
from enum import Enum
import sys
import cbor


class SwComponent(int, Enum):
    Type = 1
    MeasurementValue = 2
    Version = 4
    SignedId = 5
    MeasurementDescription = 6


def create_sw_component_data(sw_type, sw_version, sw_measurement_type,
                             sw_measurement_value, sw_signer_id):
    key_values = {
        SwComponent.Type: sw_type,
        SwComponent.Version: sw_version,
        SwComponent.SignedId: sw_signer_id,
        SwComponent.MeasurementDescription: sw_measurement_type,
        SwComponent.MeasurementValue: sw_measurement_value,
    }

    if len(key_values) > 11:
        print('Error: There are more than 11 claims in the '
              'list of sw component claims.', file=sys.stderr)
        sys.exit(1)

    return cbor.dumps(key_values)


print(create_sw_component_data(0xa0, 0xa1, 0xa2, 0xa3, 0xa4))
```

Running this results in `b'\xa5\x01\x18\xa0\x04\x18\xa1\x05\x18\xa4\x06\x18\xa2\x02\x18\xa3'` which looks correct.

boot/bootutil/include/bootutil/boot_record.h
```diff
@@ -0,0 +1,91 @@
...
+ * @param[in] size        length of added data
+ * @param[in] data        pointer to data
+ *
+ * @return Returns error code as specified in @ref shared_memory_err_t
+ */
+enum shared_memory_err_t
+boot_add_data_to_shared_area(uint8_t        major_type,
+                             uint16_t       minor_type,
+                             size_t         size,
+                             const uint8_t *data);
```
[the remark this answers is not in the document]
davidvincze: Currently there is none, I'll make it `static`. Probably new functions will be defined when there is a need to share other specific information and the boot_status.h can also be extended accordingly.

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,228 @@
...
+                      const struct flash_area *fap)
+{
+
+    struct image_tlv_iter it;
+    uint32_t offset;
+    uint16_t len;
+    uint16_t type;
+    size_t record_len = 0;
+    uint8_t image_hash[32]; /* SHA256 - 32 Bytes */
+    uint8_t buf[MAX_BOOT_RECORD_SZ];
```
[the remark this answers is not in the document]
davidvincze: It's a target/platform dependent parameter and should come from the target code (it will be described in the  PORTING.md file).

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,228 @@
...
+    uint32_t hash_found = 0;
+    uint16_t ias_minor;
+    int32_t res;
+    enum shared_memory_err_t res2;
+
+    /* Manifest data is concatenated to the end of the image.
+     * It is encoded in TLV format.
+     */
+
+    res = bootutil_tlv_iter_begin(&it, hdr, fap, IMAGE_TLV_ANY, false);
```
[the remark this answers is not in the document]
davidvincze: May I ask why do you think it would be better to use `int` over `uint32_t`? I'm just curious.:)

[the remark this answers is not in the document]
d3zd3z: I think it is more of a convention. It is common throughout this code (and a lot of other code) to use 'int' for things that return small range of things, such as errors. It also helps on things like 64-bit platforms, where int would be a natural register size (64-bits) whereas uint32_t might require extra instructions for masking and such.

d3zd3z:
One question I want to make sure we think about is what other uses are there for this data beyond just PSA? I know of at least one other vendor that has some stuff they want to communicate. What if only some of the data ends up in the attestation token, will they have to parse it out of CBOR?

boot/bootutil/src/boot_record.c
davidvincze: The hash is not appended the boot record, it overwrites the last 32 bytes (sizeof(image_hash)) of the boot record that is in the buffer ('buf' is pointing to the first element of this array, so this memcpy is writing 32 bytes from this element: buf[record_len - sizeof(image_hash)]).

utzig: Ah got it, so you are adding 32 byte `\0` string and overriding it here from the read sha256 TLV! :+1:

boot/bootutil/src/boot_record.c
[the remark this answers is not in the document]
davidvincze: We could use the uint_fastN_t, int_leastN_t ..etc. types as well.  I always felt that these types are too neglected.:)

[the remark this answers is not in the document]
utzig: > We could use the uint_fastN_t, int_leastN_t ..etc. types as well. I always felt that these types are too neglected.:)

`int` looks good enough to me, and not neglected!

scripts/imgtool/boot_record.py
davidvincze: The result may depend on the version of Python and environment (Linux/Windows). At the time I was working on it, the order of items in the resulting binary was mixed up. This is not a requirement of a dictionary to keep the order of its items.

utzig: It **is a requirement** of the language since 3.7, and it was available in 3.6 (but not a requirement). So, even though I'd doubt you have such an old Python, you can still rely on `OrderedDict` in this case.

davidvincze: I should have added the details of my environment to the comments.

davidvincze: If I see it correctly, the order of dictionary items can be guaranteed from Python 3.7.
This cbor Python package iterates over the dictionary with `for k,v in d.items()` (over a `view` object) and from Python 3.7 the dictionary order is guaranteed to be insertion order according to the Python documentation: https://docs.python.org/3.7/library/stdtypes.html#dict-views

davidvincze: I didn't see your last message.. (should have pressed F5 :))
I was playing with `OrderedDict` but I cannot exactly recall why we chose this "hacky" solution at the end. Let me do another round of test with `OrderedDict`, it only requires Python 3.1. But it can only retain the order of items (when passed to the constructor or the update() method) from Python version 3.6.
As far as I know the imgtool doesn't require a minimum Python version (only that it must be version 3), does it? Would it be acceptable to assume a minimum version of 3.7? If yes, then I believe it could use a normal `Dict` as you suggested.

utzig: > Would it be acceptable to assume a minimum version of 3.7? If yes, then I believe it could use a normal Dict as you suggested.

It is 3.6 actually, it was added with https://github.com/python/cpython/commit/742da040db28e1284615e88874d5c952da80344e; I **think** that yes we could depend on this version, although that would be my personal take, I am hopeful that nobody in 2020 is still using Python < 3.6!

davidvincze: FYI I can confirm that it mixes up the order of items in the dictionary with Python 3.5. I can create an OrderedDict object and add the items to it one-by-one and it retains the order but when the CBOR package iterates over this dictionary with `for k,v in d.items()`. The .items() returns a `view` object and then it invokes its iterator and I think that's the point where the order changes (the order is guaranteed from Python 3.7). However, I tested it with Python 3.6 and it looks okay (in different environments).

Finally I would rather use an OrderedDict (as the order preserving nature of `dict` was just an implemetation detail and became mandatory with 3.7) as I consider this a safer solution.

utzig: > FYI I can confirm that it mixes up the order of items in the dictionary with Python 3.5.

I mean srsly, is anyone still using 3.5? :rofl:

> Finally I would rather use an OrderedDict (as the order preserving nature of dict was just an implemetation detail and became mandatory with 3.7) as I consider this a safer solution. But I can still check the used Python version from the script and require a minimum version of 3.7 :)

`OrderedDict` should be fine! No need to check versions, we can update in 10 years!

scripts/imgtool/boot_record.py
davidvincze: These SW component property IDs are specified by PSA (in the PSA Attestation API document).
The ID `3` is Reserved.

scripts/imgtool/boot_record.py
davidvincze: > I mean srsly, is anyone still using 3.5? 🤣

Neither do I think it will be a real issue, but who knows.. I've seen more surprising things. :)
Fortunately on Ubuntu for example one is not able to install 3.5 from the default package repositories.

utzig: I would just use `dict` and update `setup.py` / `main.py` to depend and enforce a dependency on 3.6. But that's just my personal take; we'd find out if anyone does actually need <3.6, also sometimes it's nice to force people to upgrade their 10 year old tech! :-) But feel free to do it as you find more appropriate.

boot/bootutil/src/boot_record.c
davidvincze: Yes.

utzig:
@davidvincze The code looks good, I will take a look at the PSA docs tomorrow. Please also add a commit updating `imgtool_version` to `1.6.0a1`.

scripts/imgtool/boot_record.py
d3zd3z: Why not just something like:
```
ky = list(dict.keys())
ky.sort()
for k in ky:
     use k and dict[k]
```
then there is no dependency or weird structure. You could also sort on something else.

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,241 @@
...
+#include "bootutil_priv.h"
+#include "bootutil/image.h"
+#include "flash_map_backend/flash_map_backend.h"
+
+/**
+ * @enum shared_data_err_t
+ *
+ * @brief Return values for adding data entry to shared memory area
+ */
+enum shared_memory_err_t {
```
d3zd3z: This is a fairly different convention from the rest of MCUBoot. The value is local to just this file, but it does end up resulting in needing an 'rc2' later. The convention is still that '0' is success, and I wonder if we'd be better just using int, and defines for the errors themselves.

boot/bootutil/include/bootutil/boot_status.h
```diff
@@ -0,0 +1,150 @@
...
+#define MAJOR_POS  12      /* 12 bit */
+#define MINOR_MASK 0xFFF   /* 12 bit */
+
+#define SET_TLV_TYPE(major, minor) \
+        (((uint16_t)((major) & MAJOR_MASK) << MAJOR_POS) \
+        | ((minor) & MINOR_MASK))
+#define GET_MAJOR(tlv_type) ((uint16_t)(tlv_type) >> MAJOR_POS)
+#define GET_MINOR(tlv_type) ((tlv_type) & MINOR_MASK)
+
+/* Magic value which marks the beginning of shared data area in memory */
```
d3zd3z: This seems like a fairly weak protection of this shared data (imagine uninitialized memory). I wonder if we shouldn't either have a longer magic value, or hash the magic value, and verify that the hash is correct?

nvlsianpu: This TLV is landing in shared memory. If only secure domain has write access here - then non-secure one could trusted that if magic value is right, the TLV is OK.
@davidvincze  Am I right.

nvlsianpu:
> MCUBOOT_SHARED_DATA_BASE an .. SIZE. I wonder how you want to configure for zephyr - this would be dedicated ram partition I believe.

@hakonfam ^^ any thought?

@ioannisg This is port of on of TFM fork feature, so you might be interested.

boot/bootutil/include/bootutil/boot_status.h
d3zd3z: I'm concerned about uninitialized memory having the magic value. If it is random, then 1:65536 devices with spurious data seems rather high.

scripts/imgtool/boot_record.py
davidvincze: It is not really about sorting but keeping one of the properties as the last item in the given "container". And it is also a requirement to encode these properties in a map/dictionary (so right before encoding this container, it must be a kind of dictionary - can be a descendant of a `dict`).

I've modified the imgtool to enforce a minimum Python version of 3.6. Does it sound acceptable to you David?

boot/bootutil/src/boot_record.c
davidvincze: Yes, you are right.
I've eliminated this new `enum` type from the PR and now it uses only defines and `int` as a return type.

davidvincze:
> > The list of shared measurements is based on the recommendations of Arm's Platform Security Architecture (PSA) (https://developer.arm.com/architectures/security-architectures/platform-security-architecture) because another purpose of this patch is to support compliance with it. By encoding part of the boot data at build time the increase in code size can be minimized and the CBOR encoded data can be used directly later in the creation of an initial attestation token which is also required by PSA.
> 
> Could you please provide direct links to the relevant documents?

@utzig if you're still interested (or anyone else), you can directly download the PSA documents from here: https://developer.arm.com/architectures/security-architectures/platform-security-architecture

The `xy` API doc can be downloaded from the "What is the `xy` API?" section if you click on `Learn more`.

boot/bootutil/include/bootutil/boot_status.h
davidvincze: Yes, @nvlsianpu  is right.
That 1:65536 ratio really is troublesome. I think the only reason to check whether this area is already initialized is the fact that there can be multiple images. I think there should be an extra function call before calling `boot_save_boot_status()` function to initialize the shared data area. It means no one else can use this area before mcuboot. It sound acceptable to me (at 8:00 PM).

utzig:
> > > The list of shared measurements is based on the recommendations of Arm's Platform Security Architecture (PSA) (https://developer.arm.com/architectures/security-architectures/platform-security-architecture) because another purpose of this patch is to support compliance with it. By encoding part of the boot data at build time the increase in code size can be minimized and the CBOR encoded data can be used directly later in the creation of an initial attestation token which is also required by PSA.
> > 
> > 
> > Could you please provide direct links to the relevant documents?
> 
> @utzig if you're still interested (or anyone else), you can directly download the PSA documents from here: https://developer.arm.com/architectures/security-architectures/platform-security-architecture

This is the same link as before, isn't it possible to just directly point to the relevant PDF or whatever?

davidvincze:
> One question I want to make sure we think about is what other uses are there for this data beyond just PSA? I know of at least one other vendor that has some stuff they want to communicate. What if only some of the data ends up in the attestation token, will they have to parse it out of CBOR?

Originally we didn't encode this data during the image signing process. It was changed to help reducing the code size (TF-M currently can deal with both raw and encoded data but we marked the previous solution as depreacted in that project).  It could be a solution, to include both in MCUboot.

- Optionally MCUboot could share all kinds of image properties one by one in a separate TLV (this would enable a general measured boot mechanism - the list of properties could be based on the PSA's recommendations but could be expanded)
- New option for imgtool to provide a path of any arbitrary/custom binary data (can be the CBOR encoded boot record data) and add it to the image as a new type of TLV. If MCUboot finds this in the TLV area then it simply copies the binary data (with the corresponding TVL headers) to the shared memory.

What do you think about it @d3zd3z ?

Of course from TF-M's perspective it would make things a bit more complicated but I think it's acceptable to make the data sharing from MCUboot more generic.

boot/bootutil/include/bootutil/boot_status.h
tamasban: Hashing the data seems a bit overkill. Basically we do not want to protect against malicious modification. Shared memory must be only secure accessible. Hashing would require to have the same hashing capability in the runtime firmware, which might not the case. I  think the best would be to increase the size of the magic value. @d3zd3z What do you think how many bits would provide enough protection? 128?

d3zd3z: 128 is a bit excessive, 64 bits should be sufficient for a magic value. 32 might even be enough.
It is a little better, because this value will only matter in a configuration where somehow someone configures an mcuboot without shared data and a TFM that has it. It shouldn't happen, but might.

d3zd3z:
I think this approach is reasonable. There is a place to put additional data if someone wanted to do their own thing (say without TFM, but their own custom secure side).

boot/bootutil/include/bootutil/boot_status.h
tamasban: 32 bit seems reasonable, because as you wrote miss configuration also needs to happen. TF-M and MCUBoot has the same config option to turn on/off shared data support at compile time. So issue could happen if miss-configuration happens. i.e.: TF-M expects shared data, but bootloader does not provide it and additionally the random data in the shared memory equals with the magic number. TF-M v1.0 release is very close, so if magic value will be changed then this cause an incompatibility. It can be handled in later TF-M version, but might better to avoid. @d3zd3z: Do you have any idea to avoid?

davidvincze:
> I think this approach is reasonable. There is a place to put additional data if someone wanted to do their own thing (say without TFM, but their own custom secure side).

I would like to update my last comments. I would rather drop the first bullet point and provide all these data in the current way (calling the `boot_save_boot_status()` function which copies and updates the CBOR encoded data). This could represent the general part of the measured boot mechanism and **in addition** I would add a new function to the interface which could be implemented per target OS (mcuboot port) and can be used to add any additional data to the shared memory area using the general `boot_add_data_to_shared_area()` function.

If it's needed we could also move/hide the content of `boot_save_boot_status()` into this target specific function and modify the `imtool` as I previously described in the second point.
But this could happen in a later time if it'll be needed. At the moment probabaly more people just want to use it than to start implementing it.

davidvincze:
> > > > The list of shared measurements is based on the recommendations of Arm's Platform Security Architecture (PSA) (https://developer.arm.com/architectures/security-architectures/platform-security-architecture) because another purpose of this patch is to support compliance with it. By encoding part of the boot data at build time the increase in code size can be minimized and the CBOR encoded data can be used directly later in the creation of an initial attestation token which is also required by PSA.
> > > 
> > > 
> > > Could you please provide direct links to the relevant documents?
> > 
> > 
> > @utzig if you're still interested (or anyone else), you can directly download the PSA documents from here: https://developer.arm.com/architectures/security-architectures/platform-security-architecture
> 
> This is the same link as before, isn't it possible to just directly point to the relevant PDF or whatever?

It's the official source of the latest documents. Unfortunately I couldn't find direct links to just open these documents.

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,233 @@
...
+    uint8_t *next_tlv;
+    uint16_t boot_data_size;
+    uintptr_t tlv_end, offset;
+
+    boot_data = (struct shared_boot_data *)MCUBOOT_SHARED_DATA_BASE;
+
+    /* Check whether first time to call this function. If does then initialise
+     * shared data area.
+     */
+    if (shared_memory_init_done == false) {
```
d3zd3z: This is declared as a bool. I think it is clearer to just say: `if (!shared_memory_init_done) {`.

boot/bootutil/src/boot_record.c
```diff
@@ -0,0 +1,233 @@
...
+    tlv_entry.tlv_type = SET_TLV_TYPE(major_type, minor_type);
+    tlv_entry.tlv_len  = size;
+
+    if (!boot_u16_safe_add(&boot_data_size, boot_data->header.tlv_tot_len,
+                           SHARED_DATA_ENTRY_SIZE(size))) {
+        return SHARED_MEMORY_GEN_ERROR;
+    }
+
+    /* Verify overflow of shared area */
+    if (boot_data_size > MCUBOOT_SHARED_DATA_SIZE) {
```
d3zd3z: If we get an overflow, do we want to keep the partially filled shared data? Or should we invalidate the shared block?

davidvincze: Yes we want to keep it. We should not invalidate the block, there can be valid data in it. It only says that "sorry, this piece of data doesn't fit into the shared area, not enough space left." The only thing the `boot_add_data_to_shared_area()` may do prior to this check is initializing the shared area.

PR #677, opened 2020-03-04
Cypress: custom protected tlv

ihos-cypress:
Signed-off-by: Ihor Slabkyy <ihor.slabkyy@cypress.com>


utzig:
Signed-off-by email should be `ihos@`.

> Add a way to specify custom TLVs, which will be placed in the protected area. Tag, length and value come from command line.

Please add the text to the commit message.

utzig:
> Signed-off-by: Ihor Slabkyy [ihos@cypress.com](mailto:ihos@cypress.com)

Not sure if you added your signed-off-by manually, but we expect the folowing format: `Signed-off-by: Author Name <author@email.com>`, you are missing the `<>` characters.

scripts/imgtool/main.py
```diff
@@ -277,7 +279,21 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
             # FIXME
             raise click.UsageError("Signing and encryption must use the same "
                                    "type of key")
-    img.create(key, enckey, dependencies)
+
+    # Get list of custom protected TLVs from the command-line
+    custom_tlv = {}
```
utzig: `custom_tlvs` is a better suited name

scripts/imgtool/main.py
```diff
@@ -277,7 +279,21 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
...
+    for tlv in cptlv:
+        tag = int(tlv[0], 16) if tlv[0].startswith('0x') else int(tlv[0])
+        if tag in custom_tlv:
+            raise ValueError('Custom TLV %s already exists.' % hex(tag))
+        elif tag in image.TLV_VALUES.values():
+            raise ValueError('Custom TLV %s conflicts with predefined TLV.' % hex(tag))
```
utzig: raise `click.UsageError` for both errors

scripts/imgtool/main.py
```diff
@@ -277,7 +279,21 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
...
+        else:
+            length = tlv[1]
+            value = tlv[2]
+            custom_tlv[tag] = struct.pack(length, int(value))
```
utzig: I am not entirely sure what you are trying to achieve here, but this is probably incorrect, what you should do is pass a tuple like this `custom_tlvs[tag] = (length, value)` and leave packing to be done in the creation routine which can deal with endianness.

scripts/imgtool/main.py
```diff
@@ -277,7 +279,21 @@ def sign(key, align, version, header_size, pad_header, slot_size, pad,
...
+    custom_tlv = {}
+    for tlv in cptlv:
+        tag = int(tlv[0], 16) if tlv[0].startswith('0x') else int(tlv[0])
+        if tag in custom_tlv:
+            raise ValueError('Custom TLV %s already exists.' % hex(tag))
+        elif tag in image.TLV_VALUES.values():
+            raise ValueError('Custom TLV %s conflicts with predefined TLV.' % hex(tag))
+        else:
+            length = tlv[1]
+            value = tlv[2]
```
utzig: I would convert both to `int` already: `length, value = int(tlv[1], 0), int(tlv[2], 0)`

scripts/imgtool/image.py
```diff
@@ -255,9 +258,10 @@ def ecies_p256_hkdf(self, enckey, plainkey):
             format=PublicFormat.UncompressedPoint)
         return cipherkey, ciphermac, pubk
 
-    def create(self, key, enckey, dependencies=None):
+    def create(self, key, enckey, dependencies=None, custom_protected_tlv=None):
         self.enckey = enckey
 
+        dependencies_num = 0
```
utzig: Why was this added? To me it looks like it is only initialized and use if `dependencies is not None` which should be OK.

scripts/imgtool/image.py
```diff
@@ -255,9 +258,10 @@ def ecies_p256_hkdf(self, enckey, plainkey):
             format=PublicFormat.UncompressedPoint)
         return cipherkey, ciphermac, pubk
 
-    def create(self, key, enckey, dependencies=None):
+    def create(self, key, enckey, dependencies=None, custom_protected_tlv=None):
```
utzig: `custom_tlvs=None` should be OK.

scripts/imgtool/image.py
```diff
@@ -275,6 +279,12 @@ def create(self, key, enckey, dependencies=None):
             # Add the size of the TLV info header
             protected_tlv_size += TLV_INFO_SIZE
 
+        if custom_protected_tlv:
+            if protected_tlv_size == 0:
+                protected_tlv_size = TLV_INFO_SIZE
+            for tag, value in custom_protected_tlv.items():
```
utzig: If passing a tuple you could use `for _, (length, _) in custom_tlvs.items()`, and the size would be `TLV_SIZE + length`

scripts/imgtool/image.py
```diff
@@ -97,12 +97,15 @@ def __init__(self, endian, magic=TLV_INFO_MAGIC):
     def __len__(self):
         return TLV_INFO_SIZE + len(self.buf)
 
-    def add(self, kind, payload):
+    def add(self, kind, payload, is_custom=False):
```
d3zd3z: I'm more inclined to just have the callers lookup the kinds, pass the numeric kind values into this function, and not need extra complexity of the `is_custom` argument.

scripts/imgtool/main.py
```diff
@@ -252,13 +253,14 @@ def convert(self, value, param, ctx):
 @click.option('--align', type=click.Choice(['1', '2', '4', '8']),
               required=True)
 @click.option('-k', '--key', metavar='filename')
+@click.option('--cptlv', required=False, nargs=3, default=[], multiple=True, help='Custom protected TLV')
```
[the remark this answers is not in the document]
d3zd3z: I agree with the assumption about nobody adding a non-protected TLV. The only non protected TLV entries that make sense are those that have to do with the signature itself.

scripts/imgtool/main.py
ihos-cypress: The idea here was to specify the length in python struct format: 'B' for byte, 'I' for uint.
Example: `--custom-tlv '0x81' 'I' '0xAABBCCDD'`

However, now I see another two ways:
 - specify length as integer `--custom-tlv '0x81' '4' '0xAABBCCDD'`
 - do not specify the length at all and get get it from the value length
In both cases try to convert value to numeric type if there is 0x prefix, otherwise convert to char array (`s` format from the https://docs.python.org/3.7/library/struct.html)

scripts/imgtool/image.py
```diff
@@ -275,6 +279,12 @@ def create(self, key, enckey, dependencies=None):
             # Add the size of the TLV info header
             protected_tlv_size += TLV_INFO_SIZE
 
+        if custom_protected_tlv:
+            if protected_tlv_size == 0:
+                protected_tlv_size = TLV_INFO_SIZE
```
davidvincze, line 284: Please remove this, the `protected_tlv_size` is later increased by TLV_INFO_SIZE if necessary (protected_tlv_size != 0).

utzig:
@ihos-cypress Please rebase on latest master, squash both commits into one, and also reformat your commit message with at most 72 columns.

utzig:
@ihos-cypress A commit is not a PR, the message should say just: "Add a way to specify custom TLVs.". Also please format the remaining sentence to be 72 columns of witdth.

scripts/imgtool/main.py
```diff
@@ -305,7 +311,24 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
...
         key.pad_sig = True
 
-    img.create(key, public_key_format, enckey, dependencies, boot_record)
+    # Get list of custom protected TLVs from the command-line
+    custom_tlvs = {}
+    for tlv in custom_tlv:
+        tag = int(tlv[0], 0)
+        if tag in custom_tlvs:
+            raise click.UsageError('Custom TLV %s already exists.' % hex(tag))
+        elif tag in image.TLV_VALUES.values():
```
utzig, line 320: This could be an `if`.

scripts/imgtool/main.py
```diff
@@ -305,7 +311,24 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
...
+            raise click.UsageError(
+                'Custom TLV %s conflicts with predefined TLV.' % hex(tag))
+        else:
```
utzig, line 323: This `else` can be removed and the block below dedented.

scripts/imgtool/main.py
```diff
@@ -305,7 +311,24 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
...
+            value = tlv[1]
+            if value.startswith('0x'):
+                custom_tlvs[tag] = bytes.fromhex(value[2:])
```
utzig, line 326: You could also raise a helpful message if `len(value[2:])` is odd.

scripts/imgtool/image.py
```diff
@@ -328,8 +331,20 @@ def create(self, key, public_key_format, enckey, dependencies=None,
             # Add the size of the TLV info header
             protected_tlv_size += TLV_INFO_SIZE
 
-        # At this point the image is already on the payload, this adds
-        # the header to the payload as well
+        if custom_tlvs:
+            for _, value in custom_tlvs.items():
```
utzig, line 335: Could be written `for value in custom_tlvs.values():`

utzig:
@ihos-cypress The code looks good now, but while doing a bit of testing the image is not validated. I will try to do a bit of debugging, but I am short on time right now.

scripts/imgtool/image.py
```diff
@@ -328,8 +331,20 @@ def create(self, key, public_key_format, enckey, dependencies=None,
             # Add the size of the TLV info header
             protected_tlv_size += TLV_INFO_SIZE
 
-        # At this point the image is already on the payload, this adds
-        # the header to the payload as well
+        if custom_tlvs:
```
utzig, line 334: This block of code must happen before the previous check for `if protected_tlv_size != 0` above to correctly set the size.

ihos-cypress: Done

d3zd3z:
I'm ok with this current approach, but if we want anything more complex, we will need to find a better way to specify it. The only thing that is a little potentially confusing is that '0x12345678' is not the integer it looks like, but the bytes 0x12, 0x34, 0x56, 0x78. But, if this meets your current needs, we can come up with something more complex if needed.

PR #679, opened 2020-03-10
fix #675 ECDSA verification fail when padding is added

### LINKED ISSUE #675: ecdsa verification fail
imgtool.py can't verify signature when its length is 70 or 71 bytes when ECDSA P256 is used.
It is caused by padding of signature to 72 bytes.

    def sign(self, payload):
        # To make fixed length, pad with one or two zeros.
        sig = self.raw_sign(payload)
        sig += b'\000' * (self.sig_len() - len(sig))
        return sig

But in image.py, the sign is called before signature TLV is added. Could we remove this padding?

pakls:
The sign method may add 1 or 2 '\0' character padding to ensure the signature TLV is 72 bytes. But the verification function didn't strip the paddings before invoking crypto libraries. It causes failure when the signature is 70 or 71 bytes.

### DIFF
--- a/scripts/imgtool/keys/ecdsa.py
+++ b/scripts/imgtool/keys/ecdsa.py
@@ -61,6 +61,8 @@ class ECDSA256P1Public(KeyClass):
         return 72
 
     def verify(self, signature, payload):
+        # strip possible paddings added during sign
+        signature = signature[:signature[1] + 2]
         k = self.key
         if isinstance(self.key, ec.EllipticCurvePrivateKey):
             k = self.key.public_key()


d3zd3z:
This will soon be a duplicate of a change I am about to push.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -61,6 +61,8 @@ def sig_len(self):
         return 72
 
     def verify(self, signature, payload):
+        # strip possible paddings added during sign
+        signature = signature[:signature[1] + 2]
```
d3zd3z: I don't quite see how this will work, the length in ASN.1 is a number of elements, not a number of bytes.

d3zd3z:
#680 should make this unnecessary.

scripts/imgtool/keys/ecdsa.py
pakls: I'm not an expert of ASN.1. Per my knowledge, the second byte (or more, if bit-7 of first byte is set) is length in the case of OCTET STRING.

I have dumped a signature and the observation is the same. With regard to the padding option (default adding it, or not adding on user's request) in future versions, verify function in imgtool.py still need to be functional for both of them. No matter what is supported in MCUboot boot loader in actual embedded systems.

pakls: For example, an ECDSA signature I've added is:
304502206961ebb1....
30 45 (30=SEQUENCE, 45=69 bytes)
   02 20 (INTEGER 32 bytes)
      69 61 eb b1 ....

pakls:
> #680 should make this unnecessary.

Nope. It is not the case. When you sign with option to add padding (for compatibility with previous MCUboot), the verify function will fail as it does now if there is padding.

So, I believe this patch is still necessary.

d3zd3z:
This should be eliminated by merging #680.

d3zd3z:
> Nope. It is not the case. When you sign with option to add padding (for compatibility with previous MCUboot), the verify function will fail as it does now if there is padding.

Don't pass the option. We're trying to get rid of padding, not add more things to work around it being there. I'm not understanding how you could have both an old version of MCUboot and a new version of MCUboot which would have a workaround for this problem?

scripts/imgtool/keys/ecdsa.py
d3zd3z: The signature is an ASN.1 sequence, not just an octet string. The length is encoded as you mentioned (in octets, I was getting confused working with CBOR recently).

pakls:
I'm not going to use this option or argument myself.

Whoever is using an earlier MCUBoot with padding, he/she shall have trouble verify their image on host due to the padding issue. This patch addresses that. He/she will be able to verify on the host machine with this patch.

utzig:
@d3zd3z @pakls Did a bit of testing here and I can confirm it fixes the issue for versions `<= 1.5.0`; `master` does not need this fix anymore, but at the same this won't break `master` so I'd rather merge it. I am assuming there are no other side-effects, can you think of any @d3zd3z?

utzig:
> We wouldn't want to back port this to old versions of MCUboot. If this was really an issue for older versions of MCUboot, it would be better to backport the patch that removes the requirement for the padding.

I can think of two approaches, either `imgtool` supporting older versions, or not. In the later case, then we should probably release `imgtool` with this fix as 1.5.1 or 1.5.0.1, whatever semver dictates!

d3zd3z:
I don't see any benefit to merging this. This change would allow old versions of mcuboot to work with newer versions of imgtool (that didn't specify this option, the purpose of the option being to support older versions of mcuboot). It offers no benefit to the current version of mcuboot, so there is no reason to put it on master. It might fix an old version of mcuboot, if we were to release that, but why would we then not just apply the same fix that makes master work.

This change adds extra code, and complexity, and doesn't solve a problem in any branch that it would be applied to.

Am I misunderstanding something?

d3zd3z:
I guess if we wanted to bring something like this in, we should at least compute the length correctly (if the high bit is set, using multiple bytes), or at least detect and fail (an ECDSA signature shouldn't be that long).

d3zd3z:
Oh, let's just put this in. I now agree with Fabio that it at least shouldn't hurt anything.

PR #680, opened 2020-03-10
Remove default padding from ECDSA signatures

d3zd3z:
Continuing from #621, make non padded ECDSA signatures the default. The `--pad-sig` flag is added to imgtool to be able to make images for old versions of MCUboot.

### DIFF
--- a/scripts/imgtool/keys/ecdsa.py
+++ b/scripts/imgtool/keys/ecdsa.py
@@ -52,12 +52,17 @@ class ECDSA256P1Public(KeyClass):
         return "ECDSA256"
 
     def sig_len(self):
-        # The DER encoding depends on the high bit, and can be
-        # anywhere from 70 to 72 bytes.  Because we have to fill in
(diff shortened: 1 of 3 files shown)


utzig:
@d3zd3z Needs rebase!

scripts/imgtool/keys/ecdsa.py
```diff
@@ -140,7 +146,10 @@ def raw_sign(self, payload):
...
-        # To make fixed length, pad with one or two zeros.
         sig = self.raw_sign(payload)
-        sig += b'\000' * (self.sig_len() - len(sig))
-        return sig
+        if self.pad_sig:
+            # To make fixed length, pad with one or two zeros.
+            sig += b'\000' * (self.sig_len() - len(sig))
+            return sig
+        else:
+            return sig
```
utzig: I think you could just dedent this `return` and remove the `else` and `return` above.

PR #683, opened 2020-03-16
zephyr: cleanup of ARM core before application boot

nvlsianpu:
This patch is needed as MCUBoot should be able to chain-load any
application, not only these built using zephyr.

Introduced cleanup on ARM core control register.
Might be required as for instance the application assumes
that it starts with thread mode configured as by default, not
according to zephyr-rtos configuration.

MCUBoot disables interrupt before application chain-load used
basepr register. This Patch introduce additional celenup on
NVIC register.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -120,6 +120,11 @@ config BOOT_SIGNATURE_KEY_FILE
 	  with the public key information will be written in a format expected by
 	  MCUboot.
 
+config MCUBOOT_CLEANUP_ARM_CORE
+	bool "Perform core cleanup before chain-load the application"
+	depends on CPU_CORTEX_M
+	default y
+
 config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
--- a/boot/zephyr/arm_cleanup.c
+++ b/boot/zephyr/arm_cleanup.c
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2020 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr.h>
+
+void cleanup_arm_nvic(void) {
+	/* Allow any pending interrupts to be recognized */
+	__ISB();
+	__disable_irq();
+
+	/* Disable NVIC interrupts */
+	for (u8_t i = 0; i < ARRAY_SIZE(NVIC->ICER); i++) {
+		NVIC->ICER[i] = 0xFFFFFFFF;
+	}
+	/* Clear pending NVIC interrupts */
+	for (u8_t i = 0; i < ARRAY_SIZE(NVIC->ICPR); i++) {
+		NVIC->ICPR[i] = 0xFFFFFFFF;
+	}
+}
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -44,6 +44,10 @@ const struct boot_uart_funcs boot_funcs = {
 #include <usb/class/usb_dfu.h>
 #endif
 
+#if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
+#include <arm_cleanup.h>
+#endif
+
 #if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE)
 #ifdef CONFIG_LOG_PROCESS_THREAD
 #warning "The log internal thread for log processing can't transfer the log"\
@@ -126,7 +130,13 @@ static void do_boot(struct boot_rsp *rsp)
     /* Disable the USB to prevent it from firing interrupts */
     usb_disable();
 #endif
+#if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
+    cleanup_arm_nvic(); /* cleanup NVIC registers */
+#endif
     __set_MSP(vt->msp);
+#if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
+    __set_CONTROL(0x00); /* application will configures core on its own */
+#endif
     ((void (*)(void))vt->reset)();
 }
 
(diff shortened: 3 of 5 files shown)


nvlsianpu:
@hakonfam @ktaborowski

boot/zephyr/Kconfig
```diff
@@ -120,6 +120,11 @@ config BOOT_SIGNATURE_KEY_FILE
 	  with the public key information will be written in a format expected by
 	  MCUboot.
 
+config CLEANUP_ARM_CORE
```
hakonfam: As this is defined in mcuboot, I assume it will only be used in mcuboot source code, should we prefix this with "MCUBOOT_" ?

boot/zephyr/Kconfig
```diff
@@ -120,6 +120,11 @@ config BOOT_SIGNATURE_KEY_FILE
...
+	bool "does core cleanup before chain-load the application"
```
hakonfam:
```suggestion
	bool "Perform core cleanup before chain-load the application"
```

boot/zephyr/main.c
```diff
@@ -125,8 +129,14 @@ static void do_boot(struct boot_rsp *rsp)
 #ifdef CONFIG_USB
     /* Disable the USB to prevent it from firing interrupts */
     usb_disable();
+#endif
+#if CONFIG_CLEANUP_ARM_CORE
+    cleanup_arm_nvic(); /* cleanup NVIC registers */
 #endif
     __set_MSP(vt->msp);
+#if CONFIG_CLEANUP_ARM_CORE
+    __set_CONTROL(0x00); /* application will configures core on its own */
```
hakonfam: Is there a reason why this is not in a function inside the cleanup_arm module?

nvlsianpu: Better to execute mcuboot call `cleanup_arm_nvic()` without so crucial reconfiguration of the core.
So, This is done intentionally after MSP have been set. Setting control CPU register is something which should be very last operation before jump to the application.

boot/zephyr/Kconfig
nvlsianpu: prefix added

boot/zephyr/Kconfig
nvlsianpu: fixed

boot/zephyr/main.c
hakonfam: makes sense.

ioannisg:
> This seems like a good change to me.
> 
> @ioannisg, what do you think?

I am a bit late for the party, here, but I'll leave my comments, anyways - for future improvements, if applicable.

boot/zephyr/arm_cleanup.c
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr.h>
+
+void cleanup_arm_nvic(void) {
+	/* Allow any pending interrupts to be recognized */
+	__ISB();
+	__disable_irq();
```
ioannisg: How do you do disable IRQ - setting PRIMASK to 1? If so, ISB is redundant here.

boot/zephyr/main.c
ioannisg: MSR CONTROL needs to be followed by ISB - that's a recommendation from ARM, in particular if it writes to the SPSEL.

boot/zephyr/main.c
```diff
@@ -125,8 +129,14 @@ static void do_boot(struct boot_rsp *rsp)
 #ifdef CONFIG_USB
     /* Disable the USB to prevent it from firing interrupts */
     usb_disable();
+#endif
+#if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
+    cleanup_arm_nvic(); /* cleanup NVIC registers */
 #endif
     __set_MSP(vt->msp);
```
ioannisg: __set_MSP is not part of the change in this patch, however, I have a question here: **does do_boot() function uses MSP or PSP?** If it uses MSP there might be an issue here. `vt `is a local variable saved in stack. After __set_MSP is changing the stack pointer, normally you  should not rely on the compiler properly de-referencing `vt`, it usually works but some strange compiler settings might make it not work.

I would consider it a dangerous thing, @d3zd3z @nvlsianpu .

mbolivar-nordic:
> > @ioannisg, what do you think?
> 
> I am a bit late for the party, here, but I'll leave my comments, anyways - for future improvements, if applicable.

@d3zd3z @nvlsianpu do you plan on addressing these? They seem concerning to me.

boot/zephyr/main.c
d3zd3z: I think the only way to properly fix that would be to write the set-msp and the jump to reset in assembly. It is concerning to me that another patch has added a __set_CONTROL(0x00) call in between the MSP set and the reset, which might cause 'vt' to get written to the wrong place as well. Given that this is merged, I think this should get its own issue, though.

d3zd3z:
@mbolivar-nordic I'll make a separate issue to track them, since this is already merged.

boot/zephyr/main.c
d3zd3z: Thinking about this a little more, I don't think this will ever be a problem. Currently the _set calls are inline assembly. If they were ever changed to cause registers to be spilled to the stack, a lot more code than MCUboot would break.

mbolivar-nordic:
> @mbolivar-nordic I'll make a separate issue to track them, since this is already merged.

Thanks!

boot/zephyr/main.c
ioannisg: @d3zd3z , _set_MSP is inline assembly, that's correct. But vt in line 140 is accessed from C code, using the stack pointer, that is changed via _set_MSP.

The implementation of _set_MSP is as follows:
```
__STATIC_FORCEINLINE void __set_MSP(uint32_t topOfMainStack)
{
  __ASM volatile ("MSR msp, %0" : : "r" (topOfMainStack) : );
}
```
As you can see there is no compiler barrier, either, so stack-pointer based accesses later in the function do not take into account the stack pointer change.

boot/zephyr/main.c
nvlsianpu: addressed in https://github.com/JuulLabs-OSS/mcuboot/pull/856

PR #687, opened 2020-03-18
mynewt: Update latest tags in repository.yml

sjanc:
0-latest and 1-latest should point to latest stable release.

Signed-off-by: Szymon Janc <szymon.janc@codecoup.pl>


sjanc:
I'm not sure about 1.0-latest,   should it point to 1.0.0 instead?

utzig:
> I'm not sure about 1.0-latest, should it point to 1.0.0 instead?

I don't think `1.0-latest` is required.

PR #693, opened 2020-03-24
zephyr: Wrap sleep value with K_MSEC

nordic-krch:
Zephyr timeout API is changing and will use opaque value (k_timeout_t)
instead of raw values. K_MSEC is used to convert raw milliseconds value
to k_timeout_t.

This change is backward compatible so can be merged independently.

Signed-off-by: Krzysztof Chruscinski <krzysztof.chruscinski@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -55,7 +55,7 @@ const struct boot_uart_funcs boot_funcs = {
 #else
 #include <logging/log_ctrl.h>
 
-#define BOOT_LOG_PROCESSING_INTERVAL 30 /* [ms] */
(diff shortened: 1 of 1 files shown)


nvlsianpu:
@nordic-krch Should this be merged after the zephyr umbrella PR?

nordic-krch:
hm, don't know what is umbrella PR. In general, this is backward compatible since `K_MSEC` macro exists now `#define K_MSEC(ms)     (ms)`

nordic-krch:
it would be good to get that fairly soon because big PR relies on it (https://github.com/zephyrproject-rtos/zephyr/pull/23382) and we would like to merge it asap due rebasing cost.

PR #698, opened 2020-03-30
zephyr: boot_serial: Use K_MSEC() with k_sleep

carlescufi:
k_sleep() requires explicit conversion of the unit since it takes an
opaque type starting with:
https://github.com/zephyrproject-rtos/zephyr/pull/23382


carlescufi:
CC @nashif and @andyross

PR #700, opened 2020-03-31
zephyr: Use K_MSEC in k_sleep call

nordic-krch:
Zephyr timeout API is changing and will use opaque value (k_timeout_t)
instead of raw values. K_MSEC is used to convert raw milliseconds value
to k_timeout_t.

This change is backward compatible so can be merged independently.

Signed-off-by: Krzysztof Chruscinski <krzysztof.chruscinski@nordicsemi.no>


nordic-krch:
@nvlsianpu there was one more place where change was needed.

nvlsianpu:
already applied by #698

PR #701, opened 2020-03-31
scripts: assemble.py: switch to using devicetree_legacy_unfixed.h

henrikbrixandersen:
Switch to using the devicetree_legacy_unfixed.h header for extracting flash partition information.

Zephyr has switched to using a new gen_defines.py script and a new set of macros/defines for devicetree code generation for all nodes except flash partitions, which are still in the legacy format.

Signed-off-by: Henrik Brix Andersen <henrik@brixandersen.dk>

### DIFF
--- a/scripts/assemble.py
+++ b/scripts/assemble.py
@@ -50,7 +50,7 @@ class Assembly():
     def find_slots(self, bootdir):
         offsets = {}
         sizes = {}
-        with open(os.path.join(bootdir, 'zephyr', 'include', 'generated', 'devicetree_unfixed.h'), 'r') as fd:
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
Late to the party, but I think it would be nice to try using devicetree_legacy_unfixed.h first, then fall back on devicetree_unfixed.h if that does not exist.

henrikbrixandersen:
> Late to the party, but I think it would be nice to try using devicetree_legacy_unfixed.h first, then fall back on devicetree_unfixed.h if that does not exist.

How come? I believe other changes were pushed to mcuboot master, that prevents it from compiling with older versions of Zephyr?

mbolivar-nordic:
> How come? I believe other changes were pushed to mcuboot master, that prevents it from compiling with older versions of Zephyr?

Sigh. I was unaware of that.

PR #704, opened 2020-04-01
Add HW key support for image validation

davidvincze:
Enable the public key(s) (used for image authentication) to be removed from MCUboot and be appended to the image instead. In this case the key or its hash must be provisioned to the device and MCUboot must be able to retrieve the key-hash from the hardware to compare it with the calculated hash of the public key from the image manifest. After validating the key it can be used for image authentication. This way MCUboot can be independent from the public key(s) (they are not embedded in it).
This feature was already available in TF-M's MCUboot fork, this PR covers its integration with the original MCUboot code.

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -74,6 +74,7 @@ struct flash_area;
  *   2nd one is the actual signature.
  */
 #define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
+#define IMAGE_TLV_PUBKEY            0x02   /* public key */
 #define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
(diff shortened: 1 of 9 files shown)


boot/bootutil/include/bootutil/image.h
```diff
@@ -74,6 +74,7 @@ struct flash_area;
  *   2nd one is the actual signature.
  */
 #define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
+#define IMAGE_TLV_KEY               0x02   /* public key */
```
d3zd3z, line 77: I wonder if we should call this PUBKEY. I realize that KEYHASH already doesn't have that, and we aren't going to be storing the private key. It just feels weird to call it just "KEY" though.

davidvincze: I understand what you mean, I've renamed it to `PUBKEY`.

Issue #699, opened 2020-03-30
Building MCUBoot for Zephyr fails on multiple ARM boards


mnkp:
Building latest version of MCUBoot for Zephyr, e.g.
```
west build -b stm32f0_disco bootloader/mcuboot/boot/zephyr/
```
is causing compile time error.
```
bootloader/mcuboot/boot/zephyr/arm_cleanup.c:15:34: error: 'NVIC' undeclared (first use in this function)
   15 |  for (u8_t i = 0; i < ARRAY_SIZE(NVIC->ICER); i++) {
```
The issue can be seen on multiple ARM boards and was introduced by #683.

The `arm_cleanup.c` source file relies on CMSIS, however it includes only `zephyr.h` which is used to provide kernel API, not necessarily give access to CPU internals.

It seems like at present it is possible to compile MCUBoot only on those ARM boards which enable MPU since cmsis becomes available through some implicit header file dependencies.

PR #705, opened 2020-04-02
zephyr: arm_cleanup.c: use <.../cortex_m/cmsis.h>

mnkp:
Use <arch/arm/aarch32/cortex_m/cmsis.h> include instead of <zephyr.h> to
get access to Cortex-M CMSIS API. zephyr.h provides kernel API, does not
guarantee access to CPU internals.

Fixes: #699 "Building MCUBoot for Zephyr fails on multiple ARM boards"

### DIFF
--- a/boot/zephyr/arm_cleanup.c
+++ b/boot/zephyr/arm_cleanup.c
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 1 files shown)


nvlsianpu:
Zephyr CI kicker: https://github.com/zephyrproject-rtos/zephyr/pull/24010

Issue #205, opened 2018-02-05
Implement dual-slot (non swap) image upgrade


d3zd3z:
Implement a mechanism that allows two different images to be deployed, one that runs in slot 0 and one that runs in slot 1. MCUboot's responsibility will then be to track which one to run, not to handle the complexity of swapping them.

leodesigner:
+1 
The similar way as it done in ESP8266 OTA update. It would be very nice feature during development to preserve flash write count. In this case we must prepare two firmwares linked to the slot0 and slot1. And later during the OTA the right image will be chosen for non used currently slot.

tamasban:
We have an implementation, it will be published next week on Lianro gerrit TF-M repo. I will post a link for the review here.

leodesigner:
Great !
It is really good option. I hope it will be easy to add to the Zephyr cmake linking option for slot1.

uday-agarwal:
@tamasban can you please post the link if you have it?

tamasban:
@uday-agarwal
Hi, 
We have a forked MCUBoot repo in TF-M. Currently the non-swapping update only up-streamed to there:
https://git.trustedfirmware.org/trusted-firmware-m.git/tree/bl2/ext/mcuboot

The two repo is currently diverged but will be synced.

Tamas

uday-agarwal:
Thanks a lot Tamas.

d3zd3z:
I believe this is called "noswap" in the TF-M code. Assigning to Tamas for tracking.

nvlsianpu:
@hakonfam I received some interest on this feature request - what do you think, do you see any obstacles?
Implementation shouldn't be so complex I believe - so I would provide  a POC.

hakonfam:
> @hakonfam I received some interest on this feature request - what do you think, do you see any obstacles?
> Implementation shouldn't be so complex I believe - so I would provide a POC.

We already have this feature downstream in NCS. It will be generalized in what is called "jerryboot" to which there is an open PR.

Laczen:
@hakonfam, how do you handle encryption in such a setup?

hakonfam:
> @hakonfam, how do you handle encryption in such a setup?

Currently it is not supported.
But the the way I think it should work is that it would be up to the component storing the update to perform the decryption.
The decryption would be performed using an API exposed from the bootloader.
This allows the application to read the cleartext update, which is OK imo, since its an updated version of itself.

utzig:
[...]

PR #712, opened 2020-04-07
imgtool: Set RAM_LOAD flag in header when needed

davidvincze:
Set the RAM_LOAD flag in the image header when a load address was passed
to the imgtool script, indicating that the image should be loaded into
RAM and run from there.

### DIFF
--- a/scripts/imgtool/__init__.py
+++ b/scripts/imgtool/__init__.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-imgtool_version = "1.6.0a1"
+imgtool_version = "1.6.0a2"
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -49,6 +49,7 @@ MAX_SW_TYPE_LENGTH = 12  # Bytes
 IMAGE_F = {
         'PIC':                   0x0000001,
         'NON_BOOTABLE':          0x0000010,
+        'RAM_LOAD':              0x0000020,
         'ENCRYPTED':             0x0000004,
 }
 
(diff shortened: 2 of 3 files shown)


davidvincze:
@hakonfam you might be interested in this PR since you added the `--load-addr` option to `imgtool`.

nvlsianpu:
Shouldn't this be flag for `Load address for image when it is in its primary slot or when it should run from RAM.`

RAM load ind in place execution are options that exclude one another.

might be also desired for https://github.com/JuulLabs-OSS/mcuboot/issues/205

@hakonfam I Need your voice here.

davidvincze:
@nvlsianpu Sorry, are you saying that the load address might be required for in place exection as well?

davidvincze:
> Shouldn't this be flag for `Load address for image when it is in its primary slot or when it should run from RAM.`
> 
> RAM load ind in place execution are options that exclude one another.
> 
> might be also desired for #205
> 
> @hakonfam I Need your voice here.

@nvlsianpu Sorry, are you saying that the load address might be required for in place exection as well?

davidvincze:
From image.h:
```
/*
 * Indicates that this image should be loaded into RAM instead of run
 * directly from flash.  The address to load should be in the
 * ih_load_addr field of the header.
 */
#define IMAGE_F_RAM_LOAD                 0x00000020
```

nvlsianpu:
After rethinking, not necessary but maybe.
Of course the flag is for ram load only.

What you are doing here looks good, just want to consult this: Hakon had a conception on that, let's wait on his answer.

nvlsianpu:
@davidvincze https://github.com/JuulLabs-OSS/mcuboot/pull/535#issuecomment-526715495
Looks like you can do what you wish here, excuse me for distracting you.

d3zd3z:
I don't think this makes any sense for #205, each image would just be built to run in place, using the mechanisms from that load.
Now, if we are loading images into RAM, swap is kind of pointless, as we can run either image without having to move them around.

utzig:
@davidvincze You might want to add a new commit bumping `imgtool` to 1.6.0-alpha3, because the commit in this PR was discarded since I added one with the same changes in the PR merged previously.

davidvincze:
> @davidvincze You might want to add a new commit bumping `imgtool` to 1.6.0-alpha3, because the commit in this PR was discarded since I added one with the same changes in the PR merged previously.

Thanks, I've just did: [#712](https://github.com/JuulLabs-OSS/mcuboot/pull/712)

PR #717, opened 2020-04-15
Use _Static_assert on all platforms

d3zd3z:
This macro is part of the C11 standard.  Remove the use of a specialized macro
just for Zephyr, and use the standard macro for all platforms.  This reduces
dependencies on specific versions of Zephyr

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -27,10 +27,6 @@
 #include <inttypes.h>
 #include <stdbool.h>
 
-#ifdef __ZEPHYR__
-#include <toolchain/gcc.h>
-#endif
-
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -135,13 +131,8 @@ struct image_tlv {
 #define MUST_DECRYPT(fap, idx, hdr) \
     ((fap)->fa_id == FLASH_AREA_IMAGE_SECONDARY(idx) && IS_ENCRYPTED(hdr))
 
-#ifdef __ZEPHYR__
-BUILD_ASSERT(sizeof(struct image_header) == IMAGE_HEADER_SIZE,
-	     "struct image_header not required size");
-#else
 _Static_assert(sizeof(struct image_header) == IMAGE_HEADER_SIZE,
                "struct image_header not required size");
-#endif
 
 struct enc_key_data;
 int bootutil_img_validate(struct enc_key_data *enc_state, int image_index,


nvlsianpu:
I'm not sure whether all possible zephyr platform supports that.

utzig:
> I'm not sure whether all possible zephyr platform supports that.

It could also be `#if defined(__ZEPHYR__) && __STDC_VERSION__ < 199901L` to avoid the known issues and provide retro-compatibility.

d3zd3z:
> I'm not sure whether all possible zephyr platform supports that.

I wonder if there are any platforms of concern that wouldn't be supported. Static_assert is defined as part of C11, and seems more likely to be supported than a Zephyr define that keeps changing.

d3zd3z:
I don't think it has anything to do with Zephyr version, though. It is a feature of Gcc. I'm at least not aware of any targets MCUboot is built on that don't use gcc. It was added to gcc in 2011. I don't think there has ever been a Zephyr SDK that didn't support it. I really don't know why Zephyr even has its own macro, it just passes it through to `_Static_assert`.

d3zd3z:
Any recommendations on how to keep BUILD_ASSERT? It only works in `master` which is against our normal rule of working with the latest released version of Zephyr?
Also, are we really adding support for compiler to Zephyr that don't implement recent versions of the C language?

d3zd3z:
> It could also be #if defined(__ZEPHYR__) && __STDC_VERSION__ < 199901L to avoid the known issues and provide retro-compatibility.

I have doubts that a compiler that doesn't implement a recent C standard is going to define `__STDC_VERSION__` correctly, either.

So, the real problem here is that we are using functionality of Zephyr that hasn't been released. There are no current versions of Zephyr where this BUILD_ASSERT macro is needed. What about removing the Zephyr specific define, and if/when we ever get a version of Zephyr that needs it, we put something back. Hopefully, by then, there will have been a release or two of Zephyr that makes use of BUILD_ASSERT and we won't need to conditionalize it.

tejlmand:
@mped-oticon @daor-oti one of you probably also have some thoughts on this, as I fear this is not supported in your compiler.

daor-oti:
> @mped-oticon @daor-oti one of you probably also have some thoughts on this, as I fear this is not supported in your compiler.

@tejlmand Thank you for including us, but in this case our compiler actually supports the _Static_assert. In fact, our toolchain header file defines BUILD_ASSERT as _Static_assert.

tejlmand:
> 
> 
> > @mped-oticon @daor-oti one of you probably also have some thoughts on this, as I fear this is not supported in your compiler.
> 
> @tejlmand Thank you for including us, but in this case our compiler actually supports the _Static_assert. In fact, our toolchain header file defines BUILD_ASSERT as _Static_assert.

Great, thanks for the update.

nvlsianpu:
> Any recommendations on how to keep BUILD_ASSERT? It only works in master which is against our normal rule of working with the latest released version of Zephyr?

As i remember we ensure working with master to master + zephyr has exact reference in the west manifest.

d3zd3z:
> As i remember we ensure working with master to master + zephyr has exact reference in the west manifest.

We are trying to make a release of MCUboot, and the requirement is that releases need to work with the latest release of Zephyr as well. That means we either need to wait for the next Zephyr release, or figure out how to not depend on this feature.

I'm seeing one desired compiler supporting the static assertions, though.

mbolivar-nordic:
> That means we either need to wait for the next Zephyr release, or figure out how to not depend on this feature.

What is the problem here?

MCUboot has an RTOS abstraction layer built for purposes like this (mcuboot_config.h).

Zephyr has APIs for accessing the kernel version.

It seems fairly straightforward to combine these in a way that lets bootutil not depend on a specific RTOS version or C standard level but still get static asserts, no?

d3zd3z:
> It seems fairly straightforward to combine these in a way that lets bootutil not depend on a specific RTOS version or C standard level but still get static asserts, no?

Mainly because I'm not yet convinced we are ever going to be building MCUboot with a toolchain that doesn't support `_Static_assert`. I've seen no evidence yet that we have any current platforms. It seems silly to put in an ifdef for Zephyr version that would only be needed for a few months.

d3zd3z:
I'm kind of questioning the original patch that added this Zephyr-specific build assert macro, but did so in a way that would only ever work with gcc, and hence always end up using `_Static_assert`. Given that this could wouldn't have ever built on a system that didn't have `_Static_assert`, and there don't appear to be any targets that have don't support this, I'd like to defer "fixing" this until it actually becomes a problem.

I'd like to get consent on this before I just go ahead an merge this, but given that it is blocking a release, I may merge it anyway.

utzig:
> I'd like to get consent on this before I just go ahead an merge this, but given that it is blocking a release, I may merge it anyway.

IMO this is fine, I've not seen a single practical  example of where this wouldn't work in this thread.

mbolivar-nordic:
I'm certainly not trying to block the release, but I don't understand why this is a blocker. Do what you will.

PR #716, opened 2020-04-15
Preps for 1.6.0-rc1

d3zd3z:
Update version fields, and add release notes for the 1.6.0-rc1 release.

Signed-off-by: David Brown <david.brown@linaro.org>

PR #718, opened 2020-04-16
travis: add imgtool publishing

utzig:


utzig:
This might be a bit flaky and it's kinda annoying to test correctly, but I believe it is functional. In my naivete testing it, I pushed `1.6.0rc1`, so for #716 the `imgtool` version will have to be updated to `1.6.0rc1.post1` instead.

ci/compare_versions.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
utzig: I consider `python` == `python3` because `python2` is dead! :-) I will change it though.

ci/imgtool_install.sh
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [[ $TRAVIS_PULL_REQUEST != "false" || $TRAVIS_BRANCH != "master" ]]; then
+    echo "Either a PR or not \"master\" branch, exiting"
+    exit 0
+fi
+
+pip install setuptools twine
```
d3zd3z, line 20: I wonder if this should explicitly call out `pip3`?

ci/imgtool_run.sh
```diff
@@ -0,0 +1,51 @@
...
+elif [[ $rc -eq 1 ]]; then
+    echo "Error parsing versions"
+    exit 1
+elif [[ $rc -eq 3 ]]; then
+    echo "Imgtool downgrade detected!"
+    exit 1
+fi
+
+rm -rf $DIST_DIR
+python setup.py sdist bdist_wheel
```
d3zd3z, line 49: Likewise, this should probably explicitly invoke python3.

ci/imgtool_install.sh
utzig: Done

ci/imgtool_run.sh
utzig: Done. Btw, travis already migrated a long time ago, so using just `python`/`pip` calls 3 anyway.

d3zd3z: That's good to know. I'm kind of amazed at how stubborn some of the distros are. Some are just removing 'python' entirely, and requiring python3 to always be explicit.

PR #719, opened 2020-04-17
Cypress: Add files to eliminate HAL dependency, build/compile pass

romanjoe:
Changes:

 - fixed RAM corruption when using HAL functions on CM0
 - switched to using only PDL functions
 - clean-up makefiles
 - added details about how to program hex files to the board


d3zd3z:
Ideally, this should probably have been at least 4 separate commits. Generally, if you have a list of things in the commit text, it suggests that those should probably each be a separate commit.

PR #725, opened 2020-04-24
zephyr/Kconfig: reference missing new Nordic boards keywords

nvlsianpu:
Updated boards reference names.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -313,8 +313,8 @@ if MCUBOOT_SERIAL
 
 choice
 	prompt "Serial device"
-	default BOOT_SERIAL_UART if !BOARD_NRF52840_PCA10059
-	default BOOT_SERIAL_CDC_ACM if BOARD_NRF52840_PCA10059
+	default BOOT_SERIAL_UART if !BOARD_NRF52840DONGLE_NRF52840
+	default BOOT_SERIAL_CDC_ACM if BOARD_NRF52840DONGLE_NRF52840
 
 config BOOT_SERIAL_UART
 	bool "UART"
@@ -343,9 +343,9 @@ config BOOT_SERIAL_DETECT_PORT
 
 config BOOT_SERIAL_DETECT_PIN
 	int "Pin to trigger serial recovery mode"
-	default 6 if BOARD_NRF9160_PCA10090
+	default 6 if BOARD_NRF9160DK_NRF9160
 	default 11 if BOARD_NRF52840DK_NRF52840
-	default 13 if BOARD_NRF52_PCA10040
+	default 13 if BOARD_NRF52DK_NRF52832
 	default 23 if BOARD_NRF5340_DK_NRF5340_CPUAPP || BOARD_NRF5340_DK_NRF5340_CPUAPPNS
 	help
 	  Pin on the serial detect port which triggers serial recovery mode.


nvlsianpu:
should go into the release

utzig:
I find it so funny that those changes keep coming after a release process is started even though those changes were done weeks ago in Zephyr! :rofl:

d3zd3z:
Does this work with Zephyr 2.2 or just master on Zephyr? I can try test builds on this target, but for release purposes, we should be targeting the last release of Zephyr, and ideally also work with master on Zephyr.

nvlsianpu:
> Does this work with Zephyr 2.2 or just master on Zephyr? I can try test builds on this target, but for release purposes, we should be targeting the last release of Zephyr, and ideally also work with master on Zephyr.


@d3zd3z We try to keep mcuboot compatible with zephyr-rtos master since  long of time. Fortunately since the 1.5.0 release zephyr API used by mcuboot has not changed much, although Kconfig and project configuration files are already match master configuration. USB initialization rutine was changed. as well.

As I remember we never release mcuboot compatible with a zephyr release version. Just can said the I took care so zephyr 2.2.0 references mcuboot 1.5.0 - but 2.2.0 was released after the mcuboot release - just keep compatibility.

This Patch should be merged within the release. Release should point certain SHA of zephyr-rtos.

If we want release mcuboot for a zephyr release - we should act as for mynewt (i believe 1.3.1 was release for mynewt 1.6.0).

cc @carlescufi  @utzig

nvlsianpu:
@davidvincze @utzig @d3zd3z Who can rtrigger the CI - It failed on git clone.

carlescufi:
It is non-trivial to maintain master to master compatibility @nvlsianpu @utzig and @d3zd3z. The reason is that, although Zephyr strives to maintain backwards compatibility in all cases, this is not always possible, and sometimes it only is possible through deprecation, which introduced warnings. 

This is one of the reasons we maintain https://github.com/zephyrproject-rtos/mcuboot, to ensure that upstream Zephyr always has a reference to a working MCUboot.

I think that the best we can aspire outside of that to is a table that documents which particular release of MCUBoot is compatible with a Zephyr release (eg. MCUBoot 1.6 is compatible with Zephyr 2.3, but not 2.2), but continuours master to master compatibility is unfeasible.

d3zd3z:
> It is non-trivial to maintain master to master compatibility @nvlsianpu @utzig and @d3zd3z.

The main issue is that we have to at least be compatible with one particular release of Zephyr. I'm concerned that this change will make it so there isn't any tagged version of Zephyr that MCUboot will work with. We had the same issue with the static assert macro change.

Some ways I could see solving this:

1. Stay compatible with master, and accept that the latest release will have warnings.
2. Have lots of ifdefs. We would also need to have a way to find these later and remove them as we progress through Zephyr releases. We won't actually be compatible with old releases of Zephyr, so extra defines that make it look like we are trying would just add complexity, and possibly confusion.
3. Change how the code works to not depend on the feature in Zephyr.

Right now, we kind of state that MCUboot master is compatible with Zephyr master, and that releases are compatible with the latest released version of Zephyr. I think we need to combine this, so we don't have a bunch of effort with each release: master on MCUboot needs to work with both Zephyr master, and the last released version of Zephyr. I'd also propose changing CI to build against both versions of Zephyr to check this.

d3zd3z:
I'll try restarting the CI, but looking at it, I think there might be something broken about CI, it is trying to pull your branch from the upstream repo, instead of yours.

utzig:
> I'll try restarting the CI, but looking at it, I think there might be something broken about CI, it is trying to pull your branch from the upstream repo, instead of yours.

I think the build issues can be safely ignored because the PR build passed; for some reason it's trying to also build the pushed branch, and I think that maybe @nvlsianpu had previously pushed the branch upstream instead of to his own fork. To fix this I guess you would have to close/reopen this PR.

nvlsianpu:
> and I think that maybe @nvlsianpu had previously pushed the branch upstream instead of to his own fork.

Yes, I did that accidentally.

> To fix this I guess you would have to close/reopen this PR.

Let's try it.

utzig:
> > and I think that maybe @nvlsianpu had previously pushed the branch upstream instead of to his own fork.
> 
> Yes, I did that accidentally.
> 
> > To fix this I guess you would have to close/reopen this PR.
> 
> Let's try it.

The PR build passed again but the branch build is still there (failed). I suggest you push the branch to upstream again, so we can trigger the CI for that, and once it passes the upstream branch can be deleted. I have no idea why both builds are grouped here but it must be some travis' feature, when it detects that a PR and a branch build consist of the same commits.

carlescufi:
> > It is non-trivial to maintain master to master compatibility @nvlsianpu @utzig and @d3zd3z.
> 
> The main issue is that we have to at least be compatible with one particular release of Zephyr. I'm concerned that this change will make it so there isn't any tagged version of Zephyr that MCUboot will work with. We had the same issue with the static assert macro change.
> 
> Some ways I could see solving this:
> 
> 1. Stay compatible with master, and accept that the latest release will have warnings.
> 2. Have lots of ifdefs. We would also need to have a way to find these later and remove them as we progress through Zephyr releases. We won't actually be compatible with old releases of Zephyr, so extra defines that make it look like we are trying would just add complexity, and possibly confusion.
> 3. Change how the code works to not depend on the feature in Zephyr.
> 
> Right now, we kind of state that MCUboot master is compatible with Zephyr master, and that releases are compatible with the latest released version of Zephyr. I think we need to combine this, so we don't have a bunch of effort with each release: master on MCUboot needs to work with both Zephyr master, and the last released version of Zephyr. I'd also propose changing CI to build against both versions of Zephyr to check this.

Hey David, sorry I didn't reply to this originally. I have tried to state my point of view in this comment instead:
https://github.com/JuulLabs-OSS/mcuboot/pull/728#discussion_r419525095

PR #728, opened 2020-04-30
doc/release-note zephyr compatibility note for 1.6.0

nvlsianpu:
Noted what need to revert for being compatible with
zephyr 2.2.0 and which version of zephyr running-master
might be use.

here also branch on which i reverted mentioned patches:
https://github.com/nvlsianpu/mcuboot/tree/revert-zephyr-2.2.0-compatible

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -32,6 +32,19 @@ updates as well.
   2.9.10 has an infinite loop in a certain end-of-file situation." Fix
   by updating a dependency in documentation generation.
 
+### Zephyr-RTOS compatibility
+- for compatibility with zephyr 2.2.0 need to revert commits which align MCUBoot
+  to the zephyr master branch at the release time. These commits are reverted on
+  1.6.0-revert-zephyr-2.2.0-compatible branch.
+
+- Zephyr-RTOS references (by git SHA or git tag) a compatible MCUBoot version in
+  its manifest.
+  You might use any of zephyr running-master version which references
+  the release.
+  Any Zephyr master after
+  https://github.com/zephyrproject-rtos/zephyr/commit/7a15afc).
+  should be able to work as well.
+
 ## Version 1.5.0
 
 The 1.5.0 release of MCUboot adds support for encrypted images using


docs/release-notes.md
```diff
@@ -32,6 +32,21 @@ updates as well.
   2.9.10 has an infinite loop in a certain end-of-file situation." Fix
   by updating a dependency in documentation generation.
 
+### Zephyr-Rtos compatibility
+- for compatibility with zephyr 2.2.0 need to revert following commits:
```
utzig, line 36: `for compatibility with zephyr 2.2.0 the following commits need to be reverted:`

docs/release-notes.md
```diff
@@ -32,6 +32,21 @@ updates as well.
...
+### Zephyr-Rtos compatibility
+- for compatibility with zephyr 2.2.0 need to revert following commits:  
+    f4a9a9d  
+    470e2f3 (conflict: accept current changes)  
+    1084100  
+    41c5d50  
+    fefdea20  
+    8ecb89b  
+
+- zephyr-rtos is using mcuboot exact version reference in its manifest. You
```
utzig, line 44: What you're trying to say by "exact version reference"  is linking to a specific commit sha?

docs/release-notes.md
```diff
@@ -32,6 +32,21 @@ updates as well.
...
+  might use any of zephyr running-master version which references the release and
+  very probably most of zephyr version chronologically around the release linked
+  one (the lower limit is
+  https://github.com/zephyrproject-rtos/zephyr/commit/0f9b362).
```
utzig, line 48: My understanding of this sentence is that any Zephyr `master` after the given commit should work, is that what you mean?

docs/release-notes.md
nvlsianpu: I mean https://github.com/zephyrproject-rtos/zephyr/blob/master/west.yml#L85-L86

I can rewrite:
zephyr-rtos references (by git SHA or git tag  )a compatible mcuboot version in its manifest

docs/release-notes.md
nvlsianpu: yes, any unless its will lost compatibility gain :(

docs/release-notes.md
```diff
@@ -32,6 +32,21 @@ updates as well.
   2.9.10 has an infinite loop in a certain end-of-file situation." Fix
   by updating a dependency in documentation generation.
 
+### Zephyr-Rtos compatibility
+- for compatibility with zephyr 2.2.0 the following commits need to be reverted:  
+    f4a9a9d
```
nvlsianpu, line 37: if https://github.com/JuulLabs-OSS/mcuboot/pull/729 merged need to extend list by introduced by it commits.

docs/release-notes.md
utzig: Can't you just mention something like

> Any Zephyr `master` after https://github.com/zephyrproject-rtos/zephyr/commit/0f9b362 should be able to work.

docs/release-notes.md
```diff
@@ -32,6 +32,21 @@ updates as well.
   2.9.10 has an infinite loop in a certain end-of-file situation." Fix
   by updating a dependency in documentation generation.
 
+### Zephyr-Rtos compatibility
```
davidvincze, line 35:
```suggestion
### Zephyr-RTOS compatibility
```
(capital letters)

docs/release-notes.md
davidvincze: I like Fabio's suggestion and I think it's also useful to emphasize that those zephyr versions that reference the 1.6.0 release probably work and can be used.

docs/release-notes.md
carlescufi, line 37: While I think it's very laudable to try and help users with coping with the lack of compatibility between Zephyr and MCUBoot releases, this might just not scale in the future. It's a bit fragile to have to depend on a list of commits that need to be reverted, and that might introduce other issues.
Can I suggest another approach? Instead of trying to make a particular release of MCUBoot compatible with a Zephyr one, and since this is the release notes for MCUBoot, how about we just document which Zephyr SHA this MCUBoot release is guaranteed to work with, and leave it to users to figure out what they need to do if they want to actually uses tagged releases from both projects?

As we've discussed multiple times with @nvlsianpu, @utzig and @d3zd3z, there are enough dependencies in a bootloader that builds against different RTOS to make it very hard to maintain compatibility in certain periods, like the current one where Zephyr is simply changing too much in very little time. What I see as future-proof approaches to this problem could be:

1. Remove the dependency to Zephyr, Mynewt and RIOT, and instead bundle all the code necessary to build MCUBoot in this repository
2. Not release MCUBoot at all, instead just keeping it rolling, and relying on the different RTOS to signal which particular MCUBoot SHA is compatible with each SHA in their respective histories
3. Start maintaining backwards compatibility for the RTOS that MCUBoot supports, including (a lot of) conditional compilation based on which particular release or SHA of the RTOS we are building against
4. Release, but do the minimum to document which SHA of each RTOS is supported with this release

Let me know your thoughts.

d3zd3z: > 1. Remove the dependency to Zephyr, Mynewt and RIOT, and instead bundle all the code necessary to build MCUBoot in this repository

I think this would be even harder to scale. We already have one target that builds this way. It pulls in 5 external repos in order to get enough dependencies, and has its own directory with 13k lines of code in it. The Zephyr-specific directory is about 1/3 of this, and builds with a normal Zephyr. I think no matter the issues we find with Zephyr, they are smaller than the issues we would find trying to build bare-bones.

However, I don't think this problem is particularly related to it being a bootloader, either. The bootloader is actually a pretty simple program. Given how fast Zephyr's API surface changes, I would guess more significant applications would encounter this even more so.

> 2. Not release MCUboot at all.

People seem to like seeing releases, even though, as you point out, they really aren't that helpful.

> 3. ... backward compatibility ... conditional compilation.

At least now this is possible, but it gets fairly messy.

> 4. Release, but do the minimum to document which SHA of each RTOS is supported with this release.

I think people do like to use labeled releases of Zephyr as well. On possible hybrid with this, would be to release mcuboot against master of Zephyr, and then make branches in the Zephyr-fork of MCUboot to backport to each release that is desired. I'm just not sure who has time to do this work.

Regardless, we'll need to be clear about what we do. We've stated before (not sure if it made the docs) that mcuboot/master should work with zephyr/master, and mcuboot's releases should work with the latest release of Zephyr.

One other idea, is we could make something like a v1.6.0-zephyr-2.2.0 release of MCUboot as a branch already made to work with Zephyr.

d3zd3z:
A few additional ideas:

1. We could hold off on the 1.6.0 release until the next Zephyr release. This would delay us by probably an additional month, although as I understand, the Zephyr release cycle is soon going to get a month longer.

2. We could define certain release of MCUboot that work with Zephyr, and these could maybe be correlated with Zephyr releases. The other releases would be like our current code, that only work with Zephyr master.

Thoughts?

d3zd3z:
It might even be that we have 1.6.0 now that only works with Zephyr master, and then a 1.6.1 that makes it work with Zephyr 2.3.0, if anything is needed when that comes out, otherwise, just a note that it has been tested.

davidvincze:
I would rather not wait for Zephyr and have a 1.6.0 release now and specify the SHA  it works with and later have a 1.6.1 when Zephyr 2.3.0 is released. So I would prefer the 2. option which is less strict in terms of scheduling. We could make a MCUboot release with each Zephyr releases in the future (every 2 months I think) and could also make "intermediate" releases when it "makes sense" (when a relatively significant amount of improvements and fixes have been merged since the last release), so those who "like to use labeled releases" should not wait for a new Zephyr release.

docs/release-notes.md
utzig: > On possible hybrid with this, would be to release mcuboot against master of Zephyr, and then make branches in the Zephyr-fork of MCUboot to backport to each release that is desired. I'm just not sure who has time to do this work.

I think branches for specific Zephyr versions should live here and be *downstreamed*.

> One other idea, is we could make something like a v1.6.0-zephyr-2.2.0 release of MCUboot as a branch already made to work with Zephyr.

This to me sounds like the best option, we could do releases against `master` and branches against the *latest* stable (or LTS) release.

nvlsianpu:
+1 for having 1.6.1 release for zephyr 2.3.0 (if any patch will be really needed, as probably 1.6.0 will work with the next zephyr release.

docs/release-notes.md
```diff
@@ -32,6 +32,19 @@ updates as well.
   2.9.10 has an infinite loop in a certain end-of-file situation." Fix
   by updating a dependency in documentation generation.
 
+### Zephyr-RTOS compatibility
+- for compatibility with zephyr 2.2.0 need to revert commits which align MCUBoot
+  to the zephyr master branch at the release time. These commits are reverted on
+  1.6.0-revert-zephyr-2.2.0-compatible branch.
```
nvlsianpu, line 38: this branch (name might be change) must be pushed to to this repo within the release process.

d3zd3z: Yeah, that branch name is just a little bit wordy. Revert and compatible are kind of redundant. I'd just call it v1.6.0-zephyr-2.2.x (or 2.2.1, which ever one is current when that happens.

docs/release-notes.md
```diff
@@ -32,6 +32,19 @@ updates as well.
...
+
+- Zephyr-RTOS references (by git SHA or git tag) a compatible MCUBoot version in
+  its manifest.
+  You might use any of zephyr running-master version which references
```
d3zd3z, line 42: I'm not sure this is quite right. If someone is using the Zephyr manifest, they will get Zephyr's fork of MCUboot, which will work with Zephyr. I think we should just leave this as a mention of a particular commit this works with (and probably not even 'after'.

docs/release-notes.md
carlescufi: I will reply on the main conversation, because I think that keeping branches like the one proposed might be too much overhead.

carlescufi:
> 2\. e could define certain release of MCUboot that work with Zephyr, and these could maybe be correlated with Zephyr releases. The other releases would be like our current code, that only work with Zephyr master.

I thiink this is the best option. For example, we could wait with 1.6.0 to Zephyr 2.3 feature freeze (2 days left) and then it's pretty much guaranteed nothing will change during the 2.3 stabilization phase, so that MCUBoot 1.6.0 would be compatible with Zephyr 2.3. In a similar fashion, every MCUBoot release would be coupled to a particular Zephyr release.
I don't think we should strive for Zephyr 2.2.0 in this case, because we are almost in feature freeze for 2.3

nvlsianpu:
@d3zd3z I'm closing this as part of this note was included into release-note. I think it will be worth to mention any working zephyr SHA for the release 1.6.0 (e.g. https://github.com/zephyrproject-rtos/zephyr/commit/4a5a90aef8f3acaadb5e297027e11c89b3c9a0b4 is the last one compatible unless #734 not merged)

PR #730, opened 2020-05-01
boot: bootutil: add option to disable cryptographic check of the image

arvinf:
bootutil:

- If MCUBOOT_IGNORE_SIGNATURE is defined image cryptographic signature check
is skipped during validation. SHA256 check is still retained. Enabling
this option can reduce the binary size, at the expense of greatly reduced
security. However this can be acceptable in some scenarios where only upgrade capability
is desired and cryptographic check is not required.

zephyr:

- Add Kconfig option BOOT_NO_SIGNATURE_CHECK. When y, MCUBOOT_IGNORE_SIGNATURE`
is defined for bootutil and the bootloader will not verify the cryptographic
signature of the image at boot or prior to upgrade.


utzig:
This is interesting because Mynewt works with no signature, I don't remember ever trying to build a Zephyr based bootloader without signature validation, and I *think* this is done on purpose. But since for Mynewt you can just unselect every signature method and resort to sha256 verification, it doesn't seem necessary to have to add a new config option. I was looking at how zephyr config works for this, and it uses a `choice` in the Kconfig which seems to be the reason for not supporting "no signature". I actually tried a different approach to solve the problem, and ended up with this:

```diff
diff --git a/boot/bootutil/src/image_validate.c b/boot/bootutil/src/image_validate.c
index 7e0af0d78..cc95959ce 100644
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -48,7 +48,11 @@
 #if defined(MCUBOOT_SIGN_EC) || defined(MCUBOOT_SIGN_EC256)
 #include "mbedtls/ecdsa.h"
 #endif
+
+#if defined(MCUBOOT_ENC_IMAGES) || defined(MCUBOOT_SIGN_RSA) || \
+    defined(MCUBOOT_SIGN_EC) || defined(MCUBOOT_SIGN_EC256)
 #include "mbedtls/asn1.h"
+#endif
 
 #include "bootutil_priv.h"
 
diff --git a/boot/zephyr/CMakeLists.txt b/boot/zephyr/CMakeLists.txt
index c382c528b..cda5a48e1 100644
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -141,6 +141,16 @@ if(CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256 OR CONFIG_BOOT_ENCRYPT_EC256)
   zephyr_library_compile_definitions(
     MBEDTLS_CONFIG_FILE="${CMAKE_CURRENT_LIST_DIR}/include/mcuboot-mbedtls-cfg.h"
     )
+elseif(CONFIG_BOOT_SIGNATURE_TYPE_NONE)
+  zephyr_library_include_directories(
+    ${BOOT_DIR}/zephyr/include
+    ${TINYCRYPT_DIR}/include
+    )
+
+  zephyr_library_sources(
+    ${TINYCRYPT_DIR}/source/sha256.c
+    ${TINYCRYPT_DIR}/source/utils.c
+    )
 elseif(CONFIG_BOOT_SIGNATURE_TYPE_RSA)
   # Use mbedTLS provided by Zephyr for RSA signatures. (Its config file
   # is set using Kconfig.)
diff --git a/boot/zephyr/Kconfig b/boot/zephyr/Kconfig
index 4f1b172dd..6b8836470 100644
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -62,6 +62,10 @@ choice
 	prompt "Signature type"
 	default BOOT_SIGNATURE_TYPE_RSA
 
+config BOOT_SIGNATURE_TYPE_NONE
+	bool "No signature; use only hash check"
+	select BOOT_USE_TINYCRYPT
+
 config BOOT_SIGNATURE_TYPE_RSA
 	bool "RSA signatures"
 	select BOOT_USE_MBEDTLS
diff --git a/boot/zephyr/keys.c b/boot/zephyr/keys.c
index 5dd0ca28f..41fda124e 100644
--- a/boot/zephyr/keys.c
+++ b/boot/zephyr/keys.c
@@ -41,7 +41,7 @@ extern unsigned int ecdsa_pub_key_len;
 #define HAVE_KEYS
 extern const unsigned char ed25519_pub_key[];
 extern unsigned int ed25519_pub_key_len;
-#else
+#elif !defined(CONFIG_BOOT_SIGNATURE_TYPE_NONE)
 #error "No public key available for given signing algorithm."
 #endif
 
diff --git a/boot/zephyr/prj.conf b/boot/zephyr/prj.conf
index 71e9a2a30..17826db08 100644
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -15,6 +15,7 @@ CONFIG_BOOT_UPGRADE_ONLY=n
 CONFIG_BOOT_BOOTSTRAP=n
 
 ### Default to RSA
+CONFIG_BOOT_SIGNATURE_TYPE_NONE=n
 CONFIG_BOOT_SIGNATURE_TYPE_RSA=y
 CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN=2048
 CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=n
```

What do you think?

arvinf:
If it can be done with existing defines then that's better than my solution. Doesn't this affect the signing tool and the signature block size? https://github.com/JuulLabs-OSS/mcuboot/blob/eb1ddfdaaca5871539da1cb47bf1fbe312e6ad6c/boot/bootutil/src/image_validate.c#L181

utzig:
> Doesn't this affect the signing tool and the signature block size?

If you mean `imgtool`, then you can already skip the `-k` option to `create`/`sign` command, which will generate a hash only image, and the `SIG_BUF_SIZE` should default to 32 when no signing algorithm was used in my suggested patch.

arvinf:
Thank you. Let me try that out locally and get back to you. If it meets all our needs then I'll update the PR with your changes and resubmit.

arvinf:
@utzig Please see #732

utzig:
> @utzig Please see #732

We're in release process, and once it ends I will add reviewer, but the PR looks OK to me.

arvinf:
Thank you.

PR #732, opened 2020-05-05
boot: bootutil: disable cryptographic check of the image

arvinf:
Update of #730

Bootutil:
- If MCUBOOT_SIGN_RSA, MCUBOOT_SIGN_EC, MCUBOOT_SIGN_EC256 are not
  defined cryptographic signature check is skipped during validation.
  sha256 check is still retained. This reduces the binary size, at the
  expense of greatly reduced security. However this can be acceptable
  in some scenarios where cryptographic check is not required.

Zephyr: 
- Add Kconfig option `CONFIG_BOOT_SIGNATURE_TYPE_NONE`. When y,
  The bootloader will not verify the cryptographic signature of the image
  at boot or prior to upgrade. Image integrity check is retained.

  Images would still need to be processed using `imgtool` however no key
  should be provided to skip cryptographic signature generation.


nvlsianpu:
@arvinf  What configuration should I use for nrf52840dk_nrf52840 to test this?

utzig:
> @arvinf What configuration should I use for nrf52840dk_nrf52840 to test this?

```
CONFIG_BOOT_SIGNATURE_TYPE_NONE=y
CONFIG_BOOT_SIGNATURE_TYPE_RSA=n
```

and don't use a signing key when running `imgtool`!

d3zd3z:
Verified that this configuration is still regularly tested by the sim. This change is more about making this configuration work with Zephyr.

boot/zephyr/keys.c
```diff
@@ -41,7 +41,7 @@ extern unsigned int ecdsa_pub_key_len;
 #define HAVE_KEYS
 extern const unsigned char ed25519_pub_key[];
 extern unsigned int ed25519_pub_key_len;
-#else
+#elif !defined(CONFIG_BOOT_SIGNATURE_TYPE_NONE)
```
d3zd3z, line 44: Wouldn't it be better to just remove the `#else` and the `#error`. The NONE variant is more an issue of how the configuration works with Zephyr, but it seems reasonable to just allow none of them to be defined, rather than requiring the NONE variant.

arvinf: Reasonable. I'll make the change.

nvlsianpu:
I had forgot to rebase on master before I tested It. Works well.

PR #737, opened 2020-05-13
serial_recovery: Reduce minimum number of members in packet

oyvindronningstad:
Was 3, but mcumgr sometimes sends 2 ("off" + "data", "len" is not needed
when "off" != 0). Reduce to 1 to avoid the problem, since the code
has other checks to catch this.

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>


nvlsianpu:
This is mirror of bugfix which went to NCS mcuboot here: https://github.com/NordicPlayground/fw-nrfconnect-mcuboot/pull/96

PR #739, opened 2020-05-14
Introduce direct execute-in-place (XIP) mode

davidvincze:
This PR is addressing the #205 issue.

When direct-XIP is enabled mcuboot selects the newest image
based on the image version numbers from the image header, thereafter the
selected image runs directly from its flash partition instead of moving
it. Therefore the images must be linked to be executed from the given
image slot. It means that in XIP mode either of the primary and
secondary slots can hold the active image.


d3zd3z:
First reaction is that I think we need a better name for this, since normal mode also executes in place, in fact that is why it has to move the images. I'll try to think of something.

utzig:
> First reaction is that I think we need a better name for this, since normal mode also executes in place, in fact that is why it has to move the images. I'll try to think of something.

`execute-in-place-on-secondary-slot`? :rofl:

boot/bootutil/src/bootutil_misc.c
```diff
@@ -580,6 +580,62 @@ boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
...
+/**
+ * Comparator function for `qsort` to compare version numbers.
+ *
+ * @param ver1  Pointer to an array element which holds the 1. version number
+ * @param ver2  Pointer to an array element which holds the 2. version number
+ *
+ * @note        This function must be ported if version number format has
+ *              changed in the image header.
+ *
+ * @return      if version1 >  version2  -1
```
d3zd3z, line 618: This ordering seems a bit backward. Other functions that do comparisons (such as memcmp) will return -1 if the first value is less than the second.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -166,6 +170,10 @@ struct boot_swap_state {
 
 _Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
 
+#if defined(MCUBOOT_XIP) && (BOOT_IMAGE_NUMBER != 1)
+#error "The MCUBOOT_XIP mode only supports single-image boot (MCUBOOT_IMAGE_NUMBER=1)."
```
d3zd3z, line 174: Do we plan on adding support for this at some point? I think the main constraint will be that you always have to use either both images from the first slot, or both images from the second slot.

boot/bootutil/src/loader.c
```diff
@@ -1916,3 +1914,212 @@ split_go(int loader_slot, int split_slot, void **entry)
...
+            BOOT_LOG_IMAGE_INFO(slot, hdr, &slot_state);
+
+        } else {
+            BOOT_LOG_INF("%s slot: No valid image", (slot == BOOT_PRIMARY_SLOT)
+                         ? "Primary" : "Secondary");
+        }
+    }
+
+    /* Sort the images based on version number */
+    qsort(image_versions, slot_cnt, sizeof(struct image_slot_version),
```
d3zd3z, line 1975: Given that you're only comparing two things, pulling in qsort is kind of overkill. How about just an if statement and swap? Or even just build the boot sequence array as a result of the 'if'.

boot/bootutil/src/bootutil_misc.c
d3zd3z: I realize this is to make the qsort below work, but it should at least be clearly stated that the comparison is reversed.

boot/bootutil/src/bootutil_priv.h
tamasban: I think it's possible to support multi image boot:
- Secure image can pass the address of the veneer region when passing execution to non-secure. It can pass through the registers. Usually registers just cleared before calling NS, but nothing stops us to load there valuable info there.
- Then NS image can save this address to a global variable in Reset_Handler().
- Before a veneer call, the address of the actual valid veneer function could be calculated. Just calculate the offset of the veneer region( compile time fixed veneer region address - runtime received veneer region address) and add to the original static function address.
- I think this way always the active image veneer region would be called. 
- Also Secure code need the info that which is the active NS image to configure the memory isolation hw (MPC & SAU) properly. But this data can be propagated from bootloader to active secure image through the shared data area.

boot/bootutil/src/loader.c
tamasban: > > First reaction is that I think we need a better name for this, since normal mode also executes in place, in fact that is why it has to move the images. I'll try to think of something.
> 
> `execute-in-place-on-secondary-slot`? 🤣

Original feature name was NO_SWAP in TF-M.

davidvincze:
> First reaction is that I think we need a better name for this, since normal mode also executes in place, in fact that is why it has to move the images. I'll try to think of something.

FYI, originally it was called "NO_SWAP".

boot/bootutil/src/loader.c
davidvincze: I believe originally this solution was designed to be able to handle N number of slots.
However, if we limit the slot number to 2 (which is the current limit for the whole project)
and provide a solution for more slots later if it will be ever needed, we can easily replace qsort with an 'if' and the newly introduced 'image_slot_version' structure and boot_compare_version_numbers() function could be abandoned.
I think this is the weakest and most questionable part of the whole PR.

davidvincze: @d3zd3z , @tamasban what do yo think?

boot/bootutil/src/bootutil_priv.h
d3zd3z: Do we need this complexity, or just require the two images to always be considered together?

boot/bootutil/src/loader.c
d3zd3z: Have you checked the size of the qsort implementation being brought in? I would just casually guess it probably adds 3k or so to the image. I would say, even with 'n' images, we shouldn't be calling quicksort.

In fact, why not just scan it for the newest image, and then if that isn't valid, scan it again for the next biggest. It's technically n^2 but I can't imaging having n ever be very large. (I really don't see why it would be larger than 2).

d3zd3z:
> FYI, originally it was called "NO_SWAP".

However, the code already refers in a few places to the overwrite-only upgrade as no_swap.

Maybe something like DIRECT_XIP, or something. I'm not really coming up with that much better of a name.

boot/bootutil/src/bootutil_priv.h
tamasban: If someone want to independently update the images (on arm-v8m S and NS) then I think we need this. 
Why might need independent update, I can think of:

- NS image usually more frequently updated then secure. Bandwidth constrains.
- Device operator not owning the secure image just the NS, only can update that one. Might different entity can update S image.

But I would be interested in from chip vendor perspective as well.

boot/bootutil/src/loader.c
tamasban: We can drop qsort if it has such a big flash requirements. I agree that probably always considering to have  two images is very reasonable.

tamasban:
> > FYI, originally it was called "NO_SWAP".
> 
> However, the code already refers in a few places to the overwrite-only upgrade as no_swap.
> 
> Maybe something like DIRECT_XIP, or something. I'm not really coming up with that much better of a name.

I like DIRECT_XIP. Just to have a few more option: ALWAYS_XIP, NO_COPY.

davidvincze:
A few numbers about the memory footprints (single-image boot):
|  | Overwrite-only | direct-XIP (with qsort) | direct-XIP (latest) |
| ------------- | ------------- | ------------- | ------------- |
| Flash  | 31.07 KB | 29.53 KB | 28.53 KB |
| RAM  | 48.21 KB | 23.59 KB | 23.65 KB |

tamasban:
> A few numbers about the memory footprints (single-image boot):
> 
> Overwrite-only	direct-XIP (with qsort)	direct-XIP (latest)
> Flash	31.07 KB	29.53 KB	28.53 KB
> RAM	48.21 KB	23.59 KB	23.65 KB

Are these with TF-M debug build? If so then can you check minsizerel as well.

davidvincze:
> > A few numbers about the memory footprints (single-image boot):
> > Overwrite-only	direct-XIP (with qsort)	direct-XIP (latest)
> > Flash	31.07 KB	29.53 KB	28.53 KB
> > RAM	48.21 KB	23.59 KB	23.65 KB
> 
> Are these with TF-M debug build? If so then can you check minsizerel as well.

Oh yes, sorry I forgot to mention that debug information is included in these numbers. Let me update the table above.

boot/bootutil/src/bootutil_misc.c
davidvincze: this function has been removed (along with qsort)

boot/bootutil/src/bootutil_priv.h
d3zd3z: This might be better solved by enhancing the update protocol to be able to be a delta off of the existing image.

davidvincze:
Does everyone find the "direct-XIP" name acceptable?

>  Someone deploying will have to build their application twice, and the upgrade process will have to figure out which image is appropriate to download to the device. Overall, this should be managed, and it could be argued that having the complexity implemented on larger systems is better than trying to manage that complexity on small MCU devices.

You're right. Direct-XIP requires some added complexity "somewhere else", however it's a tradeoff between providing a convenient solution and memory footprint. As we know unfortunately there won't be a "one-fits-all" solution. However, the RAM-loading (we can find a more suitable name later :)) strategy - which I want to upstream after this PR has been merged - can serve as a solution that lies between overwrite-only and direct-xip.

nvlsianpu:
cc @hakonfam

pdunaj:
Hi, I have a quick question.
How mcboot decides which image to boot? The one with a higher version or does application needs to disable the current one somehow?

nvlsianpu:
@pdunaj Mcuboot chainn-load the one valid and with a higher version

nvlsianpu:
@de-nordic Some changes miaght be complementar to https://github.com/JuulLabs-OSS/mcuboot/pull/741. I think single-slot feature can be even implemented by punting some more configurable constrains on functionality implemented in this PR.

de-nordic:
> 
> @de-nordic Some changes miaght be complementar to #741. I think single-slot feature can be even implemented by punting some more configurable constrains on functionality implemented in this PR.

Probably with some more #ifdef cutting in context_boot_go, the #741 could happen here.
The #741 cuts more aggressively as it is basically trying to check image and and go to DFU, if image is not OK. Or pass to DFU immediately if requested.

davidvincze:
> Might I ask for extend zephyr kconfig by this new feature?

Yes, of course. ;) I'll include it in the next PR update.

boot/bootutil/src/loader.c
```diff
@@ -517,18 +523,22 @@ boot_check_header_erased(struct boot_loader_state *state, int slot)
...
+ * @param req           Required minimal image version.
+ * @param ver           Version of the image to be checked.
+ * @param accept_equal  If true, an equal version number is accepted, otherwise
+ *                      if false it is considered as not sufficient.
  *
- * @return            0 if the version is sufficient, nonzero otherwise.
+ * @return              0 if the version is sufficient, nonzero otherwise.
  */
 static int
 boot_is_version_sufficient(struct image_version *req,
```
d3zd3z, line 539: Rather than a flag `accept_equal`, I think it would be cleaner to just rename this function to something like `boot_version_cmp` that works like memcmp or strcmp. It returns -1 if the first version is strictly less, 1 if strictly greater, and 0 if the versions are equal. It can then be up to the caller to interpret that as desired. This also makes the error handling cleaner, because this function doesn't return a specific error code, and that can depend on where it is used.

boot/zephyr/Kconfig
```diff
@@ -150,6 +150,10 @@ config BOOT_VALIDATE_SLOT0
 	  every boot, but can mitigate against some changes that are
 	  able to modify the flash image itself.
 
+choice
+	prompt "Alternative image upgrade modes"
+	optional
+
```
nvlsianpu, line 156:
```suggestion
help
  Override the default `swap using scratch` image upgrade mode
```

nvlsianpu:
how I supposed to sign the image for xip. I did.`/bootloader/mcuboot/scripts/imgtool.py sign -S 0x67000 -v 1.1.0 -k ./bootloader/mcuboot/root-rsa-2048.pem --align 8 -H 0x200 ./build/sample_hellow3/zephyr/zephyr.hex signed3_dx.hex` for version 1.1.0 and 1.0.0, but what i have in logg is:
``` console
*** Booting Zephyr OS build zephyr-v2.2.0-3116-g48deea52c117  ***
[00:00:00.004,669] <inf> mcuboot: Starting bootloader
[00:00:00.011,077] <inf> mcuboot: Primary   slot: version=1.0.0+0, magic=unset, image_ok=0x3
[00:00:00.020,507] <inf> mcuboot: Secondary slot: version=1.1.0+0, magic=unset, image_ok=0x3
[00:00:00.029,876] <err> mcuboot: Unable to find bootable image


```

d3zd3z:
> with --pad things starts working

This suggests that we are only allowing an image to execute if it is set as "copy done". Perhaps we need to ignore this when doing direct-xip, and just execute the images. On the other hand, the copy done could provide a possible state to be able to revert.

But, I think we should be able to run an image without needing to write anything at the end of the image.

davidvincze:
> After rethinking:
> I suggest to made image update mode selection always explicitly.
> helper PR here [davidvincze#1](https://github.com/davidvincze/mcuboot/pull/1)

Thank you, I've squashed it with the "zephyr: Add direct-xip mode to Zephyr configuration" commit.

davidvincze:
> > with --pad things starts working
> 
> This suggests that we are only allowing an image to execute if it is set as "copy done". Perhaps we need to ignore this when doing direct-xip, and just execute the images. On the other hand, the copy done could provide a possible state to be able to revert.
> 
> But, I think we should be able to run an image without needing to write anything at the end of the image.

Originally in TF-M it was checking the 16 bytes long `magic` value and the `image_ok` flags  in the trailer. If the image header was valid and at least one of the previously mentioned fields was valid/set, it marked the slot as "used" (contains a potential image).
However, the `magic` only has an actual role if the images are moved between the primary and the secondary slot, the `image_ok` flag plays a role in the revert mechanism. I agree that the examinition of the trailer should be omitted as it doesn't really matter in direct-xip mode, I will update the code in a way that it can detect an image based on a valid header only.

davidvincze:
@nvlsianpu 
Optional 'revert' mechanism implementation for direct-xip mode: https://github.com/davidvincze/mcuboot/pull/2
AFAIK I cannot create a new PR based on this one (stacked PRs). I'll open a new PR as soon as this is merged.

nvlsianpu:
@davidvincze  thanks for updates, will look soon.

boot/bootutil/src/swap_scratch.c
```diff
@@ -179,7 +179,7 @@ boot_slots_compatible(struct boot_loader_state *state)
     size_t num_sectors_secondary;
     size_t sz0, sz1;
     size_t primary_slot_sz, secondary_slot_sz;
-#ifndef MCUBOOT_OVERWRITE_ONLY
+#if !defined(MCUBOOT_OVERWRITE_ONLY) && !defined(MCUBOOT_DIRECT_XIP)
```
utzig, line 182: Could instead create a new define, eg, `#define NO_SWAP (defined(MCUBOOT_OVERWRITE_ONLY) || defined(MCUBOOT_DIRECT_XIP))` and use through out. Or create a new `swap_none.c` file and specialize it for those cases.

boot/bootutil/src/loader.c
```diff
@@ -1916,3 +1926,216 @@ split_go(int loader_slot, int split_slot, void **entry)
...
+/**
+ * Prepares the booting process. This function choose the newer image in flash
+ * as appropriate, and tells you what address to boot from.
+ *
+ * @param state                 Boot loader status information.
+ * @param rsp                   On success, indicates how booting should occur.
+ *
+ * @return                      0 on success; nonzero on failure.
+ */
+int
```
utzig, line 1978: I did not compare to the current `context_boot_go` but this looks fairly similar. Maybe deserves a medal for the copy/pasta Olympics! :-)

d3zd3z:
> A few numbers about the memory footprints (single-image boot)

Amusing coincidence that the swap code is just about the same size as qsort.

boot/bootutil/src/loader.c
```diff
@@ -1916,3 +1922,189 @@ split_go(int loader_slot, int split_slot, void **entry)
     free(sectors);
     return rc;
 }
+
+#else /* !MCUBOOT_DIRECT_XIP */
```
d3zd3z, line 1926: The comment here should be `/* MCUBOOT_DIRECT_XIP */` which describes when the code below will be compiled in (since the above was an ifndef.

boot/bootutil/src/loader.c
d3zd3z: It is fairly different, but there is a lot in common. It might be worth trying to factor this function into what is common and what is unique to each.

boot/bootutil/src/loader.c
```diff
@@ -1916,3 +1922,189 @@ split_go(int loader_slot, int split_slot, void **entry)
...
+        goto out;
+    }
+
+out:
+   for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
+       flash_area_close(BOOT_IMG_AREA(state, BOOT_NUM_SLOTS - 1 - slot));
+   }
+   return rc;
+}
+#endif /* !MCUBOOT_DIRECT_XIP */
```
d3zd3z, line 2096: Likewise, this comment should be `/* MCUBOOT_DIRECT_XIP */` since that is when the code above it will be compiled in.

d3zd3z:
See #768 for simulator enhancements to do basic testing of this setup. More tests will be coming that test various configurations.

boot/bootutil/src/loader.c
davidvincze: I think in its current state it would be better to keep them separate, I'm planning to add some extra code to the direct-xip code in the near future to support another upgrade strategy (RAM_LOAD). It would be possible (maybe a little forced) to merge the 2 context_boot_go() functions, but it would make the first even longer and would require more ifdefs.
Common parts:
- 3 local variables
- a short 'for' loop to open the image areas
- **the ends of the functions are the most similar**:
  - security counter update (HW rollback protection): can be combined with a few extra ifdefs
  - MCUBOOT_MEASURED_BOOT and MCUBOOT_DATA_SHARING: can be combined if parameterized (selected_slot)
  - Setting the 'rsp' structure: can be combined if parameterized (selected_slot +slected_image_header?)
- a short 'for' loop to close the image areas

boot/bootutil/src/loader.c
```diff
@@ -1933,3 +1939,189 @@ split_go(int loader_slot, int split_slot, void **entry)
...
+ *                       carry information about slots that contain an image.
+ *                       After return the corresponding array elements are set
+ *                       to a non-zero value if the given slots are in use
+ *                       (contain a firmware image).
+ * @param slot_cnt       The number of slots, which can contain firmware images.
+ *
+ * @return               The number of found images.
+ */
+static uint32_t
+boot_get_slot_usage(struct boot_loader_state *state, uint8_t *slot_usage,
```
de-nordic, line 1959: `slot_usage` is array, so you may directly declare it as array (`uint8_t slot_usage[]`). It is pointer to array anyway, just seems more intuitive, at least to me.

boot/bootutil/src/swap_scratch.c
davidvincze: I updated the PR so that in the end a new define was not needed.

davidvincze:
> No `imgtool` changes are required? Btw, encrypted images won't work so that may be pointed out in the help option or design document. Also slot compatibility is not required here, correct?

The fact that image encryption is not supported in direct-xip mode has been highlighted in the help option of imgtool
and in the design.rst document too.

All the flash sector related operations and checks (e.g. slot compatibility check) have been removed from the direct-xip code.

boot/bootutil/src/loader.c
```diff
@@ -586,10 +597,10 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
     if (slot != BOOT_PRIMARY_SLOT) {
         /* Check if version of secondary slot is sufficient */
-        rc = boot_is_version_sufficient(
-                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver,
-                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver);
-        if (rc != 0 && boot_check_header_erased(state, BOOT_PRIMARY_SLOT)) {
+        rc = boot_version_cmp(
+                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+        if (rc < 0 && boot_check_header_erased(state, BOOT_PRIMARY_SLOT)) {
```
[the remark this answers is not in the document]
davidvincze: I agree, but the boot_version_cmp() has been created based on the parameter seqence and return values of standard functions such as memcmp() and strcmp() .
The logic here is that the first parameter is what we want to compare, it should be "bigger, newer, better"... :)

boot/bootutil/src/loader.c
d3zd3z: >  encrypted images won't work

Our format of encrypted images are unlikely to work, but there are more and more SoCs coming out with the ability to execute encrypted code. Usually the key management is a bit more complex, because the key never is visible to the CPU. Regardless, that is the topic of a different change than this one.

davidvincze: Are there any more thoughts on this topic? What is your opinion, can I leave it this way or shall I try to merge the 2 functions?

boot/bootutil/src/loader.c
```diff
@@ -1939,3 +1931,189 @@ split_go(int loader_slot, int split_slot, void **entry)
...
+
+    return image_cnt;
+}
+
+int
+context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
+{
+    struct image_header *hdr = NULL;
+    struct image_header *selected_image_header = NULL;
+    uint8_t slot_usage[BOOT_NUM_SLOTS] = {0};
```
nvlsianpu, line 1979: initialization of array to zeros like that is compiler not mandatory extension.
Better to use memset(), or
rework boot_get_slot_usage() so it clear or set fields.

davidvincze: Fixed.

nvlsianpu: thanks, it looks good now.

xianghui-renesas:
hi, could someone explain to me how to sign an image for the direct xip upgrade mode? Suppose below is the signing command with overwrite upgrade mode, what do we need to update in order to sign the image to use direct xip upgrade mode?
imgtool.py sign -k <boot_project>/mcu-tools/MCUboot/root-ec-p256.pem --version 1.0.0+0 --header-size 0x80 --align 128 --max-align 128 --slot-size 0x8000 --max-sectors 1 --overwrite-only  --confirm --pad-header path/to/image.bin path/to/image_signed.bin

de-nordic:
> 
> 
> hi, could someone explain to me how to sign an image for the direct xip upgrade mode? Suppose below is the signing command with overwrite upgrade mode, what do we need to update in order to sign the image to use direct xip upgrade mode?
> imgtool.py sign -k <boot_project>/mcu-tools/MCUboot/root-ec-p256.pem --version 1.0.0+0 --header-size 0x80 --align 128 --max-align 128 --slot-size 0x8000 --max-sectors 1 --overwrite-only --confirm --pad-header path/to/image.bin path/to/image_signed.bin

Can you create discussion, for the topic, here: https://github.com/mcu-tools/mcuboot/discussions

PR #740, opened 2020-05-18
boot: Fix uninitialized variable warning

davidvincze:
Tiny patch before release 1.6.0.
Fix uninitialized variable warning that may occur when
compiler optimization is enabled (especially for size).

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1583,7 +1583,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 {
     size_t slot;
     struct boot_status bs;
-    int rc;
+    int rc = -1;
     int fa_id;
     int image_index;
     bool has_upgrade;


boot/bootutil/src/loader.c
```diff
@@ -1583,7 +1583,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 {
     size_t slot;
     struct boot_status bs;
-    int rc;
+    int rc = -1;
```
d3zd3z, line 1586: I wonder if we should pick a specific err, since the error codes seem to be positive numbers. Even `BOOT_EBADIMAGE`. The only path I can see through the code where rc doesn't get set is if the build is configured for zero images.

utzig: A similar patch was submitted before; I think sometimes `gcc` just confuses itself.

d3zd3z: I was just saying that in the even that gcc somehow got it right, and the code did fall through, we should probably return a meaningful error code. But, I guess I'm also ok with the code as is.

PR #741, opened 2020-05-21
zephyr: Allow single image applications

de-nordic:
The change enables use of single image applications, for those that do
not need to use multiple images on device.
Single image means that the update of application will overwrite image
of previous application.
image_0 partition (primary) is used for loading application with single
image configuration, all other image partitions are ignored.


boot/zephyr/Kconfig
```diff
@@ -58,6 +58,14 @@ config NRF_CC310_BL
 
 menu "MCUBoot settings"
 
+config SINGLE_IMAGE_DFU
```
nvlsianpu, line 61: This is equivalent to `MCUBOOT_IMAGE_NUMBER == 0.5 true case`. Shouldn't be ensure that MCUBOOT_IMAGE_NUMBER is equal to 1?

de-nordic: It could be checked.

de-nordic: Do you want the check to be done here boot/zephyr/include/target.h, same way the other checks are done?

de-nordic: OK, done it via limiting range on `UPDATEABLE_IMAGE_NUMBER`, in Kconfig, when `SINGLE_IMAGE_DFU` is selected.

nvlsianpu:
Looks very promising - will test this on Monday.

utzig:
I tested this and it worked correctly for me. Needs rebase!

de-nordic:
> 
> 
> I tested this and it worked correctly for me. Needs rebase!

Thanks!

PR #744, opened 2020-05-26
[Cypress] - Add external memory support for cypress platform

romanjoe:
Introduce support for secondary (upgrade) slot located in external memory for Cypress devices.

Features:

- External memory support implemented using SFDP protocol
- Only "OWERWRITE" bootloading scheme is supported
- Demo platform is CY8CPROTO_062_4343W with S25FL512SAGMFI010 serial flash chip on board.
- The slot size for upgrade slot is even (or smaller) to erase size of given memory module


BohdanKovalchuk:
Roman, please include one more code fix we've done in erase API recently

PR #753, opened 2020-06-12
Cypress: implement SWAP mode for PSoC6

BohdanKovalchuk:
Signed-off by Bohdan Kovalchuk <bohdan.kovalchuk@linaro.org>

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -195,11 +195,8 @@ uint32_t
 boot_status_off(const struct flash_area *fap)
 {
     uint32_t off_from_end;
-    uint8_t elem_sz;
 
-    elem_sz = flash_area_align(fap);
-
-    off_from_end = boot_trailer_sz(elem_sz);
+    off_from_end = boot_trailer_sz(BOOT_WRITE_SZ(state));
 
     assert(off_from_end <= fap->fa_size);
     return fap->fa_size - off_from_end;
@@ -457,6 +454,28 @@ boot_write_magic(const struct flash_area *fap)
     return 0;
 }
 
+
+/**
+ * Write trailer data; status bytes, swap_size, etc
+ *
+ * @returns 0 on success, != 0 on error.
+ */
+int
+cy_boot_write_trailer(const struct flash_area *fap, uint32_t off,
+        const uint8_t *inbuf, uint8_t inlen)
+{
+    int rc;
+
+    for(uint8_t i=0; i<inlen; i++)
+    {
+        rc = flash_area_write_byte(fap, off+i, inbuf);
+        if (rc != 0) {
+            return BOOT_EFLASH;
+        }
+    }
+    return 0;
+}
+
 /**
  * Write trailer data; status bytes, swap_size, etc
  *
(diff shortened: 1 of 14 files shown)


BohdanKovalchuk:
@d3zd3z , @utzig please, review this draft code so we can progress on SWAP implementation for PSoC6

please, ignore // TODO and 1362 for now since the code isn't final

I'd appreciate your inputs on design

thanks

boot/bootutil/src/bootutil_misc.c
```diff
@@ -195,11 +195,7 @@ uint32_t
 boot_status_off(const struct flash_area *fap)
 {
     uint32_t off_from_end;
-    uint8_t elem_sz;
-
-    elem_sz = flash_area_align(fap);
-
-    off_from_end = boot_trailer_sz(elem_sz);
+    off_from_end = boot_trailer_sz(BOOT_MAX_ALIGN);
```
utzig, line 198: This is not OK, we want to use the write size alignment for each particular MCU, this is not supposed to work on *your* MCU only.

boot/bootutil/src/swap_move.c
```diff
@@ -140,7 +140,7 @@ swap_read_status_bytes(const struct flash_area *fap,
     found_idx = -1;
     /* skip erased sectors at the end */
     last_rc = 1;
-    write_sz = BOOT_WRITE_SZ(state);
+    write_sz = BOOT_MAX_ALIGN;
```
utzig, line 143: Maybe you should consider making `BOOT_WRITE_SZ` return `BOOT_MAX_ALIGN` for your build?

boot/zephyr/CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ zephyr_library_sources(
   ${BOOT_DIR}/bootutil/src/image_rsa.c
   ${BOOT_DIR}/bootutil/src/image_ec256.c
   ${BOOT_DIR}/bootutil/src/image_ed25519.c
+  ${BOOT_DIR}/bootutil/src/bootutil_misc.c
```
utzig, line 102: This is not supposed to be part of this PR, is it? Maybe needs a rebase?

boot/bootutil/src/bootutil_misc.c
```diff
@@ -457,6 +453,28 @@ boot_write_magic(const struct flash_area *fap)
...
 }
 
+
+/**
+ * Write trailer data; status bytes, swap_size, etc
+ *
+ * @returns 0 on success, != 0 on error.
+ */
+int
+cy_boot_write_trailer(const struct flash_area *fap, uint32_t off,
```
utzig, line 463: `cy_` prefixes are not acceptable here. Also what would happen if a reset happens while inside the `flash_area_write_byte` loop?

BohdanKovalchuk: yes, sure

I left it intentionally to highlight which API required to be implemented in different way

I'll change it to neutral as soon as we finalize the code itself

boot/bootutil/src/swap_move.c
BohdanKovalchuk: Is my understanding correct on following?

* BOOT_WRITE_SZ()  is a value of state->write_sz;
* state->write_sz is filled in with return value of flash_are_align();
which is, in context of SWAP status table write-operations, the same as BOOT_MAX_ALIGN;

So, is it still required different approaches - macro, API and #define name for the same value?

boot/zephyr/CMakeLists.txt
BohdanKovalchuk: rebased

BohdanKovalchuk:
- closing as unacceptable implementation

PR #757, opened 2020-06-18
boot/zephyr: add option to relocate interrupts to application

e-rk:
If SW_VECTOR_RELAY is enabled, the bootloader sets
the pointer to address, where interrupts will be
forwarded.

Signed-off-by: Rafał Kuźnia <rafal.kuznia@nordicsemi.no>

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -388,6 +388,15 @@ config RECOVERY_UART_DEV_NAME
 
 endif # MCUBOOT_SERIAL
 
+config BOOT_INTR_VEC_RELOC
+	bool "Relocate the interrupt vector to the application"
+	default n
+	depends on SW_VECTOR_RELAY || CPU_CORTEX_M_HAS_VTOR
+	help
+	  Relocate the interrupt vector to the application before it is started.
+	  Select this option if application requires vector relocation,
+	  but it doesn't relocate vector in its reset handler.
+
 config UPDATEABLE_IMAGE_NUMBER
 	int "Number of updateable images"
 	default 1
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -98,6 +98,18 @@ MCUBOOT_LOG_MODULE_REGISTER(mcuboot);
 void os_heap_init(void);
 
 #if defined(CONFIG_ARM)
+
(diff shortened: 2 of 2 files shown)


e-rk:
This change is dependent on https://github.com/zephyrproject-rtos/zephyr/pull/26276.

e-rk:
@nvlsianpu I have addressed your suggestion and extended the PR to include the hardware VTOR register.

boot/zephyr/Kconfig
```diff
@@ -388,6 +388,13 @@ config RECOVERY_UART_DEV_NAME
 
 endif # MCUBOOT_SERIAL
 
+config BOOT_INTR_VEC_RELOC
+	bool "Relocate the interrupt vector to the application"
+	default n
+	depends on SW_VECTOR_RELAY || CPU_CORTEX_M_HAS_VTOR
+	help
+	  Relocate the interrupt vector to the application before it is started.
```
nvlsianpu, line 396: Add hint on why this is needed to the help message:
```suggestion
	  Relocate the interrupt vector to the application before it is started.
	  Select this option if application requires vector relocation,
	  but it doesn't relocate vector int its reset handler.
```
Select this option if application requires vector relocation, but it doesn't relocate vector int its reset handler.

e-rk: Thank you. I have added the change.

d3zd3z:
A comment says this change depends on a Zephyr commit. Is that only if this feature is enabled? In other words, is it safe to go ahead and merge this?

boot/zephyr/Kconfig
```diff
@@ -388,6 +388,15 @@ config RECOVERY_UART_DEV_NAME
 
 endif # MCUBOOT_SERIAL
 
+config BOOT_INTR_VEC_RELOC
+	bool "Relocate the interrupt vector to the application"
+	default n
+	depends on SW_VECTOR_RELAY || CPU_CORTEX_M_HAS_VTOR
```
d3zd3z, line 394: Does it make sense for this feature to be 'n' if either of these are set? I'm wondering if this should be a Kconfig, or just a define in the code based on these other defines.

e-rk:
As @nvlsianpu mentioned, it is not strictly dependent. There should be no problem with merging this first.

PR #760, opened 2020-06-29
Allow larger primary slot in swap-move

utzig:
When using swap in move mode, the current compatibility test only allows for same size and layout between slots. This results in one sector that can never be used in the secondary slot, which corresponds to the first move up operation in the primary slot. This relaxes and checking a bit, so both same size slots and a primary with one extra sector are valid.


utzig:
Tested on HW, sim test to be added.

PR #764, opened 2020-07-08
zephyr: Find board name from dts.pre.tmp file

DunderRoffe:
In scripts/assemble.py:

Use the <board_name>.dts.pre.tmp file in <build-dir>/zephyr to find the
board name.

This allows the build directory path to be set freely and not enforcing
directory structures.


utzig:
@DunderRoffe commits must be signed.

DunderRoffe:
@utzig fixed

Issue #762, opened 2020-07-03
DFU failure due to unexpected/invalid trailer in secondary slot


nwsetec:
I'm getting an occassional issue where a DFU transfer to the secondary slot (when the primary slot application is running) can't proceed because even though the secondary slot is otherwise empty (all 0xFF) there is a trailer at the end.  The trailer looks like this:

```
0x00065FE8: FFFFFF01
0x00065FF0: F395C277
0x00065FF4: 7FEFD260
0x00065FF8: 0F505235
0x00065FFC: 8079B62C
```

Bytes FFFFFF01 indicates the image is 'ok' but it is otherwise empty.

I'm using a nRF52832 SoC with an external flash that is used for the secondary slot.
The bootloader was built with mcuboot tag v1.3.1 and
Zephyr v1.14.1 (9591025942af9d8334901cc281d2b41a1747836d)
I had to revert mcuboot commit f99a4c790a6b4f508938f36e811c80ae1f25c2a3 to get it to build.
We also had to modify mcuboot to ensure an external WDT was regularly cleared.

This is a showstopper for us as if this failure occurs in the field the device can not longer receive updates.

I'm looking into implementing a workaround by modifying the mcumgr code in Zephyr 1.14.1 to add an extra check at the begining of a DFU transfer that the image trailer is valid by also checking if the 1st sector of the secondary slot is empty and if it is then we'll erase the last sector of the secondary slot and let the DFU transfer commence.

Note also that mcuboot doesn't seem to detect the invalid trailer and erase it on reboot.

I haven't been able to find any other issues that look like this issue in the mcuboot project.

utzig:
What external flash are you using?

nwsetec:
@utzig a Spansion S25FL216K

utzig:
> Note also that mcuboot doesn't seem to detect the invalid trailer and erase it on reboot.

[...]

PR #765, opened 2020-07-09
Fix erase of possible trailer leftover after reset

utzig:
This fixes an issue where an image might be erased, but a trailer left behind. It can happen if the image in the secondary slot did not pass validation, in which case the whole slot is erased. If during the erase operation, a reset occurs, parts of the slot might have been erased while some have not. The concerning part is the trailer because it might disable a new image from being loaded through mcumgr; so just get rid of the trailer here, if the header is erased.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -578,6 +578,23 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
     hdr = boot_img_hdr(state, slot);
     if (boot_check_header_erased(state, slot) == 0 ||
         (hdr->ih_flags & IMAGE_F_NON_BOOTABLE)) {
+
+#if defined(MCUBOOT_SWAP_USING_SCRATCH) || defined(MCUBOOT_SWAP_USING_MOVE)
+        /*
+         * This fixes an issue where an image might be erased, but a trailer
+         * be left behind. It can happen if the image is in the secondary slot
+         * and did not pass validation, in which case the whole slot is erased.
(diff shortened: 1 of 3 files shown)


utzig:
Under test - should fix #762

utzig:
> I wonder how hard a sim test of this would be.

Not too hard I suppose, but needs to be a new test.

d3zd3z:
I'm starting to look at adding some more tests for some of our newer functionality. I can include this change if you'd like.

utzig:
@d3zd3z Test was added, if you want to take a look ;-)

With reverted commit:

```
test dependency_combos ... ok
test bad_secondary_slot ... ok
test downgrade_prevention ... ok
test norevert ... ok
test norevert_newimage ... ok
test perm_with_random_fails ... ok
ERROR 2020-07-09T22:53:33Z: bootsim::image: Expected trailer on secondary slot to be erased
test basic_revert ... ok
test secondary_trailer_leftover ... FAILED
test status_write_fails_complete ... ok
test status_write_fails_with_reset ... ok
test perm_with_fails ... ok
```

utzig:
> I'm starting to look at adding some more tests for some of our newer functionality. I can include this change if you'd like.

Lol, sorry I've only seen this now!

utzig:
Tested OK on HW, no side-effects that I could find. @nvlsianpu Care to take a look?

Fixes #762

nvlsianpu:
I'm looking into, thanks for fast response.

PR #767, opened 2020-07-10
assemble: Allow use of ZEPHYR_BASE environment var

utzig:
Make `-z` flag optional, so if it is not provided rely on the ZEPHYR_BASE environemnt variable to find the Zephyr tree.

### DIFF
--- a/scripts/assemble.py
+++ b/scripts/assemble.py
@@ -22,6 +22,7 @@ import argparse
 import errno
 import io
 import re
+import os
 import os.path
 import sys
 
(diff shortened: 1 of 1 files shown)


scripts/assemble.py
```diff
@@ -111,19 +112,27 @@ def main():
...
+    parser.add_argument('-z', '--zephyr-base',
             help='Zephyr base containg the Zephyr repository')
 
     args = parser.parse_args()
 
-    sys.path.insert(0, os.path.join(args.zephyr_base, "scripts", "dts"))
+    zephyr_base = args.zephyr_base
+    if zephyr_base is None:
+        try:
+            zephyr_base = str(os.environ['ZEPHYR_BASE'])
```
d3zd3z, line 123: Is the 'str' necessary here? Environment variables should always be strings. If it wasn't a string (say a byte array), this wouldn't do what you want, anyway (`str(b'foo')` results in the string `"b'foo'"`.)

utzig: Nah, I was testing something else and forgot to remove it, thanks for pointing it out. Also fixed a typo in the help message I've missed before...

PR #768, opened 2020-07-10
Sim tests for direct xip

d3zd3z:
Currently draft, as it depends on #739 merging. Also, at this point, this test only serves to allow the tests to be run with direct-xip enabled. Most of the tests will be skipped, since most of the tests are for testing the upgrade code, which is not being used.

Work still needed:
- [ ] Write tests for the direct-xip functionality itself.


d3zd3z:
I've updated this for the direct xip code that merged. Provided the tests all pass, this should be ready as well, now.

utzig:
@d3zd3z Please rebase this!

PR #770, opened 2020-07-14
boot/zephyr: set VTOR to relay vector before chain-load

e-rk:
If software IRQ relay is enabled and VTOR is available,
the bootloader will switch from its real vector table
to the relay vector table before chain-loading to the
application.

Signed-off-by: Rafał Kuźnia <rafal.kuznia@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -22,6 +22,7 @@
 #include <drivers/timer/system_timer.h>
 #include <usb/usb_device.h>
 #include <soc.h>
+#include <linker/linker-defs.h>
 
 #include "target.h"
 
@@ -99,16 +100,9 @@ void os_heap_init(void);
 
 #if defined(CONFIG_ARM)
 
-#ifdef CONFIG_BOOT_INTR_VEC_RELOC
-
 #ifdef CONFIG_SW_VECTOR_RELAY
 extern void *_vector_table_pointer;
-#define VTOR _vector_table_pointer
-#elif CONFIG_CPU_CORTEX_M_HAS_VTOR
-#define VTOR SCB->VTOR
 #endif
-
-#endif /* CONFIG_BOOT_INTR_VEC_RELOC */
 
 struct arm_vector_table {
     uint32_t msp;
@@ -147,8 +141,20 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 
 #ifdef CONFIG_BOOT_INTR_VEC_RELOC
-    VTOR = vt;
+#if defined(CONFIG_SW_VECTOR_RELAY)
+    _vector_table_pointer = vt;
+#ifdef CONFIG_CPU_CORTEX_M_HAS_VTOR
+    SCB->VTOR = (uint32_t)__vector_relay_table;
 #endif
+#elif defined(CONFIG_CPU_CORTEX_M_HAS_VTOR)
+    SCB->VTOR = (uint32_t)vt;
+#endif /* CONFIG_SW_VECTOR_RELAY */
+#else /* CONFIG_BOOT_INTR_VEC_RELOC */
+#if defined(CONFIG_CPU_CORTEX_M_HAS_VTOR) && defined(CONFIG_SW_VECTOR_RELAY)
+    _vector_table_pointer = _vector_start;
+    SCB->VTOR = (uint32_t)__vector_relay_table;
+#endif
+#endif /* CONFIG_BOOT_INTR_VEC_RELOC */
 
     __set_MSP(vt->msp);
 #if CONFIG_MCUBOOT_CLEANUP_ARM_CORE


boot/zephyr/main.c
```diff
@@ -146,9 +140,18 @@ static void do_boot(struct boot_rsp *rsp)
...
 #endif
 
-#ifdef CONFIG_BOOT_INTR_VEC_RELOC
-    VTOR = vt;
+#if defined(CONFIG_SW_VECTOR_RELAY) && defined(CONFIG_BOOT_INTR_VEC_RELOC)
+    _vector_table_pointer = vt;
+#endif
+
+#if defined(CONFIG_CPU_CORTEX_M_HAS_VTOR)
+#if defined(CONFIG_SW_VECTOR_RELAY)
```
ioannisg, line 148: strictly speaking, you need an `&& defined(CONFIG_BOOT_INTR_VEC_RELOC)` here.

e-rk: No, `CONFIG_BOOT_INTR_VEC_RELOC` is only for having the mcuboot configure whatever mechanism (sw/hw) to the application's vector table, if the booted application does not do it itself. Otherwise, the application itself is expected to configure the VTOR/_vector_table_pointer (see line 151 for this case).

ioannisg: @e-rk, if you have the following configuration:
CONFIG_CPU_HAS_VTOR=y
CONFIG_SW_VECTOR_RELAY=y
then, you set SCB.VTOR = __vector_relay_table
But then _vector_table_pointer will only be set if CONFIG_BOOT_INTR_VEC_RELOC is defined. If CONFIG_BOOT_INTR_VEC_RELOC is not set, _vector_table_pointer  is undefined. I wanted to avoid this non-deterministic case.

ioannisg: The original code sets _vector_table_pointer only if CONFIG_BOOT_INTR_VEC_RELOC is set. So you are introducing a logical change here. At minimum the commit message shall explain this.

e-rk: I see now where you are coming from. Adding `&& defined(CONFIG_BOOT_INTR_VEC_RELOC)` should do no harm.

Regarding setting the `_vector_table_pointer` I'm confused a little bit. Both before and after change the pointer is assigned when both `CONFIG_BOOT_INTR_VEC_RELOC` and `CONFIG_SW_VECTOR_RELAY` are set.
`SCB->VTOR` is assigned differently and it is mentioned in the commit message. 

In summary, the final outcomes are (without applying your suggestion and assuming https://github.com/zephyrproject-rtos/zephyr/pull/26276 is merged):
  * with `CONFIG_SW_VECTOR_RELAY=y`, `CONFIG_BOOT_INTR_VEC_RELOC=n` and `CONFIG_CPU_HAS_VTOR=n`
      * Before:
        `_vector_table_pointer = bootloader` ( set in `prep_c.c`)
      * After:
        `_vector_table_pointer = bootloader` ( set in `prep_c.c`)
   * with `CONFIG_SW_VECTOR_RELAY=y`, `CONFIG_BOOT_INTR_VEC_RELOC=y` and `CONFIG_CPU_HAS_VTOR=n` 
      * Before:
        `_vector_table_pointer = application` ( set in `main.c`) 
      * After:
        `_vector_table_pointer = application` ( set in `main.c`)
   * with `CONFIG_SW_VECTOR_RELAY=y`, `CONFIG_BOOT_INTR_VEC_RELOC=n` and `CONFIG_CPU_HAS_VTOR=y` 
      * Before:
       `VTOR = bootloader` (`prep_c.c`)
        `_vector_table_pointer = unassigned`
      * After:
        `VTOR = relay` (`main.c`)
        `_vector_table_pointer = unassigned`
        (Zephyr application sets the VTOR anyway)
   *  with `CONFIG_SW_VECTOR_RELAY=y`, `CONFIG_BOOT_INTR_VEC_RELOC=y` and `CONFIG_CPU_HAS_VTOR=y` 
      * Before:
       `VTOR = bootloader` (`prep_c.c`)
        `_vector_table_pointer = application`
      * After:
        `VTOR = relay` (`main.c`)
        `_vector_table_pointer = application`
   *  with `CONFIG_SW_VECTOR_RELAY=n`, `CONFIG_BOOT_INTR_VEC_RELOC=n` and `CONFIG_CPU_HAS_VTOR=y` 
      * Before:
       `VTOR = bootloader` (`prep_c.c`)

      * After:
        `VTOR = bootloader` (`prep_c.c`)
   *  with `CONFIG_SW_VECTOR_RELAY=n`, `CONFIG_BOOT_INTR_VEC_RELOC=y` and `CONFIG_CPU_HAS_VTOR=y` 
      * Before:
       `VTOR = application` (`main.c`)

      * After:
        `VTOR = application` (`main.c`)

(unless I have made a mistake somewhere)

boot/zephyr/main.c
```diff
@@ -146,9 +140,18 @@ static void do_boot(struct boot_rsp *rsp)
     cleanup_arm_nvic(); /* cleanup NVIC registers */
 #endif
 
-#ifdef CONFIG_BOOT_INTR_VEC_RELOC
-    VTOR = vt;
+#if defined(CONFIG_SW_VECTOR_RELAY) && defined(CONFIG_BOOT_INTR_VEC_RELOC)
```
nvlsianpu, line 143: Maybe this could be some more legible, as  CONFIG_BOOT_INTR_VEC_RELOC is common here:
```suggestion
#ifdef CONFIG_BOOT_INTR_VEC_RELOC
#if defined(CONFIG_SW_VECTOR_RELAY)
    _vector_table_pointer = vt;
#if defined(CONFIG_CPU_CORTEX_M_HAS_VTOR)
    /* Set VTOR to point to the relay vector table before the chain-load. */
    SCB->VTOR = (uint32_t)__vector_relay_table;
#endif
#elif  defined(CONFIG_CPU_CORTEX_M_HAS_VTOR)
    SCB->VTOR = vt;
#endif /* CONFIG_BOOT_INTR_VEC_RELOC */
```

e-rk: Done.

boot/zephyr/main.c
```diff
@@ -147,8 +141,15 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 
 #ifdef CONFIG_BOOT_INTR_VEC_RELOC
-    VTOR = vt;
+#if CONFIG_SW_VECTOR_RELAY
```
nvlsianpu, line 144: #ifdef

e-rk: Eyup, forgot about that one. Done.

utzig:
Please rebase on latest master.

boot/zephyr/main.c
```diff
@@ -147,8 +141,15 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 
 #ifdef CONFIG_BOOT_INTR_VEC_RELOC
-    VTOR = vt;
+#if defined(CONFIG_SW_VECTOR_RELAY)
+    _vector_table_pointer = vt;
+#ifdef CONFIG_CPU_CORTEX_M_HAS_VTOR
+    SCB->VTOR = (uint32_t)__vector_relay_table;
 #endif
+#elif defined(CONFIG_CPU_CORTEX_M_HAS_VTOR)
```
ioannisg, line 149: This is weird @e-rk looks both the ifdef and elif have the same check

utzig: > This is weird @e-rk looks both the ifdef and elif have the same check

The `elif` refers to the `#if defined(CONFIG_SW_VECTOR_RELAY)` above, so not the same checks.

ioannisg: yes, realized that, sorry for the noise.

boot/zephyr/main.c
ioannisg: @e-rk @utzig @nvlsianpu I see the following Kconfig configuration

```
config BOOT_INTR_VEC_RELOC
	bool "Relocate the interrupt vector to the application"
	default n
	depends on SW_VECTOR_RELAY || CPU_CORTEX_M_HAS_VTOR
	help
	  Relocate the interrupt vector to the application before it is started.
	  Select this option if application requires vector relocation,
	  but it doesn't relocate vector in its reset handler.

```

So, unless I miss something, this is not ON by default. It can be set on any Mainline Cortex-M, and on Baseline (if SW_VECTOR_RELAY is enabled). But the default is OFF.

[When building BL with SW_VECTOR_RELAY]

   - In CPUs without VTOR: there is no problem, because Zephyr boot code will anyway set the vector_table_pointer address to the BL's vector start. So the BL will always be able to service its IRQs.
Before jumping to Application, BOOT_INTR_VEC_RELOC controls whether the BL updates the vector_table_pointer_address, or not. If BOOT_INTR_VEC_RELOC=n, BL will do nothing (we expect the Application to update the pointer)

   - In CPUs with VTOR: Zephyr boot code will set VTOR to BL's vector table, so BL can service it's own interrupts. Before jumping to Application, BOOT_INTR_VEC_RELOC controls whether the BL updates the VTOR to point to the relay vector table, so if this is left to OFF there will be no update and the Application cannot get interrupts forwarded.

ioannisg: To summarize, this is the logic I think we should follow:

1. CORTEX_M_HAS_VTOR=n:

   - Case 1.1: `SW_VECTOR_RELAY=n`
      No relaying. prep_c.c does nothing. BL jump software will also do nothing. Obviously, we rely on Application to use HW mechanism to relocate interrupt vector table. In this case BOOT_INTR_VEC_RELOC option is not available.

   - Case 1.2: `SW_VECTOR_RELAY=y`
     In this case BOOT_INTR_VEC_RELOC is available and we have the following options:
     - Case 1.2.1: `BOOT_INTR_VEC_RELOC=n`
        Zephyr creates relay-vector-table _and_ vector-table. relay-vector-table is used by BL. At prep_c.c vector-table-pointer is set to BL's vector-table, so BL can have IRQs. Nothing is done before jump. We rely on Application to re-configure vector-table-pointer
     - Case 1.2.2: `BOOT_INTR_VEC_RELOC=y`
      Zephyr creates relay-vector-table _and_ vector-table. relay-vector-table is used by BL. At prep_c.c vector-table-pointer is set to BL's vector-table, so BL can have IRQs. Before jump we do the job and re-configure vector-table-pointer to point to Applications' vector-table.

This is already working ok, and this PR does not change this functionality.

2. CORTEX_M_HAS_VTOR=y:

    BOOT_INTR_VEC_RELOC and SW_VECTOR_RELAY options are available. We have the following cases:
    - Case 2.1: SW_VECTOR_RELAY=n
      No relaying. Zephyr only creates vector-table. prep_c.c set VTOR to BL's vector-table so BL can have IRQs.
          - Case 2.1.1: BOOT_INTR_VEC_RELOC=n
             Do nothing before jump
          - Case 2.1.2: BOOT_INTR_VEC_RELOC=y
             Set VTOR to Application's vector table before jump.
    - Case 2.2: SW_VECTOR_RELAY=y (this case was made possible with zephyrproject-rtos/zephyr#26276)
       We want to relay interrupts. Zephyr will create relay-vector-table and vector-table. prep_c.c will set VTOR to the
      vector-table, so BL can have IRQs.
          - Case 2.2.1: BOOT_INTR_VEC_RELOC=n
             Currently, we do nothing before jump. This is a problem because VTOR must be re-located to the relay-table :)
          - Case 2.2.2: BOOT_INTR_VEC_RELOC=y
            Before jump, set VTOR to the relay-vector-table, and vector-table-pointer to the Applications' vector-table.

e-rk: I see. Do you think for `CORTEX_M_HAS_VTOR=y`, `SW_VECTOR_RELAY=y` and `BOOT_INTR_VEC_RELOC=n` we should reroute interrupts from direct `VTOR->bootloader vector` to `VTOR->relay vector->bootloader vector`?

ioannisg: No, that's the point. When BOOT_INTR_VEC_RELOC=n the BL does not need to care about the Application at all. It should **only** set VTOR to the relay table address right before the jump. In this case the Application will be responsible for adapting the vector table pointer (and this will solve the legacy use-case of yours)
Is my thinking ok?

e-rk: After some discussion, we decided to go with the `VTOR->relay vector->bootloader vector` flow for `CORTEX_M_HAS_VTOR=y && SW_VECTOR_RELAY=y && BOOT_INTR_VEC_RELOC=n`

ioannisg:
LGTM now
@utzig FYI

ioannisg:
@utzig @d3zd3z , maybe this could be merged now?
I ran Zephyr CI with this PR here: https://github.com/zephyrproject-rtos/zephyr/pull/27012

utzig:
> @utzig @d3zd3z , maybe this could be merged now?
> I ran Zephyr CI with this PR here: [zephyrproject-rtos/zephyr#27012](https://github.com/zephyrproject-rtos/zephyr/pull/27012)

You should be able to merge it.

PR #776, opened 2020-07-24
SW counter measures against HW level fault injection

tamasban:
Adding SW counter measures against fault injection attacks in HW. These attacks could be clock or power glitching, EM pulse, laser beam, temperature change.  The goal of the attacker is is to execute unauthenticated code on the device.  The expectation of the attacker that due to these injected faults the CPU will change its normal behaviour and in this way create a benefit for the attacker to bypass secure boot and image authentication. The potential differences in the CPU behaviour could be one or more instruction skip, modifying register content, corrupting memory read, corrupting instruction decoding, resetting bits in program status register. 

The SW counter measure are not guaranteeing 100% protection against all HW attack but make harder to bypass image authentication. The SW counter measures are disabled by default they can be enabled with MCUBOOT_FIH_PROFILE_* macros. Currently there are defined 4 protection level: none, moderate, medium, max. These are enabling more and more SW counter measures. These are only added to the generic part of MCUboot files in `boot/bootutil/src`. Currently only the critical call path is hardened where a corruption of a single `if (ret)` statement can be accept an unauthenticated image. The crypto (mbed-crypto, tinycrypt) libraries are independent projects from MCUboot so they are not changed. They might contains similar feature or not, need to be checked individually.

Main hardening code are in :

- boot/bootutil/include/bootutil/fault_injection_hardening.h
- boot/bootutil/src/fault_injection_hardening.c

### DIFF
--- a/boot/bootutil/include/bootutil/fault_injection_hardening.h
+++ b/boot/bootutil/include/bootutil/fault_injection_hardening.h
@@ -1,0 +1,356 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Copyright (c) 2020 Arm Limited
+ */
+
+#ifndef __FAULT_INJECTION_HARDENING_H__
+#define __FAULT_INJECTION_HARDENING_H__
+
+/* Fault injection mitigation library.
+ *
+ * Has support for different measures, which can either be enabled/disabled
+ * separately or by defining one of the MCUBOOT_FIH_PROFILEs.
+ *
+ * NOTE: These constructs against fault injection attacks are not guaranteed to
+ *       be secure for all compilers, but execution is going to be correct and
+ *       including them will certainly help to harden the code.
+ *
+ * FIH_ENABLE_DOUBLE_VARS makes critical variables into a tuple (x, x ^ msk).
+ * Then the correctness of x can be checked by XORing the two tuple values
+ * together. This also means that comparisons between fih_ints can be verified
+ * by doing x == y && x_msk == y_msk.
+ *
+ * FIH_ENABLE_GLOBAL_FAIL makes all while(1) failure loops redirect to a global
+ * failure loop. This loop has mitigations against loop escapes / unlooping.
+ * This also means that any unlooping won't immediately continue executing the
+ * function that was executing before the failure.
+ *
+ * FIH_ENABLE_CFI (Control Flow Integrity) creates a global counter that is
+ * incremented before every FIH_CALL of vulnerable functions. On the function
+ * return the counter is decremented, and after the return it is verified that
+ * the counter has the same value as before this process. This can be used to
+ * verify that the function has actually been called. This protection is
+ * intended to discover that important functions are called in an expected
+ * sequence and neither of them is missed due to an instruction skip which could
+ * be a result of glitching attack. It does not provide protection against ROP
+ * or JOP attacks.
+ *
+ * FIH_ENABLE_DELAY causes random delays. This makes it hard to cause faults
+ * precisely. It requires an RNG. An mbedtls integration is provided in
+ * fault_injection_hardening_delay_mbedtls.h, but any RNG that has an entropy
+ * source can be used by implementing the fih_delay_random_uchar function.
+ *
+ * The basic call pattern is:
+ *
+ * fih_int fih_rc = FIH_FAILURE;
+ * FIH_CALL(vulnerable_function, fih_rc, arg1, arg2);
+ * if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
+ *     FIH_PANIC;
+ * }
+ *
+ * Note that any function called by FIH_CALL must only return using FIH_RETURN,
+ * as otherwise the CFI counter will not be decremented and the CFI check will
+ * fail causing a panic.
+ */
+
+#include "mcuboot_config/mcuboot_config.h"
+
+#if defined(MCUBOOT_FIH_PROFILE_HIGH)
+
+#define FIH_ENABLE_DELAY         /* Requires an entropy source */
+#define FIH_ENABLE_DOUBLE_VARS
+#define FIH_ENABLE_GLOBAL_FAIL
+#define FIH_ENABLE_CFI
+
+#elif defined(MCUBOOT_FIH_PROFILE_MEDIUM)
+
+#define FIH_ENABLE_DOUBLE_VARS
+#define FIH_ENABLE_GLOBAL_FAIL
+#define FIH_ENABLE_CFI
+
+#elif defined(MCUBOOT_FIH_PROFILE_LOW)
+
+#define FIH_ENABLE_GLOBAL_FAIL
+#define FIH_ENABLE_CFI
+
+#else
+#define MCUBOOT_FIH_PROFILE_OFF
+#endif /* MCUBOOT_FIH_PROFILE */
+
+#ifdef FIH_ENABLE_DELAY
+#include "fault_injection_hardening_delay_rng.h"
+#endif /* FIH_ENABLE_DELAY */
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/* Non-zero success value to defend against register resets. Zero is the most
+ * common value for a corrupted register so complex bit-patterns are used
+ */
+#ifndef MCUBOOT_FIH_PROFILE_OFF
+#define FIH_POSITIVE_VALUE 0x1AAAAAAA
+#define FIH_NEGATIVE_VALUE 0x15555555
+#else
+#define FIH_POSITIVE_VALUE 0
+#define FIH_NEGATIVE_VALUE -1
+#endif
+
+/* A volatile mask is used to prevent compiler optimization - the mask is xored
+ * with the variable to create the backup and the integrity can be checked with
+ * another xor. The mask value doesn't _really_ matter that much, as long as
+ * it has reasonably high hamming weight.
+ */
+#define _FIH_MASK_VALUE 0xBEEF
+
+#ifdef FIH_ENABLE_DOUBLE_VARS
+
+/* All ints are replaced with two int - the normal one and a backup which is
+ * XORed with the mask.
+ */
+extern volatile int _fih_mask;
+typedef volatile struct {
+    volatile int val;
+    volatile int msk;
+} fih_int;
+
+#else
+
+typedef int fih_int;
+
+#endif /* FIH_ENABLE_DOUBLE_VARS */
+
+extern fih_int FIH_SUCCESS;
+extern fih_int FIH_FAILURE;
+
+#ifdef FIH_ENABLE_GLOBAL_FAIL
+/* Global failure handler - more resistant to unlooping. noinline and used are
+ * used to prevent optimization
+ */
+__attribute__((noinline)) __attribute__((used))
+void fih_panic_loop(void);
+#define FIH_PANIC fih_panic_loop()
+#else
+#define FIH_PANIC while (1) {}
+#endif  /* FIH_ENABLE_GLOBAL_FAIL */
+
+/* NOTE: For functions to be inlined outside their compilation unit they have to
+ * have the body in the header file. This is required as function calls are easy
+ * to skip.
+ */
+#ifdef FIH_ENABLE_DELAY
+#include "bootutil/bootutil_log.h"
+
+/* Delaying logic, with randomness from a CSPRNG */
+__attribute__((always_inline)) inline
+int fih_delay(void)
+{
+    unsigned char delay;
+    int foo = 0;
+    volatile int rc;
+
+    delay = fih_delay_random_uchar();
+
+    for (volatile int i = 0; i < delay; i++) {
+        foo++;
+    }
+
+    rc = 1;
+
+    /* rc is volatile so if it is the return value then the function cannot be
+     * optimized
+     */
+    return rc;
+}
+
+#else
+
+__attribute__((always_inline)) inline
+int fih_delay_init(void)
+{
+    return 1;
+}
+
+__attribute__((always_inline)) inline
+int fih_delay(void)
+{
+    return 1;
+}
+#endif /* FIH_ENABLE_DELAY */
+
+#ifdef FIH_ENABLE_DOUBLE_VARS
+
+__attribute__((always_inline)) inline
+void fih_int_validate(fih_int x)
+{
+    if (x.val != (x.msk ^ _fih_mask)) {
+        FIH_PANIC;
+    }
+}
+
+/* Convert a fih_int to an int. Validate for tampering. */
+__attribute__((always_inline)) inline
+int fih_int_decode(fih_int x)
+{
+    fih_int_validate(x);
+    return x.val;
+}
+
+/* Convert an int to a fih_int, can be used to encode specific error codes. */
+__attribute__((always_inline)) inline
+fih_int fih_int_encode(int x)
+{
+    fih_int ret = {x, x ^ _fih_mask};
+    return ret;
+}
+
+/* Standard equality. If A == B then 1, else 0 */
+__attribute__((always_inline)) inline
+int fih_eq(fih_int x, fih_int y)
+{
+    fih_int_validate(x);
+    fih_int_validate(y);
+    return (x.val == y.val) && fih_delay() && (x.msk == y.msk);
+}
+
+__attribute__((always_inline)) inline
+int fih_not_eq(fih_int x, fih_int y)
+{
+    fih_int_validate(x);
+    fih_int_validate(y);
+    return (x.val != y.val) && fih_delay() && (x.msk != y.msk);
+}
+
+#else
+
+/* NOOP */
+__attribute__((always_inline)) inline
+void fih_int_validate(fih_int x)
+{
+    (void) x;
+    return;
+}
+
+/* NOOP */
+__attribute__((always_inline)) inline
+int fih_int_decode(fih_int x)
+{
+    return x;
+}
+
+/* NOOP */
+__attribute__((always_inline)) inline
+fih_int fih_int_encode(int x)
+{
+    return x;
+}
+
+__attribute__((always_inline)) inline
+int fih_eq(fih_int x, fih_int y)
+{
+    return x == y;
+}
+
+__attribute__((always_inline)) inline
+int fih_not_eq(fih_int x, fih_int y)
+{
+    return x != y;
+}
+#endif /* FIH_ENABLE_DOUBLE_VARS */
+
+/* C has a common return pattern where 0 is a correct value and all others are
+ * errors. This function converts 0 to FIH_SUCCESS and any other number to a
+ * value that is not FIH_SUCCESS
+ */
+__attribute__((always_inline)) inline
+fih_int fih_int_encode_zero_equality(int x)
+{
+    if (x) {
+        return FIH_FAILURE;
+    } else {
+        return FIH_SUCCESS;
+    }
+}
+
+#ifdef FIH_ENABLE_CFI
+extern fih_int _fih_cfi_ctr;
+#endif /* FIH_ENABLE_CFI */
+
+fih_int fih_cfi_get_and_increment(void);
+void fih_cfi_validate(fih_int saved);
+void fih_cfi_decrement(void);
+
+/* Label for interacting with FIH testing tool. Can be parsed from the elf file
+ * after compilation. Does not require debug symbols.
+ */
+#define FIH_LABEL(str) __asm volatile ("FIH_LABEL_" str "_%=:" ::);
+
+/* Main FIH calling macro. return variable is second argument. Does some setup
+ * before and validation afterwards. Inserts labels for use with testing script.
+ *
+ * First perform the precall step - this gets the current value of the CFI
+ * counter and saves it to a local variable, and then increments the counter.
+ *
+ * Then set the return variable to FIH_FAILURE as a base case.
+ *
+ * Then perform the function call. As part of the funtion FIH_RET must be called
+ * which will decrement the counter.
+ *
+ * The postcall step gets the value of the counter and compares it to the
+ * previously saved value. If this is equal then the function call and all child
+ * function calls were performed.
+ */
+#define FIH_CALL(f, ret, ...) \
+    do { \
+        FIH_LABEL("FIH_CALL_START"); \
+        FIH_CFI_PRECALL_BLOCK; \
+        ret = FIH_FAILURE; \
+        if (fih_delay()) { \
+            ret = f(__VA_ARGS__); \
+        } \
+        FIH_CFI_POSTCALL_BLOCK; \
+        FIH_LABEL("FIH_CALL_END"); \
+    } while (0)
+
+/* FIH return changes the state of the internal state machine. If you do a
+ * FIH_CALL then you need to do a FIH_RET else the state machine will detect
+ * tampering and panic.
+ */
+#define FIH_RET(ret) \
+    do { \
+        FIH_CFI_PRERET; \
+        return ret; \
+    } while (0)
+
+
+#ifdef FIH_ENABLE_CFI
+/* Macro wrappers for functions - Even when the functions have zero body this
+ * saves a few bytes on noop functions as it doesn't generate the call/ret
+ *
+ * CFI precall function saves the CFI counter and then increments it - the
+ * postcall then checks if the counter is equal to the saved value. In order for
+ * this to be the case a FIH_RET must have been performed inside the called
+ * function in order to decrement the counter, so the function must have been
+ * called.
+ */
+#define FIH_CFI_PRECALL_BLOCK \
+    fih_int _fih_cfi_saved_value = fih_cfi_get_and_increment()
+
+#define FIH_CFI_POSTCALL_BLOCK \
+        fih_cfi_validate(_fih_cfi_saved_value)
+
+#define FIH_CFI_PRERET \
+        fih_cfi_decrement()
+#else
+#define FIH_CFI_PRECALL_BLOCK
+#define FIH_CFI_POSTCALL_BLOCK
+#define FIH_CFI_PRERET
+#endif  /* FIH_ENABLE_CFI */
+
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+
+#endif /* __FAULT_INJECTION_HARDENING_H__ */
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -3,7 +3,7 @@
  *
  * Copyright (c) 2017-2019 Linaro LTD
  * Copyright (c) 2016-2019 JUUL Labs
- * Copyright (c) 2019 Arm Limited
+ * Copyright (c) 2019-2020 Arm Limited
  *
  * Original license:
  *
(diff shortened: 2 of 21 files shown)


boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,339 @@
...
+#include "mcuboot_config/mcuboot_config.h"
+
+#if defined(MCUBOOT_FIH_PROFILE_MAX)
+
+#define FIH_ENABLE_DELAY         /* Requires an entropy source */
+#define FIH_ENABLE_DOUBLE_VARS
+#define FIH_ENABLE_GLOBAL_FAIL
+#define FIH_ENABLE_CFI
+
+#elif defined(MCUBOOT_FIH_PROFILE_MEDIUM)
```
RcColes, line 62: IMO Using both `MODERATE` and `MEDIUM` could be very confusing. I think something like `OFF / LOW / MEDIUM / HIGH` would be better.

adrianlshaw: Agreed

utzig:
This needs rebase. Also I would appreciate if you could submit me links to papers/&c that explain why those changes are made the way they are!

tamasban:
> This needs rebase. Also I would appreciate if you could submit me links to papers/&c that explain why those changes are made the way they are!

Here is two white paper in the topic and also the QEMU based testing (will be another PR) can prove the usefulness of these patterns:

- https://www.cl.cam.ac.uk/~rja14/Papers/whatyouc.pdf

- https://www.riscure.com/uploads/2018/11/201708_Riscure_Whitepaper_Side_Channel_Patterns.pdf

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: Fixed.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,339 @@
...
+
+    rc = 1;
+
+    /* rc is volatile so if it is the return value then the function cannot be
+     * optimized
+     */
+    return rc;
+}
+#else
+__attribute__((always_inline)) inline
```
soby-mathew, line 163: minor: The `always inline` attribute is still an indication that a function must be inlined. Macros are the better way to ensure inlining. Same comment below.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * CFI precall function saves the CFI counter and then increments it - the
+ * postcall then checks if the counter is equal to the saved value. In order for
+ * this to be the case a FIH_RET must have been performed inside the called
+ * function in order to decrement the counter, so the function must have been
+ * called.
+ */
+#ifdef FIH_ENABLE_CFI
+#define FIH_CFI_PRECALL_BLOCK \
+    fih_int _fih_cfi_saved_value = fih_cfi_get_and_increment()
```
soby-mathew, line 322: Just some blue sky thinking, not sure whether it can be done or not. If the increment can be done based on the function being invoked i.e some way to hash the name of the function into a unique number, then we can confirm if an increment is called for a function call, then the decrement in the same function need to be invoked for it to be validated. 

This would ensure that if the call is subverted to a different location and the RET from that different function will fail the check on return back to the caller.

soby-mathew (commented):
just some suggestion. Otherwise looks good to me.

boot/bootutil/include/bootutil/fault_injection_hardening.h
RcColes: There's a deliberate tradeoff between usability and performance going on here; Macros are certainly be best way to ensure things are always inlined but they come with some tradeoffs. They tend to cause issues with debuggers / debug symbols and also things like code-completion / code formatters etc. All the compilers I tested did respect the `always_inline`, but I do agree that it could be an issue in theory.

boot/bootutil/include/bootutil/fault_injection_hardening.h
RcColes: It would be an interesting idea, but maybe too difficult in practice. The C preprocessor does have access to the function name via `__func__`, but I'm not sure if you could evaluate a hash function or similar inside the preprocessor. I'd avoid doing the hash inside actual code, as that'd cause a fairly large size increase due to the inlining.

In general, most functions don't have an FIH_RET and it's pretty hard to control the value of register corruption (and hence jump location). More targeted attacks like ion beams could probably target the lower bits deliberately, which would allow jumps to be slightly more precise.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,339 @@
...
+ * loop. This loop has mitigations against loop escapes / unlooping. This also
+ * means that any unlooping won't immediately continue executing the function
+ * that was executing before the failure.
+ *
+ * CFI (Control Flow Integrity) creates a global counter that is incremented
+ * before every FIH_CALL of vulnerable functions. On the function return the
+ * counter is decremented, and after the return it is verified that the counter
+ * has the same value as before this process. This can be used to verify that
+ * the function has actually been called.
+ *
```
adrianlshaw, line 34: There should be some description here that describes how this only provides a specific subset of CFI. This is checking that a specific sequence of calls are made within the same block and that the adversary is someone performing glitching. It does not cover other types of CFI such as ROP protection or JOP protection.

boot/bootutil/include/bootutil/fault_injection_hardening.h
adrianlshaw: This would be ideal. A similar alternative might be easier: take the function address and derive a number from it. 

It's worth noting that fault injection techniques are unlikely to produce precise values for a return address. If precise return addresses are possible then it is time to consider more comprehensive ROP/JOP protection, which I believe is beyond the scope of the attacks/adversaries considered in this PR.

adrianlshaw: I see Raef beat me to it :)

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: Your are right. So far they were inlined. I would keep in this way for simplicity. If one want to use this feature then the compiled code must be verified by test and visual inspection. So functions not be inlined should be caught by these steps.

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: Might `sizeof(__func__)` would add some randomness.

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: Fixed

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,343 @@
...
+
+#ifdef FIH_ENABLE_CFI
+extern fih_int _fih_cfi_ctr;
+#endif /* FIH_ENABLE_CFI */
+
+fih_int fih_cfi_get_and_increment(void);
+void fih_cfi_validate(fih_int saved);
+void fih_cfi_decrement(void);
+
+/* Label for interacting with FI testing tool. Can be parsed from the elf file
```
nvlsianpu, line 273:
```suggestion
/* Label for interacting with FIH testing tool. Can be parsed from the elf file
```

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,343 @@
...
+ * FIH_CALL then you need to do a FIH_RET else the state machine will detect
+ * tampering and panic.
+ */
+#define FIH_RET(ret) \
+    do { \
+        FIH_CFI_PRERET; \
+        return ret; \
+    } while (0)
+
+/* Macro wrappers for functions - Even when the functions have zero body this
```
nvlsianpu, line 315: move macro doc right above the macro (after `ifdef FIH_ENABLE_CFI`)

utzig:
testing some message in GH slack integration, will remove...

utzig:
Please rebase.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,343 @@
...
+/* Fault injection mitigation library.
+ *
+ * Has support for different measures, which can either be enabled/disabled
+ * separately or by defining one of the MCUBOOT_FIH_PROFILEs.
+ *
+ * NOTE: These constructs against fault injection attacks are not guaranteed to
+ *       be secure for all compilers, but execution is going to be correct and
+ *       including them will certainly help to harden the code.
+ *
+ * DOUBLE_VARS makes critical variables into a tuple (x, x ^ msk). Then the
```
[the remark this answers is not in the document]
tamasban: ack

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,343 @@
...
+
+#ifndef __FAULT_INJECTION_HARDENING_H__
+#define __FAULT_INJECTION_HARDENING_H__
+
+/* Fault injection mitigation library.
+ *
+ * Has support for different measures, which can either be enabled/disabled
+ * separately or by defining one of the MCUBOOT_FIH_PROFILEs.
+ *
+ * NOTE: These constructs against fault injection attacks are not guaranteed to
```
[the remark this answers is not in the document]
tamasban: There is another PR, which is a test suite on top of QEMU:
https://github.com/JuulLabs-OSS/mcuboot/pull/789

This can be used with any compiler to validate the hardening code. It does not provide a 100% test coverage against FI on the entire mcuboot binary. It testes only that code sections which are between the FIH_CALL_START and FIH_CALL_END macros.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+/* Convert a fih_int to an int. Validate for tampering. */
+__attribute__((always_inline)) inline
+int fih_int_decode(fih_int x)
+{
+    fih_int_validate(x);
+    return x.val;
+}
```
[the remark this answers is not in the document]
tamasban: ack

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: ack

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: ack

utzig:
@tamasban You don't really need to keep `ack`ing, do you? I get one email for every time you ack something. Why not just fix the problem and push?

tamasban:
> @tamasban You don't really need to keep `ack`ing, do you? I get one email for every time you ack something. Why not just fix the problem and push?

understood, sorry I get to used to it in Gerrit workflow.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -107,6 +108,38 @@ static const struct boot_swap_table boot_swap_tables[] = {
...
+ * @param s1    The first  memory region to compare.
+ * @param s2    The second memory region to compare.
+ * @param n     The amount of bytes to compare.
+ *
+ * @note        This function does not comply with the specification of memcmp,
+ *              so should not be considered a drop-in replacement.
+ *
+ * @return      FIH_SUCCESS if memory regions are equal, otherwise FIH_FAILURE
+ */
+fih_int boot_secure_memequal(const void *s1, const void *s2, size_t n)
```
utzig, line 123: I can see what you're doing but when I see `secure_memequal` I expect a function that is constant time, which is not the case here, maybe a more appropriate name would be helpful, maybe settle on `fih` like `boot_fih_memequal` (sorry for the lack of inspiration!)

boot/bootutil/src/bootutil_misc.c
```diff
@@ -107,6 +108,38 @@ static const struct boot_swap_table boot_swap_tables[] = {
...
+{
+    size_t i;
+    uint8_t *s1_p = (uint8_t*) s1;
+    uint8_t *s2_p = (uint8_t*) s2;
+    fih_int ret = FIH_FAILURE;
+
+    for (i = 0; i < n; i++) {
+        if (s1_p[i] != s2_p[i]) {
+            goto out;;
```
utzig, line 132: can remove one `;`

utzig:
This should get back around 40 bytes for someone who doesn't want to have the fault-injection protections:

```
diff --git a/boot/bootutil/src/bootutil_misc.c b/boot/bootutil/src/bootutil_misc.c
index 9d80b5a06..a48b1431f 100644
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -120,6 +120,12 @@ static const struct boot_swap_table boot_swap_tables[] = {
  *
  * @return      FIH_SUCCESS if memory regions are equal, otherwise FIH_FAILURE
  */
+#ifdef MCUBOOT_FIH_PROFILE_OFF
+inline int boot_secure_memequal(const void *s1, const void *s2, size_t n)
+{
+    return memcmp(s1, s2, n);
+}
+#else
 fih_int boot_secure_memequal(const void *s1, const void *s2, size_t n)
 {
     size_t i;
@@ -139,6 +145,7 @@ fih_int boot_secure_memequal(const void *s1, const void *s2, size_t n)
 out:
     FIH_RET(ret);
 }
+#endif
 
 static int
 boot_magic_decode(const uint32_t *magic)
```

boot/bootutil/src/image_validate.c
```diff
@@ -449,44 +464,55 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
...
+                goto out;
             }
 
             /* Compare the new image's security counter value against the
              * stored security counter value.
              */
-            if (img_security_cnt < security_cnt) {
-                /* The image's security counter is not accepted. */
-                return -1;
+            fih_rc = fih_int_encode_zero_equality(img_security_cnt <
```
nvlsianpu, line 485:
```suggestion
  if (img_security_cnt < fih_int_decode(security_cnt)) {
    fih_rc = FIH_SUCCESS;
  } else {
    fih_rc = FIH_FAILURE;
  }
```
just think that usage _encode_ funct here is overcomplicated

boot/bootutil/src/loader.c
```diff
@@ -386,13 +387,14 @@ boot_write_status(const struct boot_loader_state *state, struct boot_status *bs)
...
  */
-static int
+static fih_int
 boot_image_check(struct boot_loader_state *state, struct image_header *hdr,
                  const struct flash_area *fap, struct boot_status *bs)
 {
     TARGET_STATIC uint8_t tmpbuf[BOOT_TMPBUF_SZ];
     uint8_t image_index;
-    int rc;
+    int rc = -1;
```
nvlsianpu, line 396:
```suggestion
    int rc;
```

boot/bootutil/src/loader.c
```diff
@@ -386,13 +387,14 @@ boot_write_status(const struct boot_loader_state *state, struct boot_status *bs)
...
+    fih_int fih_rc = FIH_FAILURE;
```
nvlsianpu, line 397:
```suggestion
    fih_int fih_rc ;
```
always assigned by FIH_CALL(bootutil_img_validate...

boot/bootutil/src/loader.c
```diff
@@ -574,19 +575,20 @@ boot_version_cmp(const struct image_version *ver1,
...
+static fih_int
 boot_validate_slot(struct boot_loader_state *state, int slot,
                    struct boot_status *bs)
 {
     const struct flash_area *fap;
     struct image_header *hdr;
     int area_id;
-    int rc;
+    fih_int fih_rc = FIH_FAILURE;
+    int rc = 0;
```
nvlsianpu, line 586:
```suggestion
    int rc ;
```

boot/bootutil/src/loader.c
```diff
@@ -643,16 +646,14 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         BOOT_LOG_ERR("Image in the %s slot is not valid!",
                      (slot == BOOT_PRIMARY_SLOT) ? "primary" : "secondary");
 #endif
-        rc = 1;
+        fih_rc = fih_int_encode(1);
         goto out;
     }
 
-    /* Image in the secondary slot is valid. */
-    rc = 0;
```
nvlsianpu, line 651: shouldn't a success code be encoded to `fih_rc`?

boot/bootutil/src/loader.c
```diff
@@ -675,7 +676,7 @@ boot_update_security_counter(uint8_t image_index, int slot,
 {
     const struct flash_area *fap = NULL;
     uint32_t img_security_cnt;
-    int rc;
+    int rc = 0;
```
nvlsianpu, line 679:
```suggestion
    int rc;
```

boot/bootutil/src/loader.c
tamasban: It is intended to protect against attack when the PC is tried to abuse such a way (manipulating lower bits with EM pulse) to jump at the end of the function and just returns without performing any operation. Very unlikely and hard to achieve in practice. But during a certification process, code reviewer can mandate to have negative return value set by default.

utzig: Yes, all variables should be initialized to the error value, otherwise abuse and unintended early exit would result in success returns. I wonder if those could not just be initialized at the beginning of the function instead of where they were declared, but I don't have strong opinions about it anymore.

boot/bootutil/src/loader.c
tamasban: You are right, and `fih_int_encode()` does exactly this: Converting an int to fih_int. You can check: fault_injection_hardening.h:203

boot/bootutil/src/loader.c
nvlsianpu: Ok, I'm fine with this clarification - although all that might be compiled out as optimization.

boot/bootutil/src/image_validate.c
tamasban: Yeah, tend to agree.  I think both version has the same level of resistance against FI. With the usage of `fih_int_encode_zero_equality` the code is a bit more concise and if  `fih_int_encode_zero_equality` would have adding some random delay (calling `fih_delay()`) then it would be more robust. But I can accept your version as well. Let me know which one do you prefer.

nvlsianpu: This was not a strong suggestion from my side.

boot/bootutil/src/loader.c
nvlsianpu: I see` line 637:  FIH_CALL(boot_image_check, fih_rc, state, hdr, fap, bs);` may set `fih_rc` to FIH_SUCCESS.

So this function now returns either `FIH_FAILURE` (previous `-1`), `FIH_SUCCESS` (previous `0`), and the value which is encoded by fih_int_encode(1) call (previous `1`).

I can't see initialization for either `FIH_FAILURE` and `FIH_SUCCESS`, so what the warranty that all value listed here are distinguishable?

boot/bootutil/src/loader.c
```diff
@@ -574,19 +575,20 @@ boot_version_cmp(const struct image_version *ver1,
  *         1 if no bootloable image was found
```
nvlsianpu, line 575: @returns description need to be upadted.

nvlsianpu (changes requested):

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -0,0 +1,356 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/* Non-zero success value to defend against register resets. Zero is the most
+ * common value for a corrupted register so complex bit-patterns are used
+ */
+#ifndef MCUBOOT_FIH_PROFILE_OFF
+#define FIH_POSITIVE_VALUE 0x1AAAAAAA
```
nvlsianpu, line 94: so these two value are problematic if would be a case as previous integer return-value, not a case for this code, but hypothetically might be an issue.

nvlsianpu: I men that `FIH_POSITIVE_VALUE== fih_int_encoded(0x1AAAAAAA)`...

boot/bootutil/src/loader.c
tamasban: FIH_FAILURE and FIH_SUCCESS are just normal integer values if MCUBOOT_FIH_PROFILE is OFF. Actually their value is 0 and 1 in this case. Complex value are used only if some level of FI protection is applied.  In this case might the returns values are colliding: FIH_FAILURE  and 1 (no bootable image). will double check.

boot/bootutil/include/bootutil/fault_injection_hardening.h
tamasban: Not really got it. Please elaborate.

nvlsianpu: `FIH_POSITIVE_VALUE supersedes what before was `0`. FIH_NEGATIV_VALUE` supersedes what was an negative error-code before.
If an interface will try to use return value in a manner where encodes a range of integers as successful results, and uses `FIH_NEGATIV_VALUE` for encoding an error - this will cause misunderstanding, and probably an issue.

fih_int with either `FIH_NEGATIV_VALUE` and `FIH_POSITIVE_VALUE`  should be use like boolean type.

tamasban: Arbitrary error code can be returned, you can use `fih_int_encode()` and `fih_int_decode` to covert them.
There is an example in: loader.c:726
If MCUBOOT_FIH_PROFILE is OFF then these are just normal integers. I have updated the PR to preserve the original error codes. So distinction can be done on the application layer (where `boot_go` or `split_go` returns) .

boot/zephyr/Kconfig
```diff
@@ -288,6 +288,45 @@ config BOOT_SHARE_DATA
...
+	bool "Maximum level hardening against hardware level fault injection"
+	select BOOT_FIH_PROFILE_HIGH
+	help
+	  Maximum level hardening: Long global fail loop to avoid break out,
+	  control flow integrity check to discover discrepancy in expected code
+	  flow, double variables to discover register or memory corruption, random
+	  delays to make code execution less predictable. Random delays requires an
+	  entropy source.
+
+endcoice
```
utzig, line 328: typo: `endchoice`

boot/zephyr/Kconfig
```diff
@@ -297,6 +297,45 @@ config BOOT_SHARE_DATA
 	bool "Save application specific data in shared memory area"
 	default n
 
+choice
+	prompt "Fault injection hardening profile"
+	default BOOT_FIH_PROFILE_OFF
+
+config BOOT_FIH_PROFILE_OFF
+	bool "No hardening against hardware level fault injection"
+	select BOOT_FIH_PROFILE_OFF
```
nvlsianpu, line 306:
```suggestion
```

boot/zephyr/Kconfig
```diff
@@ -297,6 +297,45 @@ config BOOT_SHARE_DATA
...
+	help
+	  No hardening in SW against hardware level fault injection: power or
+	  clock glitching, etc.
+
+config BOOT_FIH_PROFILE_LOW
+	bool "Moderate level hardening against hardware level fault injection"
+	select BOOT_FIH_PROFILE_LOW
```
nvlsianpu, line 313:
```suggestion
```

boot/zephyr/Kconfig
```diff
@@ -297,6 +297,45 @@ config BOOT_SHARE_DATA
...
+	help
+	  Moderate level hardening: Long global fail loop to avoid break out,
+	  control flow integrity check to discover discrepancy in expected code
+	  flow.
+
+config BOOT_FIH_PROFILE_MEDIUM
+	bool "Medium level hardening against hardware level fault injection"
+	select BOOT_FIH_PROFILE_MEDIUM
```
nvlsianpu, line 321:
```suggestion
```

boot/zephyr/Kconfig
```diff
@@ -297,6 +297,45 @@ config BOOT_SHARE_DATA
...
+	help
+	  Medium level hardening: Long global fail loop to avoid break out,
+	  control flow integrity check to discover discrepancy in expected code
+	  flow, double variables to discover register or memory corruption.
+
+config BOOT_FIH_PROFILE_HIGH
+	bool "Maximum level hardening against hardware level fault injection"
+	select BOOT_FIH_PROFILE_HIGH
```
nvlsianpu, line 329:
```suggestion
```

nvlsianpu (commented):
remove selecting kconfig property by itself

utzig:
> remove selecting kconfig property by itself

Yes, and there are other issues, I will squash in fixes.

tamasban:
> > remove selecting kconfig property by itself
> 
> Yes, and there are other issues, I will squash in fixes.

Thanks, for the fixes @utzig!

nvlsianpu (approved):

utzig (approved):
It looks OK, I am a bit bothered that it increases all binaries by 100 bytes or so, which is not that bad but would be better to not add anything if set to `_OFF`. Also for Tinycrypt users `uECC_vli_equal` has to be updated and same for fiat's `CRYPTO_memcmp`, but it can be done in a follow up PR before releasing the next version.

utzig:
@romanjoe Are you able to check this PR for Cypress today?

romanjoe:
@utzig yes, i'll do right now

romanjoe:
First issue i have faced while trying to compile is error message:
`mcuboot/boot/cypress/../bootutil/include/bootutil/crypto/ecdsa_p256.h:17:6: error: #error "One crypto backend must be defined: either CC310 or TINYCRYPT"
     #error "One crypto backend must be defined: either CC310 or TINYCRYPT"`

cypress use `#define MCUBOOT_USE_MBED_TLS`

I assume this change is introduced with https://github.com/JuulLabs-OSS/mcuboot/pull/798, but we did not rebase our internal developing to include this PR yet, so i assume this issue is present in master as well, i need to resolve this issue first

utzig:
@romanjoe Please drop your commits and afterwards @tamasban can rebase.

romanjoe:
This PR fixes compilation for cypress on master: https://github.com/JuulLabs-OSS/mcuboot/pull/825 - please rebase
@utzig @tamasban i also added additional fix regarding cypress in this PR

tamasban:
> This PR fixes compilation for cypress on master: #825 - please rebase
> @utzig @tamasban i also added additional fix regarding cypress in this PR

I have squashed your fix and did the rebase.

romanjoe:
@utzig @tamasban i checked latest with cypress and it works fine, thank you

romanjoe (approved):

PR #781, opened 2020-08-04
boot: zephyr: Use private key

MarekPieta:
Change adds using own private key for MCUBoot bootloader.


MarekPieta:
@nvlsianpu, @hakonfam, could you review, please?

PR #782, opened 2020-08-05
Change imgtool version to 1.7.0a1

ihos-cypress:
Signed-off-by: Ihor Slabkyy <ihor.slabkyy@cypress.com>

### DIFF
--- a/scripts/imgtool/__init__.py
+++ b/scripts/imgtool/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Linaro Limited
+# Copyright 2017-2020 Linaro Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


scripts/imgtool/__init__.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-imgtool_version = "1.6.0"
+imgtool_version = "1.7.0"
```
utzig, line 15: Are we releasing MCUBoot `1.7.0`?

ihos-cypress: I don't know about any plan to release MCUBoot. Suppose if backward compatibility wasn't broken it will be 1.7.0.

utzig: No, we are not releasing `1.7.0` so this can't be `1.7.0`... update it to `1.7.0a1`!

PR #787, opened 2020-08-19
Turn off cache for Cortex M7 before rebooting

arvid-r:
On Cortex M7, and NXP i.MX RT 1020 specifically, MCUBOOT will fail after loading a new image if the instruction cache is enabled. This works for us and might help others, but I need help to figure out where to do this, and what ifdefs to use to turn it on. Also, I'm not sure if the ARM_MPU_Disable thing is really needed.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -128,6 +128,13 @@ static void do_boot(struct boot_rsp *rsp)
     vt = (struct arm_vector_table *)(flash_base +
                                      rsp->br_image_off +
                                      rsp->br_hdr->ih_hdr_size);
+
+#ifdef CONFIG_CPU_CORTEX_M7
+    /* Disable instruction cache and data cache before chain-load the application */
+    SCB_DisableDCache();
+    SCB_DisableICache();
+#endif
+
     irq_lock();
 #ifdef CONFIG_SYS_CLOCK_EXISTS
     sys_clock_disable();


arvid-r:
Any comments on this?

utzig:
> Any comments on this?

I guess most people don't look at "WIP" PRs!

arvid-r:
Fair enough. What would be the best way forward then? I don't know enough about MCUBOOT to judge if this PR is the right fix for this problem, but I know that it fixed my problem (which was very difficult to debug), so I would be happy to share it with the community.

arvid-r:
Well, I can at least investigate if the MPU thing is needed. I'll to that and update this PR if needed.

utzig:
> Well, I can at least investigate if the MPU thing is needed. I'll do that and update this PR if needed.

Please sign off your email, and add some text describing your particular problem. For those "not sure if needed" comments you can prefix with `XXX`.

boot/zephyr/main.c
```diff
@@ -128,6 +128,14 @@ static void do_boot(struct boot_rsp *rsp)
     vt = (struct arm_vector_table *)(flash_base +
                                      rsp->br_image_off +
                                      rsp->br_hdr->ih_hdr_size);
+
+#ifdef CONFIG_CPU_CORTEX_M_SOMETHING
```
utzig, line 132: There's probably a way of testing if the features, instruct/data cache, MPU, etc exist in a target MCU.

arvid-r: Changed now!

arvid-r:
I removed the MPU part, which wasn't required, which I found out after some testing. Also, the code is now enabled only if we are running on Cortex M7.

boot/zephyr/main.c
```diff
@@ -128,6 +128,13 @@ static void do_boot(struct boot_rsp *rsp)
     vt = (struct arm_vector_table *)(flash_base +
                                      rsp->br_image_off +
                                      rsp->br_hdr->ih_hdr_size);
+
+#ifdef CONFIG_CPU_CORTEX_M7
+    /* Disable instruction cache and data cache before re-boot */
```
nvlsianpu, line 133:
```suggestion
    /* Disable instruction cache and data cache before chain-load the application */
```

nvlsianpu (commented):
looks OK to me. Although I think the patch might land right before call to the application's reset handler.
@ioannisg can you take a look at this.

boot/zephyr/main.c
ioannisg: "chain-loading the application".

Otherwise the suggestion by @nvlsianpu is good.

boot/zephyr/main.c
ioannisg: @utzig I guess the CMSIS needs to provide the information by enabling the __ICACHE_PRESENT and __DCACHE_PRESENT macros. In the absence of definitions, the calls to these CMSIS functions are suppressed.

ioannisg (approved):
@arvid-r Would be nice to test this PR with Zephyr CI, before merging that in.

boot/zephyr/main.c
arvid-r: Thanks! Will update.

arvid-r:
> looks OK to me. Although I think the patch might land right before call to the application's reset handler.
> @ioannisg can you take a look at this.

Hm... I don't understand what you mean here. Is this a problem? Should I change or investigate anything?

arvid-r:
> @arvid-r Would be nice to test this PR with Zephyr CI, before merging that in.

Yes! Is there a procedure for doing so? Or should I create a dummy PR over at Zephyr, and later drop it?

ioannisg:
> > @arvid-r Would be nice to test this PR with Zephyr CI, before merging that in.
> 
> Yes! Is there a procedure for doing so? Or should I create a dummy PR over at Zephyr, and later drop it?

You could create a PR in Zephyr that sets the manifest file pointer for mcuboot to point to this remote and this PR head.

boot/zephyr/main.c
arvid-r: Do you think I should add something like:
```#if defined(__ICACHE_PRESENT) && __ICACHE_PRESENT``` and ```#if defined(__DCACHE_PRESENT) && __DCACHE_PRESENT```?

arvid-r:
Here we go:
https://github.com/zephyrproject-rtos/zephyr/pull/28142
This is a bit beyond my GitHub skills though, so maybe this is not what you ment.

utzig:
@arvid-r Please rebase on latest `master`!

arvid-r:
Hm... CI fail over at Zephyr. I guess they expect a special patched MCUBOOT tree these days, or could it be something else?

utzig:
> Hm... CI fail over at Zephyr. I guess they expect a special patched MCUBOOT tree these days, or could it be something else?

Rebase

arvid-r:
I think I did

arvid-r:
Last commit after rebase, but before my change, is 009a150422543f94c826250942a990fb48d2a95e

utzig:
> Last commit after rebase, but before my change, is [009a150](https://github.com/JuulLabs-OSS/mcuboot/commit/009a150422543f94c826250942a990fb48d2a95e)

Ah, correct, but the last time I meant rebase the Zephyr PR, because buildkite does not run automatically when you rebase the MCUBoot PR, so you have to rebase here and there as well, sorry if it was not clear!

arvid-r:
Oh, I see. I manually forced it to re-run the tests, but maybe that doesn't pull the external files again. This is so complicated (in not complaining way) ;)

arvid-r:
Pipeline running again ...

utzig (approved):

nvlsianpu (approved):

utzig:
@arvid-r Thanks for the patch!

PR #792, opened 2020-09-02
zephyr: struct device const-ification

utzig:
Follows up the change added to Zephyr with: https://github.com/zephyrproject-rtos/zephyr/pull/24873


utzig:
West test: https://github.com/zephyrproject-rtos/zephyr/pull/27976

PR #795, opened 2020-09-10
boot: zephyr: Fix build issue for multiple conf files

nvlsianpu:
Change fixes build issue that occurs if CONF_FILE contains multiple
file paths instead of single file path.

Signed-off-by: Marek Pieta <Marek.Pieta@nordicsemi.no>

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -231,17 +231,30 @@ if(CONFIG_MCUBOOT_SERIAL)
     )
 endif()
 
-# CONF_FILE points to the KConfig configuration file of the bootloader
-get_filename_component(CONF_DIR ${CONF_FILE} DIRECTORY)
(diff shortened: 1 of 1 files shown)


nvlsianpu:
cc @MarekPieta

utzig:
This fixes my local issues building with `west`, but I wonder why the Zephyr buildkite runs have not failed, any idea?

utzig (approved):

utzig:
> This fixes my local issues building with `west`, but I wonder why the Zephyr buildkite runs have not failed, any idea?

At a glance it doesn't look like any of the Zephyr tests build MCUboot itself, they only use some of its provided API, so there's not that much point in submitting a build PR there.

nvlsianpu:
User discovered this on disco_l475_iot1 which is not used in zephyr CI.

MarekPieta (approved):

PR #796, opened 2020-09-10
Separate Zephyr sample build

d3zd3z:
Some changes to the test build and run process so that it is possible to build the tests and run them on separate machines (or separate containers). Without arguments, `run-tests.go` works as it did before, but now, the `tests-compile.go` can be used to build and assemble the results into a zip file, which can be transported to another system, and used with the new `-prebuilt` argument to `run-tests.go` to use the prebuilt tests rather than building them.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/JuulLabs-OSS/mcuboot
+
+go 1.14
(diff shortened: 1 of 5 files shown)


utzig (approved):
LGTM

PR #785, opened 2020-08-17
Replace ext/tinycrypt with submodule.

archseer:
`ext/tinycrypt` hasn't been updated in 3 years, and if I remember correctly, I needed https://github.com/intel/tinycrypt/commit/28b33ec081bd24bcaedd93e66ade3432d89ac270 to port mcuboot to a non-gcc compiler/platform.

PR #798, opened 2020-09-14
Abstract crypto primitives to simplify porting.

archseer:
Opening this up as a preview. As mentioned previously in #785, we've extracted the crypto primitives in a similar matter to `include/bootutil/sha256.h` (& `MCUBOOT_USE_CC310`) to simplify porting to native accelerated functions.

- `_init` and `_drop` functions were added that are simply compiled out in some circumstances (tinycrypt). We need these hooks to initialize and release the native accelerator.

Should these be `void` or do we want to handle the return value? `mbedtls` contexts use `void`, I imagine failing to initialize these primitives is unrecoverable so we could simply assert OK in our port.

There's a few other changes included:

- `hal_flash.h` is no longer imported. It was unused in the codebase, but porting to a new platform would fail during compilation due to the missing headers.
- `assert.h` was removed from various files and instead `bootutil_priv.h` will now either import `mcuboot_assert.h` or fall back to `assert.h`. We had issues during port because importing `assert.h` on the TI compiler would pull in the wrong implementation (debug mode CIO logging) and inflate the firmware size.


boot/bootutil/src/encrypted.c
```diff
@@ -766,25 +786,7 @@ boot_encrypt(struct enc_key_data *enc_state, int image_index,
 
     enc = &enc_state[rc];
     assert(enc->valid == 1);
-    for (i = 0; i < sz; i++) {
```
archseer, line 769: This block was simplified since it seemed to reimplement AES CTR mode by hand.

archseer:
Should be ready for review. \cc @utzig ?

[There is a single failure](https://travis-ci.org/github/JuulLabs-OSS/mcuboot/jobs/727275855) because `MBEDTLS_CIPHER_MODE_CTR` is not set when building for mynewt (similar to https://github.com/JuulLabs-OSS/mcuboot/pull/798/commits/8735ad373daa04ee0319f80d1e56c1e73be45e13), not sure where to modify that.

boot/bootutil/include/bootutil/crypto/aes_ctr.h
```diff
@@ -0,0 +1,150 @@
...
+    int rc;
+    rc = tc_aes128_set_encrypt_key(ctx, k);
+    if (rc != TC_CRYPTO_SUCCESS) {
+        return -1;
+    }
+    return 0;
+}
+
+static inline int _bootutil_aes_ctr_crypt(bootutil_aes_ctr_context *ctx, uint8_t *counter, const uint8_t *in, uint32_t inlen, uint32_t blk_off, uint8_t *out)
+{
```
d3zd3z, line 104: This is a pretty large function to be inline. With any type of cache, performance would probably be better just making this a function and calling it. Function call overhead is pretty light on Cortex-M, and avoiding code duplication can same noticeable space.

boot/bootutil/include/bootutil/crypto/aes_kw.h
```diff
@@ -0,0 +1,141 @@
...
+typedef mbedtls_nist_kw_context bootutil_aes_kw_context;
+static inline void bootutil_aes_kw_init(bootutil_aes_kw_context *ctx)
+{
+    (void)mbedtls_nist_kw_init(ctx);
+}
+
+static inline void bootutil_aes_kw_drop(bootutil_aes_kw_context *ctx)
+{
+    // config defines MBEDTLS_PLATFORM_NO_STD_FUNCTIONS so no need to free
+    // (void)mbedtls_nist_kw_free(ctx);
```
d3zd3z, line 46: Likewise with the free on this one.

boot/bootutil/include/bootutil/crypto/aes_kw.h
```diff
@@ -0,0 +1,141 @@
...
+        return -1;
+    }
+    return 0;
+}
+
+/*
+ * Implements AES key unwrapping following RFC-3394 section 2.2.2, using
+ * tinycrypt for AES-128 decryption.
+ */
+static inline int bootutil_aes_kw_unwrap(bootutil_aes_kw_context *ctx, const uint8_t *wrapped_key, uint32_t wrapped_key_len, uint8_t *key, uint32_t key_len)
```
d3zd3z, line 93: Again, probably best as a regular, non-inline, function.

boot/bootutil/src/encrypted.c
```diff
@@ -51,79 +43,49 @@
...
 {
-    mbedtls_nist_kw_context kw;
-    int rc;
-    size_t olen;
-
-    mbedtls_nist_kw_init(&kw);
+    return _compare(a, b, size);
+}
+#else
+static inline int bootutil_constant_time_compare(const uint8_t *a, const uint8_t *b, size_t size)
```
d3zd3z, line 53: Any reason this should be inline? Function call overhead is far from being a significant factor in execution time of mcuboot (nearly all time is spent in crypto primitives), and inline only serves to make the code larger.

d3zd3z (changes requested):
Overall, looks like a good change. I mostly think we just need to not use inline for things that are more than a few lines long.

boot/bootutil/include/bootutil/crypto/aes_ctr.h
```diff
@@ -0,0 +1,150 @@
...
+typedef mbedtls_aes_context bootutil_aes_ctr_context;
+static inline void bootutil_aes_ctr_init(bootutil_aes_ctr_context *ctx)
+{
+    (void)mbedtls_aes_init(ctx);
+}
+
+static inline void bootutil_aes_ctr_drop(bootutil_aes_ctr_context *ctx)
+{
+    // config defines MBEDTLS_PLATFORM_NO_STD_FUNCTIONS so no need to free
+    // (void)mbedtls_aes_free(ctx);
```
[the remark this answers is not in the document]
archseer: It fails to compile if `MBEDTLS_PLATFORM_NO_STD_FUNCTIONS` is set because it can't find the symbols. I think the mbedtls config should be standardized somewhere because we leak memory on platforms where it's not set ([cypress](https://github.com/JuulLabs-OSS/mcuboot/blob/4650f2011bd2f017c29a7722bf0b8d7a2693366b/boot/cypress/MCUBootApp/config/mcuboot_crypto_config.h#L199) for example)

boot/bootutil/src/encrypted.c
archseer: Agreed, I'll drop `inline` on all of these.

archseer:
I've dropped `inline` on the functions mentioned. Which config file do I need to modify to fix the mynewt test? It's missing `MBEDTLS_CIPHER_MODE_CTR`.

nvlsianpu:
cc @hakonfam @oyvindronningstad

oyvindronningstad:
This is definitely an improvement. I'm curious whether you considered standardizing on one of the existing APIs, such as mbedTLS, instead? That way, mbedTLS needs no glue layer, and all other crypto can create a glue layer for mbedTLS instead of the new crypto API.

archseer:
Yeah, in most cases the interface matches mbedtls, but we wrapped it again with inlinable functions just in case (so the wrappers will simplify away during compilation). In some places there's also no mbedtls equivalent (x25519, ed25519).

My first attempt at porting was actually to try and use the `MBEDTLS_*_ALT` shims and plug our code straight into mbedtls but it ended up being far more complicated (some paths also used tinycrypt exclusively)

utzig (commented):
I rebuilt all my targets using this and the results are good, aka smaller binaries, for all but the tinycrypt based encrypted images. I assume it is because you added `ctr_mode.c` which has to use work-arounds to properly support working as a stream cipher. Can be fixed later though. Please squash https://github.com/JuulLabs-OSS/mcuboot/pull/798/commits/821505e57512892f2cecd337409de53ab2489a62 into the proper commits.

d3zd3z:
@oyvindronningstad wrote

> This is definitely an improvement. I'm curious whether you considered standardizing on one of the existing APIs, such as mbedTLS, instead? That way, mbedTLS needs no glue layer, and all other crypto can create a glue layer for mbedTLS instead of the new crypto API.

I'm not sure that is the best idea, though, since it seems that mbed TLS may be changing their API. It may be useful to, at some point, settle on something like the PSA crypto API, but we might want to wait for mbed TLS to move to that first.

boot/bootutil/src/encrypted.c
archseer: Ah, that explains it. I was wondering about it but couldn't find any comments about it.

archseer:
I've gone ahead and made the changes.

utzig:
> I've gone ahead and made the changes.

Also because of the rebase you need to add your signed-off-by to the last commit.

archseer:
Pushed again.

boot/bootutil/include/bootutil/crypto/ecdh_x25519.h
```diff
@@ -0,0 +1,57 @@
...
+
+#if (defined(MCUBOOT_USE_TINYCRYPT)) != 1
+    #error "One crypto backend must be defined: TINYCRYPT"
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// TODO: also || MBDED_TLS
```
utzig, line 23: Also this comment should be `/* ... */`, and just to mention it, mbed TLS happens to have imported x25519 from the Everest project some time ago: https://github.com/ARMmbed/mbedtls/blob/development/3rdparty/everest/library/x25519.c. So this can be updated later, in another PR.

archseer: I realized the X25519 implementation is sourced from `fiat-crypto` so I've also added it for mbedtls and removed that comment.

utzig (approved):
LGTM, very nice refactoring!

utzig:
@d3zd3z I believe your comments were addressed, please approve/merge if no more pending issues remain.

d3zd3z (approved):

PR #799, opened 2020-09-14
[DNM] Add gpio based fallback as proof-of-concept

arvid-r:
Just for discussion. This shows a way to chain-load a fixed fallback image based on GPIO input on startup.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -49,6 +49,8 @@
 #include "bootutil/enc_key.h"
 #endif
 
+#include <drivers/gpio.h>
+
 #include "mcuboot_config/mcuboot_config.h"
 
 MCUBOOT_LOG_MODULE_DECLARE(mcuboot);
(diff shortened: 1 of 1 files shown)


arvid-r:
`loader.c` probably isn't a good place for doing the zephyr specific GPIO stuff and so on, but I mostly want to bring this type of feature for discussion

nvlsianpu:
> loader.c probably isn't a good place for doing the zephyr specific GPIO stuff and so on, but I mostly want to bring this type of feature for discussion

You can add an API which should be implemented by the OS port. For feature introduction the API implementation for zephyr is enough.

arvid-r:
> > loader.c probably isn't a good place for doing the zephyr specific GPIO stuff and so on, but I mostly want to bring this type of feature for discussion
> 
> You can add an API which should be implemented by the OS port. For feature introduction the API implementation for zephyr is enough.

I see. Where are these APIs typically declared?

arvid-r:
> IMO even the fallback image should be signed.

Sounds reasonable. And then the signature checked if `MCUBOOT_VALIDATE_PRIMARY_SLOT` is defined? Goes against the name, but aligns with the purpose I guess.

nvlsianpu:
API - We are using API function which have same name among ports, same number of parameters and close types. But it is not supper strict so each OS-port might use other mcuboot code compatible prototypes.
For instance: https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/zephyr/include/flash_map_backend/flash_map_backend.h

arvid-r:
Thanks. Makes sense. I guess something like `boot/zephyrinclude/gpio.h` would be a reasonable start then.

nvlsianpu:
> Thanks. Makes sense. I guess something like boot/zephyrinclude/gpio.h would be a reasonable start then.

Actually was thinking about something more dedicated - introduction of such low level API might be over-kill for one feature. So maybe `get_falback_gpio_state()`, `init_falback_gpio()` would be enough for now.  Mcuboot already uses gpio states bat this is implemented at the port level (and no sense to change that).

utzig:
> > Thanks. Makes sense. I guess something like boot/zephyrinclude/gpio.h would be a reasonable start then.
> 
> Actually was thinking about something more dedicated - introduction of such low level API might be over-kill for one feature. So maybe `get_falback_gpio_state()`, `init_falback_gpio()` would be enough for now. Mcuboot already uses gpio state if the serial recovery is enabled, but this is implemented at the OS port level (and no sense to change that).

What we could also think about is having some form of `setup`/`teardown` callbacks, like in testing frameworks; we could then add callback functions that could be configured as part of the build as needed, so for instance this particular option of reading a GPIO and making a decision would be one of those possible callbacks to do in a `setup` stage. If done correctly, it should allow having extra functions that are not for general use, but more oriented to specific use-cases, and still be part of the distribution. And it should probably live in a separate file or directory.

PR #801, opened 2020-09-15
zephyr: renamed single-image mode to single-slot mode

### LINKED ISSUE #800: Single-image boot is ambiguous
The terms "single-image boot" and the "single image application" of Zephyr are confused.
The original suggestion comes from @nvlsianpu (https://github.com/JuulLabs-OSS/mcuboot/pull/793#discussion_r487111576).
It is not clear whether the MCUBoot handles only one updateable image with the "classic" primary + secondary image slot model or there is only 1 image slot in total.
Would it make sense to have this single slot/image mode in case of multiple independent images?
The documentation and comments should be rephrased to clearly distinguish these 2 cases.

nvlsianpu:
The terms "single-image boot" and the "single image application"
of Zephyr are confused. It might be might be understood as one-pair image
dual-bank mode in comparison to multi-image mode.

This patch replaces CONFIG_SINGLE_IMAGE_DFU by CONFIG_SINGLE_SLOT_DFU
name, which makes this mode name unambiguous.

fixes #800

zephyr CI: https://github.com/zephyrproject-rtos/zephyr/pull/28396


nvlsianpu:
@de-nordic ?

de-nordic:
I do not mind changing the name. Unfortunately I had not gave enough though to the name prior to crafting that motley of words.
Generally it should mean "The one application slot configuration" so maybe it should be rather replaced with
`SINGLE_APPLICATION_SLOT`. I think that the `DFU` is misleading here because the code does not actually provide DFU functionality, it is provided by other components of the MCUBOOT; even the source file name indicates that this is only loader (`single_loader.c`).

nvlsianpu:
@de-nordic `SINGLE_APPLICATION_SLOT` sound good to me as well. And, right DFU abbreviation  has nothing to what MCUBoot does.

boot/zephyr/Kconfig
```diff
@@ -58,7 +58,7 @@ config NRF_CC310_BL
 
 menu "MCUBoot settings"
 
-config SINGLE_IMAGE_DFU
+config SINGLE_APPLICATION_SLOT
 	bool "Single image application"
```
de-nordic, line 62: Maybe change the title here to be also more in line with the function the config enables.

de-nordic (approved):
Looks fine, compiles and works as tested on nrf52840dk.

Approved with come comments left (https://github.com/JuulLabs-OSS/mcuboot/pull/801#discussion_r488593316)

de-nordic:
@davidvincze Do you want to look at this?

boot/zephyr/Kconfig
nvlsianpu: Is it better now?

de-nordic: .Kinda sounds like there is a feature where application may span multiple slots.

nvlsianpu: Actually this was same before.

nvlsianpu: `single slot bootloader`?

de-nordic: I think that this sounds better.

davidvincze (approved):
LGTM

PR #805, opened 2020-09-18
imgtool: Fix 'custom_tlvs' argument handling

davidvincze:
When Image.create() is called without the 'custom_tlvs' argument, it
gets its default value (None). It must be checked before performing
any operations on/with it.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -1,5 +1,5 @@
 # Copyright 2018 Nordic Semiconductor ASA
-# Copyright 2017 Linaro Limited
+# Copyright 2017-2020 Linaro Limited
 # Copyright 2019-2020 Arm Limited
 #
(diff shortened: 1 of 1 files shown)


scripts/imgtool/image.py
```diff
@@ -375,8 +375,9 @@ def create(self, key, public_key_format, enckey, dependencies=None,
                                     )
                     prot_tlv.add('DEPENDENCY', payload)
 
-            for tag, value in custom_tlvs.items():
-                prot_tlv.add(tag, value)
+            if custom_tlvs is not None:
```
utzig, line 378: It looks OK, since you're at it, also fix the other `custom_tlvs` access on line 330 to check against `None`.

davidvincze: Fixed.

utzig (approved):

nvlsianpu (approved):

PR #806, opened 2020-09-18
imgtool: Add possibility to set confirm flag for hex files as well

romanjoe:
Imgtool does not take into account value of `confirm` flag, when signing .hex images. Because of that user can't get signed hex file for swap upgrade type `permanent`.

This PR adds `confirm` flag analysis while signing hex file format.

Signed-off-by: Roman Okhrimenko <roman.okhrimenko@cypress.com>

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -219,9 +219,12 @@ class Image():
                                                   self.save_enctlv,
                                                   self.enctlv_len)
                 trailer_addr = (self.base_addr + self.slot_size) - trailer_size
-                padding = bytes([self.erased_val] *
-                                (trailer_size - len(boot_magic))) + boot_magic
-                h.puts(trailer_addr, padding)
+                padding = bytearray([self.erased_val] * 
+                                    (trailer_size - len(boot_magic)))
+                if self.confirm and not self.overwrite_only:
+                    padding[-MAX_ALIGN] = 0x01  # image_ok = 0x01
+                padding += boot_magic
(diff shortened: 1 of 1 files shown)


scripts/imgtool/image.py
```diff
@@ -222,6 +222,8 @@ def save(self, path, hex_addr=None):
                 padding = bytes([self.erased_val] *
                                 (trailer_size - len(boot_magic))) + boot_magic
                 h.puts(trailer_addr, padding)
+                if self.confirm and not self.overwrite_only:
+                    h[trailer_addr + trailer_size - len(boot_magic) - MAX_ALIGN] = 0x01  # image_ok = 0x01
```
utzig, line 226: What is the problem of adding `image_ok` to an `overwrite_only` image? This code assumes that `IntelHex` allows byte access, wouldn't it be better to just build `padding` correctly based on the `confirm` flag and use `puts()`?

romanjoe: I have changed implementation. This `if self.confirm and not self.overwrite_only:` is the same condition as in `def pad_to(self, size):` function, which process bin files.

utzig: > This if self.confirm and not self.overwrite_only: is the same condition as in def pad_to(self, size): function, which process bin files.

Yeah, there is nothing wrong with it, it's just weird because it is irrelevant if overwite-only mode is being used.

utzig (approved):
LGTM

PR #808, opened 2020-09-22
[DNM] boot: zephyr: remove flash erased val functions

utzig:
Functions that were used to get the erased value from the flash device driver were removed on this commit, because a correct implementation is being added to Zephyr which correctly queries the flash driver.

Depends on https://github.com/zephyrproject-rtos/zephyr/pull/28519


nvlsianpu (approved):

utzig:
Already fixed.

PR #809, opened 2020-09-24
boot: bootutil: drop flash_area_read_is_empty

utzig:
Removes the current `flash_area_read_is_empty` which lacked a bit of clarity in its naming and error handling, as well as requiring an extra API in the flash map, and switches to using an internal function `bootutil_buffer_is_erased`.
    
Code that was previously using `flash_area_read_is_empty` must now be updated to do a `flash_area_read` followed by a call to `bootutil_buffer_is_erased` with the read buffer.
    
The proposal was previously discussed here: https://github.com/zephyrproject-rtos/zephyr/pull/28519

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -291,6 +291,27 @@ boot_enc_key_off(const struct flash_area *fap, uint8_t slot)
 }
 #endif
 
+bool bootutil_buffer_is_erased(const struct flash_area *area,
+                               const void *buffer, size_t len)
+{
+    size_t i;
+    uint8_t *u8b;
+    uint8_t erased_val;
+
+    if (buffer == NULL || len == 0) {
+        return false;
+    }
+
+    erased_val = flash_area_erased_val(area);
+    for (i = 0, u8b = (uint8_t *)buffer; i < len; i++) {
+        if (u8b[i] != erased_val) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
 int
 boot_read_swap_state(const struct flash_area *fap,
                      struct boot_swap_state *state)
@@ -301,18 +322,18 @@ boot_read_swap_state(const struct flash_area *fap,
     int rc;
 
     off = boot_magic_off(fap);
-    rc = flash_area_read_is_empty(fap, off, magic, BOOT_MAGIC_SZ);
+    rc = flash_area_read(fap, off, magic, BOOT_MAGIC_SZ);
     if (rc < 0) {
         return BOOT_EFLASH;
     }
-    if (rc == 1) {
+    if (bootutil_buffer_is_erased(fap, magic, BOOT_MAGIC_SZ)) {
         state->magic = BOOT_MAGIC_UNSET;
     } else {
         state->magic = boot_magic_decode(magic);
     }
 
     off = boot_swap_info_off(fap);
-    rc = flash_area_read_is_empty(fap, off, &swap_info, sizeof swap_info);
+    rc = flash_area_read(fap, off, &swap_info, sizeof swap_info);
     if (rc < 0) {
         return BOOT_EFLASH;
     }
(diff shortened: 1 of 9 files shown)


utzig:
This is currently missing Zephyr and Cypress backend updates, I will do it in a bit...

docs/PORTING.md
```diff
@@ -119,11 +119,11 @@ int     flash_area_erase(const struct flash_area *, uint32_t off, uint32_t len);
...
 uint8_t flash_area_erased_val(const struct flash_area *);
-/*< Reads len bytes from off, and checks if the read data is erased. Returns
-    1 if empty (that is containing erased value), 0 if not-empty, and -1 on
-    failure. */
-int     flash_area_read_is_empty(const struct flash_area *fa, uint32_t off,
-                     void *dst, uint32_t len);
+/*< Verify that the data in buffer is erased according to the erase value of
+    the given flash area device. Returns true if the buffer is erased; false
+    if it is not erased, or buffer==NULL, or len==0. */
+bool    flash_area_buffer_is_erased(const struct flash_area *fa,
```
pabigot, line 125: Why does every back-end have to implement this?  Aren't they all the same?  Could it be put into common code somewhere?  `buffer_is_erased(data, size, erase_value)`?

utzig: Done, thanks for the hint!

boot/bootutil/src/bootutil_misc.c
```diff
@@ -248,6 +248,25 @@ boot_enc_key_off(const struct flash_area *fap, uint8_t slot)
...
+{
+    size_t i;
+    uint8_t *u8b;
+
+    if (buffer == NULL || len == 0) {
+        return false;
+    }
+
+    for (i = 0, u8b = (uint8_t *)buffer; i < len; i++) {
+        if (u8b[i] != flash_area_erased_val(area)) {
```
pabigot, line 262: Might be worth caching the return value of `flash_area_erased_val` on entry.  Some implementations might actually require a function in each compare.

pabigot (commented):
One potential optimization, otherwise looks good.

boot/bootutil/src/bootutil_misc.c
utzig: Done.

pabigot (approved):

docs/PORTING.md
d3zd3z: It isn't always the same. For example, the LPC55S69 has a command in the flash controller that is able to do this read check with different voltage thresholds to be able to check truly erased vs written as erased.

d3zd3z (changes requested):
I think we should be keeping `flash_area_read_is_empty`, and just remove it from here when it becomes available in Zephyr. There is at least one device that has a better implementation that is able to distinguish between truly erased flash, partially erased flash (say with a powerdown), and data that was written as the erased value.

docs/PORTING.md
utzig: > to be able to check truly erased vs written as erased.

As long as it is erased, "truly" or "written", shouldn't it be the same from our perspective?

pabigot:
> I think we should be keeping `flash_area_read_is_empty`, and just remove it from here when it becomes available in Zephyr.

I (for one) don't want it in Zephyr.  It's a complex way to verify that you can write new data into an area without the write being corrupted, as required to support updating a firmware image.  I think this approach is a significant improvement.

I wouldn't expect any erasing done in the context of normal mcuboot operations to support gradients of "erased".  That's something that should be done with an API that's designed to support special security needs.  If there's a use case for mcuboot to be able to securely erase an image, that should be a dedicated operation that can be implemented where the hardware supports it, and isn't tightly coupled with a read operation that's expected to provide useful data that may need to be updated.

Please reconsider your NAK.

de-nordic:
>
> It's a complex way to verify that you can write new data into an area without the write being corrupted, as required to support updating a firmware image. 
>
I think the intent was for the function to be used to find out whether the area can be written or should be erased first.
The only way to check whether write of has not been corrupted is to attempt reading data back and applying some method of verification against source.

>
> There is at least one device that has a better implementation that is able to distinguish between truly erased flash, partially erased flash (say with a powerdown), and data that was written as the erased value.

Shouldn't we then use this as a hint that we should fully erase the area again?

nvlsianpu:
> There is at least one device that has a better implementation that is able to distinguish between truly erased flash, partially erased flash (say with a powerdown), and data that was written as the erased value.

@d3zd3z  Is this leverage in mcuboot yet? Something like this would required in the flash driver API support.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -248,6 +248,27 @@ boot_enc_key_off(const struct flash_area *fap, uint8_t slot)
 }
 #endif
 
+bool bootutil_buffer_is_erased(const struct flash_area *area,
```
nvlsianpu, line 251: For me this name is not descriptive, maybe you can consider:
`bootutil_read_check_buff_is_erased()`

nvlsianpu (commented):
> There is at least one device that has a better implementation that is able to distinguish between truly erased flash, partially erased flash (say with a powerdown), and data that was written as the erased value.

Something like this would required support in the OS/port flash driver API.
Can not this be sported using conditional compilation in the future?

boot/bootutil/src/bootutil_misc.c
pabigot: Why?  It doesn't read anything.  It just checks the buffer against the erased value for the flash area specified.

pabigot: I do agree some documentation explaining what this is *intended* to do is important.

de-nordic (approved):

boot/bootutil/src/bootutil_misc.c
nvlsianpu: right

nvlsianpu (approved):

boot/bootutil/src/bootutil_misc.c
utzig: It could be `bootutil_check_buff_is_erased`, although if I had to write `buff` I would just use `buf` instead. :-P

I like the current naming, it feels quite clear already, and I added a comment header explaining what it does and returns.

nvlsianpu: It's ok now

utzig:
> > There is at least one device that has a better implementation that is able to distinguish between truly erased flash, partially erased flash (say with a powerdown), and data that was written as the erased value.
> 
> Something like this would required support in the OS/port flash driver API.
> Can not this be sported using conditional compilation in the future?

I thought the same yesterday, but then I completely fail to see how can erased have multiple meanings, or what difference does it for us, since we check writes the we did. If it could somehow be abused by a third party for security reasons, so it can today by just writing garbage. Also if 100% of the backends do the same, on has to ask what the purpose of the abstraction really is. I think we can update/improve this when the time arrives to actually see something that doesn't match the current implementation.

d3zd3z:
> I think the intent was for the function to be used to find out whether the area can be written or should be erased first.
The only way to check whether write of has not been corrupted is to attempt reading data back and applying some method of verification against source.

Thinking about this some more, I now feel that this API isn't really that useful. There is only one device I'm aware of that would have a meaningful implementation, and using it is dubious anyway. If an application is unsure of the erase status of a block of data, it should proceed by erasing that area, not trying to figure out some gradation of erasedness.

d3zd3z (approved):

utzig:
@romanjoe The cypress port still has a `flash_area_read_is_empty` implementation, I suggest you drop it now.

PR #810, opened 2020-09-24
Create github workflow files for CI

d3zd3z:
These duplicate the existing travis builds using the github workflow. Since most of the work is done by support scripts, this is mostly about running these scripts.

### DIFF
--- a/.github/workflows/mynewt.yaml
+++ b/.github/workflows/mynewt.yaml
@@ -1,0 +1,28 @@
+# For development, trigger this on any push.
+on: [push, pull_request]
+
+# Eventually, we'll want to limit this to pull requests.
+# on: ???
+
+jobs:
+  environment:
+    name: Mynewt build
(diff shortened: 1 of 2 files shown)


utzig:
Can you throw `.travis.yml` away?

d3zd3z:
> Can you throw .travis.yml away?

definitely, but I was wondering if we wanted to keep both active to make sure we get a chance to make sure that the actions are running correctly.

utzig:
> > Can you throw .travis.yml away?
> 
> definitely, but I was wondering if we wanted to keep both active to make sure we get a chance to make sure that the actions are running correctly.

we can compare the performance as well, ok, let's leave it!

utzig (approved):
LGTM

utzig:
I will just go ahead and merge, because workflows only run once they exist in a branch, so I can continue some testing...

PR #812, opened 2020-09-25
workflows: run on PR only

utzig:
Remove option to run on push because it is triggering two runs for each test for all PRs.


utzig:
@d3zd3z Can you take a look?

d3zd3z (approved):

PR #813, opened 2020-09-28
zephyr: Improve signature kconfig

hakonfam:

PR #823, opened 2020-10-01
zephyr/Kconfig: not use cc310 by default

### LINKED ISSUE #821: `No such file or directory: NRFXLIB_DIR` error while building for nRF SoC with cc310 HW
raveled by #813 as now `CONFIG_BOOT_USE_NRF_CC310_BL=y` by default.

https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/zephyr/CMakeLists.txt#L64-L69
this cause cmake build stage error:

/home/anpu/codedev/zephyrproject/zephyr/cmake/extensions.cmake:1579 (message):
  No such file or directory: NRFXLIB_DIR:
  '/home/anpu/codedev/zephyrproject/zephyr/../nrfxlib'
Call Stack (most recent call first):
  CMakeLists.txt:66 (assert_exists)
/ext/nrf/README.md contains description, but this is buried

nvlsianpu:
Usage of cc310 by defaults gives poor user
experience out of the box as it requires manual
cloning of nrfxlib repository.

fixes #821


nvlsianpu:
@hakonfam @sigvartmh @MarekPieta please review as #813 revealed that actually having cc310 enabled by default is problematic while using outside NCS.

boot/zephyr/CMakeLists.txt
```diff
@@ -63,7 +63,15 @@ set(NRF_DIR "${MCUBOOT_DIR}/ext/nrf")
 
 if(CONFIG_BOOT_USE_NRF_CC310_BL)
 set(NRFXLIB_DIR ${ZEPHYR_BASE}/../nrfxlib)
-assert_exists(NRFXLIB_DIR)
+if(NOT EXISTS ${NRFXLIB_DIR})
+  message(FATAL_ERROR "
+    ------------------------------------------------------------------------
+    No such file or directory: ${NRFXLIB_DIR}
+    Current configuration enables nRF CC310 crypto accelerator hardware
+    by `BOOT_USE_NRF_CC310_BL` option. Please follow `ext/nrf/README.md` guide
```
utzig, line 71: "**The** current configuration enables ..."; "... accelerator hardware **with** ..."; and I would use `CONFIG_BOOT_USE_NRF_CC310_BL` instead

boot/zephyr/CMakeLists.txt
```diff
@@ -63,7 +63,15 @@ set(NRF_DIR "${MCUBOOT_DIR}/ext/nrf")
...
+    to fix your setup.
```
utzig, line 72: "... fix your setup, or disable the HW accelerator with `-DCONFIG_BOOT_ECDSA_TINYCRYPT=y`."

boot/zephyr/Kconfig
```diff
@@ -68,7 +68,6 @@ config SINGLE_APPLICATION_SLOT
 
 choice
 	prompt "Signature type"
-	default BOOT_SIGNATURE_TYPE_ECDSA_P256 if HAS_HW_NRF_CC310
```
utzig, line 71: I don't think you should disable this default config, since I would assume most users use Zephyr with west (or should!). As long as the message states clearly how to fix the problem, it should be fine for everyone.

boot/zephyr/CMakeLists.txt
MarekPieta: Maybe we could even write:

fix your setup or use tinycrypt instead of the HW accelerator. To use the tinycrypt set `CONFIG_BOOT_ECDSA_TINYCRYPT` to `y`.

boot/zephyr/Kconfig
MarekPieta: Maybe we could mention in the readme that `you can also use west` (in case user is not aware about this)?

utzig: I guess I was too fast at commenting before, but it doesn't seem that Zephyr's west brings in `nrfxlib`, so you can disregard my previous comment.

MarekPieta (approved):

utzig (approved):

PR #825, opened 2020-10-01
Add ifdefs in crypto abstract function for ec256 to enable compilation when mbedtls used as backend

romanjoe:
Cypress use `MCUBOOT_USE_MBED_TLS` backend for crypto operation and `MCUBOOT_SIGN_EC256 ` algo. After recent change https://github.com/JuulLabs-OSS/mcuboot/pull/798 where abstraction of crypto functions was introduced, this combination is no longer work and `cypress` code can not be compiled.

This fix puts back ifdef `MCUBOOT_USE_TINYCRYPT` or `MCUBOOT_USE_CC310` in `image_ec256p.c` to exclude new implementation from build if `MCUBOOT_USE_MBED_TLS` and `MCUBOOT_SIGN_EC256` are defined in mcuboot_config.h until `cypress` port mbedtls backend to new abstraction crypto.
 
Signed-off-by: Roman Okhrimenko <roman.okhrimenko@cypress.com>

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/ecdsa_p256.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdsa_p256.h
@@ -13,7 +13,8 @@
 #include "mcuboot_config/mcuboot_config.h"
 
 #if (defined(MCUBOOT_USE_TINYCRYPT) + \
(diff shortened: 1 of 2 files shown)


utzig (approved):
LGTM, thanks!

Issue #738, opened 2020-05-14
512 byte flash alignment test


M1cha:
with the feature `large-write` enabled, sim creates tests for devices with an alignment of 512 bytes. That ends up being a 0-byte alignment because `flash_area_align` returns `uint8_t`.

utzig:
This should have been fixed by #826, thanks!

PR #826, opened 2020-10-02
sim: add bootstrap test

### LINKED ISSUE #706: Bootstrap does not work when CONFIG_BOOT_SWAP_USING_MOVE=y
I am trying to bootstrap MCUBoot using Slot1 but it does not seem to work when CONFIG_BOOT_SWAP_USING_MOVE=y. 

It seems like it fails on checking the magic for Slot 0. This is expected as there is nothing in Slot0 but I do not understand why it is checked.

Is this a bug in MCUBoot or have I misunderstood how this is supposed to work?
It works fine when CONFIG_BOOT_SWAP_USING_MOVE=n

RTOS: Zephyr
Target board: particle_boron
mcuboot conf:
```
CONFIG_BOOT_BOOTSTRAP=y
CONFIG_BOOT_SWAP_USING_MOVE=y
```

utzig:
This should fix bootstrap in swap-move along with other fixes.

@d3zd3z After some issues I've found working on this PR, I think we should just remove the `MCUBOOT_OVERWRITE_ONLY_FAST` option and make it the default behavior when doing overwrite-only, what do you think?

Fixes #706, #738

### DIFF
--- a/.github/workflows/sim.yaml
+++ b/.github/workflows/sim.yaml
@@ -14,7 +14,7 @@ jobs:
         - "sig-rsa overwrite-only large-write,sig-ecdsa overwrite-only large-write,multiimage overwrite-only large-write"
         - "sig-rsa validate-primary-slot,sig-ecdsa validate-primary-slot,sig-rsa multiimage validate-primary-slot"
         - "enc-kw overwrite-only large-write,enc-rsa overwrite-only large-write"
-        - "sig-rsa enc-rsa validate-primary-slot,swap-move enc-rsa sig-rsa validate-primary-slot"
+        - "sig-rsa enc-rsa validate-primary-slot,swap-move enc-rsa sig-rsa validate-primary-slot bootstrap"
         - "sig-rsa enc-kw validate-primary-slot bootstrap,sig-ed25519 enc-x25519 validate-primary-slot"
         - "sig-ecdsa enc-kw validate-primary-slot"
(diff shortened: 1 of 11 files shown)


d3zd3z (approved):

erik-johnson-nimbelink:
I don't think commit 738c8b7 is quite right. If I'm not mistaken, "BOOT_CURR_IMG(state)" doesn't mean primary/secondary, but rather which image is being checked. So, if trying to bootstrap an image that isn't the first one -- which I happen to be currently doing -- this will fail. Removing that check lets me bootstrap my second image.

Wouldn't it be sufficient to only have the erased header check there?

utzig:
> Wouldn't it be sufficient to only have the erased header check there?

Yes I agree, it was a mess-up on my part, I think it's safe to remove the check you mentioned. Have you tried bootstrapping a multi-image test in the simulator?

erik-johnson-nimbelink:
I haven't used the simulator, and we're behind quite a bit from master since we're close to release. But, for what it's worth, I did try a series of real-world tests, and removing the check did allow me to bootstrap the primary slot in my second image.

utzig:
> I haven't used the simulator, and we're behind quite a bit from master since we're close to release. But, for what it's worth, I did try a series of real-world tests, and removing the check did allow me to bootstrap the primary slot in my second image.

Ok, what MCU did you use for testing? Please send a PR.

PR #824, opened 2020-10-01
boot: zephyr: reset SPLIM registers before boot

hakonfam:
to avoid causing stack overflow in the image being booted.

Signed-off-by: Håkon Øye Amundsen <haakon.amundsen@nordicsemi.no>

PR #828, opened 2020-10-05
boot: zephyr: Disable HW stack protection

mniestroj:
Some Zephyr boards enable HW stack protection by default, which
propagates in mcuboot build. Bootloader works fine with that option, but
the problem arises when we jump to application code. HW stack protection
catches application code during early init, which results in
"unresponsive" application.

Disable HW stack protection, so it doesn't make any harm for application
code.

Signed-off-by: Marcin Niestroj <m.niestroj@grinn-global.com>

### DIFF
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -33,3 +33,5 @@ CONFIG_FLASH=y
 CONFIG_LOG=y
 ### Ensure Zephyr logging changes don't use more resources
 CONFIG_LOG_DEFAULT_LEVEL=0
+
+CONFIG_HW_STACK_PROTECTION=n


utzig:
Please rebase over latest master.

nvlsianpu:
@mniestroj Did you checked code with this patch? https://github.com/JuulLabs-OSS/mcuboot/pull/824/files

nvlsianpu:
The patch #824 can work only for cortex M33. M4 doesn't have SPLIM register.

mniestroj:
Yes, I checked. #824 doesn't help with nRF52840.

nvlsianpu:
@ioannisg Need you help for implement stack guard disabling for case when SPLIM register are not in usage.

ioannisg:
> @ioannisg Need you help for implement stack guard disabling for case when SPLIM register are not in usage.
> Something similar in implementation to #824

Sure, I can provide that.

nvlsianpu:
@mniestroj We planed to implement  `disable HW stack protection` in the runtime soon.
This patch will be merged as temporary hack.

nvlsianpu (approved):

mniestroj:
>We planed to implement disable HW stack protection in the runtime soon.
This patch will be merged as temporary hack.

Hack for "feature not implemented"? :) I would rather call it a (temporary) solution.

PR #830, opened 2020-10-06
bootutil: fix swap-move brick with padded image in primary slot

### LINKED ISSUE #829: Brick when using BOOT_SWAP_USING_MOVE and reset happens during images swap
**Describe the bug**
Interrupted swap of images when **swap_move** strategy is selected, can end up in "bricked" state. Possible workaround is to use **swap_scratch** strategy, which doesn't seem to have this specific issue.

**To Reproduce**
See https://github.com/zephyrproject-rtos/zephyr/issues/28933 for reproducing using Zephyr RTOS.

**Expected behavior**
Swap of images should continue.

**Impact**
Remote firmware upgrade is not safe, i.e. can result in bricked devices in the field.

**Environment (please complete the following information):**
 - MCU software: Zephyr RTOS
 - OS: Linux
 - Toolchain: Zephyr SDK 0.11.4
 - aac99a14c9cdc8684fd4fbced130d3065a75ffea (tip of master) + https://github.com/JuulLabs-OSS/mcuboot/pull/828

**Additional context**
It seems that problems start here (https://github.com/JuulLabs-OSS/mcuboot/blob/aac99a14c9cdc8684fd4fbced130d3065a75ffea/boot/bootutil/src/loader.c#L1084):
```
        /*
         * If a swap was under way, the swap_size should already be present
         * in the trailer...
         */
        rc = boot_read_swap_size(image_index, &bs->swap_size);
        assert(rc == 0);
```
mcuboot reads `0xFFFFFFFF` after swap was interrupted, because this is what was contained originally in `app.hex` (signed and confirmed) image trailer and that was never updated by mcuboot. This value is passed as `copy_size` to `swap_run()` and results in infinite loop there (https://github.com/JuulLabs-OSS/mcuboot/blob/aac99a14c9cdc8684fd4fbced130d3065a75ffea/boot/bootutil/src/swap_move.c#L448):
```
    while (1) {
        sz += sector_sz;
        /* Skip to next sector because all sectors will be moved up. */
        g_last_idx++;
        if (sz >= copy_size) {
            break;
        }
    }
```

utzig:
When the image in the primary slot is padded, the boot source is considered the primary slot; this results in skipping the typical initialization of the trailer, which ends up bricking the device. As it is fine to always initialize the trailer in the primary slot when starting a new upgrade the extra check was removed.

Should fix #829 when done.

### DIFF
--- a/boot/bootutil/src/swap_move.c
+++ b/boot/bootutil/src/swap_move.c
@@ -315,13 +315,11 @@ boot_move_sector_up(int idx, uint32_t sz, struct boot_loader_state *state,
     old_off = boot_img_sector_off(state, BOOT_PRIMARY_SLOT, idx - 1);
 
     if (bs->idx == BOOT_STATUS_IDX_0) {
-        if (bs->source != BOOT_STATUS_SOURCE_PRIMARY_SLOT) {
-            rc = swap_erase_trailer_sectors(state, fap_pri);
-            assert(rc == 0);
+        rc = swap_erase_trailer_sectors(state, fap_pri);
+        assert(rc == 0);
 
-            rc = swap_status_init(state, fap_pri, bs);
-            assert(rc == 0);
-        }
+        rc = swap_status_init(state, fap_pri, bs);
+        assert(rc == 0);
 
         rc = swap_erase_trailer_sectors(state, fap_sec);
         assert(rc == 0);


utzig:
@d3zd3z @nvlsianpu This should fix the bricking for which issues exists here and on Zephyr. If the CI passes, I will drop the simulator commit padding the primary image, because I don't think it is necessary in the long run. I don't have a very good recollection of why I added this `if` in the first place but I somewhat recall it being required before I wrote `fixup_revert` (https://github.com/JuulLabs-OSS/mcuboot/blob/master/boot/bootutil/src/swap_move.c#L393).

@mniestroj Let me know if you can confirm that this PR fixes your brick issue!

mniestroj (approved):
I confirm this fixes #829. LGTM.

mniestroj (approved):

nvlsianpu (approved):

PR #836, opened 2020-10-07
Temporarily disable travis test jobs

matetothpal:
This PR disables (almost) all of the travis test jobs. 

Please note that the 'imgtool' test is not disabled, as in case of an empty matrix the CI seem to report failure.


utzig:
`imgtool` does the publishing to pypi.org, it was not migrated to workflows yet, so it's better to leave it here.

matetothpal:
Ah, OK. Then the chang should be Ok.

utzig (approved):

PR #838, opened 2020-10-09
boot/zephyr: add nrfx watchdog kick during boot region copy

nvlsianpu:
copy of https://github.com/zephyrproject-rtos/mcuboot/pull/34

This fixes soft-bricks that we have seen as a result
of the bootloader being interrupted by the watchdog.

Signed-off-by: Emil Hammarstrom <emil.hammarstrom@assaabloy.com>

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -495,6 +495,16 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION
 
 endchoice
 
+config BOOT_WATCHDOG_FEED
+	bool "Feed the watchdog while doing swap"
+	default y if SOC_FAMILY_NRF
+	imply NRFX_WDT
+	imply NRFX_WDT0
+	imply NRFX_WDT1
+	help
+	  Enables implementation of MCUBOOT_WATCHDOG_FEED() macro which is
+	  used to feed watchdog while doing time consuming operations.
+
 endmenu
 
 config MCUBOOT_DEVICE_SETTINGS
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -154,6 +154,7 @@
 
 #endif /* !__BOOTSIM__ */
 
+#if CONFIG_BOOT_WATCHDOG_FEED
 #if CONFIG_NRFX_WDT
 #include <nrfx_wdt.h>
 
@@ -180,12 +181,18 @@
 #endif /* defined(CONFIG_NRFX_WDT0) && defined(CONFIG_NRFX_WDT1) */
 
 #else /* CONFIG_NRFX_WDT */
-
+#warning "MCUBOOT_WATCHDOG_FEED() is no-op"
+/* No vendor implementation, no-op for historical reasons */
 #define MCUBOOT_WATCHDOG_FEED()         \
     do {                                \
-        /* TODO: to be implemented */   \
+    } while (0)
+#endif /* CONFIG_NRFX_WDT */
+#else  /* CONFIG_BOOT_WATCHDOG_FEED */
+/* Not enabled, no feed activity */
+#define MCUBOOT_WATCHDOG_FEED()         \
+    do {                                \
     } while (0)
 
-#endif /* CONFIG_NRFX_WDT */
+#endif /* CONFIG_BOOT_WATCHDOG_FEED */
 
 #endif /* __MCUBOOT_CONFIG_H__ */
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -154,9 +154,38 @@
 
 #endif /* !__BOOTSIM__ */
 
+#if CONFIG_NRFX_WDT
+#include <nrfx_wdt.h>
+
+#define FEED_WDT_INST(id)                                    \
+    do {                                                     \
+        nrfx_wdt_t wdt_inst_##id = NRFX_WDT_INSTANCE(id);    \
+        for (uint8_t i = 0; i < NRF_WDT_CHANNEL_NUMBER; i++) \
+        {                                                    \
+            nrf_wdt_reload_request_set(wdt_inst_##id.p_reg,  \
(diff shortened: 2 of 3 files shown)


utzig (approved):

pabigot (commented):
I don't think this is an ideal nor a robust fix.

For Nordic hardware the watchdog would only be running when MCUBOOT starts in a situation where something caused a non-watchdog reset while the watchdog was active, since most reset causes don't clear the watchdog configuration.

In this situation you have *no idea* when the watchdog will actually fire, because you don't know when it was last fed: it might reset at any instant.  The only robust solution is to detect a running watchdog as soon as MCUBOOT gains control, and spin in place doing no harm until the watchdog fires.  Then you'll return to mcuboot in a clean environment in which you can do what you need to do regardless of how the application (or mcuboot) configures the watchdog.

If this isn't done, then you have no guarantee, even if the watchdog feed code is implemented, that the watchdog still won't fire before the first time it's fed.

Also: from what I can tell the sole invocation of `MCUBOOT_WATCHDOG_FEED` is at the bottom of the loop in `boot_copy_region()` **after** data has already been written.  To do the best you can to satisfy the needs of an already running watchdog it should be done at the **top** of the loop---as well as at the mcuboot entrypoint function, and before/after trying to erase pages, and any other place that might take more than some documented duration to reach.

Even then if somebody uses a very short-duration timeout it still might fail.

eHammarstrom:
@pabigot The spin behavior would not be of interest to us. It would delay the fw upgrade process for too long. But it might be of interest to others.

The other two wdt feeds you mention are good points and should be raised in a separate issue as the wdt feed macro is used by others already.

Note that this commit message is now misleading should have been updated in the previous PR (and now both @nvlsianpu). The soft bricks are related to mcuboot 1.3 and the message carried over from a migration, one of the bricked images have been reviewed by @utzig.

I think this is a good addition for the NRF chips as we do not want the wdt reset reason to be watchdog after a successful fw upgrade. If we are in the midst of swapping the reset is unnecessary, even if MCUBoot continues where it left off.

boot/zephyr/Kconfig
```diff
@@ -495,6 +495,10 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION
 
 endchoice
 
+config BOOT_WATCHDOG_FEED
+	bool "Feed the watchdog while doing swap"
+	depends on NRFX_WDT
```
pabigot, line 500: Why should this depend on a specific vendor implementation of the watchdog?

If this is needed it should probably default to `y` to match current behavior.

nvlsianpu: yes, and no

Indeed implementation allows what you suggested but for other platforms the feed is not implemented.
I think better to not suggest to user that something is implemented while is not.

pabigot: It's not implemented upstream.  You have no idea what patches are applied downstream.

pabigot: A good reason to have this option is in a case where it is implemented, but is not desired.  For that it should not have a dependency on a vendor, and should default to enabled.

pabigot: Alternatively, leave it defaulted to disabled and change the code so that if it's selected and there isn't a vendor-specific definition of `MCUBOOT_WATCHDOG_FEED` it hits an `#error` or other build-time assert.  Only when it's not selected should the existing no-op empty definition be used.

Is that more reasonable?

nvlsianpu: I would set this to `y` as you initially suggested.

pabigot: Yes, that's the safe approach to match existing behavior.  Sorry, I'm jumping between too many things and not paying enough attention to subtleties.

nvlsianpu:
This patch enables implementation of MCUBOOT_WATCHDOG_FEED() macro which is used to feed watchdog while doing time consuming operation.
User case is that that mcuboot started while watchdog have been running (started by the application).
Wisely designed application should feed the watchdog just before it reset the SoC.

I think mcuboot should feed watchdog right after startup additionally.

This patch is to simple to provide solution for lot of possible watchdog hw implementation (STM32 window WDT for instance).
It just enables usage what is already implemented

> Even then if somebody uses a very short-duration timeout it still might fail.

That a fact, for such a case need to identify timing consume routines and put `feed` where needed. I would add such call to `MCUBOOT_WATCHDOG_FEED()` once needed, once we encounter such a need.

> In this situation you have no idea when the watchdog will actually fire, because you don't know when it was last fed: it might reset at any instant. The only robust solution is to detect a running watchdog as soon as MCUBOOT gains control, and spin in place doing no harm until the watchdog fires. Then you'll return to mcuboot in a clean environment in which you can do what you need to do regardless of how the application (or mcuboot) configures the watchdog.

I also have been considered re-entering to the mcuboot via watchdog reset - but this looks like as time consuming and definitely as another feature to me.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -154,9 +154,38 @@
 
 #endif /* !__BOOTSIM__ */
 
+#if CONFIG_NRFX_WDT && CONFIG_BOOT_WATCHDOG_FEED
```
pabigot, line 157: This seems imbalanced.  There really are two conditions where we have to default: enabled without a vendor implementation, and disabled.  I think the high-level structure should be:
```
#if CONFIG_BOOT_WATCHDOG_FEED
#if VENDOR
...
#elif OTHER_VENDOR
...
#else /* VENDORS */
/* No vendor implementation, no-op for historical reasons but should be a configuration error */
#define MCUBOOT_WATCHDOG_FEED() do { } while (0)
#endif /* VENDORS */
#else /* CONFIG_BOOT_WATCHDOG_FEED */
 /* Not enabled, no feed activity */
#define MCUBOOT_WATCHDOG_FEED() do { } while (0)
#endif /* CONFIG_BOOT_WATCHDOG_FEED */
```
because as proposed the entry condition doesn't pair correctly with the top-level else, and the preprocessor conditional nesting is more complex than it needs to be.

pabigot (approved):
This looks good now, thanks.  Somebody does need to add the feed at the point mcuboot starts, if it's going to be robust, but that can be done separately.

PR #840, opened 2020-10-19
boot/boot_serial: allow unaligned last chunk

nvlsianpu:
- 1.
The last data packet might be non aligned to multiple of
the flash write-block-size. This cause that the `image upload`
mcumgr command mighty fail if the device flash driver doesn't
support one-byte write-block-size.

This patch complements the last write operation to aligned chunk
so it meet the requirements.

- 2.
.Added check for ensure than received data chunk doesn't
cross expected image size..

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -51,6 +51,7 @@
 #include <os/os_malloc.h>
 
 #include <bootutil/image.h>
+#include <bootutil/bootutil.h>
 
 #include "boot_serial/boot_serial.h"
 #include "boot_serial_priv.h"
@@ -312,11 +313,17 @@ bs_upload(char *buf, int len)
         rc = 0;
         goto out;
     }
-    if (curr_off + img_blen < img_size) {
-        rem_bytes = img_blen % flash_area_align(fap);
-        if (rem_bytes) {
-            img_blen -= rem_bytes;
-        }
+
+    if (curr_off + img_blen > img_size) {
+        rc = MGMT_ERR_EINVAL;
+        goto out;
+    }
+
+    rem_bytes = img_blen % flash_area_align(fap);
+
+    if ((curr_off + img_blen < img_size) && rem_bytes) {
+        img_blen -= rem_bytes;
+        rem_bytes = 0;
     }
 
 #ifdef CONFIG_BOOT_ERASE_PROGRESSIVELY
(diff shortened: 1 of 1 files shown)


nvlsianpu:
cc @hakonfam @de-nordic

boot/boot_serial/src/boot_serial.c
```diff
@@ -312,10 +313,13 @@ bs_upload(char *buf, int len)
...
         goto out;
     }
+
+    rem_bytes = img_blen % flash_area_align(fap);
+
     if (curr_off + img_blen < img_size) {
-        rem_bytes = img_blen % flash_area_align(fap);
         if (rem_bytes) {
             img_blen -= rem_bytes;
+            rem_bytes = 0;
```
de-nordic, line 322: The `rem_bytes` are needed only when `curr_off + img_blen >= img_size`, actually `==`, so this `if` could be reversed to just get `rem_bytes` only when the last part of image is written. Right?

boot/boot_serial/src/boot_serial.c
```diff
@@ -337,7 +341,33 @@ bs_upload(char *buf, int len)
...
+
+        if (img_blen) {
+            memcpy(wbs_aligned, img_data, rem_bytes);
+            memset(wbs_aligned + rem_bytes, flash_area_erased_val(fap),
+                   sizeof(wbs_aligned) - rem_bytes);
+            rc = flash_area_write(fap, curr_off, wbs_aligned, flash_area_align(fap));
+        }
+
+    }
+    else {
```
de-nordic, line 367: I do not remember what is the style for `mcuboot`, so dumb question: shouldn't these be in one line?

boot/boot_serial/src/boot_serial.c
nvlsianpu: I think, that to be precisely all transfer should be aligned, but the last one must not.

nvlsianpu: @de-nordic 
Introduced a check for the size-alignment of the rest of image blocks as
All of their sizes have to be aligned to
multiple of the flash write-block-size. This should be ensured
as well, so fare code has been cutting blocks to meet the requirement.

boot/boot_serial/src/boot_serial.c
```diff
@@ -312,10 +313,14 @@ bs_upload(char *buf, int len)
         rc = 0;
         goto out;
     }
-    if (curr_off + img_blen < img_size) {
-        rem_bytes = img_blen % flash_area_align(fap);
+
+    rem_bytes = img_blen % flash_area_align(fap);
+
+    if (((curr_off + img_blen < img_size) && rem_bytes) ||
+        (curr_off + img_blen > img_size)) {
```
de-nordic, line 320: It this gets true, but `rem_bytes == 0`, then it still should be an error, right?

de-nordic: Lets assume:
```
img_blen = 16
curr_off = 4
img_size = 16
flash_area_align(fap) = 4
rem_bytes = 16 % 4 => rem_bytes = 0;
```
now
```
cur_off + img_len > img_size => 4 + 16 > 16 is true
```
but `rem_bytes == 0` which means that
```
if (rem_bytes) {
        img_blen -= rem_bytes;
        rc = MGMT_ERR_EINVAL;
        goto out;
}
```
will not happen because `rem_bytes == 0`

nvlsianpu: Indeed ,I forgoten about the line below - It has been tested on another branch where it was removed.

de-nordic (approved):

d3zd3z (approved):

Issue #713, opened 2020-04-10
Multiple SoCs have flash configurations unsupported by MCUboot


d3zd3z:
Some newer devices have flash configurations that are not supported currently by MCUboot. This issue attempts to collect these in one place to help with the design of any solutions to handle this situation.

Known devices:
- LPC55S69, 512-byte writes, 512-byte erase
- PSOC6x, 512-byte writes, 512-byte erase
- SAM-X70, 16-byte writes, 128k-byte erase

jameswalmsley:
STM32H7 (43) 32 bytes write, 128K erase

weinholtendian:
A further complication with STM32H7 is that the internal flash has an integrated ECC. Writing to the same flash word twice has a high probability of causing an ECC error, and if it's a double ECC error then a read results in a bus error. Does MCUboot rely on being able to write to the same flash word more than once?

utzig:
> Does MCUboot rely on being able to write to the same flash word more than once?

No, it only writes once to any "word΅ which is the alignment size for the flash supplied by the OS.

weinholtendian:
I have MCUboot working on a PoC level on STM32H7. This required changing both BOOT_MAX_ALIGN and BOOT_MAGIC_SZ to 32. The boot_img_magic array had to be changed as well (one copy in bootutil_misc.c, one in image.py, one in Zephyr's mcuboot.c, maybe more) because it's only 16 bytes which is less than the alignment. All in all it's simple changes, but just changing BOOT_MAX_ALIGN for everyone isn't backwards compatible.

Is there any good reason to not make BOOT_MAX_ALIGN configurable? A user could set it to e.g. 32, which would then cause the magic to be padded to 32 bytes. (Or 16, or 512?). Everyone else would keep it at 8 and keep compatibility.

jameswalmsley:
I'd like to confirm also that I have been running MCUBoot on STM32H7 for
12 months now.

We made the same changes as you described above.

We did find its possible to get an ECC error if we lost power during an
image swap, and so MCUBoot would cause a hard-fault during swap-resume.

We solved this by:

  1. Using a watchdog if an ECC fault is triggered.
  2. If boot-reason is due to a watchdog in the bootloader then we
     erase the image, and scratch areas.
  3. We provide a recovery mode, where we wait for a repair image over
     DFU.
  4. Just before we boot the application we set a flag to say the
     firmware was booted. If the firmware triggers WD then we don't
     cause recovery.
[...]

PR #841, opened 2020-10-19
Boot: Introduce new swap method using status partition

romanjoe:
This pull request introduces new swap mode suitable for devices with large minimum write/erase sizes. New mode is based on existing swap move algorithm but uses specially managed flash area called `status partition` instead of image trailer to store sectors statuses while swapping, as well as other data located in trailer.

This mode is developed using Cypress PSoC6 as hardware platform. PSoC6 has minimum write/erase size of 512 bytes, swap mode using  status partition can be easily configured for use with platforms which have different min write/erase size.

Supported:
* swap modes: test, revert, permanent
* power safe swapping (power loss, random resets)
* configurable minimum write/erase size
* flash wear elimination approach (configurable)
* swap status partition size calculations are automated and based on BOOT_MAX_IMG_SECTORS and min write/erase size
* single and multi image mcuboot configurations supported
* existing imgtool compatible

Considerations:
* only tested on cypress hardware
* sim does not support it yet, but @d3zd3z hopes to add it

Any comments, suggestions are welcome.


romanjoe:
Extracted watchdog related commit and created separate PR for that.

https://github.com/mcu-tools/mcuboot/pull/931

d3zd3z:
> Many of the newer automotive rated MCU's seem to have ECC flash now so it'd be useful in the future.

Is 512-byte write/erase common? We've been trying to track what is in devices in #713, and it'd be nice if you have any better insight into what devices are really being used.

The code in this PR basically set set up to work with devices with small erase units (basically the same as the write size). At least the Psoc and NXP LPC devices are like that. The others with ECC generally seem to do the ECC across a smaller range of bytes, such as 8 bytes.

elcritch:
> Is 512-byte write/erase common? We've been trying to track what is in devices in #713, and it'd be nice if you have any better insight into what devices are really being used.

It seems to be the case for the NXP chips and maybe the TI chips. However, I have a couple of  STM32H7 based boards (like [NUCLEO-H755ZI-Q](https://www.st.com/en/evaluation-tools/nucleo-h755zi-q.html)) that appear to have 32B write/128kB erase [according to their flash manual](https://www.st.com/content/ccc/resource/training/technical/product_training/group0/1f/4f/b7/85/91/e0/4d/2f/STM32H7-Memory-Flash_FLASH/files/STM32H7-Memory-Flash_FLASH.pdf/_jcr_content/translations/en.STM32H7-Memory-Flash_FLASH.pdf). So likely those chips will need to use the current trailer method. It'd be wasteful but the upside is that those chips have 1-2k sectors at that 128k sector size (1-2MB) so in theory the trailer method should work? 

> The code in this PR basically set set up to work with devices with small erase units (basically the same as the write size). At least the Psoc and NXP LPC devices are like that. The others with ECC generally seem to do the ECC across a smaller range of bytes, such as 8 bytes.

It does seem more common, however, perhaps it'll need to be configured on a board-by-board case? I'm hoping to test this out more this week, but not sure if I'll be able get to it yet.

d3zd3z:
@elcritch you may be interested in #949 which is working on making the slightly-larger write sizes workable.

elcritch:
> @elcritch you may be interested in #949 which is working on making the slightly-larger write sizes workable.

Thanks! I'm hoping to getting back to testing MCUBoot soon. The larger write would be helpful on some of these ECC flash units.

d3zd3z:
@romanjoe Have you guys gathered any statistics on how much wear this change introduces, specifically to a fairly large device with 512-byte pages? It seems like this would result in a few thousand write/erase cycles on just a small number of page.

PR #844, opened 2020-10-20
boot: Add vendor defined TLV range and macros

agross-korg:
This patch adds macro definitions for proprietary vendor TLV types.  The
intent of this patch is to set aside a certain range of TLVs for use by
various vendors.

Signed-off-by: Andy Gross <andy.gross@juul.com>

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -89,6 +89,16 @@ struct flash_area;
 #define IMAGE_TLV_DEPENDENCY        0x40   /* Image depends on other image */
 #define IMAGE_TLV_SEC_CNT           0x50   /* security counter */
 #define IMAGE_TLV_BOOT_RECORD       0x60   /* measured boot record */
+					   /*
+					    * vendor reserved TLVs at xxA0-xxFF,
+					    * where xx denotes the upper byte
+					    * range.  Examples:
+					    * 0x00a0 - 0x00ff
+					    * 0x01a0 - 0x01ff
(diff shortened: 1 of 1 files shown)


utzig:
I don't think the macros are required, but I agree we should document reserved ranges of TLV numbers that will never be touched by MCUboot itself, so can be safely used by vendors. 0XA0 - 0xxF (on the lower bits where x could be up to F) looks good to me. @d3zd3z @nvlsianpu have we discusses this in a previous meeting?

d3zd3z:
The reason I suggested the macros was because our ranges will end up non-contiguous, because of the history, and the field being little endian, where the existing values all have patterns in the low byte.

Realistically, though, I'm not sure the macros will ever get used by this code, so maybe comments are enough. But, the macros do pretty definitely describe how we want to allocate these values. It makes compilation slightly longer, but won't add any code.

agross-korg:
If we forgo the macros, i'd just need to expand the comments to give the pattern for the reserved areas.  Not a big deal.

nvlsianpu (approved):

nvlsianpu:
@utzig @d3zd3z ?

utzig:
> @utzig @d3zd3z ?

I'm not a big fan of the macros tbh, @agross-korg do you mind removing the macros? Also the range could be 0xA0 - 0xFF maybe?

agross-korg:
Removed macros and updated the ranges and commit message to reflect the changes.

utzig (approved):
LGTM

PR #847, opened 2020-10-22
zephyr: Kconfig: change default for MCUBOOT_CLEANUP_ARM_CODE

ioannisg:
Zephyr has introduced an option to perform the cleanup
of ARM core HW registers during early boot, when the
firmware is chain-loaded by MCUboot. Therefore, MCUboot
does not need to perform the same cleanup before jumping
to the application image. The patch relies on the fact
that building MCUboot with Zephyr implies loading also
a Zephyr-based application firmware. If this is not the
case, the application developer needs to enable the
MCUBOOT_CLEANUP_ARM_CODE Kconfig option manually, in the
project configuration.

Signed-off-by: Ioannis Glaropoulos <Ioannis.Glaropoulos@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -131,12 +131,6 @@ static void do_boot(struct boot_rsp *rsp)
                                      rsp->br_image_off +
                                      rsp->br_hdr->ih_hdr_size);
 
-#ifdef CONFIG_CPU_CORTEX_M7
-    /* Disable instruction cache and data cache before chain-load the application */
-    SCB_DisableDCache();
-    SCB_DisableICache();
-#endif
-
     irq_lock();
 #ifdef CONFIG_SYS_CLOCK_EXISTS
     sys_clock_disable();
@@ -147,6 +141,15 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 #if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
     cleanup_arm_nvic(); /* cleanup NVIC registers */
+
+#ifdef CONFIG_CPU_CORTEX_M7
+    /* Disable instruction cache and data cache before chain-load the application */
+    SCB_DisableDCache();
+    SCB_DisableICache();
+#endif
+
+#if CONFIG_CPU_HAS_ARM_MPU
+    z_arm_clear_arm_mpu_config();
 #endif
 
 #if defined(CONFIG_BUILTIN_STACK_GUARD) && \
@@ -157,6 +160,8 @@ static void do_boot(struct boot_rsp *rsp)
     __set_PSPLIM(0);
     __set_MSPLIM(0);
 #endif
+
+#endif /* CONFIG_MCUBOOT_CLEANUP_ARM_CORE */
 
 #ifdef CONFIG_BOOT_INTR_VEC_RELOC
 #if defined(CONFIG_SW_VECTOR_RELAY)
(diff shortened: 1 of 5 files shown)


ioannisg:
Zephyr PR for ARM core registers' cleanup at early boot: https://github.com/zephyrproject-rtos/zephyr/pull/29400

nvlsianpu (commented):
LGTM

We shouldn't merge this until https://github.com/zephyrproject-rtos/zephyr/pull/29400 is not merged.

d3zd3z (approved):

ioannisg:
> LGTM
> 
> We shouldn't merge this until [zephyrproject-rtos/zephyr#29400](https://github.com/zephyrproject-rtos/zephyr/pull/29400) is not merged.

@nvlsianpu I am confused by the above.

IMHO, the policy should be: Do not merge this  until zephyrproject-rtos/zephyr#29400 is concluded that it will be merged (e.g. the review is completed).

ioannisg:
@nvlsianpu I am reverting 4dc12f4, since I added in this PR the logic to cleanup the MPU before jumping to the application.

Note that the above is not done when we rely on the application to perform the clean up (this is what my Zephyr PR does). So in either case we should be safe to run MCU boot with HW Stack Protection ON.

nvlsianpu:
> @nvlsianpu I am confused by the above.

> IMHO, the policy should be: Do not merge this until zephyrproject-rtos/zephyr#29400 is concluded that it will be merged (e.g. the review is completed).

Not a problem - This PR can be merge in a second after zephyr PR. What worries me is that 4dc12f4 was for instance bugfix for bug registered in zephyr-rtos project. Let's merge both same day (sequence is negligible), so no one won't encounter an issue.

ioannisg:
> > @nvlsianpu I am confused by the above.
> 
> > IMHO, the policy should be: Do not merge this until [zephyrproject-rtos/zephyr#29400](https://github.com/zephyrproject-rtos/zephyr/pull/29400) is concluded that it will be merged (e.g. the review is completed).
> 
> Not a problem - This PR can be merge in a second after zephyr PR. What worries me is that [4dc12f4](https://github.com/JuulLabs-OSS/mcuboot/commit/4dc12f4807099f0c210cad350e1439971677b9c0) was for instance bugfix for bug registered in zephyr-rtos project. Let's merge both same day (sequence is negligible), so no one won't encounter an issue.

This PR changes the policy of MCUBOOT_CLEANUP_ARM_CORE, assuming the Zephyr changes will get through. The Zephyr PR, though, includes a manifest update for MCUboot, so it really needs the MCUboot upstream and zephyr module to both be updated first. So I agree this should be done atomically (and I can volunteer for this since I seem to have merge rights to all 3 repos ;) )

boot/zephyr/main.c
```diff
@@ -147,6 +141,15 @@ static void do_boot(struct boot_rsp *rsp)
...
     cleanup_arm_nvic(); /* cleanup NVIC registers */
+
+#ifdef CONFIG_CPU_CORTEX_M7
+    /* Disable instruction cache and data cache before chain-load the application */
+    SCB_DisableDCache();
+    SCB_DisableICache();
+#endif
+
+#if CONFIG_CPU_HAS_ARM_MPU
+    z_arm_clear_arm_mpu_config();
```
nvlsianpu, line 152: can you add `__week` z_arm_clear_arm_mpu_config(void).
Thanks to that mcuboot would be zephyr backward compatible?

ioannisg: @nvlsianpu , yes, sure this could be done, hang on... :)

nvlsianpu (approved):

PR #852, opened 2020-10-27
[DNM] Allow incompatible slots when using swap-move

utzig:
Update swap with move strategy to use a build time size configuration to use for moving up and swapping, so incompatible slots, eg primary with 4096 byte sectors and secondary with 8192 sectors can work together. Sectors must still be all linear with the same size on each slot, both have to have sectors with sizes that are multiples, etc.

WIP - To test on sim: `MCUBOOT_SWAP_MOVE_SIZE=8192 cargo test --features "swap-move sig-rsa validate-primary-slot"`


utzig:
@d3zd3z @nvlsianpu While testing this PR I had the same issues we discussed in the meeting this morning while running on the simulator, `perm_with_random_fails` sometimes fails, I will investigate this but we should not release a new version before it is known why this is happening.

utzig:
> @d3zd3z @nvlsianpu While testing this PR I had the same issues we discussed in the meeting this morning while running on the simulator, `perm_with_random_fails` sometimes fails, I will investigate this but we should not release a new version before it is known why this is happening.

I believe, although I'm not 100% sure, this is an issue with the test itself, not with the bootloader, I'll see into fixing it tomorrow. It only happens when using `swap-move`.

PR #853, opened 2020-10-28
boot: Fix FIH_LABEL/FIH_CALL macros for IAR

TTornblom:
The IAR toolchain does not handle the "%=" template. Emulate it with
a combination of __LINE__ and __COUNTER__

Signed-off-by: TTornblom <thomas.tornblom@iar.com>

### DIFF
--- a/boot/bootutil/include/bootutil/fault_injection_hardening.h
+++ b/boot/bootutil/include/bootutil/fault_injection_hardening.h
@@ -284,7 +284,11 @@ void fih_cfi_decrement(void);
 /* Label for interacting with FIH testing tool. Can be parsed from the elf file
  * after compilation. Does not require debug symbols.
  */
+#if defined(__ICCARM__)
+#define FIH_LABEL(str, lin, cnt) __asm volatile ("FIH_LABEL_" str "_" #lin "_" #cnt "::" ::);
+#else
 #define FIH_LABEL(str) __asm volatile ("FIH_LABEL_" str "_%=:" ::);
+#endif
 
 /* Main FIH calling macro. return variable is second argument. Does some setup
  * before and validation afterwards. Inserts labels for use with testing script.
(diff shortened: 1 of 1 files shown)


utzig (approved):

tamasban (approved):

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -284,7 +284,11 @@ void fih_cfi_decrement(void);
 /* Label for interacting with FIH testing tool. Can be parsed from the elf file
  * after compilation. Does not require debug symbols.
  */
+#if defined(__ICCARM__)
+#define FIH_LABEL(str, lin, cnt) __asm volatile ("FIH_LABEL_" str "_" #lin "_" #cnt "::" ::);
```
RcColes, line 288: Could `#lin` in this case be replaced by `__LINE__`? If we could get some combination that would allow global uniqueness with the same function signature (only one argument), then it would make this change a lot more simple.

I'm not sure of the exact list of symbols allowed in these labels, but `"FIH_LABEL_" str "_" __func__ "_line_"` might be a good options for uniqueness.

TTornblom: I've tried both of these suggestions and for some reason they don't work.

__func__ or __FUNCTION__ apparently are variables, not pre-processor macros, so they expand to themselves when used as that.

RcColes: Ah yes - three are three, of which two are macros and one is a function. I got it the wrong way round - `__FILE__` and `__LINE__` are the two that are macros. I'd suspect that some of the symbols in the filepaths might make a symbol invalid though

RcColes: I think it's fine to merge in its current form anyway - seems like this might be more trouble than it's worth

PR #855, opened 2020-10-28
Fix upgrade issue in swap-move

utzig:
A previous fix for allowing padded images in the primary slot, ended up causing an issue that would fail to upgrade under the right circumstances. The issue was caused when the following set of steps happened after an upgrade was detected:

1) trailer is erased on the primary slot
2) status is written to the primary slot
3) trailer is erased on the secondary slot
4) reset
1) trailer is erased on the primary slot
2) status partially written or not written to primary slot (no magic)
3) reset

This would result in images stored in the same slots they were initially, aka no upgrade, which would fail the simulator test for upgraded images.

The previous padded images fix was reverted and the status source was upgraded to also check that the secondary slot has a valid magic in it, so that there's never a circumstance where there is no trailer in any of the slots while an upgrade operation is being decided on.

### DIFF
--- a/boot/bootutil/src/swap_move.c
+++ b/boot/bootutil/src/swap_move.c
@@ -264,6 +264,7 @@ int
 swap_status_source(struct boot_loader_state *state)
 {
     struct boot_swap_state state_primary_slot;
+    struct boot_swap_state state_secondary_slot;
     int rc;
     uint8_t source;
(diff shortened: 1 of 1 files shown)


d3zd3z (approved):
LGTM

nvlsianpu (approved):

PR #856, opened 2020-10-29
zephyr: added ISB after cortex control was set

nvlsianpu:
After modifying the CONTROL register, ISB function should be
used to ensure the effect of the change applies to
subsequent code.
Due to the simple nature of the Cortex-M3, Cortex-M4, Cortex-M0+,
Cortex-M0, and Cortex-M1 pipeline,
omission of the ISB instruction does not cause any problem.
But might cause problem on a more advanced core.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -182,6 +182,7 @@ static void do_boot(struct boot_rsp *rsp)
     __set_MSP(vt->msp);
 #if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
     __set_CONTROL(0x00); /* application will configures core on its own */
+    __ISB();
 #endif
     ((void (*)(void))vt->reset)();
 }


utzig:
> But might cause problem on a more advanced core.

Which one? How do you know that you have to call ISB after changing CONTROL? Could you point me to where this is written in the ARM docs?

utzig:
> > But might cause problem on a more advanced core.
> 
> Which one? How do you know that you have to call ISB after changing CONTROL? Could you point me to where this is written in the ARM docs?

I am fine with the change btw, just curious! :-)

nvlsianpu:
"Software must use an ISB barrier instruction to ensure a write to the CONTROL register takes effect before the next
instruction is executed." B 1.4.4 in ARMv7-M Architecture Reference Manual https://developer.arm.com/documentation/ddi0403/ed

ioannisg (approved):
Agree with the change, I confirm that isb is needed after setting the Control. However I think we need to address the issue of changing the SP being used, and then referencing a stack variable. I had left a comment about this in the original pr

utzig (approved):

nvlsianpu:
> Agree with the change, I confirm that isb is needed after setting the Control. However I think we need to address the issue of changing the SP being used, and then referencing a stack variable. I had left a comment about this in the original pr

ok, in another PR. Do you think making `vt` either static or global will works?

ioannisg:
> > Agree with the change, I confirm that isb is needed after setting the Control. However I think we need to address the issue of changing the SP being used, and then referencing a stack variable. I had left a comment about this in the original pr
> 
> ok, in another PR. Do you think making `vt` either static or global will works?

@nvlsianpu I think the safest thing to do is to store the vt on a register, so it is always accessible no matter what the compiler chooses to do.,

As a side note, I think this file behaves as an ARM-only, although nothing in the directory or file name suggests that. This probably needs to be handled.

nvlsianpu:
> As a side note, I think this file behaves as an ARM-only, although nothing in the directory or file name suggests that. This probably needs to be handled.

Not really - there are here 3 versions of do_boot() functions - for ARM, Xtensa and x86. Although ARM has much better support here than the others.

ioannisg:
> > As a side note, I think this file behaves as an ARM-only, although nothing in the directory or file name suggests that. This probably needs to be handled.
> 
> Not really - there are here 3 versions of do_boot() functions - for ARM, Xtensa and x86. Although ARM has much better support here than the others.

Right, my bad, I noticed that now. So we need to think how scalable this solution is (i.e. duplicating the do_boot() function), considering we might want to support also risc-v and/or ARC here...

PR #863, opened 2020-11-05
Introduce utility function to retrieve primary application version.

AGlass0fMilk:
This commit introduces a utility function that allows the application to retrieve the version number stored in the application header info for the primary slot.

Wasn't sure if this functionality was already available from mcuboot's APIs, but I did not find it when I looked.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -837,3 +837,30 @@ done:
     flash_area_close(fap);
     return rc;
 }
+
+/**
+ * Populates the version information of the
+ * currently installed primary application
+ *
+ * @param[in] version Destination version structure buffer
+ * @return 0 on success; nonzero on failure.
+ */
+int boot_get_current_version(struct image_version *version)
+{
+    struct boot_loader_state state;
+    struct boot_status status;
+    struct image_header hdr;
+    int rc;
+
+    assert(version != NULL);
+
+    rc = boot_read_image_header(&state, 0, &hdr, &status);
+    if (rc != 0) {
+        return rc;
+    }
+
+    /** Copy the header's version struct over into the caller-supplied buffer */
+    memcpy(version, &hdr.ih_ver, sizeof(struct image_version));
+
+    return 0;
+}
--- a/boot/bootutil/include/bootutil/bootutil.h
+++ b/boot/bootutil/include/bootutil/bootutil.h
@@ -62,6 +62,8 @@ extern "C" {
 #define BOOT_MAX_ALIGN          8
 
 struct image_header;
+struct image_version;
+
 /**
  * A response object provided by the boot loader code; indicates where to jump
  * to execute the main image.
@@ -102,6 +104,8 @@ int boot_swap_type(void);
 
 int boot_set_pending(int permanent);
 int boot_set_confirmed(void);
+
+int boot_get_current_version(struct image_version *version);
 
 #define SPLIT_GO_OK                 (0)
 #define SPLIT_GO_NON_MATCHING       (-1)


boot/bootutil/src/bootutil_misc.c
```diff
@@ -837,3 +837,28 @@ boot_set_confirmed(void)
...
+ * Populates the version information of the
+ * currently installed primary application
+ *
+ * @param[in] version Destination version structure buffer
+ * @return 0 on success; nonzero on failure.
+ */
+int boot_get_current_version(struct image_version *version)
+{
+    struct boot_loader_state state; // Dummy state
+    struct boot_status status;      // Dummy status
```
utzig, line 851: We don't use C++-style line comments. Also these comments are not very relevant so they don't need to be here.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -837,3 +837,28 @@ boot_set_confirmed(void)
...
+    struct image_header hdr;
+    int rc;
+
+    rc = boot_read_image_header(&state, 0, &hdr, &status);
+    if(rc != 0) {
```
utzig, line 856: Extra space between `if (`

boot/bootutil/src/bootutil_misc.c
```diff
@@ -837,3 +837,28 @@ boot_set_confirmed(void)
...
+        return rc;
+    }
+
+    // Copy the header's version struct over into the caller-supplied buffer
```
utzig, line 860: Use `/* ... */`

boot/bootutil/src/bootutil_misc.c
```diff
@@ -837,3 +837,28 @@ boot_set_confirmed(void)
...
+    memcpy(version, &hdr.ih_ver, sizeof(struct image_version));
```
utzig, line 861: What if someone passed in `version` equal to `NULL`?

utzig (changes requested):
I am not sure it is a good idea to add this to the bootloader itself, this should probably live in some supporting code in the OS. But please fix the issues while we give it some extra thought.

AGlass0fMilk:
> I am not sure it is a good idea to add this to the bootloader itself, this should probably live in some supporting code in the OS. But please fix the issues while we give it some extra thought.

The bootloader already provides utility functions to the application such as `boot_set_pending` and `boot_set_confirmed`. I think it makes sense to provide an additional utility function to read the primary slot version. 

The linker should take care of removing this if it is unused in the bootloader.

https://github.com/mcu-tools/mcuboot/blob/fe67e8a83ad3cfbfbd332f30dd9f24b63a5a865b/boot/bootutil/include/bootutil/bootutil.h#L105-L106

I will address your other comments soon.

AGlass0fMilk:
@d3zd3z @nvlsianpu  What do you think about adding this

`int boot_get_current_version(struct image_version *version); `

to the API in bootutil.h? 

For Mbed we made it so that one can build mcuboot as a library for the application (by excluding the main in mcuboot). This allows applications to implement the update mechanism. 

Therefore, they must be able to call things like `boot_set_confirmed` and `boot_set_pending`. I think it makes sense to have a way to get the primary slot header version info alongside these functions.

nvlsianpu:
I see no obstacles.
Zephyr-RTOS implements such features in its code-base, but at some point it might turn to use what is provided here as well.

AGlass0fMilk:
@utzig @nvlsianpu I addressed the requested changes above.

Let me know if there's anything else this PR needs.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -837,3 +837,30 @@ boot_set_confirmed(void)
...
+ * @return 0 on success; nonzero on failure.
+ */
+int boot_get_current_version(struct image_version *version)
+{
+    struct boot_loader_state state;
+    struct boot_status status;
+    struct image_header hdr;
+    int rc;
+
+    assert(version != NULL);
```
utzig, line 855: This is not OK. You should fail when someone passes `version` == `NULL`, not cause a fault. Also `assert` could be opted out by defining `NDEBUG` which would make this a no-opt. Building with `NDEBUG` is currently unsupported by MCUboot, but the issue was already raised in the Mynewt community, so until we fix it, at least don't add new code which depends on `assert`.

utzig (changes requested):
This can be better thought about after 1.7.0 is released. Also depending on the build configuration, an image might be running on the secondary slot, and this PR is making the assumption it will always run on the primary. At least it needs to get some #ifdefery for that case.

boot/bootutil/src/bootutil_misc.c
AGlass0fMilk: I have removed the call to `assert` and instead now return `BOOT_EBADARGS` if `NULL` is passed in.

AGlass0fMilk:
@utzig I have updated the API so the user can provide the slot they want to get the version of.

Do you see a straightforward way for the application to check what slot it is running from? Then we could introduce another function like `boot_get_current_slot` or something along those lines.

I still think this kind of functionality (parsing the mcuboot headers) should be centralized in mcuboot itself instead of having each OS implement its own method as is currently done by zephyr. This minimizes the chance for future incompatibility and reduces code duplication.

d3zd3z:
> to the API in bootutil.h?

While this is useful within the bootloader itself, it really isn't intended to be an API between the application and the bootloader itself. Currently, the code is duplicated between MCUboot and Zephyr. There is a pending PR to extract this code so it can be more easily shared, but it would still be duplicated.

If we wanted a cleaner API, we would need to think about how to do it in such a way that there isn't link-time dependencies between the application and the bootloader. We've thought about this in the past, and had some ideas of either a table of function pointers, or a single function that takes arguments to determine what to do (similar to ioctl). It would need to be at a well-known address, and have some magic associated with it to make sure that it was present, and in a version that was compatible between the app and the bootloader. It would also be useful to have this able to query slot configuration, and capabilities compiled into the code.

nvlsianpu:
@AGlass0fMilk Do you want to continue this patch?
Now code which is aimed to be the chain-loaded application's interface should be implemented by boootutil_public.c/h  files.

boot/bootutil/src/bootutil_misc.c
d3zd3z: There is an assert, which suggests the API is not intended to cover this case. If assertions aren't tested, this will provoke undefined behavior.

PR #864, opened 2020-11-05
Mbed-OS patches - temporarily disable TinyCrypt backend

AGlass0fMilk:
This PR introduces a few patch changes to the Mbed-OS port including:

- Add a link to the Mbed-OS readme from the respository's main README.
- Temporarily disable building with the TinyCrypt backend for Mbed-OS builds

An explanation of the TinyCrypt change:

TinyCrypt uses a modified version of micro-ecc. Another version of micro-ecc is also used by the Mbed Cordio BLE stack. When building mcuboot for a target with BLE enabled, this causes multiple-defined symbol errors during linking. Due to the nature of Mbed's current build system, it is difficult to fix this.

Mbed will soon release a more flexible cmake-based build system that will make it possible to exclude these TinyCrypt files from an application build that may also link Cordio BLE sources.

Until then, this commit temporarily disables the use of TinyCrypt with Mbed-OS and excludes its sources from the build to avoid this build error. 

See AGlass0fMilk/mbed-mcuboot-blinky#2 for more details.

### DIFF
--- a/.mbedignore
+++ b/.mbedignore
@@ -15,3 +15,5 @@ ext/mbedtls/*
 ext/mbedtls-asn1/*
 ext/nrf/*
 ext/tinycrypt/tests/*
+ext/tinycrypt/*
+ext/tinycrypt-sha512/*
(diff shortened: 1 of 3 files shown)


nvlsianpu (approved):

nvlsianpu:
cc @LDong-Arm

utzig (approved):

LDong-Arm (approved):

PR #869, opened 2020-11-10
ci: move imgtool publishing to GH workflows

utzig:
Remove simulator and Mynewt tests from Travis config; disable imgtool publishing on Travis; update scripts to work on both Travis and GH; add GH workflow for publishing.


utzig:
@d3zd3z @nvlsianpu I tested this with a PR on my own fork because secrets are not shared with forks; short of knowing what will happen when the version is updated, it seems to work correctly.

d3zd3z (approved):

utzig:
I just commented out the lines that should not run, so it should be easy to get back with #789. I also removed the token from Travis so even some hiccup should not allow it to publish there.

nvlsianpu (approved):

utzig:
Alright, travis is no more, #789 can add it back.

PR #871, opened 2020-11-10
boot: add support for custom user code to handle bootloader events

arvinf:
An event callback is invoked at each stage of the bootloader
lifecycle to indicate progress, errors, and image swap operations.
The user application can provide the function `bootloader_event`
to receives these events and performs design specific tasks such
as updating UI indicators to show image swap progress or bootloader
error.

`USER_ADDON_FILES` can be used to pass a list of user source files,
with custom code to handle events, to cmake be built as part of the
bootloader.

This feature enables users to customize some aspects of the bootloader
for their own application and board without needing to modify
the bootloader source code and keeping it in sync the master repo.

### DIFF
--- a/boot/bootutil/include/bootutil/bootloader_events.h
+++ b/boot/bootutil/include/bootutil/bootloader_events.h
@@ -1,0 +1,46 @@
+/*
+ *  Copyright (C) 2020, IP-Logix Inc.
+ *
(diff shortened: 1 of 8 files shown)


utzig:
The simulator fails because your code does not build at least one configuration, you have to fix it so that it works properly no matter which config options the user selected.

PR #876, opened 2020-11-12
ci: update workflows names and add more badges

utzig:
Fix workflows names on yaml root to be able to use them to refer to the workflow. Add badges for pypi, workflows and license.

### DIFF
--- a/.github/workflows/imgtool.yaml
+++ b/.github/workflows/imgtool.yaml
@@ -3,19 +3,20 @@ on:
     branches:
       - master
 
+name: imgtool
+
 jobs:
   environment:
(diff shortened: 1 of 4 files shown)


utzig:
@nvlsianpu @d3zd3z I believe this will add proper badges for workflows (and a few extras!). The README preview is not shows imgtool workflow correctly, but I think it's a hiccup.

utzig:
> The README preview is not shows imgtool workflow correctly, but I think it's a hiccup.

Duh, probably because it does not run on PRs!

nvlsianpu (approved):

Issue #873, opened 2020-11-11
Useless test in LOAD_IMAGE_DATA 


TTornblom:
Just noticed that there is a test on the return value from memcpy() in the LOAD_IMAGE_DATA() macro where the return value is compared to the first argument to memcpy().

This is a useless test as memcpy() always return its first argument, there is no other return value defined for this function.

utzig:
What would you suggest do be done instead?

TTornblom:
I'm not sure.

If MCUBOOT_RAM_LOAD is not defined, then it will use flash_area_read() instead of memcpy(), and I assume that will return some kind of error status if it fails, but if it is defined, then this test will always fail, so perhaps something like:

`#define LOAD_IMAGE_DATA(hdr, fap, start, output, size)       \
    (memcpy((output),(void*)((hdr)->ih_load_addr + (start)), \
            (size)), 0)
`

This will give a compiler a better chance of eliminating the dead code

tamasban:
Thanks @TTornblom for the proposal! Tbh never seen such a syntax, but it works with ARMCLANG and GNUARM as well. It spares around 50bytes of code. I will test it on Musca-A, and open PR later if works.

davidvincze:
I have never seen anything like this before: `rc = (memcpy(...), 0);`
Nor was I sure the syntax is valid. :D

utzig:
> Nor was I sure the syntax is valid. :D Does it have a name?

```
6.5.17 Comma operator

Syntax

1 expression:
       assignment-expression
            expression , assignment-expression

Semantics

2 The left operand of a comma operator is evaluated as a void expression; there is a sequence point
between its evaluation and that of the right operand. Then the right operand is evaluated; the result
has its type and value.114)
```

Source: http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1570.pdf

tamasban:
https://github.com/mcu-tools/mcuboot/pull/877

Tested on Musca-A with GNUARM.

davidvincze:
PR#877 has been merged which provides a fix, so I'm closing this issue.

PR #877, opened 2020-11-12
boot: Fix LOAD_IMAGE_DATA macro

tamasban:
If RAM_LOAD is defined then the return value of memcpy() is
always compared against its first parameter. By definition
memcpy() returns with its first paramter (destination) so
the not equal check is always false. The fix replaces this
runtime check with a comma operator assigning 0 to the
variable at build time, as a result compiler can be done
dead code elimination much better.

Signed-off-by: Tamas Ban <tamas.ban@arm.com>
Change-Id: I8d0d6e68a931661fa19d395556beb20470d74fb1


tamasban:
Fix issue #873

utzig (approved):

nvlsianpu:
Doesn't this be a problem: `warning: right-hand operand of comma expression has no effect [-Wunused-value]`
This will be problem when this code will be compiled for the zephyr port (not yet!).

nvlsianpu (approved):

PR #882, opened 2020-11-14
zephyr: add an option to use GPIO to enable USB DFU.

jmgao:
Signed-off-by: Josh Gao <josh@jmgao.dev>

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -355,9 +355,15 @@ config BOOT_FIH_PROFILE_HIGH
 
 endchoice
 
-config BOOT_WAIT_FOR_USB_DFU
+choice BOOT_USB_DFU
+	prompt "USB DFU"
+	default BOOT_USB_DFU_NO
+
+config BOOT_USB_DFU_NO
+	prompt "Disabled"
+
+config BOOT_USB_DFU_WAIT
 	bool "Wait for a prescribed duration to see if USB DFU is invoked"
-	default n
 	select USB
 	select USB_DFU_CLASS
 	select IMG_MANAGER
@@ -365,6 +371,58 @@ config BOOT_WAIT_FOR_USB_DFU
 	  If y, MCUboot waits for a prescribed duration of time to allow
 	  for USB DFU to be invoked. Please note DFU always updates the
 	  slot1 image.
+
+config BOOT_USB_DFU_GPIO
+	bool "Use GPIO to detect whether to trigger DFU mode"
+	select USB
+	select USB_DFU_CLASS
+	select IMG_MANAGER
+	help
+	  If y, MCUboot uses GPIO to detect whether to invoke USB DFU.
+
+endchoice
+
+config BOOT_USB_DFU_WAIT_DELAY_MS
+	int "USB DFU wait duration"
+	depends on BOOT_USB_DFU_WAIT
+	default 12000
+	help
+	  Milliseconds to wait for USB DFU to be invoked.
+
+if BOOT_USB_DFU_GPIO
+
+config BOOT_USB_DFU_DETECT_PORT
+	string "GPIO device to trigger USB DFU mode"
+	default GPIO_0 if SOC_FAMILY_NRF
+	help
+	  Zephyr GPIO device that contains the pin used to trigger
+	  USB DFU.
+
+config BOOT_USB_DFU_DETECT_PIN
+	int "Pin to trigger USB DFU mode"
+	default 6 if BOARD_NRF9160DK_NRF9160
+	default 11 if BOARD_NRF52840DK_NRF52840
+	default 13 if BOARD_NRF52DK_NRF52832
+	default 23 if BOARD_NRF5340_DK_NRF5340_CPUAPP || BOARD_NRF5340_DK_NRF5340_CPUAPPNS
+	help
+	  Pin on the DFU detect port that triggers DFU mode.
+
+config BOOT_USB_DFU_DETECT_PIN_VAL
+	int "USB DFU detect pin trigger value"
+	default 0
+	range 0 1
+	help
+	  Logic value of the detect pin that triggers USB DFU mode.
+
+config BOOT_USB_DFU_DETECT_DELAY
+	int "Serial detect pin detection delay time [ms]"
+	default 0
+	help
+	  Used to prevent the bootloader from loading on button press.
+	  Useful for powering on when using the same button as
+	  the one used to place the device in bootloader mode.
+
+endif # BOOT_USB_DFU_GPIO
 
 config ZEPHYR_TRY_MASS_ERASE
 	bool "Try to mass erase flash when flashing MCUboot image"
@@ -453,7 +511,7 @@ config BOOT_SERIAL_DETECT_PORT
 	string "GPIO device to trigger serial recovery mode"
 	default GPIO_0 if SOC_FAMILY_NRF
 	help
-	  Zephyr GPIO device which contains the pin used to trigger
+	  Zephyr GPIO device that contains the pin used to trigger
 	  serial recovery mode.
 
 config BOOT_SERIAL_DETECT_PIN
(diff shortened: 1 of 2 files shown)


nvlsianpu (commented):
Can you add decryption on what and why were reworked/introduced?
(To the commit message as well)

jmgao:
Here's the updated commit message, is this what you were looking for?

```
zephyr: add an option to use GPIO to enable USB DFU.

Devices with a physical reset button might prefer to use it to enter USB
DFU mode, instead of always entering it with a timeout. Extract the
existing CONFIG_BOOT_SERIAL_DETECT detection code and use it to enter
DFU mode when CONFIG_BOOT_USB_DFU_GPIO is enabled.

This commit depends on zephyrproject-rtos/zephyr#30015, which changes
wait_for_usb_dfu from a nullary function that waits for a compile-time
fixed amount of time, to one that takes a timeout.

Signed-off-by: Josh Gao <josh@jmgao.dev>
```

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
 	  for USB DFU to be invoked. Please note DFU always updates the
 	  slot1 image.
 
+config BOOT_USB_DFU_WAIT_DELAY_MS
+	int
+	depends on BOOT_USB_DFU_WAIT
+	default 12000
+	help
+	  The duration to wait for USB DFU to be invoked.
```
de-nordic, line 378: I know that the config identifier states the unit is MS, but maybe it is worth mentioning that these are milliseconds in the help.

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
...
+
+config BOOT_USB_DFU_GPIO
+	bool "Use GPIO to detect whether to trigger DFU mode"
+	select USB
+	select USB_DFU_CLASS
+	select IMG_MANAGER
+	help
+	  If y, MCUboot uses GPIO to detect whether to invoke USB DFU.
+	  Please note DFU always updates the slot1 image.
```
de-nordic, line 387: I do not think that this comment is needed here.

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
...
+
+endchoice
+
+config BOOT_USB_DFU_DETECT_PORT
+	string "GPIO device to trigger USB DFU mode"
+	depends on BOOT_USB_DFU_GPIO
+	default GPIO_0 if SOC_FAMILY_NRF
+	help
+	  Zephyr GPIO device which contains the pin used to trigger
```
de-nordic, line 396: "contains the pin (...)" is restrictive clause, which identifies a specific GPIO, use "that" instead of "which".

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
...
+
+config BOOT_USB_DFU_DETECT_PIN
+	int "Pin to trigger USB DFU mode"
+	depends on BOOT_USB_DFU_GPIO
+	default 6 if BOARD_NRF9160DK_NRF9160
+	default 11 if BOARD_NRF52840DK_NRF52840
+	default 13 if BOARD_NRF52DK_NRF52832
+	default 23 if BOARD_NRF5340_DK_NRF5340_CPUAPP || BOARD_NRF5340_DK_NRF5340_CPUAPPNS
+	help
+	  Pin on the DFU detect port which triggers DFU mode.
```
de-nordic, line 407: "that" instead of "which"

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
...
+
+config BOOT_USB_DFU_DETECT_PIN_VAL
+	int "USB DFU detect pin trigger value"
+	depends on BOOT_USB_DFU_GPIO
+	default 0
+	range 0 1
+	help
+	  Logic value of the detect pin which triggers USB DFU mode.
```
de-nordic, line 415: "that" instead of "which"

boot/zephyr/main.c
```diff
@@ -313,6 +313,40 @@ void zephyr_boot_log_stop(void)
...
+#if defined(CONFIG_MCUBOOT_SERIAL) || defined(CONFIG_BOOT_USB_DFU_GPIO)
+static bool detect_pin(const char* port, int pin, uint32_t expected_value) {
+    int rc;
+    struct device const *detect_port;
+    uint32_t detect_value = !expected_value;
+
+    detect_port = device_get_binding(port);
+    __ASSERT(detect_port, "Error: Bad port for GPIO boot detection.\n");
+
+    /* The default presence value is 0 which would normally be
```
de-nordic, line 325: The explanation is hard to understand.

boot/zephyr/Kconfig
jmgao: Previously, this blurb was only in the `BOOT_WAIT_FOR_USB_DFU` config option: where should it go after a second way to enable USB DFU is added?

boot/zephyr/Kconfig
```diff
@@ -364,6 +370,50 @@ config BOOT_WAIT_FOR_USB_DFU
...
+	select IMG_MANAGER
+	help
+	  If y, MCUboot uses GPIO to detect whether to invoke USB DFU.
+	  Please note DFU always updates the slot1 image.
+
+endchoice
+
+config BOOT_USB_DFU_DETECT_PORT
+	string "GPIO device to trigger USB DFU mode"
+	depends on BOOT_USB_DFU_GPIO
```
de-nordic, line 393: You may enclose all of the configs that depend on `BOOT_USB_DFU_GPIO` in if/endif block (https://docs.zephyrproject.org/latest/guides/kconfig/tips.html?highlight=kconfig#factoring-out-common-dependencies)

boot/zephyr/main.c
```diff
@@ -313,6 +313,40 @@ void zephyr_boot_log_stop(void)
...
+     * active-low, but historically the raw value was checked so we'll
+     * use the raw interface.
+     */
+    rc = gpio_pin_configure(detect_port, pin,
+#ifdef GPIO_INPUT
+                            GPIO_INPUT | GPIO_PULL_UP
+#else
+                            GPIO_DIR_IN | GPIO_PUD_PULL_UP
+#endif
+           );
```
de-nordic, line 335: I do not know whether mcuboot style allows that but it does not look nice.

jmgao: Essentially this entire function of code was extracted from below.

boot/zephyr/Kconfig
de-nordic: I mean the "Please note DFU always updates the slot1 image".
The problem with this comment is that you may change where the image will be loaded without knowing this comment exists, which will make it misleading. For example I am doing something like this here https://github.com/zephyrproject-rtos/zephyr/pull/29974.

jmgao: Yeah, I meant that as well. Do you mean that I should just delete the line here (and in `BOOT_USB_DFU_WAIT`)?

boot/zephyr/main.c
de-nordic: My bad.

boot/zephyr/Kconfig
de-nordic: Yes, I think so. The option does not affect which Image is selected for upload.

jmgao:
Resolved all of the comments with the latest revision.

de-nordic (approved):
Looks OK

jmgao:
Rebased onto tip of tree.

boot/zephyr/Kconfig
```diff
@@ -453,7 +511,7 @@ config BOOT_SERIAL_DETECT_PORT
 	string "GPIO device to trigger serial recovery mode"
 	default GPIO_0 if SOC_FAMILY_NRF
 	help
-	  Zephyr GPIO device which contains the pin used to trigger
```
d3zd3z, line 456: Any particular reason to change this? Both 'which' and 'that' would be fine here.

jmgao: I was [asked to change the phrasing](https://github.com/mcu-tools/mcuboot/pull/882#discussion_r526215347) in the new config variable, so I changed the original one to match.

d3zd3z: Ok. That's fine. I guess I've been around too much non-American English. The that for restrictive clauses is an American thing.

d3zd3z (approved):

PR #884, opened 2020-11-20
Cypress: Integrate encryption with mbedTLS

BohdanKovalchuk:
Signed-off-by: Bohdan Kovalchuk <bohd@cypress.com>


BohdanKovalchuk:
@d3zd3z , @utzig please, review

Thanks

BohdanKovalchuk:
This one should be merged before  #841.

boot/cypress/MCUBootApp/encrypted_mbedtls.c
```diff
@@ -0,0 +1,457 @@
...
+            for (j = 16; j > 0; --j) {
+                if (++nonce[j - 1] != 0) {
+                    break;
+                }
+            }
+        }
+
+        u8 = *buf;
+        *buf++ = u8 ^ blk[blk_off];
+        blk_off = (blk_off + 1) & 0x0f;
```
utzig, line 444: This block is not required because MbedTLS has proper streaming for CTR with `mbedtls_aes_crypt_ctr`

utzig (changes requested):
The code under `boot/cypress/MCUBootApp/encrypted_mbedtls.c`, which apart from the new stuff (EC+HKDF+HMAC with MbedTLS), is mostly copypasta, should instead be added to `bootutil/src/encrypted.c` (with the proper `bootutil/crypto` updates). Also the EC256 encrypted images with MbedTLS should also run on the simulator.

utzig:
@BohdanKovalchuk I fast glimpse over the code, it seems to be going in the right direction, but before a proper review please fix your commits so the simulator can be properly run.

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -56,6 +64,72 @@ static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *c
...
+{
+    mbedtls_mpi_init(&ctx->z);
+    mbedtls_mpi_init(&ctx->d);
+
+    mbedtls_ecp_group_init( &ctx->grp );
+    mbedtls_ecp_point_init( &ctx->P );
+
+    if (0 != mbedtls_ecp_group_load( &ctx->grp, MBEDTLS_ECP_DP_SECP256R1 )) {
+        mbedtls_ecp_group_free( &ctx->grp );
+        mbedtls_ecp_point_free( &ctx->P );
```
utzig, line 85: Don't need to free MPI's `z` and `d` here?

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -12,10 +12,18 @@
...
+#if (defined(MCUBOOT_USE_MBED_TLS) + \
+     defined(MCUBOOT_USE_TINYCRYPT)) != 1
+    #error "One crypto backend must be defined: either MBED_TLS or TINYCRYPT"
 #endif
 
+#if defined(MCUBOOT_USE_MBED_TLS)
+    #include <mbedtls/ecp.h>
+    #include <mbedtls/ecdh.h>
+
+#    define EC_TAG_INDEX        (65)
```
utzig, line 24: I believe this symbol is not defined for Tinycrypt anymore?

boot/bootutil/src/encrypted.c
```diff
@@ -536,7 +536,7 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
      */
     bootutil_ecdh_p256_init(&ecdh_p256);
 
-    rc = bootutil_ecdh_p256_shared_secret(&ecdh_p256, &buf[EC_PUBK_INDEX], private_key, shared);
+    rc = bootutil_ecdh_p256_shared_secret(&ecdh_p256, buf, private_key, shared);
```
utzig, line 539: `&buf[EC_PUBK_INDEX]` != `buf`, can this work? I don't understand why you are trying to change the structure of the data in the buffer.

utzig (changes requested):

boot/bootutil/src/encrypted.c
BohdanKovalchuk: there is a call to mbedtls_ecp_point_read_binary() inside _shared_secret()

It does a PK check for [0]-element to be 0x04, which seems to be different behavior from tinycrypt

I can suggest to make conditional-compile call(s) here - mbedTLS will take whole buff, tinycrypt from 1st element

utzig: I see, interesting that the library function would check for uncompressed at this place, but then what you should do is to leave the code as it was and instead set `EC_PUBK_INDEX` for MbedTLS to `0` (and leave it as `1` for Tinycrypt).

BohdanKovalchuk: sounds reasonable

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
BohdanKovalchuk: z and d are only initialized with values by mbedtls_mpi_init(), no allocation

boot/bootutil/src/encrypted.c
BohdanKovalchuk: how do you fee about following approach:

- make bootutil_ecdh_p256_shared_secret() accept beginning of buffer, so complete raw TLV will be passed in
- move if (buf[0] != 0x04) under tinycrypt-specific implementation of bootutil_ecdh_p256_shared_secret();
- mbedTLS-specific implementation will not have this, because internal functions did that

BohdanKovalchuk: same approach for bootutil_ecdh_x25519_shared_secret()

and define different EC_PUBK_INDEX  in ecdh_x25519.h, ecdh_ec256.h respectively, so moved out of encrypted.c

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -56,6 +64,72 @@ static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *c
...
+    mbedtls_mpi z;
+    mbedtls_mpi d;
+} bootutil_ecdh_p256_context;
+
+static inline void bootutil_ecdh_p256_init(bootutil_ecdh_p256_context *ctx)
+{
+    mbedtls_mpi_init(&ctx->z);
+    mbedtls_mpi_init(&ctx->d);
+
+    mbedtls_ecp_group_init( &ctx->grp );
```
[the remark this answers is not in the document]
BohdanKovalchuk: fixed those

boot/bootutil/src/encrypted.c
```diff
@@ -438,8 +438,8 @@ boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
 #elif defined(MCUBOOT_ENCRYPT_EC256)
 #    define EXPECTED_ENC_TLV    IMAGE_TLV_ENC_EC256
 #    define EC_PUBK_INDEX       (1)
-#    define EC_TAG_INDEX        (65)
-#    define EC_CIPHERKEY_INDEX  (65 + 32)
+/*  #    define EC_TAG_INDEX        (65) *//**/
+/*  #    define EC_CIPHERKEY_INDEX  (65 + 32)  *//**/
```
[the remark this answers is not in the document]
BohdanKovalchuk: moved defines appropriately

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -56,6 +64,72 @@ static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *c
...
+    mbedtls_ecp_point_init( &ctx->P );
+
+    if (0 != mbedtls_ecp_group_load( &ctx->grp, MBEDTLS_ECP_DP_SECP256R1 )) {
```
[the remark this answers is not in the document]
BohdanKovalchuk: reworked

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
BohdanKovalchuk: fixed this

boot/bootutil/src/encrypted.c
```diff
@@ -527,16 +527,17 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
     }
 
     /* is EC point uncompressed? */
-    if (buf[0] != 0x04) {
-        return -1;
-    }
+//    if (buf[0] != 0x04) {
+//        return -1;
+//    }
```
utzig, line 532: I might have mentioned it before but commented out code is not acceptable, fix the ocurrences and I will review further.

utzig (changes requested):

boot/bootutil/src/encrypted.c
BohdanKovalchuk: cleaned

boot/bootutil/include/bootutil/crypto/ecdh_x25519.h
```diff
@@ -41,7 +43,7 @@ static inline int bootutil_ecdh_x25519_shared_secret(bootutil_ecdh_x25519_contex
     int rc;
     (void)ctx;
 
-    rc = X25519(z, sk, pk);
+    rc = X25519(z, sk, &pk[EC_PUBK_INDEX]);
```
utzig, line 46: Please revert this to use use just `pk`.

boot/bootutil/src/encrypted.c
```diff
@@ -526,17 +523,12 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
...
-    if (buf[0] != 0x04) {
-        return -1;
-    }
-
     /*
      * First "element" in the TLV is the curve point (public key)
      */
     bootutil_ecdh_p256_init(&ecdh_p256);
 
-    rc = bootutil_ecdh_p256_shared_secret(&ecdh_p256, &buf[EC_PUBK_INDEX], private_key, shared);
```
utzig, line 539: Revert to using `&buf[EC_PUBK_INDEX]`

boot/bootutil/src/encrypted.c
```diff
@@ -564,7 +556,7 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
 
     bootutil_ecdh_x25519_init(&ecdh_x25519);
 
-    rc = bootutil_ecdh_x25519_shared_secret(&ecdh_x25519, &buf[EC_PUBK_INDEX], private_key, shared);
+    rc = bootutil_ecdh_x25519_shared_secret(&ecdh_x25519, buf, private_key, shared);
```
utzig, line 559: Revert to using &buf[EC_PUBK_INDEX]

utzig (changes requested):
Looking more closely now, I would prefer to leave `EC_PUBK_INDEX`, `EC_TAG_INDEX` and `EC_CIPHERKEY_INDEX` in `encrypted.c`. This avoids breaking the abstraction in crypto, where the functions are just supposed to receive the data for the EC points, keys, etc, without having to know about inner structure of how those are stored. Also using those defines makes it clear that this buffer contains more data than just one particular "thing".

boot/bootutil/include/bootutil/crypto/hmac_sha256.h
```diff
@@ -76,6 +85,55 @@ static inline int bootutil_hmac_sha256_finish(bootutil_hmac_sha256_context *ctx,
...
+        rc = mbedtls_md_hmac_starts(ctx, key, key_size);
+    }
+    return rc;
+}
+
+static inline int bootutil_hmac_sha256_update(bootutil_hmac_sha256_context *ctx, const void *data, unsigned int data_length)
+{
+    int rc;
+
+    rc = mbedtls_md_hmac_update(ctx, data, data_length);
```
utzig, line 119: You could just return this function call as well.

boot/bootutil/include/bootutil/crypto/hmac_sha256.h
```diff
@@ -76,6 +85,55 @@ static inline int bootutil_hmac_sha256_finish(bootutil_hmac_sha256_context *ctx,
...
+}
+
+static inline int bootutil_hmac_sha256_finish(bootutil_hmac_sha256_context *ctx, uint8_t *tag, unsigned int taglen)
+{
+    int rc;
+    (void)taglen;
+    /*
+     * HMAC the key and check that our received MAC matches the generated tag
+     */
+    rc = mbedtls_md_hmac_finish(ctx, tag);
```
utzig, line 131: No need for `rc`, can just return the function call.

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -43,19 +50,89 @@ static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *c
...
+}
+
+static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *ctx, const uint8_t *pk, const uint8_t *sk, uint8_t *z)
+{
+    int rc;
+
+    rc = mbedtls_ecp_point_read_binary(&ctx->grp,
+                                       &ctx->P,
+                                       pk,
+                                       EC_TAG_OFFSET);
```
utzig, line 107: `EC_TAG_OFFSET` has no meaning here, this should be called `EC_PUBK_LEN`, or `EC256_PUBK_LEN`.

utzig (changes requested):
A few left-overs. Please squash the commits and rebase on latest `master`.

boot/bootutil/include/bootutil/crypto/hmac_sha256.h
```diff
@@ -76,6 +85,48 @@ static inline int bootutil_hmac_sha256_finish(bootutil_hmac_sha256_context *ctx,
...
+}
+
+static inline int bootutil_hmac_sha256_set_key(bootutil_hmac_sha256_context *ctx, const uint8_t *key, unsigned int key_size)
+{
+    int rc;
+
+    rc = mbedtls_md_setup(ctx, mbedtls_md_info_from_string("SHA256"), 1);
+    if (rc == 0) {
+        rc = mbedtls_md_hmac_starts(ctx, key, key_size);
+    }
```
utzig, line 111: One final nitpick would be to change this block from checking success to the "ïf fail, return early" typical pattern used through-out the code.

utzig (commented):
But how do we know this implementation works?

boot/bootutil/include/bootutil/crypto/hmac_sha256.h
BohdanKovalchuk: I prefer to do not use early or multiple returns in a functions as it is against MISRA.

Will fix if it is essential for the project.

BohdanKovalchuk: I run a smoke test on an actual HW each time before commit.

mbedTLS which is a platform-agnostic SW library so should work in any case.

utzig: > I prefer to do not use early or multiple returns in a functions as it is against MISRA.
> 
> Will fix if it is essential for the project.

I honestly did not know, sad. Anyway good taste would implicate in using the same style that is already employed, not whatever style one came up with because of this or that, but I'll leave it at your discression.

> I run a smoke test on an actual HW each time before commit.
> 
> mbedTLS is a platform-agnostic SW library so should work in any case.

Well, I don't want to sit here and test it every time someone changes the code, also I don't want you to do it, so I guess my question was a bit of a tricky question for which your correct answer would have been something like: "Ha, I got this neat idea, why don't we run it on the simulator so we know it's always working?", for which I would answer: "Awesome idea! Add a new commit doing this!".

d3zd3z: > I prefer to do not use early or multiple returns in a functions as it is against MISRA.

It's not actually against MISRA. Not having early returns, and a goto to the end of the function are only recommendations, not requirements (for example, a requirement is to not have gotos to arbitrary parts of the code).

Computer science has learned quite a bit since 1998. One of those things is that early return, or a goto for cleanup results in cleaner, and more understandable code than code that uses what ends up being a convoluted set of if statements to try and work around that kind of restriction. There are numerous projects that follow this, including the Linux kernel, and even the Zephyr project (which seeks MIRSA compliance). Several newer programming languages (Rust and Go, specifically) encourage the use of early returns to make for clearer code.

So, yes, please use early return, or a goto to cleanup code if needed. It makes the code easier to understand, and will not be any kind of hindrance to MISRA compliance, if desired.

boot/cypress/MCUBootApp/keys.c
```diff
@@ -174,3 +174,23 @@ struct bootutil_key bootutil_keys[1] = {
 };
 const int bootutil_key_cnt = 1;
 #endif /* !MCUBOOT_HW_KEY */
+
+unsigned char enc_priv_key[] = {
```
d3zd3z, line 178: Can you elaborate on where this private key came from? I'm curious why this is different than the equivalent value used on other platforms.

boot/cypress/keys/enc-ec256-priv.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
d3zd3z, line 1: Does this need to be a separate copy of the same key that is in the root directory of the project? Can the build files just refer to that file instead of another copy?

d3zd3z (changes requested):
There seem to be a mixture of two things in this commit. One is adding support for using MBED TLS for the encryption. The other is adding this support to the Cypress platform. Ideally, we'd like to be able to enable this support on at least one other platform (say Zephyr). We should also get the sim to be able to build with this configuration (I don't think additional tests need to be written, as it does the same thing, but this would be an important test of both building and running this code.

From this, I envision 4 commits. One that adds the Mbed TLS support to the encryption code. Another that adds cypress build support, another that adds Zephyr build support, and an additional one that adds sim support. Let us know if you need help with the sim support.

boot/cypress/keys/enc-ec256-priv.pem
BohdanKovalchuk: yes, you're right - it should refer to common folder

boot/cypress/MCUBootApp/keys.c
BohdanKovalchuk: those are same as enc-ec256-*.pem

removed duplication inside \cypress-folder

boot/bootutil/include/bootutil/crypto/hmac_sha256.h
BohdanKovalchuk: did not have an intent to argue or insist, just explained the motivation

surely, updated to match overall style

d3zd3z:
We will also have to resolve the conflicts with 918. I may look into pulling the first commit of this series into #918 as that may be easier than trying to resolve this after the fact.

romanjoe:
Rebased this PR to latest changes in master, as well as fixes for cypress in https://github.com/mcu-tools/mcuboot/pull/957

romanjoe:
@d3zd3z could you please review/comment on this one again since #918 is merged now and this PR is rebased to it, what needs to be done here if needed - we are dependent on this one from https://github.com/mcu-tools/mcuboot/pull/962

d3zd3z (approved):

utzig (approved):
LGTM

PR #885, opened 2020-11-21
boot: zephyr: always clear SPLIM registers when present

hakonfam:
This fixes a bug where zephyr images would not boot correctly
on M33 cores.

PR-847 (commit 70af708b85b3) made clearing SPLIM registers
configurable. The registers would not be cleared if the architecture
had support for cleaning the state of the core before booting.

However, if the SPLIM register is not cleared before booting the
next image that image will get a fault when jumping to the routine
which cleans the core.

In other words we must clear the SPLIM registers when it is present.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -152,6 +152,8 @@ static void do_boot(struct boot_rsp *rsp)
     z_arm_clear_arm_mpu_config();
 #endif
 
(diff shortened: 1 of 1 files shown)


nvlsianpu (approved):

hakonfam:
This will be resolved in zephyr instead.

https://github.com/zephyrproject-rtos/zephyr/pull/30190

PR #888, opened 2020-11-25
Post v1.7.0 action

nvlsianpu:
Switch version info to 1.8.0-dev


nvlsianpu:
@utzig I believe no sense of upgrade imgtool version - this will trigger unwanted pip release right?

utzig:
> @utzig I believe no sense of upgrade imgtool version - this will trigger unwanted pip release right?

Yes, we can eventually do some alpha/beta for relevant new stuff, but no point in this particular case. Also I think you don't need to update the Mynewt metadata at the moment.

utzig (approved):

PR #889, opened 2020-11-27
zephyr: use minimal CBPRINTF implementation

nvlsianpu:
Addition of cbprintf capability in zephyr
https://github.com/zephyrproject-rtos/zephyr/pull/29876
cause flash footprint rise by ~4 KB. Selecting
CBPRINTF_NANO=y mitigates the footprint rise.

### DIFF
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -33,3 +33,5 @@ CONFIG_FLASH=y
 CONFIG_LOG=y
 ### Ensure Zephyr logging changes don't use more resources
 CONFIG_LOG_DEFAULT_LEVEL=0
+### Decrease footprint by ~4 KB in comparison to CBPRINTF_COMPLETE=y
+CONFIG_CBPRINTF_NANO=y


nvlsianpu:
cc @pabigo

utzig (approved):

tejlmand (approved):

boot/zephyr/prj.conf
```diff
@@ -33,3 +33,4 @@ CONFIG_FLASH=y
 CONFIG_LOG=y
 ### Ensure Zephyr logging changes don't use more resources
 CONFIG_LOG_DEFAULT_LEVEL=0
+CBPRINTF_NANO=y
```
tejlmand, line 36:
```suggestion
CONFIG_CBPRINTF_NANO=y
```

tejlmand (changes requested):

tejlmand (approved):

boot/zephyr/prj.conf
utzig: @tejlmand Would it be too hard to enable the build system to prefix options with `CONFIG_` if it's not already there?

nvlsianpu: fixed

I have been testing a few version of the modification before, so in finally i commit patch with obvious bug.
It is a pity that not possible `imply` Kconfig` bool property while it is a `choice` part.

pabigot (approved):
This looks fine, though I'm confused by the statement that `CBPRINTF_NANO=y` increases the footprint by 1 KiBy.  Relative to before cbprintf was added?  Because for that case the footprint should be essentially unchanged, or even be reduced (nano is the same as the old printk, but also replaces minimal libc printf, if anything is using libc printf capabilities).

nvlsianpu:
^^ fine, will correct the message

boot/zephyr/prj.conf
tejlmand: > @tejlmand Would it be too hard to enable the build system to prefix options with `CONFIG_` if it's not already there?

No, but I don't like to start anticipating what user intended if he forgets `CONFIG_`.

Could be a simple mistake like: `COFNIG_x` and then it would be wrong to prepend `CONFIG_` also.

After all, the following error is printed with invalid conf files:
```
<path>/prj.conf:2: warning: ignoring malformed line 'CBPRINTF_NANO=y'

error: Aborting due to Kconfig warnings
```

PR #892, opened 2020-12-01
zephyr: extracted bootutil_public library

nvlsianpu:
Part of code of boot/bootutil/ is re-implemented in zephyr-rtos
repository.
As some code are defined here and there it becomes problem when
need to include it with outstanding feature in a build.
It is possible to mitigate problem using #fdefry - but this was
rather temporary hack.

This patch introduce new module which is common for MCUBoot build
and application build.

[DNM] - Do not merge. I will merge this in synch with zephyr adaptation patch https://github.com/zephyrproject-rtos/zephyr/pull/30370


boot/bootutil/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  ../zephyr/include
+)
+
+zephyr_library()
+zephyr_library_sources(
+  src/bootutil_misc.c
+    )
+zephyr_library_link_libraries(MCUBOOT_BOOTUTIL)
+target_link_libraries(MCUBOOT_BOOTUTIL INTERFACE zephyr_interface)
+endif()
```
d3zd3z, line 20: This file is missing a newline at the end.

boot/zephyr/CMakeLists.txt
```diff
@@ -109,10 +109,12 @@ zephyr_library_sources(
   ${BOOT_DIR}/bootutil/src/image_rsa.c
   ${BOOT_DIR}/bootutil/src/image_ec256.c
   ${BOOT_DIR}/bootutil/src/image_ed25519.c
-  ${BOOT_DIR}/bootutil/src/bootutil_misc.c
```
d3zd3z, line 112: I don't think it makes sense to move all of this functionality into a library, especially one shared in other places.  Do we know what functions actually need to be used by the targeted image? I' thinking of something like `bootutil_buffer_is_erased` where the flash operations on the guest os might not even be the same.

d3zd3z (changes requested):

boot/zephyr/CMakeLists.txt
nvlsianpu: Ok, I can extract then these a few common functions to a common file and share only this file.

nvlsianpu: `bootutil_buffer_is_erased()` uses OS primitives, so see no problem with it.
looks like only a few function are mcuboot's only:
boot_find_status()
boot_read_swap_size()
boot_write_swap_size(
boot_enc_key_off()
boot_write_enc_key()
boot_read_enc_key()
boot_write_copy_done()

I would move them to separate file.

d3zd3z (approved):
I'm ok with this as is. We just need to keep in mind that this now becomes part of our public API, and we have to be careful making any changes to it.

utzig:
Depends on #902

ci/fih_test_docker/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN \
 # Clone TF-M and dependencies
 RUN mkdir -p /root/work/tfm &&\
     cd /root/work/tfm  &&\
-    git clone https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git -b TF-Mv1.2-RC1 &&\
```
nvlsianpu, line 42: @utzig , @RcColes What I am supposed to do with this line now?

utzig: I assume wait for @RcColes to submit a PR with the new trusted-firmware-m tag!

RcColes: Yes, just working on getting the new TF-M commit hash now. Ideally should be by end of today. Do you think it'd be better to have it as a separate commit/PR or is it fine to just add it in here?

utzig: > Do you think it'd be better to have it as a separate commit/PR or is it fine to just add it in here?

A separate PR would be better, but both are OK IMO.

RcColes: PR created at #903

boot/zephyr/CMakeLists.txt
```diff
@@ -113,6 +113,9 @@ zephyr_library_sources(
   ${BOOT_DIR}/bootutil/src/fault_injection_hardening.c
   )
 
+# library which might be common sourec code for MCUBoot and an application
```
utzig, line 116: typo: sourec -> source

utzig (approved):
LGTM

utzig:
Btw, while looking at the Zephyr complementary PR I started to wonder if a public library/API should not be prefixed with `mcuboot_` instead of `boot_`, although I think this one is OK for now, because changing the naming would involve checking and fixing all ports, as well as mcumgr...

nvlsianpu:
I would keep the naming for now.

utzig:
> although I think this one is OK for now

yes, that's what "although I think this one is OK for now" means! :-)

utzig:
> I would keep the naming for now.

yes, that's what "although I think this one is OK for now" means! :-)

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -0,0 +1,184 @@
...
+ * @brief Read the image swap state
+ *
+ * @param flash_area_id Id of flash parttition from which trailer will be read.
+ * @param state Struture for holding swap state.
+ *
+ * @return 0 on success, nonzero errno code on fail.
+ */
+int
+boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state);
+
```
Sherryzhang2, line 174: Considering the OTA case, the tfm firmware update partition needs to get the image version information. This can be achieved by reading the shared memory between the bl2 and the tfm, but I think adding such an interface like boot_query_image_info() in bl2 is more directly and user friendly to get the image version or image hash for the tfm partitions. Also some users download and writes the image directly in the nonsecure side, in this case, the nonsecure does not have access to the shared memory of tfm and bl2, so it is necessary to add such image query API.

Sherryzhang2: Also, I think, the public API like boot_write_secondary_image or simply flash_area_write should be added to the public library. These APIs are used by the firmware update service in TF-M or the nonsecure firmware update application. The firmware update service, as well as the nonsecure firmware update application, should not all the flash_area_write in the mcuboot src, instead, it should call the public library directly which should cover all the functionality for the firmware update.
If this PR is just to extract the currently existing public APIs, I can create a new PR to add the 'query' and 'write' functions to the public library.

nvlsianpu: Yes: this PR is mainly for extract the current existing public APIs.
I'm not sure why do you mean by `flash_area_write` here. flash_area_write() is on of the porting functions.
https://github.com/mcu-tools/mcuboot/blob/master/docs/PORTING.md

`boot_write_secondary_image()` Is this function for collecting the image to the second image bank? Such functionalities are supposed (and are) implemented by the application. See no point to implement this here.

Sherryzhang2: Make sense. It seems the flash_area_write and other flash operations are implemented in TF-M, not in the mcuboot repo.

Sherryzhang2 (approved):
Looks good to me.

de-nordic (approved):

ci/fih_test_docker/Dockerfile
nvlsianpu: @RcColes Looks like after merging that Travis on master failed as this PR CI before fix.
Can you help with that?

PR #894, opened 2020-12-02
Add reference counting to Mbed OS flash backend

AGlass0fMilk:
The internal functions of mcuboot are not consistent in opening/closing flash areas and sometimes nested calls happen to `flash_area_open` and `flash_area_close`. With the previous implementation, a nested call to `flash_area_close` would deinitialize the underlying `BlockDevice`. This could cause subsequent flash operations on an "open" flash area to fail.

This PR adds a simple open counter for each flash area and ensures the underlying `BlockDevice` is initialized and deinitialized appropriately. The `BlockDevice` is only initialized when transitioning from an open count of 0 to 1. The `BlockDevice` is only deinitialized when the open count falls to 0.

### DIFF
--- a/boot/mbed/src/flash_map_backend.cpp
+++ b/boot/mbed/src/flash_map_backend.cpp
@@ -53,6 +53,8 @@ static mbed::BlockDevice* flash_map_bd[FLASH_AREAS] = {
 
 static struct flash_area flash_areas[FLASH_AREAS];
 
+static unsigned int open_count[FLASH_AREAS] = {0};
+
 int flash_area_open(uint8_t id, const struct flash_area** fapp) {
 
     *fapp = &flash_areas[id];
@@ -77,17 +79,40 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
             return -1;
     }
 
+    open_count[id]++;
+    MCUBOOT_LOG_DBG("flash area %d open count: %d (+)", id, open_count[id]);
+
     fap->fa_id = id;
     fap->fa_device_id = 0; // not relevant
 
     mbed::BlockDevice* bd = flash_map_bd[id];
     fap->fa_size = (uint32_t) bd->size();
-    return bd->init();
+
+    /* Only initialize if this isn't a nested call to open the flash area */
+    if (open_count[id] == 1) {
+        MCUBOOT_LOG_DBG("initializing flash area %d...", id);
+        return bd->init();
+    } else {
+        return 0;
+    }
 }
 
 void flash_area_close(const struct flash_area* fap) {
-    mbed::BlockDevice* bd = flash_map_bd[fap->fa_id];
-    bd->deinit();
+    uint8_t id = fap->fa_id;
+    /* No need to close an unopened flash area, avoid an overflow of the counter */
+    if (!open_count[id]) {
+        return;
+    }
+
+    open_count[id]--;
+    MCUBOOT_LOG_DBG("flash area %d open count: %d (-)", id, open_count[id]);
+    if (!open_count[id]) {
+        /* mcuboot is not currently consistent in opening/closing flash areas only once at a time
+         * so only deinitialize the BlockDevice if all callers have closed the flash area. */
+        MCUBOOT_LOG_DBG("deinitializing flash area block device %d...", id);
+        mbed::BlockDevice* bd = flash_map_bd[id];
+        bd->deinit();
+    }
 }
 
 /*
@@ -96,8 +121,8 @@ void flash_area_close(const struct flash_area* fap) {
 int flash_area_read(const struct flash_area* fap, uint32_t off, void* dst, uint32_t len) {
     mbed::BlockDevice* bd = flash_map_bd[fap->fa_id];
 
-    // Note: The address must be aligned to bd->get_read_size(). If MCUBOOT_READ_GRANULARITY
-    // is defined, the length does not need to be aligned.
+    /* Note: The address must be aligned to bd->get_read_size(). If MCUBOOT_READ_GRANULARITY
+       is defined, the length does not need to be aligned. */
 #ifdef MCUBOOT_READ_GRANULARITY
     uint32_t read_size = bd->get_read_size();
     if (read_size == 0) {
@@ -121,7 +146,7 @@ int flash_area_read(const struct flash_area* fap, uint32_t off, void* dst, uint3
         else {
             int ret = bd->read(dst, off, len);
             if (ret != 0) {
-                MCUBOOT_LOG_ERR("Read failed: fa_id %d offset 0x%x len 0x%x", fap->fa_id, off, len);
+                MCUBOOT_LOG_ERR("Read failed: fa_id %d offset 0x%x len 0x%x (%d)", fap->fa_id, off, len, ret);
                 return ret;
             }
         }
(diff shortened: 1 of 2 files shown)


AGlass0fMilk:
@trowbridgec Maybe this is related to issues you were seeing?

utzig:
> The internal functions of mcuboot are not consistent in opening/closing flash areas and sometimes nested calls happen to `flash_area_open` and `flash_area_close`.

What functions are inconsistent? Wouldn't it be better to just make them consistent then?

AGlass0fMilk:
> What functions are inconsistent? Wouldn't it be better to just make them consistent then?

I added trace calls to the `flash_area_open` and `flash_area_close` implementations and got the following output on a typical bootup:

```
[INFO][BL]: Starting MCUboot                                                                                                                    
[DBG ][MCUb]: flash area 0 open count: 1 (+)                                                                                                    
[DBG ][MCUb]: initializing flash area 0...                                                                                                      
[DBG ][MCUb]: flash area 1 open count: 1 (+)                                                                                                    
[DBG ][MCUb]: initializing flash area 1...                                                                                                      
[DBG ][MCUb]: flash area 2 open count: 1 (+)                                                                                                    
[DBG ][MCUb]: initializing flash area 2...                                                                                                      
[DBG ][MCUb]: flash area 0 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 0 open count: 1 (-)                                                                                                    
[DBG ][MCUb]: flash area 1 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 1 open count: 1 (-)                                                                                                    
[DBG ][MCUb]: flash area 0 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 0 open count: 1 (-)                                                                                                    
[DBG ][MCUb]: flash area 2 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 2 open count: 1 (-)                                                                                                    
[INFO][MCUb]: Primary image: magic=good, swap_type=0x2, copy_done=0x1, image_ok=0x3                                                             
[INFO][MCUb]: Scratch: magic=unset, swap_type=0x1, copy_done=0x3, image_ok=0x3                                                                  
[INFO][MCUb]: Boot source: none                                                                                                                 
[DBG ][MCUb]: flash area 0 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 0 open count: 1 (-)                                                                                                    
[DBG ][MCUb]: flash area 1 open count: 2 (+)                                                                                                    
[DBG ][MCUb]: flash area 1 open count: 1 (-)                                                                                                    
[INFO][MCUb]: Swap type: test
```
In many places throughout mcuboot's internal code there are situations where a given flash area will be opened by multiple nested calls at the same time. I'm not sure how other operating systems handle this. I initially missed this because for some `BlockDevice` implementations `init` and `deinit` will not cause problems reading/writing to/from flash. I recently started working with a BlockDevice that actually puts the underlying flash device into sleep mode when `deinit` is called and it was causing problems with mcuboot's update process.

It's not a bug unless this is unintended operation of mcuboot (opening a flash area multiple times). It should be clearly documented in the flash backend porting documentation if this behavior should be expected.

I was originally going to open an issue but I decided to just implement this fix instead. I decided to apply this patch because it seems like a large effort to rearrange the mcuboot code itself to ensure that a flash area is only opened by one caller at a time. This is a relatively simple fix with minimal RAM/ROM implications.

I _did_ want to start the conversation regarding expected flash backend behavior though :grin:

d3zd3z:
It is probably worth spending a little time figuring out if the nested usage in the current code is difficult to fix. Otherwise, I think this is a reasonable thing to do. Another thing that could result from this, would be to add something similar to the sim, along with a way to query that the counter makes it to zero on all normal exit paths.

Really, what this points out is that on most platforms, flash_area_close doesn't actually do anything. So, this is good to be aware of.

utzig:
> It is probably worth spending a little time figuring out if the nested usage in the current code is difficult to fix. Otherwise, I think this is a reasonable thing to do. Another thing that could result from this, would be to add something similar to the sim, along with a way to query that the counter makes it to zero on all normal exit paths.

I added a reference counting to the sim and it does not track the issues because all non-reset tests end up with a matching open/close. I think the main reason is on the sim everything behaves, so there are no error cases, etc. Also I fixed a some issues with closes after an error in the open, and open with no matching close; just have to test it a bit before submitting a PR.

AGlass0fMilk:
I think this is suitable as a stop-gap patch until the nested calls can be fixed.

To be clear though: The reference counters _do_ return to 0 at the end of mcuboot's operation. So from that standpoint mcuboot does a fine job ultimately closing all flash areas. It's just that there may be nested calls where a flash area is opened several times and should not be "truly closed" when closed just once.

The "correct" fix would probably be to add parameters to internal functions so they may pass around an open flash area handle by reference.

Is it worth fixing just so it aligns with the "de facto" concept of opening file handles only once? Maybe, since we want to maintain the reputation of this codebase :smile:  Or could we just document this behavior and/or add reference counting to mcuboot itself?

Depends on how much work we want to do. Unfortunately, I'm too busy at the moment to do refactoring.

It's interesting this hasn't been an issue for any other operating systems. 

> Really, what this points out is that on most platforms, flash_area_close doesn't actually do anything. So, this is good to be aware of.

I agree.

d3zd3z:
>  Or could we just document this behavior and/or add reference counting to mcuboot itself?

It is probably worth at least documenting, or figuring out the right behavior. But, so far Mbed OS seems to be the only platform where this has mattered, so it makes sense to start with this fix there.

nvlsianpu (commented):
Patch looks fine to me.

AGlass0fMilk:
> Patch looks fine to me.

Let's get that merge!

nvlsianpu (approved):

nvlsianpu:
@d3zd3z Have you any objection to merge this?

boot/mbed/src/flash_map_backend.cpp
```diff
@@ -77,17 +79,40 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
...
+        return;
+    }
+
+    open_count[id]--;
+    MCUBOOT_LOG_DBG("flash area %d open count: %d (-)", id, open_count[id]);
+    if (!open_count[id]) {
+        /* mcuboot isn't very consistent on opening/closing flash areas
+         * so only deinitialize the BlockDevice if all callers have closed the flash area. */
+        MCUBOOT_LOG_DBG("deinitializing flash area block device %d...", id);
+        mbed::BlockDevice* bd = flash_map_bd[fap->fa_id];
```
utzig, line 113: Can use `id` here instead of `fap->fa_id`

boot/mbed/src/flash_map_backend.cpp
```diff
@@ -77,17 +79,40 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
...
+
     fap->fa_id = id;
     fap->fa_device_id = 0; // not relevant
 
     mbed::BlockDevice* bd = flash_map_bd[id];
     fap->fa_size = (uint32_t) bd->size();
-    return bd->init();
+
+    /* Only initialize if this isn't a nested call to open the flash area */
+    if(open_count[id] == 1) {
```
utzig, line 92: Please add a space between `if`  and `(` for consistency

utzig (changes requested):

utzig:
@AGlass0fMilk Please update https://github.com/mcu-tools/mcuboot/blob/master/docs/PORTING.md#flash-map including a small description of this issue and that it might need to be worked around.

AGlass0fMilk:
> @AGlass0fMilk Please update https://github.com/mcu-tools/mcuboot/blob/master/docs/PORTING.md#flash-map including a small description of this issue and that it might need to be worked around.

Done:

> **Note:** As of writing, it is possible that mcuboot will open a flash area multiple times simultaneously (through nested calls to `flash_area_open`). As a result, mcuboot may call `flash_area_close` on a flash area that is still opened by another part of mcuboot. As a workaround when porting, it may be necessary to implement a counter of the number of times a given flash area has been opened by mcuboot. The `flash_area_close` implementation should only truly deinitialize the underlying flash area when the open counter is decremented to 0. See [this GitHub PR](https://github.com/mcu-tools/mcuboot/pull/894/) for a more detailed discussion.

utzig (approved):
LGTM

d3zd3z (approved):

PR #895, opened 2020-12-02
boot: zephyr: Default to LOG_MINIMAL

kyberdin:
Enables LOG_MINIMAL in the default build.

Ref: NCSDK-7206

Signed-off-by: Stephen Stauts <stephen.stauts@nordicsemi.no>

### DIFF
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -31,6 +31,7 @@ CONFIG_FLASH=y
 # CONFIG_I2C is not set
 
 CONFIG_LOG=y
+CONFIG_LOG_MINIMAL=y
 ### Ensure Zephyr logging changes don't use more resources
 CONFIG_LOG_DEFAULT_LEVEL=0
 ### Decrease footprint by ~4 KB in comparison to CBPRINTF_COMPLETE=y


boot/zephyr/prj.conf
```diff
@@ -31,6 +31,7 @@ CONFIG_FLASH=y
 # CONFIG_I2C is not set
 
 CONFIG_LOG=y
+CONFIG_LOG_MINIMAL=y
```
utzig, line 34: What is the rationale? What changes using this config? Wouldn't it be better to add a reason to the commit message?

kyberdin: Yes, I should have put more info in the commit message. Sorry about that - I will update the commit message.

The rationale is that it saves us ~2-4kB of flash usage, depending on the device, without any cost to the end user. We ran the change across CI without any breaks.

utzig: Thanks! But if there is no cost to the end user why would anyone ever have to enable the option instead of it just being the default in Zephyr? :-P

kyberdin: Sorry if I am slow today, but I'm not sure that I follow. This change seeks to make it the default for Zephyr, as the current default uses up 2kB+ of memory that it doesn't need to in order to use MCUBoot with logging functionality. So without any cost to the end user, this change is a benefit and opens up more free space for its partition.

utzig: The change is fine to me, but my question could be rephrased "Why would anyone ever not enable this option since there is no loss of functionality?"

kyberdin: I see, thanks for clarifying. I should also clarify my "loss" for the user being in terms of significant feature loss or breaks in the default configuration, e.g. the ability to use logging macros for output in the bootloader.

The `kconfig` documentation of `LOG_MINIMAL` outlines what is no longer present by enabling it:

> Hexdump macros are also supported, with a small
	  amount of code pulled in if used. Build time filtering is supported,
	  but not runtime filtering. There are no timestamps, prefixes,
	  colors, or asynchronous logging, and all messages are simply
	  sent to printk().

It could be argued that the above are nice-to-haves rather than hard requirements for logging output. In that case, it would make sense to have their inclusion be a choice by the end user, while defaulting to a state that reduces the bootloader's memory footprint.

So to answer your question, an end user would not enable this option if they prefer a bootloader with nice-to-have logging features over minimum output functionality using the logging macros.

utzig (approved):

nvlsianpu (approved):

PR #896, opened 2020-12-08
boot: zephyr: cleanup NXP MPU configuration before boot

henrikbrixandersen:
Cleanup the NXP MPU configuration before chain-loading the application.

Signed-off-by: Henrik Brix Andersen <hebad@vestas.com>

### DIFF
--- a/boot/zephyr/arm_cleanup.c
+++ b/boot/zephyr/arm_cleanup.c
@@ -5,6 +5,11 @@
  */
 
 #include <arch/arm/aarch32/cortex_m/cmsis.h>
+#include <toolchain.h>
+
+#if CONFIG_CPU_HAS_NXP_MPU
+#include <fsl_sysmpu.h>
+#endif
 
 void cleanup_arm_nvic(void) {
 	/* Allow any pending interrupts to be recognized */
@@ -31,6 +36,20 @@ __weak void z_arm_clear_arm_mpu_config(void)
 
 	for (i = 0; i < num_regions; i++) {
 		ARM_MPU_ClrRegion(i);
+	}
+}
+#elif CONFIG_CPU_HAS_NXP_MPU
+__weak void z_arm_clear_arm_mpu_config(void)
+{
+	int i;
+
+	int num_regions = FSL_FEATURE_SYSMPU_DESCRIPTOR_COUNT;
+
+	SYSMPU_Enable(SYSMPU, false);
+
+	/* NXP MPU region 0 is reserved for the debugger */
+	for (i = 1; i < num_regions; i++) {
+		SYSMPU_RegionEnable(SYSMPU, i, false);
 	}
 }
 #endif
--- a/boot/zephyr/include/arm_cleanup.h
+++ b/boot/zephyr/include/arm_cleanup.h
@@ -13,7 +13,7 @@
  */
 void cleanup_arm_nvic(void);
 
-#if defined(CONFIG_CPU_HAS_ARM_MPU)
+#if defined(CONFIG_CPU_HAS_ARM_MPU) || defined(CONFIG_CPU_HAS_NXP_MPU)
 /**
  * Cleanup all ARM MPU region configuration
  */
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -148,7 +148,7 @@ static void do_boot(struct boot_rsp *rsp)
     SCB_DisableICache();
 #endif
 
-#if CONFIG_CPU_HAS_ARM_MPU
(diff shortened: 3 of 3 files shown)


boot/zephyr/arm_cleanup.c
```diff
@@ -33,4 +38,18 @@ __weak void z_arm_clear_arm_mpu_config(void)
 		ARM_MPU_ClrRegion(i);
 	}
 }
+#elif CONFIG_CPU_HAS_NXP_MPU
+__weak void z_arm_clear_arm_mpu_config(void)
```
utzig, line 42: Wouldn't it make more sense to have this function in the Zephyr tree? Is `CONFIG_CPU_HAS_ARM_MPU` also defined to NXP?

henrikbrixandersen: Yes, that could make sense - but the same could be set for the existing `z_arm_clear_arm_mpu_config()` in the same file? `CONFIG_CPU_HAS_ARM_MPU` is not defined for the NXP MPUs.

utzig: > Yes, that could make sense - but the same could be said for the existing `z_arm_clear_arm_mpu_config()` in the same file?

Which is itself a copy of https://github.com/zephyrproject-rtos/zephyr/blob/master/arch/arm/core/aarch32/cortex_m/scb.c#L47, I assume.

utzig (approved):
LGTM

boot/zephyr/arm_cleanup.c
nvlsianpu: z_arm_clear_arm_mpu_config() is implemented here as I encounter  the case wen mcuboot have to chainnload a legacy none-zephyr application.

nvlsianpu: So normally chain loaded application should does MPU cleanup. @henrikbrixandersen Shouldn't this be moved to the app? Or you encountered similar case as I had?

henrikbrixandersen: No. The issue, I was facing, was a Zephyr-based MCUboot built with `CONFIG_HW_STACK_PROTECTION=y` where the application would sometimes (depending on the stack sizes used in the application) crash with an MPU fault even before reaching the NXP MPU setup code of the application image.

Disabling the MPU and clearing the MPU regions prior to booting the application solves this issue.

nvlsianpu: Name which match zephyr's analog function for NXP mpu be used here instead of `z_arm_clear_arm_mpu_config`. Otherwise this function will be overridden by implementation dedicated to ARM MPU.

nvlsianpu (changes requested):

boot/zephyr/arm_cleanup.c
nvlsianpu: BTW - there is no a NXP MPU clear function in the zephyr-rtos right now.

nvlsianpu: I would change function name name for reflecting that it works on NXP mpu.

ioannisg (commented):
Does MCUboot have any way to implement SoC-specific functionality outside the mainstream ARM architecture core functionality? If we don't, then, I guess this approach should be acceptable.

This is something we would need to do in Zephyr as well.

boot/zephyr/arm_cleanup.c
ioannisg: > No. The issue, I was facing, was a Zephyr-based MCUboot built with `CONFIG_HW_STACK_PROTECTION=y` where the chain-loaded application would sometimes (depending on the stack sizes used in the application) crash with an MPU fault even before reaching the NXP MPU setup code of the application image.
> 
> Disabling the MPU and clearing the MPU regions prior to booting the application solves this issue.

@henrikbrixandersen the chain-loaded application could enable CONFIG_INIT_ARCH_HW_AT_BOOT, that would do the clearing during early Zephyr boot. Or are you assuming an non-Zephyr-based chain-loaded firmware?

d3zd3z:
> Does MCUboot have any way to implement SoC-specific functionality outside the mainstream ARM architecture core functionality? If we don't, then, I guess this approach should be acceptable.

I agree that this is something somewhat missing in MCUboot, but I'm unsure how much should be here vs in Zephyr.

In order to be even moderately secure, MCUboot itself needs to be protected from being modified by the application that it runs. How this is done depends a lot on the particular MCU. In some cases, the MPU or equivalent can be configured to protect this. In other cases, there is protection more directly configured into the flash controller, and certain regions can be marked as read-only.

But, I'm find with this change for now, until we come up with some better way to handle this.

d3zd3z (approved):

nvlsianpu (approved):

PR #899, opened 2020-12-14
Update FIH test docker flow on Travis

utzig:
Moved Dockerimage and build script to a separate directory. This should be used to build a new docker image and this image should be pushed to Docker hub. Source repos were removed from the image and only tools remain. The Travis config file was updated to pull and cache the used docker image instead of rebuilding all the time. The running script now clones trusted-firmware-m at the right place (sibiling to mcuboot clone) and runs the docker images mapping the VM's directory to the internal container.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,10 @@ install:
 script:
   - ./ci/${TEST}_run.sh
 
+cache:
+  directories:
+  - docker
+
 notifications:
   slack:
     rooms:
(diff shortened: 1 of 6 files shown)


d3zd3z (approved):
Have we tried any of the docker-based tests in GH actions?

utzig:
> Have we tried any of the docker-based tests in GH actions?

Not yet.

d3zd3z (approved):

PR #900, opened 2020-12-15
bootutil: Fix wrong decryption attempt on invalid images

jamike:
With swap, and encryption active : virgin primary slots on flash
having 0xff erased pattern are detected as encrypted, and install
from secondary slot fails, when primary slot is fully erased

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -91,6 +91,7 @@ boot_read_image_headers(struct boot_loader_state *state, bool require_all,
         struct boot_status *bs)
 {
     int rc;
+    bool hdr_ok = false;
     int i;
 
     for (i = 0; i < BOOT_NUM_SLOTS; i++) {
@@ -100,22 +101,24 @@ boot_read_image_headers(struct boot_loader_state *state, bool require_all,
         {
             rc = boot_read_image_header(state, i, boot_img_hdr(state, i), bs);
         }
-        if (rc != 0) {
-            /* If `require_all` is set, fail on any single fail, otherwise
-             * if at least the first slot's header was read successfully,
-             * then the boot loader can attempt a boot.
-             *
-             * Failure to read any headers is a fatal error.
-             */
-            if (i > 0 && !require_all) {
-                return 0;
-            } else {
-                return rc;
-            }
-        }
+
+        /* Failure to read any headers is a fatal error. */
+        if (rc > 0)
+            return rc;
+
+        /* If `require_all` is set, fail on any single fail */
+        if (rc != 0 && require_all)
+            return rc;
+
+        /*  Header present */
+        if (rc == 0)
+            hdr_ok = true;
     }
 
-    return 0;
+    /* If at least one slot's header was read successfully,
+     * then the boot loader can attempt to install or boot
+     */
+    return  hdr_ok ? 0 : -1 ;
 }
 
 /**
--- a/boot/bootutil/src/swap_move.c
+++ b/boot/bootutil/src/swap_move.c
@@ -104,6 +104,9 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
 
     /* We only know where the headers are located when bs is valid */
     if (bs != NULL && out_hdr->ih_magic != IMAGE_MAGIC) {
+        /*  Image is not valid, unset all flags to avoid further
+         *  unexpected behaviour */
+        out_hdr->ih_flags = 0;
         rc = -1;
         goto done;
     }
--- a/boot/bootutil/src/swap_scratch.c
+++ b/boot/bootutil/src/swap_scratch.c
@@ -73,6 +73,13 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
         rc = BOOT_EFLASH;
     }
 
+    if (out_hdr->ih_magic != IMAGE_MAGIC) {
+       /*  Image is not valid, unset all flags to avoid further
+        *  unexpected behaviour */
+        out_hdr->ih_flags = 0;
+        rc = -1;
+    }
+
     return rc;
 }
 


d3zd3z (approved):

d3zd3z (changes requested):
I'm now beginning to think that this fix might not be in the right place. Really, the code should never be getting to a point where it is checking `IS_ENCRYPTED`, unless it has already decided that the header is valid. Have you found the path through the code that causes this, and I think there is either some earlier check missing, or we need to fix a check that isn't working right.

boot/bootutil/include/bootutil/image.h
```diff
@@ -142,7 +142,7 @@ struct image_tlv {
     uint16_t it_len;    /* Data length (not including TLV header). */
 };
 
-#define IS_ENCRYPTED(hdr) ((hdr)->ih_flags & IMAGE_F_ENCRYPTED)
+#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags & (IMAGE_F_ENCRYPTED)) && (hdr->ih_magic == IMAGE_MAGIC))
```
utzig, line 145: I would suggest trying something like:

```
#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags & IMAGE_F_ENCRYPTED) && ((hdr)->ih_flags != 0xffffffff))
```

d3zd3z: This kind of implies that we won't ever use all of the flag bits. That might be a reasonable thing to do, especially if we decide to make a bit reserved, and indicate it must always be 0.

Additionally, not all flash devices use 0xffffffff as the erased value. Is there a particular reason to not check the magic? Arguably, that should be done first by the code using this, rather than every time the flags are checked, though.

jamike: A possible fix is to clear ih_flags after having read it from flash . A  check of  hdr magic is added in function boot_read_image_header from swap_move.c or swap_scratch.c, if magic is not present all ih_flags are cleared.

jamike: Updated patch fixes installation of encrypted file in secondary slot to an empty primary slot in SWAP_USING_MOVE  and SWAP_USING_SCRATCH

boot/bootutil/src/swap_move.c
```diff
@@ -104,8 +104,9 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
 
     /* We only know where the headers are located when bs is valid */
     if (bs != NULL && out_hdr->ih_magic != IMAGE_MAGIC) {
-        rc = -1;
-        goto done;
+        /*  image is not valid, unset all flags to avoid futher
+        *  unexpected behaviour */
```
gustavonihei, line 108:
```suggestion
         *  unexpected behaviour */
```
nit: fixing code alignment

boot/bootutil/src/swap_move.c
```diff
@@ -104,8 +104,9 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
 
     /* We only know where the headers are located when bs is valid */
     if (bs != NULL && out_hdr->ih_magic != IMAGE_MAGIC) {
-        rc = -1;
-        goto done;
+        /*  image is not valid, unset all flags to avoid futher
```
gustavonihei, line 107:
```suggestion
        /*  image is not valid, unset all flags to avoid further
```
Typo in `further`

boot/bootutil/src/swap_scratch.c
```diff
@@ -73,6 +73,12 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
         rc = BOOT_EFLASH;
     }
 
+    if (out_hdr->ih_magic != IMAGE_MAGIC) {
+       /*  image is not valid, unset all flags to avoid futher
```
gustavonihei, line 77:
```suggestion
       /*  Image is not valid, unset all flags to avoid further
```
nit: Typo in `further` and capital initial on `Image`.

boot/bootutil/src/swap_move.c
```diff
@@ -104,8 +104,9 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
 
     /* We only know where the headers are located when bs is valid */
     if (bs != NULL && out_hdr->ih_magic != IMAGE_MAGIC) {
-        rc = -1;
```
nvlsianpu, line 107: why -1 is not returned anymore?

boot/bootutil/src/loader.c
```diff
@@ -107,7 +107,7 @@ boot_read_image_headers(struct boot_loader_state *state, bool require_all,
              *
              * Failure to read any headers is a fatal error.
              */
-            if (i > 0 && !require_all) {
+            if (rc < 0 && !require_all) {
```
gustavonihei, line 110: Could you please explain why this change is required?
I understand the need to check for a negative `rc`, but considering the above comment, the check `i > 0` seems to be required.

boot/bootutil/src/swap_move.c
```diff
@@ -104,7 +104,10 @@ boot_read_image_header(struct boot_loader_state *state, int slot,
...
+        /*  Image is not valid, unset all flags to avoid further
+         *  unexpected behaviour */
+        out_hdr->ih_flags = 0;
+        rc=-1;
```
gustavonihei, line 110:
```suggestion
        rc = -1;
```

boot/bootutil/src/swap_move.c
gustavonihei: So you've added the `rc = -1` back to the Move implementation. Shouldn't Swap Scratch also return `-1` in this scenario?

boot/bootutil/src/loader.c
jamike: With i >0 an error is return if primary slot has no valid header, it prevents from parsing the secondary slot header, and installing an image on a secondary slot. Your remark is valid, and I have  another fix that I should update soon, there is more changes... Let's see automatic test result and then the review comment.

PR #901, opened 2020-12-15
Enable support for building mcuboot for Mbed with direct-xip

AGlass0fMilk:
This commit introduces changes that allow users to build for other non-swap type update methods (overwrite only, swap using move, direct xip, or RAM loading). Changes include:

- Adding configuration options relating to XIP
- Updating the Mbed flash map backend to be compatible with XIP updates
- Add default secondary_bd in internal flash for XIP on Mbed OS.

### DIFF
--- a/boot/mbed/mbed_lib.json
+++ b/boot/mbed/mbed_lib.json
@@ -160,6 +160,20 @@
             "help": "Share data (NOT TESTED)",
             "macro_name": "MCUBOOT_DATA_SHARING",
             "value": null
+        },
+        "direct-xip": {
+            "help": "Enable ability to boot update candidates in-place.",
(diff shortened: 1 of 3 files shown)


AGlass0fMilk:
@LDong-Arm In the future, I would like to look into having a more flexible way of defining the number of flash slots available. The `mcuboot.xip-secondary-slot-address` parameter doesn't seem like the ideal solution to me. How does this parameter change for applications built to run in the secondary slot?

For now though, I think this is a good incremental step to support XIP until we refactor the flash backend.

utzig (approved):
LGTM

nvlsianpu (approved):

PR #902, opened 2020-12-16
bootutil: Add cmake build file

RcColes:
In order to allow other projects to include the bootutil files more
easily. Allows renaming and moving of bootutil files without breaking
external projects' file lists (if they include this cmake file instead
of directly listing the files they use). Prevents an issue where
moving/renaming bootutil files breaks the FIH CI test.

### DIFF
--- a/boot/bootutil/CMakeLists.txt
+++ b/boot/bootutil/CMakeLists.txt
@@ -1,0 +1,35 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2020, Arm Limited. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+#------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


utzig:
@RcColes Thanks for this, do you already have a trusted-firmware-m "PR" as well?

RcColes:
> @RcColes Thanks for this, do you already have a trusted-firmware-m "PR" as well?

This one would have to be merged first, due to needing the commit hash in order to update the tfm dependency list. I can sort one out for testing purposes if that'd be useful, but the actual merges need to be serialized in order to get final commit hashes.

utzig (approved):

nvlsianpu (approved):

nvlsianpu:
Thanks @RcColes Merged!

PR #904, opened 2021-01-05
Fix ptest

d3zd3z:
Now that we are running tests git github workflows instead of Travis, the ptest utility to test off-platform does nothing. It also crashes due to a bug provoked by newer version of Rust. This PR fixes these so that ptest will once again run the same sets of simulator tests that CI runs.


utzig (approved):
LGTM

nvlsianpu (approved):

PR #906, opened 2021-01-05
ci: Pin tf-m-tests version in FIH test

RcColes:
To avoid problems where the latest master of tf-m-tests is incompatible
with the fixed version used by the FIH test. Bump the version of TF-M
used slightly to a commit that allows the version of tf-m-tests to be
easily overridden.


utzig:
Thanks for fixing the issue, but would it be possible to use tags so we don't need to rely on SHAs for these things?

tamasban (approved):

utzig (approved):

PR #909, opened 2021-01-11
imgtool: Use stable version numbering

de-nordic:
Change versioning and update version to 1.7.0.1

### DIFF
--- a/scripts/imgtool/__init__.py
+++ b/scripts/imgtool/__init__.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-imgtool_version = "1.8.0a1"
+imgtool_version = "1.7.0.1"


de-nordic:
@utzig Something like this?

utzig:
Thanks, this is due to some discussion that happened on https://github.com/zephyrproject-rtos/zephyr/pull/30695.

utzig:
Btw, downgrade is currently not allowed: https://github.com/mcu-tools/mcuboot/blob/master/ci/imgtool_run.sh#L45, so `1.8.0a1` must be first deleted before this is merged.

I believe the rationale for using `alpha` was that it was not a "proper" release, it was untested, and there could be broken things. In practice it seems to be quite stable, no bugs that I can remember in any `alpha` release. Also maybe we should publish downgrades, so `alphas` could be used as well, if something "unstable" ever shows up.

utzig (approved):
LGTM

nvlsianpu (approved):

d3zd3z (changes requested):
Four digit version numbers are not allowed by the semantic versioning spec, which is what we follow. What are we trying to accomplish here? https://semver.org/

d3zd3z:
We need to figure out how to be more consistent in keeping the version numbers consistent between imgtool and mcuboot. I notice we have things like 1.7.0a1 on pypy. Unfortunately, we decided in another change to use 1.8.0-dev for mcuboot. Since semver makes the prerelease tags just alphabetical, this kind of precludes us from using 'a' (presumably alpha).

I think a deeper problem here is that development shouldn't really be using imgtool out of pypi at all. Zephyr shouldn't be trying to sync up a version of imgtool out of pypi, it should be using the one in the mcuboot tree that it is using. If it isn't using mcuboot, then why would it need imgtool?

So, I don't think we should be pushing a whole bunch of releases of imgtool to pypi, we should change Zephyr to use the in-tree version of imgtool, then compatibility and versions will _never_ be a problem.

de-nordic:
> I think a deeper problem here is that development shouldn't really be using imgtool out of pypi at all. Zephyr shouldn't be trying to sync up a version of imgtool out of pypi, it should be using the one in the mcuboot tree that it is using. If it isn't using mcuboot, then why would it need imgtool?
> 
> So, I don't think we should be pushing a whole bunch of releases of imgtool to pypi, we should change Zephyr to use the in-tree version of imgtool, then compatibility and versions will _never_ be a problem.

I agree on this and will as around what is the reason for current state.
For most time I have been using Zephyr I have actually been signing the binaries by hand, by invoking the imgtool from cloned mcuboot repo; I have been quite surprised when started using west for signing and my "debug prints" stopped appearing.

PR #910, opened 2021-01-12
bootutil: Modify boot_add_data_to_shared_area function scope

Sherryzhang2:
Modify the function boot_add_data_to_shared_area from static to
global as it is used in the user interface boot_save_shared_data.

### DIFF
--- a/boot/bootutil/src/boot_record.c
+++ b/boot/bootutil/src/boot_record.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 Arm Limited
+ * Copyright (c) 2018-2021 Arm Limited
  * Copyright (c) 2020 Linaro Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jamike (commented):
I suggest to update interface file ( boot_record.h )

Sherryzhang2:
Make sense. Fixed. @jamike

jamike (approved):

boot/bootutil/src/boot_record.c
```diff
@@ -44,18 +44,8 @@
...
- * @param[in] major_type  TLV major type, identify consumer
- * @param[in] minor_type  TLV minor type, identify TLV type
- * @param[in] size        length of added data
- * @param[in] data        pointer to data
- *
- * @return                0 on success; nonzero on failure.
- */
-static int
+/* See in boot_record.h */
+int
```
utzig, line 48: You should add an abort early check for `data == NULL` to the function as well.

Sherryzhang2: Fixed.

utzig (approved):

utzig:
I have just gone ahead and merged this because @davidvincze is on vacation. Sorry for the wait.

Issue #913, opened 2021-01-12
Separate build of sim and mcuboot code


d3zd3z:
Currently, CI invokes runs the simulator in numerous configurations. Each of these are run through an invocation of `cargo test` with differing `--features` passed in. Although most of the dependencies do not change with the feature flags, this does cause a new test image to be built for each configuration.

In a large part, the simulator has already been made to be independent from the features defined, querying most information from the capability queries. There are a few conditional compilations that have crept in since this was done that will have to be fixed to implement this.

What I propose here is to separate the build of MCUboot from the simulator. The simulator will be built once. MCUboot itself will be recompiled for each feature configuration, and collected into a series of .so files. The simulator would then just be given an .so to test, determining capabilities dynamically.

Steps:
- [ ] Benchmark test. Need to make sure this is even worth doing. Catalog compile time for the tests, compile time for the C code in mcuboot-sys, and the rest as time spent running tests.
- [ ] Remove any remaining compile-time decisions in the simulator code
- [ ] Implement dynamic loading of mcuboot code in the test
- [ ] Modify the workflow and scripts to run these configurations the new way.

d3zd3z:
First impression, after adding some timing code to the test building and running is that this change wouldn't help very much. It might help with disk space is all (the sim build directory typically grows to about 4GiB). The tests typically take 10-20 seconds to build (other than the first one to run). The test runs range from a second or two, to a bunch of them that take 7-10 minutes. This change would be shaving maybe 5-10 seconds off of each test, which is fairly insignificant compared to the test runs themselves.

I'll push the test separation and ptest log changes through, though, but I'm closing this ticket.

PR #914, opened 2021-01-12
Allow logging from all test, not just failures.

d3zd3z:
A few testing changes to help benchmark test runs to figure out where time is being spent. The first patch splits the build and run parts of the tests, while adding a `time` command to show the amount of time taken by the test. The second patch enhances ptest to check for an environment variable `PTEST_LOG_ALL`, which will cause the results of all tests to be output, instead of just failures. These changes helped determine that #913 was not worth pursuing.


utzig (approved):
LGTM

nvlsianpu (approved):

PR #921, opened 2021-01-15
ci: Update TF-M version

RcColes:
To fix a regression caused by f68473814f17a299ff8766bc3cf86e014552f479,
where an older TF-M version was used that didn't support the bootutil
cmake.


utzig (approved):
LGTM

nvlsianpu (approved):
Thanks for the path!

PR #926, opened 2021-01-21
imgtool: fix validation with protected TLVs

### LINKED ISSUE #924: imgtool.py cannot verify the img with TLV protect info
I use command as below to sign the image  with  TLV Prot info：security count：
`imgtool.py sign -k root-rsa-2048.pem -e big -S 0x28000 -H 0x0400 --pad-header -v "1.0.0" -s auto --align 8 Project.bin test_sign.bin`
Then i use 'verify' command to verify the signed image, failed with log "Invalid TLV info magic; is this an MCUboot image?" 

After i check the function verify in image.py, it seem dont adapt to image include TLV_PROT_INFO_MAGIC
```
        tlv_info = b[header_size+img_size:header_size+img_size+TLV_INFO_SIZE]
        magic, tlv_tot = struct.unpack('HH', tlv_info)
        if magic != TLV_INFO_MAGIC:
            return VerifyResult.INVALID_TLV_INFO_MAGIC, None, None

```

utzig:
After the change to support protected TLVs, the `verify` command was not updated with proper support. Fix it by skipping any protected TLV found, and fix the size of the hashed/signed region to also include the protected TLV area.

Fixes part of #924

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -544,16 +544,22 @@ class Image():
         if magic != IMAGE_MAGIC:
             return VerifyResult.INVALID_MAGIC, None, None
 
-        tlv_info = b[header_size+img_size:header_size+img_size+TLV_INFO_SIZE]
+        tlv_off = header_size + img_size
+        tlv_info = b[tlv_off:tlv_off+TLV_INFO_SIZE]
         magic, tlv_tot = struct.unpack('HH', tlv_info)
+        if magic == TLV_PROT_INFO_MAGIC:
+            tlv_off += tlv_tot
+            tlv_info = b[tlv_off:tlv_off+TLV_INFO_SIZE]
+            magic, tlv_tot = struct.unpack('HH', tlv_info)
+
         if magic != TLV_INFO_MAGIC:
             return VerifyResult.INVALID_TLV_INFO_MAGIC, None, None
 
         sha = hashlib.sha256()
-        sha.update(b[:header_size+img_size])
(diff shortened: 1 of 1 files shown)


nvlsianpu (approved):

scripts/imgtool/image.py
```diff
@@ -534,31 +534,38 @@ def pad_to(self, size):
         self.payload += pbytes
 
     @staticmethod
-    def verify(imgfile, key):
+    def verify(imgfile, key, endian="little"):
```
d3zd3z, line 537: I'm not quite sure what happened, but the comments in image.h, and the way we did the TLV, both the image header and the TLV and trailer are all supposed to be little endian. I think we should fix the code to deal with this on big-endian platforms. Either that, or we should change those headers. But, I think I prefer the image header and TLV being the same no matter the endianness of the platform running the code.

d3zd3z (changes requested):
I'll change this to request changes, because we really do need to either obey the image.h, and do everything as little endian, or change the comments there to match what we really do. I prefer keeping the images the same no matter the target.

scripts/imgtool/image.py
utzig: Yeah, I agree, but `create` is broken then as well, so someone with access to the HW must confirm it is not working. In any case I dropped that commit and will open a new PR only with that commit.

utzig: Btw, I am quite sure there are big endian users already, so I would assume at least the basics are functional.

PR #927, opened 2021-01-21
imgtool: add endianness support to verify command

### LINKED ISSUE #924: imgtool.py cannot verify the img with TLV protect info

utzig:
Add new cli option `-e` to specify endianness, similarly to how it is done for `create`/`sign` and use proper struct unpacking based on chosen endiannes.

This partly fixes #924 and must be rebased after #926 is merged.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -534,11 +534,12 @@ class Image():
         self.payload += pbytes
 
     @staticmethod
-    def verify(imgfile, key):
+    def verify(imgfile, key, endian="little"):
         with open(imgfile, "rb") as f:
             b = f.read()
 
-        magic, _, header_size, _, img_size = struct.unpack('IIHHI', b[:16])
+        e = STRUCT_ENDIAN_DICT[endian]
+        magic, _, header_size, _, img_size = struct.unpack(e + 'IIHHI', b[:16])
         version = struct.unpack('BBHI', b[20:28])
 
         if magic != IMAGE_MAGIC:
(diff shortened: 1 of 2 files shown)


utzig:
@d3zd3z Regarding the endianness issue pointed out at #926, I think you should re-consider at least while we are not at 1.8.0. The `create`/`sign` command already does the "wrong" thing, by creating images that don't follow the specification as described in `image.h`, so leaving `verify` broken in relation to how image creation works is even worse. I would suggest fixing the magic fields for real while we are in preparation for 1.8.0, because it will break compatibility for existing big endian users. Also I think there will need to be updates to `bootutil` and `mcumgr` (maybe others?). I will work on a PR with the proper fix.

d3zd3z:
I guess my suggestion, though, is that if we're going to make the fields endian-specific, we should fix the image.h to reflect that, especially since we're already doing it that way.

utzig:
> I guess my suggestion, though, is that if we're going to make the fields endian-specific, we should fix the image.h to reflect that, especially since we're already doing it that way.

Agreed. I am a bit surprised that #926 alone fixed #924, so I'll leave this PR "unfinished' and "RFC" until I can take a better look at all the "dependencies".

PR #928, opened 2021-01-22
boot: zephyr: adding indication LED and startup delay

jaredwolff:
Add LED support and debounce for serial detect pin operation.

Signed-off-by: Jared Wolff <hello@jaredwolff.com>


nvlsianpu (changes requested):

nvlsianpu (changes requested):

nvlsianpu:
button denouncing should be optional - as you hold button preset while reset the board, button signal shouldn't bounce in most cases.

boot/zephyr/main.c
```diff
@@ -376,12 +426,50 @@ void main(void)
...
+#else
+            rc = gpio_pin_read(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN,
+                            &detect_value);
+#endif
+            __ASSERT(rc >= 0, "Error of the reading the detect pin.\n");
+
+            uint32_t delta = k_uptime_get() -  timestamp;
+
+            /* If not pressed OR if pressed > debounce period stop loop*/
+            if( (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL && delta >=
```
nvlsianpu, line 449: if you really expected environmental noises, this check looks strange to me.

boot/zephyr/main.c
```diff
@@ -376,12 +426,50 @@ void main(void)
...
     if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL &&
         !boot_skip_serial_recovery()) {
-        BOOT_LOG_INF("Enter the serial recovery mode");
-        rc = boot_console_init();
-        __ASSERT(rc == 0, "Error initializing boot console.\n");
-        boot_serial_start(&boot_funcs);
-        __ASSERT(0, "Bootloader serial process was terminated unexpectedly.\n");
-    }
+            
+        k_sleep(K_MSEC(50));
```
nvlsianpu, line 430: this is part of debounce time as well.

boot/zephyr/main.c
nvlsianpu: I harsh environment medina value can be used as expected value.
In office environment signal should be stable after denounce time so even one sample might be used.

d3zd3z: How are we expecting this button to be used? I thought the button was to be held down before reset, in which case, I wouldn't expect any bounce.

d3zd3z (commented):
The other question, should we be defining something like a chosen node to set which LED to use for the led, rather than just using LED0?

boot/zephyr/main.c
jaredwolff: Another main use case for this button is that it's used to wake the device up from power down. Without this "debounce"  the device goes into Bootloader mode. It's an extra confirmation that the end user actually wants to go into the bootloaer.

jaredwolff: I would also agree we can make this optional. It's particular to the nRF9160 Feather series of boards. Not everyone cares about that extra startup time.

jaredwolff:
> The other question, should we be defining something like a chosen node to set which LED to use for the led, rather than just using LED0?

A agree here too. Wasn't sure it would make more sense to create a button definition in the DTS called `bootloader_led`. Then it's up to the board creator to define and use it.

boot/zephyr/Kconfig
```diff
@@ -435,6 +435,20 @@ config BOOT_SERIAL_CDC_ACM
 
 endchoice
 
+config MCUBOOT_INDICATION_LED
+	bool "Turns on LED indication when device is in DFU"
+	default n
+	help
+		Using LED0, device activates the LED while in bootloader mode.
```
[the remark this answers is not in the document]
jaredwolff: Should be fixed now

boot/zephyr/Kconfig
```diff
@@ -435,6 +435,20 @@ config BOOT_SERIAL_CDC_ACM
...
+
+config MCUBOOT_INDICATION_LED_POLARITY
```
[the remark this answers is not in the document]
jaredwolff: Changed this so the board needs a bootloader-led0 defined in the alias section. I hope that's what you were referring to here. Also removed any references to LED0 in the notes.

boot/zephyr/main.c
```diff
@@ -43,6 +43,51 @@ const struct boot_uart_funcs boot_funcs = {
...
+
+const static struct device *led;
+
+void led_init(void)
+{
+    
+  led = device_get_binding(LED0_GPIO_LABEL);
+  if (led == NULL)
+  {
+    printk("Didn't find LED device %s\n", LED0_GPIO_LABEL);
```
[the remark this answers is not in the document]
jaredwolff: Should be fixed now

boot/zephyr/main.c
```diff
@@ -43,6 +43,51 @@ const struct boot_uart_funcs boot_funcs = {
...
+#define LED0_NODE DT_ALIAS(led0)
+
+#if DT_NODE_HAS_STATUS(LED0_NODE, okay) && DT_NODE_HAS_PROP(LED0_NODE, gpios)
+#define LED0_GPIO_LABEL DT_GPIO_LABEL(LED0_NODE, gpios)
+#define LED0_GPIO_PIN DT_GPIO_PIN(LED0_NODE, gpios)
+#define LED0_GPIO_FLAGS (GPIO_OUTPUT | FLAGS_OR_ZERO(LED0_NODE))
+#else 
+/* A build error here means your board isn't set up to drive an LED. */
+#error "Unsupported board: led0 devicetree alias is not defined"
+#define LED0	""
```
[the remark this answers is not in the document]
jaredwolff: Should be removed.

boot/zephyr/main.c
jaredwolff: Just pushed a fix for this. By default there is no boot delay. If there is one then this is added in.

boot/zephyr/main.c
jaredwolff: Only used now when `CONFIG_BOOT_SERIAL_DETECT_DELAY` is defined > 0. Ensures there's no bouncing for that first 50 ms.

boot/zephyr/main.c
```diff
@@ -105,6 +105,49 @@ static inline bool boot_skip_serial_recovery()
...
+{
+    
+  led = device_get_binding(LED0_GPIO_LABEL);
+  if (led == NULL)
+  {
+    BOOT_LOG_ERR("Didn't find LED device %s\n", LED0_GPIO_LABEL);
+    return;
+  }
+
+  gpio_pin_configure(led, LED0_GPIO_PIN, LED0_GPIO_FLAGS);
```
nvlsianpu, line 146:
```suggestion
  gpio_pin_configure(led, LED0_GPIO_PIN, LED0_GPIO_FLAGS);
  gpio_pin_set(led, LED0_GPIO_PIN, 0);
```

boot/zephyr/Kconfig
```diff
@@ -466,6 +474,14 @@ config BOOT_SERIAL_DETECT_PIN_VAL
 	  Logic value of the detect pin which triggers serial recovery
 	  mode.
 
+config BOOT_SERIAL_DETECT_DELAY
+	int "Prevents immediate boot into MCUBoot firmware."
```
nvlsianpu, line 478: maybe "Serial detect pin denounce time [ms]"

nvlsianpu (changes requested):
With requested change in main.c this patchworks well.

boot/zephyr/main.c
jaredwolff: Added this guy in a separate commit.

boot/zephyr/Kconfig
jaredwolff: Ahh yea that would make more sense huh. I pushed a tweak to this. Let me know if it looks ok.

nvlsianpu (approved):

nvlsianpu:
@d3zd3z I'm OK with merging that.

boot/zephyr/main.c
```diff
@@ -105,6 +105,50 @@ static inline bool boot_skip_serial_recovery()
...
+#error "Unsupported board: led0 devicetree alias is not defined"
+#endif
+
+const static struct device *led;
+
+void led_init(void)
+{
+    
+  led = device_get_binding(LED0_GPIO_LABEL);
+  if (led == NULL)
```
utzig, line 140: Regarding style, the opening bracket should be on the same line.

utzig (approved):

boot/zephyr/main.c
jaredwolff: @utzig would you like me to fix this? I just tried pushing another commit but it's not being added to this PR. maybe because it's already approved?

nvlsianpu: You should squash patch to current commit and force push there

jaredwolff: Got it. Thanks @nvlsianpu. Done! :)

PR #932, opened 2021-01-26
boot: zephyr: fix Zephyr build

gmarull:
CONFIG_LOG_MINIMAL has now become CONFIG_LOG_MODE_MINIMAL.

Signed-off-by: Gerard Marull-Paretas <gerard@teslabs.com>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -52,7 +52,7 @@ const struct boot_uart_funcs boot_funcs = {
 #endif
 
 #if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) && \
-    !defined(CONFIG_LOG_MINIMAL)
+    !defined(CONFIG_LOG_MODE_MINIMAL)
 #ifdef CONFIG_LOG_PROCESS_THREAD
 #warning "The log internal thread for log processing can't transfer the log"\
          "well for MCUBoot."
@@ -259,7 +259,7 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 
 #if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
-    !defined(CONFIG_LOG_PROCESS_THREAD) && !defined(CONFIG_LOG_MINIMAL)
+    !defined(CONFIG_LOG_PROCESS_THREAD) && !defined(CONFIG_LOG_MODE_MINIMAL)
 /* The log internal thread for log processing can't transfer log well as has too
  * low priority.
  * Dedicated thread for log processing below uses highest application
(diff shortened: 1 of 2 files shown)


boot/zephyr/main.c
```diff
@@ -52,7 +52,7 @@ const struct boot_uart_funcs boot_funcs = {
 #endif
 
 #if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) && \
-    !defined(CONFIG_LOG_MINIMAL)
+    !defined(CONFIG_LOG_MODE_MINIMAL)
```
d3zd3z, line 55: Do you know the specifics of when this was changed in Zephyr, and what we need to do to support multiple Zephyr versions. One possibility is to conditionalize on either definition. e.g. `&& !(defined(CONFIG_LOG_MINIMAL) || defined(CONFIG_LOG_MODE_MINIMAL))`.

d3zd3z: In light of this, we could also conditionalize this once, and define a local definition that would keep all of this from needing to be expanded in every place we care about it (which, admittedly is only two at this point).

gmarull: https://github.com/zephyrproject-rtos/zephyr/pull/31404 seem to introduce the change. To be honest I'm not sure if multiple versions have to be supported, Zephyr maintainers or authors of the changes should be able to tell.

henrikbrixandersen: There is also the `prj.conf` file change, which cannot easily support both.

boot/zephyr/main.c
```diff
@@ -259,7 +259,7 @@ static void do_boot(struct boot_rsp *rsp)
 #endif
 
 #if defined(CONFIG_LOG) && !defined(CONFIG_LOG_IMMEDIATE) &&\
-    !defined(CONFIG_LOG_PROCESS_THREAD) && !defined(CONFIG_LOG_MINIMAL)
+    !defined(CONFIG_LOG_PROCESS_THREAD) && !defined(CONFIG_LOG_MODE_MINIMAL)
```
nvlsianpu, line 262: CONFIG_LOG_MINIMAL is still emitted by zephyr, change here is not required, or maybe
```suggestion
    !defined(CONFIG_LOG_PROCESS_THREAD) && !defined(CONFIG_LOG_MINIMAL) && !defined(CONFIG_LOG_MODE_MINIMAL)
```

boot/zephyr/main.c
nvlsianpu: >In light of this, we could also conditionalize this once, and define a local definition that would keep all of this from needing to be expanded in every place we care about it (which, admittedly is only two at this point).

more is not expected

I encourage to keep backward compatibility in the code.

CONFIG_LOG_MINIMAL is still emitted by zephyr, change here is not required or better check can be just extended by new define check.

nvlsianpu (changes requested):

nvlsianpu:
I will update this patch.

gmarull:
Thanks for taking it @nvlsianpu

PR #939, opened 2021-02-04
fix build with encryption enabled after bootutil_publi library was introduced

nvlsianpu:
fixes #938

PR #941, opened 2021-02-04
imgtool: fix encrypting hex images

utzig:
Fixes padding hex images when encrypting. The issues stems from binaries using `bytes` and IntelHex returning `array` where `bytes` cannot be appended to, so use `.extend()` instead.

Found while testing #939

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -354,7 +354,11 @@ class Image():
         if self.enckey is not None:
             pad_len = len(self.payload) % 16
             if pad_len > 0:
-                self.payload += bytes(16 - pad_len)
+                pad = bytes(16 - pad_len)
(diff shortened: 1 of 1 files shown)


vrahane:
@utzig This might be related to the issue we were talking about the other day.

utzig:
> @utzig This might be related to the issue we were talking about the other day.

This causes a very obvious error:

```
File "/home/utzig/src/mcuboot/scripts/imgtool/image.py", line 359, in create
    self.payload += pad
TypeError: can only extend array with array (not "bytes")
```

Probably not the same issue, btw, I tested both EC256 and RSA encrypted images with Zephyr and both work 100%, so your issue might required some extra debugging.

nvlsianpu (approved):

PR #943, opened 2021-02-08
Update README.md

gudnimg:
The first sentence feels very wrong reading it.

I am also thinking the following text could be improved (or even removed):

> Currently, MCUboot works with both the Apache Mynewt and Zephyr operating systems, but more ports are planned in the future. RIOT is currently supported as a boot target with a complete port planned.

There are more OSs mentioned in the **Using MCUboot** section just below like Mbed-OS. 

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ define a common infrastructure for the bootloader, system flash layout on
 microcontroller systems, and to provide a secure bootloader that enables
 simple software upgrades.
 
-MCUboot is operating system and hardware independent and relies on
-hardware porting layers from the operating. Currently, mcuboot works
+MCUboot is hardware independent and relies on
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,8 +23,8 @@ define a common infrastructure for the bootloader, system flash layout on
 microcontroller systems, and to provide a secure bootloader that enables
 simple software upgrades.
 
-MCUboot is operating system and hardware independent and relies on
-hardware porting layers from the operating. Currently, mcuboot works
+MCUboot is hardware independent and relies on
```
d3zd3z, line 26: I'm not sure this is quite right, either, though. Although probably not commonly used, it is reasonable to build MCUboot use one operating system and the application with a different one.

PR #947, opened 2021-02-17
Adding IMAGE_F_ROM_FIXED flag

de-nordic:
The series of commits adds IMAGE_F_ROM_FIXED flag, which has already been defined in imgtool and used in mcumgr, but has been missed in mcuboot, and code that uses the flag to skip application images that have been placed in incorrect slot.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -566,6 +566,36 @@ boot_version_cmp(const struct image_version *ver1,
 }
 #endif
 
+#if defined(MCUBOOT_DIRECT_XIP)
+/**
+ * Check if image in slot has been set with specific ROM address to run from
+ * and whether the slot starts at that address.
+ *
+ * @returns 0 if IMAGE_F_ROM_FIXED flag is not set;
+ *          0 if IMAGE_F_ROM_FIXED flag is set and ROM address specified in
+ *            header matches the slot address;
+ *          1 if IMF_F_ROM_FIXED flag is set but ROM address specified in header
+ *          does not match the slot address.
+ */
+static bool
+boot_rom_address_check(struct boot_loader_state *state, size_t slot)
+{
+    const struct image_header *hdr = boot_img_hdr(state, slot);
+    uint32_t f_off = boot_img_slot_off(state, slot);
+    if (hdr->ih_flags & IMAGE_F_ROM_FIXED && hdr->ih_load_addr != f_off) {
+        BOOT_LOG_WRN("Image in %s slot at 0x%x has been built for offset 0x%x"\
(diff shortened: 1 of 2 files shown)


nvlsianpu (commented):
looks good to me

boot/bootutil/src/loader.c
```diff
@@ -566,6 +566,36 @@ boot_version_cmp(const struct image_version *ver1,
...
+ *          does not match the slot address.
+ */
+static bool
+boot_rom_address_check(struct boot_loader_state *state, size_t slot)
+{
+    const struct image_header *hdr = boot_img_hdr(state, slot);
+    uint32_t f_off = boot_img_slot_off(state, slot);
+    if (hdr->ih_flags & IMAGE_F_ROM_FIXED && hdr->ih_load_addr != f_off) {
+        BOOT_LOG_WRN("Image in %s slot at 0x%x has been built for offset 0x%x"\
+		     ", skipping", slot == 0 ? "primary" : "secondary", f_off,
```
utzig, line 587: There are tabs here, fix to use spaces.

boot/bootutil/src/loader.c
```diff
@@ -566,6 +566,36 @@ boot_version_cmp(const struct image_version *ver1,
...
+ * and whether the slot starts at that address.
+ *
+ * @returns 0 if IMAGE_F_ROM_FIXED flag is not set;
+ *          0 if IMAGE_F_ROM_FIXED flag is set and ROM address specified in
+ *            header matches the slot address;
+ *          1 if IMF_F_ROM_FIXED flag is set but ROM address specified in header
+ *          does not match the slot address.
+ */
+static bool
+boot_rom_address_check(struct boot_loader_state *state, size_t slot)
```
utzig, line 581: Couldn't this be moved to the same #ifdef block where `boot_select_or_erase` is defined?

utzig (changes requested):

boot/bootutil/src/loader.c
de-nordic: Will fix.

boot/bootutil/src/loader.c
de-nordic: It used to, but that block is included for different reasons where this function is not used which causes compilation errors on unused static.

d3zd3z (approved):
I'm ok with the change. We will need to add this to the growing list of things implemented that we don't have tests for, though.

utzig (approved):

de-nordic:
> 
> 
> I'm ok with the change. We will need to add this to the growing list of things implemented that we don't have tests for, though.

Seems that testing this flag should be my responsibility, I will write test for this, just need to get familiar with mcuboot test framework.

PR #948, opened 2021-02-17
boot/zephyr: Fix Direct-XIP boot messages

de-nordic:
The commit changes banner message to indicate that bootloader has been
built for Direct-XIP mode.
It also removes, in Direct-XIP compiled code, information on jumping
into first slot: that information is not always correct for Direct-XIP
and other messages correctly point slot that will be used.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -371,7 +371,11 @@ void main(void)
 
     MCUBOOT_WATCHDOG_FEED();
 
+#if !defined(MCUBOOT_DIRECT_XIP)
     BOOT_LOG_INF("Starting bootloader");
+#else
+    BOOT_LOG_INF("Starting Direct-XIP bootloader");
+#endif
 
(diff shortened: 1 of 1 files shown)


nvlsianpu (approved):

boot/zephyr/main.c
```diff
@@ -498,7 +502,9 @@ void main(void)
     BOOT_LOG_INF("Bootloader chainload address offset: 0x%x",
                  rsp.br_image_off);
 
+#if !defined(MCUBOOT_DIRECT_XIP)
     BOOT_LOG_INF("Jumping to the first image slot");
```
nvlsianpu, line 506: or maybe BOOT_LOG_INF("Jumping to the image slot") for direct-xip mode?

de-nordic: Will change the message as suggested.

utzig (approved):

PR #949, opened 2021-02-23
Increase minimum flash write

kjassmann-renesas:
An admittedly incomplete attempt at https://github.com/mcu-tools/mcuboot/issues/713, but basic functionality is working for us on our flash with 128 byte minimum write size and 32K erase size (no ECC).  Based on comments in https://github.com/mcu-tools/mcuboot/issues/713, the following may still need investigation/updates:
 * mcumgr and newt
 * simulator
 * other OS integration (though default settings are backward compatible)
 * ECC considerations
 * other areas, perhaps?


d3zd3z:
How large does this make the trailer area with your flash devices? One of the reason we've kept the write size somewhat small is because making it large can greatly increase the size of this area.

kjassmann-renesas:
> How large does this make the trailer area with your flash devices? One of the reason we've kept the write size somewhat small is because making it large can greatly increase the size of this area.

We removed the limit on BOOT_MAX_IMG_SECTORS.  With the 32K erase size, it's common for our applications to have very few sectors.  The trailer size is (5 * BOOT_MAX_ALIGN) + (BOOT_MAX_IMG_SECTORS * min-write-size * 3). For a 128K image, BOOT_MAX_IMG_SECTORS is 4, so the trailer is 17 * 128 = 2176 bytes.  Yes, that's a lot of mostly unused space, but it leaves over 98% of the flash for the image, so I thought it might be an acceptable trade off for some applications that want to be able to confirm or revert upgrades.

kjassmann-renesas:
> Using your branch, we can get only 32 aligned trailer if we set `DEFAULT_MAX_ALIGN = 32` in the file image.py on line 44
> 
> max_align is set in the _init_ to
> ` def __init__(..., max_align=DEFAULT_MAX_ALIGN)`
> 
> You have added
> `self.max_align = DEFAULT_MAX_ALIGN if max_align is None else int(max_align)`
> 
> What is the correct way to configure this setting?

We set max_align using a new --max_align argument to imgtool.py.  For example:

`> python3 scripts/imgtool.py sign -k root-rsa-2048.pem --header-size 0x80 --slot-size 0x8000 --max-sectors 1 --version 1.0.1 --align 128 --max-align 128 --confirm --pad-header input.bin signed.bin`

kjassmann-renesas:
> where is the correct place to configure `MCUBOOT_BOOT_MAX_ALIGN` in the project?

We are setting `MCUBOOT_BOOT_MAX_ALIGN` in mcuboot_config.h.  Should I update mcuboot_config.template.h?

d3zd3z:
> where is the correct place to configure `MCUBOOT_BOOT_MAX_ALIGN` in the project?

Ideally, it would be configured appropriately for the target platform. For Zephyr, that would be through a Kconfig option. For Mynewt, I think it would be `boot/mynewt/syscfg.yml`.

utzig:
I don't think d0378c826b022c5e750be4f34cf25468312bbfa2 is OK. While `32` really is arbitrary there is a minimum amount which is OK; it surely won't work with 1 or 2 sectors (depending on the swap mode). Maybe changing it to 4, or better 8, would be more reasonable.

I am starting to consider if we should not just get rid to this MAX_BOOT_ALIGN concept and use whatever align is provided by the flash driver and from `imgtool` parameters. `newt` already gets it from the flash driver in Mynewt, so does `mcumgr`/`newtmgr` in both Zephyr and Mynewt. I have no idea if some flash drivers might have a run-time only align size? For Mynewt I am quite sure none does, but for Zephyr I wouldn't know.

Regarding `mcumgr`, `newt`, etc I can take care of later. At a minimum we need the changes here, and that they are running on the simulator. I will take a look at the simulator changes in the weekend but I assume it is a pretty straight-forward change.

Also since ECC was mentioned, some time ago someone asked if MCUboot writes to non-erased sectors. I answered "no" at the time, but there was a mention that it indeed does. I think the reason is that they way MCUboot checks that a sector is erased is by reading it, which might not work for ECC flashes. So we have to look into that, but it is not in the scope of this PR, and it is OS specific...

kjassmann-renesas:
> I don't think [d0378c8](https://github.com/mcu-tools/mcuboot/commit/d0378c826b022c5e750be4f34cf25468312bbfa2) is OK. While `32` really is arbitrary there is a minimum amount which is OK; it surely won't work with 1 or 2 sectors (depending on the swap mode). Maybe changing it to 4, or better 8, would be more reasonable.

Could you please explain? We are testing 1 sector using a confirmed upgrade, and it's working for swap, overwrite, and overwrite fast modes. We haven't tested XIP or RAM loading.

> I am starting to consider if we should not just get rid to this MAX_BOOT_ALIGN concept and use whatever align is provided by the flash driver and from `imgtool` parameters. `newt` already gets it from the flash driver in Mynewt, so does `mcumgr`/`newtmgr` in both Zephyr and Mynewt. I have no idea if some flash drivers might have a run-time only align size? For Mynewt I am quite sure none does, but for Zephyr I wouldn't know.

I am also not aware of any flash drivers with a run-time alignment. I think the build time option may be smaller, and optimal size is very important to us.

> Regarding `mcumgr`, `newt`, etc I can take care of later. At a minimum we need the changes here, and that they are running on the simulator. I will take a look at the simulator changes in the weekend but I assume it is a pretty straight-forward change.

Thanks! I will be interested to see the simulator update.

> Also since ECC was mentioned, some time ago someone asked if MCUboot writes to non-erased sectors. I answered "no" at the time, but there was a mention that it indeed does. I think the reason is that they way MCUboot checks that a sector is erased is by reading it, which might not work for ECC flashes. So we have to look into that, but it is not in the scope of this PR, and it is OS specific...

It definitely writes to non-erased sectors after this change (in swap and overwrite fast mode at least).  We are buffering writes at the driver layer and flushing them when we see a write to a new sector.  We would prefer to update MCUboot to only write to erased sectors, but buffered writes are acceptable for now. Our flash does not have ECC.

SwissKnife64:
Using MCUboot with a STM32H7 we have had do do some fixes to get the boot loader running, but the writing of "confirmed" from the application still fails.

We had to correct the writing of the magic 16 byte pattern. The function for writing the magic pattern has to use a bigger buffer and write the padding and the magic pattern to the flash. The current implementation tries to write only 16 byte and fails silently on STM32H743 due to an error in the flash driver. 

The check for alignment in the file `drivers/flash/flash_stm32h7x.c` has an error on line 103, detected by @SpooniSpoon: 
with error `if ((offset % FLASH_NB_32BITWORD_IN_FLASHWORD * 4) != 0)`
corrected `if ((offset % (FLASH_NB_32BITWORD_IN_FLASHWORD * 4)) != 0)` with additional braces for correct calculation precedence.
_(Sorry, but not yet used to do github pull requests.)_ 

Then there is a conceptual question to solve:
MCUBoot uses  `{ byte padding[n]; byte magic[16] }`, 
and the Zephyr implementation in the application `{ byte magic[16]; byte padding[n]  }`, 
resulting in different position of magic pattern.
At MCUboot the 16 magic pattern is always at the end of the slot, at Zephyr the position of the magic pattern moves with paddings > 16 bytes.

kjassmann-renesas:
Please reopen this pull request.

scripts/imgtool/main.py
```diff
@@ -282,8 +282,12 @@ def convert(self, value, param, ctx):
               help='Specify the value of security counter. Use the `auto` '
               'keyword to automatically generate it from the image version.')
 @click.option('-v', '--version', callback=validate_version,  required=True)
-@click.option('--align', type=click.Choice(['1', '2', '4', '8']),
+@click.option('--align', type=click.Choice(['1', '2', '4', '8', '16', '32',
+              '64', '128', '256', '512', '1024', '2048', '4096']),
               required=True)
+@click.option('--max-align', type=click.Choice(['1', '2', '4', '8', '16', '32',
```
mnkp, line 288: This extra option seems unnecessary. Makes `imgtool` less intuitive for the end user. We can calculate the value of `max_align`. It should be equal to `align` if the value of `align` is greater than 8 or otherwise it should be 8. Also, without the option it will be easier to refactor mcuboot code, if the need arises.

mnkp (changes requested):
Grepping for `flash_area_write` reveals a few more places where the change is required
```
boot/bootutil/src/bootutil_misc.c:    rc = flash_area_write(fap, off, bs->enckey[slot], BOOT_ENC_KEY_SIZE);
boot/bootutil/src/bootutil_public.c:    rc = flash_area_write(fap, off, boot_img_magic, BOOT_MAGIC_SZ);
```
That's in the functions writing the magic and encryption keys. They need to be adapted to use `boot_write_trailer` or similar approach.

mnkp:
> > I am starting to consider if we should not just get rid to this MAX_BOOT_ALIGN concept and use whatever align is provided by the flash driver and from `imgtool` parameters. `newt` already gets it from the flash driver in Mynewt, so does `mcumgr`/`newtmgr` in both Zephyr and Mynewt. I have no idea if some flash drivers might have a run-time only align size? For Mynewt I am quite sure none does, but for Zephyr I wouldn't know.
> 
> I am also not aware of any flash drivers with a run-time alignment. I think the build time option may be smaller, and optimal size is very important to us.

Since MCUboot supports storing images in an external flash what should happen if the flash alignment of the external and internal flash are different? Should `BOOT_MAX_ALIGN` be always the larger of the two, should it be specific to the flash the image is written to?

d3zd3z:
> Since MCUboot supports storing images in an external flash what should happen if the flash alignment of the external and internal flash are different? Should `BOOT_MAX_ALIGN` be always the larger of the two, should it be specific to the flash the image is written to?

I think it works if the max align is set to the larger of the two devices.

elcritch:
> For the LPC55S69, originally, reads from unwritten data would result in a fault. We ended up working around this by using a flash command to read that would return an error status, and in that case, we would put bogus data into the buffer. That flash device has commands to distinguish between corrupt data and erased flash, but it isn't currently being used.

Is the workaround for LPC55S69 included in this PR, or is it something done in the application layer? I'm working with LPC55S69 on Zephyr. The bootloader and application image boots if I adjust it to use `imgtool sign --align 8`. However I haven't tried upgrading OTA however and it seems like that's where the bogus data / buffer would come into play? I'm willing to port changes over to Zephyr if the fix requires it.

d3zd3z:
> Is the workaround for LPC55S69 included in this PR, or is it something done in the application layer? I'm working with LPC55S69 on Zephyr. The bootloader and application image boots if I adjust it to use imgtool sign --align 8. However I haven't tried upgrading OTA however and it seems like that's where the bogus data / buffer would come into play? I'm willing to port changes over to Zephyr if the fix requires it.

The workaround for the LPC55S69 is entirely within the Zephyr driver, and makes the flash device behave a little more like a non-ECC device. However, since it still has an "align" of 512, MCUboot is only supported in overwrite-only mode, not with swap.

elcritch:
> The workaround for the LPC55S69 is entirely within the Zephyr driver, and makes the flash device behave a little more like a non-ECC device. However, since it still has an "align" of 512, MCUboot is only supported in overwrite-only mode, not with swap.

Ok, that makes sense. Could you share the Zephyr changes available for that? I'm curious as to how the ECC flash is different.

Sherryzhang2:
What's the status of this PR? Any comments on this PR?

utzig:
@kjassmann-renesas Could you rebase this on latest master?

kjassmann-renesas:
> @kjassmann-renesas Could you rebase this on latest master?

@utzig We are working on migrating these changes to the latest master. We will update this MR or issue a new one soon.

michaelthomasj:
> > > @kjassmann-renesas Could you rebase this on latest master?
> 
> > 
> 
> > @utzig We are working on migrating these changes to the latest master. We will update this MR or issue a new one soon.
> 
> 
> 
> Do you mind me taking over this PR?

Fabio, 
I'll update this MR with one rebased to 1.8 in the next day or so.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -71,7 +71,17 @@ extern "C" {
 /** Swapping encountered an unrecoverable error */
 #define BOOT_SWAP_TYPE_PANIC    0xff
 
+#define BOOT_MAGIC_SZ           16
```
alexdolzhenkovbdl, line 74: Remove `#define BOOT_MAGIC_SZ (sizeof boot_img_magic)`? (at line 90)

d3zd3z:
This does need to be rebased. It is pulling in older version of numerous commits, and includes a merge commit.

Also, please look for lines with whitespace at the end. At least the last of the commits introduces a bunch.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -71,7 +71,17 @@ extern "C" {
...
+
+#ifdef MCUBOOT_BOOT_MAX_ALIGN
+#define BOOT_MAX_ALIGN          MCUBOOT_BOOT_MAX_ALIGN
+#define BOOT_MAGIC_ALIGN_SIZE \
+    ((((BOOT_MAGIC_SZ - 1) / BOOT_MAX_ALIGN) + 1) * BOOT_MAX_ALIGN)
```
gustavonihei, line 79: This expression for aligning may be extracted to a generic macro to improve code comprehension. See:
https://github.com/mcu-tools/mcuboot/blob/main/boot/espressif/port/esp_mcuboot.c#L28
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/const.h#L31

boot/bootutil/src/bootutil_public.c
```diff
@@ -316,14 +316,30 @@ int
...
+    uint8_t magic[BOOT_MAGIC_ALIGN_SIZE];
+    uint8_t erased_val;   
 
     off = boot_magic_off(fap);
+    /* image_trailer structure was modified with additional padding such that 
+     * the pad+magic ends up in a flash minimum write region. The address 
+     * returned by boot_magic_off() is the start of magic which is not the 
+     * start of the flash write boundary and thus writes to the magic will fail. 
+     * To account for this change, write to magic is first padded with 0xFF 
+     * before writing to the trailer. */
```
gustavonihei, line 330:
```suggestion

    /* image_trailer structure was modified with additional padding such that
     * the pad+magic ends up in a flash minimum write region. The address
     * returned by boot_magic_off() is the start of magic which is not the
     * start of the flash write boundary and thus writes to the magic will fail.
     * To account for this change, write to magic is first padded with 0xFF
     * before writing to the trailer.
     */
```
1) Missing blank line before the beginning of the block comment.
2) Dangling whitespaces at the end of each line.
3) For consistency, the terminator for the block comment should be placed in a new line.

boot/bootutil/src/bootutil_public.c
```diff
@@ -316,14 +316,30 @@ int
 boot_write_magic(const struct flash_area *fap)
 {
     uint32_t off;
+    uint32_t pad_off;
     int rc;
+    uint8_t magic[BOOT_MAGIC_ALIGN_SIZE];
+    uint8_t erased_val;
```
gustavonihei, line 322:
```suggestion
    uint8_t erased_val;
```
Dangling whitespaces

d3zd3z:
Be careful, the latest patch adds a lot of extraneous trailing whitespace. This will definitely need to be rebased, with the changes squashed into a cleaner history.

boot/bootutil/include/bootutil/bootutil_public.h
gustavonihei:
```suggestion
#define BOOT_MAGIC_SZ           (sizeof boot_img_magic)
```
Actually the most correct would be to bring line 90 upwards.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -157,7 +157,7 @@ boot_magic_off(const struct flash_area *fap)
 static inline uint32_t
 boot_image_ok_off(const struct flash_area *fap)
 {
-    return boot_magic_off(fap) - BOOT_MAX_ALIGN;
+    return (boot_magic_off(fap) - BOOT_MAX_ALIGN) & ~(BOOT_MAX_ALIGN - 1);
```
gustavonihei, line 160: Should we apply this same patch to  https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/bootutil_public.c#L143?

boot/bootutil/src/bootutil_public.c
```diff
@@ -316,14 +316,30 @@ int
...
+
+    memset(&magic[0], erased_val, sizeof(magic));
+    memcpy(&magic[BOOT_MAGIC_ALIGN_SIZE - BOOT_MAGIC_SZ], boot_img_magic, BOOT_MAGIC_SZ);
 
     BOOT_LOG_DBG("writing magic; fa_id=%d off=0x%lx (0x%lx)",
-                 flash_area_get_id(fap), (unsigned long)off,
-                 (unsigned long)(flash_area_get_off(fap) + off));
-    rc = flash_area_write(fap, off, boot_img_magic, BOOT_MAGIC_SZ);
+                 fap->fa_id, (unsigned long)off,
+                 (unsigned long)(fap->fa_off + off));
```
gustavonihei, line 340: This change should be reverted. Flash Area members should be accessed through getters.

gustavonihei:
@kjassmann-renesas @utzig Do you mind if I take over this PR? It has been open for too long, and this is holding some other PRs from our side.
To avoid losing the PR comment history, it might be interesting to continue the work on the same branch. Then I'll need permission to force push to it in order to properly rebase and squash the commits.

kjassmann-renesas:
> @kjassmann-renesas @utzig Do you mind if I take over this PR? It has been open for too long, and this is holding some other PRs from our side. To avoid losing the PR comment history, it might be interesting to continue the work on the same branch. Then I'll need permission to force push to it in order to properly rebase and squash the commits.

@gustavonihei, we have already handed this off to @utzig. No issues on our end if he wants to transfer it to you.

boot/bootutil/src/bootutil_public.c
```diff
@@ -316,14 +316,30 @@ int
...
+     * the pad+magic ends up in a flash minimum write region. The address 
+     * returned by boot_magic_off() is the start of magic which is not the 
+     * start of the flash write boundary and thus writes to the magic will fail. 
+     * To account for this change, write to magic is first padded with 0xFF 
+     * before writing to the trailer. */
+    pad_off = off & ~(BOOT_MAX_ALIGN - 1);
+    
+    erased_val = flash_area_erased_val(fap);
+
+    memset(&magic[0], erased_val, sizeof(magic));
```
nvlsianpu, line 335: set size should be (BOOT_MAGIC_ALIGN_SIZE - BOOT_MAGIC_SZ)

nvlsianpu (changes requested):

boot/bootutil/src/bootutil_priv.h
```diff
@@ -109,16 +109,24 @@ struct boot_status {
  *  |                 Encryption key 0 (16 octets) [*]              |
  *  |                                                               |
  *  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ *  |  0xff padding as needed (BOOT_MAX_ALIGN - 16 EK0 octets) [*]  |
```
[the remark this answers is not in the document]
gustavonihei: I believe it stands for Encryption Key. In #1217 I've proposed a different layout in an attempt to making it easier to understand.

boot/bootutil/src/bootutil_public.c
gustavonihei: Clang static analyzer emits a warning due to the possibility of `(BOOT_MAGIC_ALIGN_SIZE - BOOT_MAGIC_SZ)` resulting in a call to `memset` with 0 size.
Since filling the `magic` buffer with the `erased_val` pattern is not wrong either, I believe there is no problem leaving it like this. Do you agree?

boot/bootutil/include/bootutil/bootutil_public.h
gustavonihei: `BOOT_MAGIC_SZ` needs to be resolved at preprocessing time due to a new change introduced in this PR and `sizeof` is only resolved at compile time.
So I had to leave it as **16**

gustavonihei:
> > @kjassmann-renesas @utzig Do you mind if I take over this PR? It has been open for too long, and this is holding some other PRs from our side. To avoid losing the PR comment history, it might be interesting to continue the work on the same branch. Then I'll need permission to force push to it in order to properly rebase and squash the commits.
> 
> @gustavonihei, we have already handed this off to @utzig. No issues on our end if he wants to transfer it to you.

I have no write permission to `renesas/mcuboot`. I created another PR at #1217.

boot/bootutil/src/bootutil_public.c
nvlsianpu: Both C99 and C11 allow to have 0 length here for size_t type parameter of <string.h> library functions.
But if Clang complains I agree that we can just do what is done here.

alexdolzhenkovbdl:
Should this PR be closed since #1217 has fixed this issue and has been merged?

kjassmann-renesas:
> Should this PR be closed since #1217 has fixed this issue and has been merged?

@alexdolzhenkovbdl, thanks for bringing this up!

I'll close this since it's merged with https://github.com/mcu-tools/mcuboot/pull/1217

PR #951, opened 2021-02-26
boot: zephyr: Switch to chosen zephyr,uart-mcumgr for serial recovery

de-nordic:
The commit changes the default device selection, for serial recovery,
from the DTS chosen zephyr,console to zephyr,uart-mcumgr.
This is more in line with Zephyr which uses the later for mcumgr
UART, which is also used for serial recovery in mcuboot.

The default configurations of Zephyr boards select the same device,
uart0, for both chosen devices; projects that do not use overlays to
change these selections will not be affected.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -483,16 +483,17 @@ config BOOT_SERIAL_DETECT_DELAY
 	  the one used to place the device in bootloader mode.
 
 # Workaround for not being able to have commas in macro arguments
-DT_CHOSEN_Z_CONSOLE := zephyr,console
+DT_CHOSEN_Z_UART_MCUMGR := zephyr,uart-mcumgr
 
 config RECOVERY_UART_DEV_NAME
 	string "UART Device Name for Recovery UART"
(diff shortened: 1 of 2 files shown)


de-nordic:
@nvlsianpu When you have a moment.

d3zd3z (approved):

nvlsianpu (changes requested):
Change of configuration source to zephyr,uart-mcumgr looks fine to me and it is reasonable.

Although this not solve issue which will be encountered if console will be enabled on same serial device. Compilation error should be still generated if configuration collides.

de-nordic:
> 
> 
> Change of configuration source to zephyr,uart-mcumgr looks fine to me and it is reasonable.
> 
> Although this not solve issue which will be encountered if console will be enabled on same serial device. Compilation error should be still generated if configuration collides.

I will try to figure this out, but the problem is that currently the UART device identifier is taken, from Kconfig option, as a string, which means that I cannot compare this at compile time. Additionally I am not able to use any of the DTS stuff as the Kconfig option, although comes from the DTS, can be modified by user.

de-nordic:
Replaced with https://github.com/mcu-tools/mcuboot/pull/1155

PR #955, opened 2021-03-03
tests: Add zephyr on-board test test_specification

PerMac:
Add the test specification describing the tests that will run in
automated test plan. The point of this document is to allow validation
of the implemented tests: if they perform the required procedure. This
specification should also give a user a better understanding of
the test process.

Signed-off-by: Maciej Perkowski <Maciej.Perkowski@nordicsemi.no>

Objective for posted this draft is to get coherent on terminology and testsuites division. This draft is obviously not 
comprehensive (it is the base). Other testsuites and testcases are planed to be added in other patches.

### DIFF
--- a/docs/test_specification.yaml
+++ b/docs/test_specification.yaml
@@ -1,0 +1,218 @@
+definitions: 
+  hello1: 
+    - A signed hello_word image
+  hello2:
+    - An upgrade image, signed and marked for upgrade 
+  good_key:
+    - A key generated with a cryptosystem set in the bootloader with a correct value
+  bad_key:
+    - A key generated with a cryptosystem set in the bootloader with an incorrect value
+  wrong_key:   
+    - A key generated with a cryptosystem not matching the algorithm set in the bootloader
+  bootloader_output:
+    - 'unable to find bootable image'
+  hello1_output:
+    - 'Hello World from hello1'
+  hello2_output:
+    - 'Hello World from hello2'
+test_suites:
+  test_rsa_variants:
+    description:
+      - Test suite that verifies that boodloader correctly handles images with RSA signatures (good and bad)
+    test_level: system
+    test_type: feature
+    test_group: rsa
+    implemented: False  
+    test_cases:
+      test_good_rsa:
+        description:
+          - Test a good image, with a good upgrade, using RSA signatures
+        test setup:
+          - clean build directories
+          - clean hexes 
+        test procedure:
+          - build bootloader with overlay-rsa.conf
+          - build and sign hello1 with the good RSA key
+          - build and sign hello2 with the good RSA key
+          - erase chip
+          - flash bootloader
+          - Verify output. Expected bootloader_output
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello2_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified
+      test_wrong_rsa:
+        description:
+          - Test a good image, with a wrong-signature upgrade, using RSA signatures.
+        test setup:
+          - clean build directories
+          - clean hexes 
+        test procedure:
+          - build bootloader with overlay-rsa.conf
+          - build and sign hello1 with the good RSA key
+          - build and sign hello2 with the bad RSA key
+          - erase chip
+          - flash bootloader
+          - Verify output. Expected bootloader_output
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello1_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified         
+      test_bad_rsa_upgrade:
+        description:
+          - Test that when configured for RSA, a wrong signature in the upgrade image will fail to upgrade.
+        test setup:
+          - clean build directories
+          - clean hexes 
+        test procedure:
+          - build bootloader with overlay-rsa.conf
+          - build and sign hello1 with the good RSA key
+          - build and sign hello2 with the wrong (ECDSA) key
+          - erase chip
+          - flash bootloader
+          - Verify output. Expected bootloader_output
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello1_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified         
+  test_ecdsa_variants:
+    description:
+      - Test suite that verifies that boodloader correctly handles images with ECDSA signatures (good and bad)
+    test_level: system
+    test_type: feature
+    test_group: ecdsa
+    implemented: False
+    test_cases:
+      test_good_ecdsa:
+        description:
+          - Test a good image, with a good upgrade, using ECDSA signatures
+        test setup:
+          - clean build directories
+          - clean hexes 
+        test procedure:
+          - build bootloader with overlay-ecdsa-p256.conf
+          - build and sign hello1 with the good ECDSA key
+          - build and sign hello2 with the good ECDSA key
+          - erase chip
+          - flash bootloader
+          - Verify output. Expected bootloader_output
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello2_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified
+      test_wrong_ecdsa:
+        description:
+          - Test a good image, with a wrong-signature upgrade, using ECDSA signatures
+        test setup:
+          - clean build directories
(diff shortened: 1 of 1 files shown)


PerMac:
I wrote this specification based on /mcuboot/samples/zephyr/makefile and /mcuboot/samples/zephyr/run-tests.sh. The description of test cases were copied from these files.

nvlsianpu:
@de-nordic Can you also look at this?

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
+definitions: 
+  hello1:
```
de-nordic, line 2: maybe just name these samples respectively:
hello_signed
hello_signed_upgrade_confirmed
hello_signed_upgrade_bad_key
hello_signed_upgrade_wrong_key
etc.

de-nordic: The unsigned image should probably by also checked as update option.

de-nordic: Apart from signing, the versioning should also be checked.

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
...
+          - Test a good image, with a good upgrade, using RSA signatures
+        test setup:
+          - clean build directories
+          - clean hexes 
+        test procedure:
+          - build bootloader with overlay-rsa.conf
+          - build and sign hello1 with the good RSA key
+          - build and sign hello2 with the good RSA key
+          - erase chip
+          - flash bootloader
```
de-nordic, line 38: This may require hardware reboot afterwards. I do not know whether this only happens with my setup but flashing new bootloader, either over old one or into clean board, causes board to hang and requires complete power down.

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
...
+          - build bootloader with overlay-rsa.conf
+          - build and sign hello1 with the good RSA key
+          - build and sign hello2 with the bad RSA key
+          - erase chip
+          - flash bootloader
+          - Verify output. Expected bootloader_output
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello1_output
```
de-nordic, line 65: I think that here you may also see log messages about  deleting the bad image.

de-nordic:
I think that "unsigned" images scenarios should also be added as it is possible to boot unsigned images.
Direct-XIP, which is already supported in mcuboot, should also be taken into account. In this case the messages will differ a little bit and the images will not be swapped.

de-nordic:
Btw. you do not have to build separate hello1 and hello2 images. The image will be the same, you may just use imgtool to sign it/encrypt/add version in which case you may use one build and create unlimited number of application images with different parameters.

nvlsianpu:
> you do not have to build separate hello1 and hello2 images. 

No, for helping to distinguish which image is running each application generates different messages, so need to build them both.

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
...
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello2_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified
+      test_wrong_rsa:
```
nvlsianpu, line 49: test_bad_rsa as `bad key` mean key not matches but algorithm is proper.

de-nordic:
> 
> No, for helping to distinguish which image is running each application generates different messages, so need to build them both.

Ah yes.

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
...
+          - flash hello1 hex
+          - Verify output. Expected hello1_output
+          - flash hello2 hex
+          - Verify output. Expected hello1_output
+          - reset chip
+          - Verify output. Expected hello1_output
+        success_criteria:
+          - Procedure complete without errors
+          - Output correctly verified         
+      test_bad_rsa_upgrade:
```
nvlsianpu, line 71: test_wrong_rsa_upgrade

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
+definitions: 
+  hello1: 
+    - A signed hello_word image
+  hello2:
+    - An upgrade image, signed and marked for upgrade 
+  good_key:
+    - A key generated with a cryptosystem set in the bootloader with a correct value
+  bad_key:
+    - A key generated with a cryptosystem set in the bootloader with an incorrect value
+  wrong_key:
```
nvlsianpu, line 10: I'm starting thinking that `wrong_key` is wrong name for such a key. No much differ from `bad_key`. Maybe `incompatible_key`.

If decided to switch term, then Terminology should be aligned for testcases name as well.

@d3zd3z Any thought on terminology.

docs/test_specification.yaml
nvlsianpu: ^^ that only for revert while dual-XIP mode is in usage, not in a image swap mode.
I see that test description should be extended by information on mcuboot configuration used.

docs/test_specification.yaml
d3zd3z: I wasn't quite sure what to call it. There are a few things that should be tested here. An obvious one (that I think is missed) is where the hash/signature is generated from the wrong image. This will fail before the signature is even verified. Then there are several things that can be wrong with the signature itself. It could be things like this where the wrong key is just used. It ends up just showing up as a bad signature. It is also possible to do things like modify a bit in the signature, which will cause the signature to fail for a possibly different reason.

I'm not sure I understand why "wrong key" is any better than "bad key". There is also the possibility of testing what happens if we have the wrong kind of key (ECDSA signature but build configured for RSA).

d3zd3z (commented):
Probably a good idea to also look at `samples/zephyr/test-compile.go`. I've modified `run-tests.go` to be able to handle having all of the build artifacts zipped up. This is useful for some CI environments where the test environment doesn't have the tools installed to build Zephyr images.

The other big thing missing from this test code is that it is pretty hard-coded for the frdm-k64 board. You can define other boards, but the offsets to the partitions is hard coded.

PerMac:
Thank you all for your comments. This specification is a starting point for creating an automated testing solution for mcuboot, that can be easily implemented into some company CI but also accessible by anyone who would like to run the tests on their boards. The goal is to make a solution that is customizable: you select which board you'd like to test and won't need to change any test code. In addition, the tests should be easy to maintain, and implementing new tests should be simple.
We were thinking about using Twister from Zephyr as a framework for such tests.  I've done some analysis of Twister code/performance and here are my findings:

General description of how the Twister performs on-target tests:

The process of testing, on a higher level, is performed by `class ProjectBuilder(FilterBuilder).process()` in twisterlib.py. It processes pipelines for tests with the following stages: cmake, build, run, report, cleanup.
The run stage calls for the handle() method of DeviceHandler class which creates threads for tests. This class has several methods which would be useful for the mcuboot test: monitor_serial, get_available_device, and run_custom_script. The method handle() does the following: finds available hardware and runner, starts monitoring serial, flashes an image, and verifies the results. It also allows for running custom pre/post scripts. However, in the available mode, only one cycle of build/flash/validate is available.

For mcuboot tests, the following is needed:

- allow for building multiple images - probably as wrapping method over run_build() in ProjectBuilder calling it multiple times
- add step for images signing - maybe using run_custom_script() method
- add a feature for running multiple cycles of monitor/flash/verify within one test - probably as a class inheriting from DeviceHandler
- add tests' descriptions/procedures which Twister will follow and mechanism for Twister to "understand" them - probably expanding testcase.yaml parsing

I think the most work would be needed for expanding Twister to allow for such tests. Then, I believe, adding test cases would be more of a copy/paste effort, with changing a few parameters, like which overlay/key to use and what output is required. Twister (and west) already handles board-specific procedures and I think it would save us some effort. 

Keeping the above in mind I think I won't respond to every comment now, but I will have them in mind when implementing the test solution. I saw that there are already more interesting lines than just the last one in outputs which can be verified. Luckily, in Twister defining what to look in the output is already implemented with regex and is very easy to use (e.g. see https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/subsys/usb/audio/headset/sample.yaml)

nvlsianpu:
I am thinking about possibility to run all test cases for various MCUboot modes. This will gives Cartesian product of configurations described by test suites and these modes (with few skips as not all possibility might be supported) of tests.
Firs example of that is to run all of these test-cases for image swap methods we have supported for the board.

PerMac:
I opened a PR to zephyr upstream, which adds Twister (test runner) features and implements mcuboot on-target tests into Twister framework. I started with a configuration for nrf52840, however, I tried to make the workflow in a way that it can be easily expanded with more configurations. I decided to stick for now with the naming convention existing in run_tests.sh and Makefile (good, wrong, bad, etc. ) so it is easier to verify if everything works as in the current procedure. The convention can be changed later on.

docs/test_specification.yaml
```diff
@@ -0,0 +1,218 @@
+definitions:
```
d3zd3z, line 1: Be sure to configure your editor to not insert whitespace at the ends of lines. There are quite a few of these in this file.

nvlsianpu:
This PR seams to be obsolete.
We will add zephyr testing in other manner as this approach requires zephyr enchantment which encountered serious disagreement in the zephyr-project,

Issue #599, opened 2019-12-02
Add HSM support to imgtool


d3zd3z:
Add support for an HSM to imgtool so that the private keys can be managed, and don't have to live in the filesystem. This can be developed using softhsm. Derived from #219.

schultetwin:
I'm willing to help do the development work for this ticket! I'd suggest the following changes:

**class KeyBackend**
A new interface will be added to `imgtool` that is `KeyBackend`. It will provide the following interface:
```
class KeyBackend:
    def key_size():
        """Returns the keysize in bits. Only used for RSA keys"""
    def public_bytes(self, encoding, format):
        """Returns the bytes of the public key in the proper format and encoding"""
    def private_bytes(self, encoding, format):
        """Returns the bytes of the private key in the proper format and encoding"""
    def sign(self, payload, signature_algorithm):
        """Signs the payload. signature_type indicates how it should be signed. i.e.: ecdsa_sh256 (still need to determine how to name these)"""
    def verify(self, signature, payload, signature_algorithm):
        """Verify the signature with the payload"""
```

The current `KeyClass` objects will take a `KeyBackend` type as part of their `__init__`.

imgtool will come with a built-in KeyBackend for "python-cryptography". It will offer a mechanism to register new plugins via: https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata. This will allow anyone to create their own HSM implementation.

**CLI API***
All commands that accept a `--key` argument will also accept a `--key-storage` argument. The valid arguments to `--key-storage` would the choices from one of the `KeyBackend` plugins that are registered?

Thoughts? I'll likely put-up a PR in a few weeks, but am more than open to throw it all out and take on a different design.

d3zd3z:
This seems like a reasonable way to approach this. I would recommend having an implementation using something like softhsm that is readily available for other developers to use and test.

nmoskopp:
@d3zd3z I have added HSM support to imgtool.py using Nitrokey HSM:
https://github.com/grandcentrix/mcuboot/commits/nitrokey-hsm-support

d3zd3z:
re HSM support:

The only thing that stands out to me is the searching all of "/usr" for the hsm library. The directory can be quite large, and there is no particular reason that something being found would actually be the desired library. I'd suggest having either an environment variable or an argument. How is this normally done for PKCS#11 python apps?

Otherwise, feel free to add a pull request to the main MCUboot repo and we can take a look at it there.

[...]

PR #956, opened 2021-03-04
imgtool: enhance password support

kurddt:
Allow to pass the private key password as an option by either
providing directly using pass:<password> or via environment using
env:<variable_name>

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -24,6 +24,7 @@ import imgtool.keys as keys
 import sys
 from imgtool import image, imgtool_version
 from imgtool.version import decode_version
+import os
 from .keys import (
     RSAUsageError, ECDSAUsageError, Ed25519UsageError, X25519UsageError)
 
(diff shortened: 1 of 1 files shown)


utzig:
I don't think this is OK. There **are** reasons why passwords should not exist in the open, so if you really want it for everyone to see, why even use a password?

Btw, if you REALLY want to give a password as an environment variable or plain text, just use `sshpass` (https://linux.die.net/man/1/sshpass):

```
sshpass -p <your-plaintext-password> -P "Enter key passphrase:" imgtool create <parameters>
```

or `-e` to get the password from `$SSHPASS`. This should be enough to deal with your issue.

kurddt:
Hi thanks for your comments, the sshpass won't work in my use case. I want to sign a zephyr application during CI and I don't think it's possible to insert a command before the imgtool invocation during the build process. I could remove the plain text option and only use the environment variable possibility

utzig:
> Hi thanks for your comments, the sshpass won't work in my use case. I want to sign a zephyr application during CI and I don't think it's possible to insert a command before the imgtool invocation during the build process. I could remove the plain text option and only use the environment variable possibility

An example of using `sshpass` in a CI can be seen here: https://github.com/nrfconnect/sdk-nrf/blob/master/.github/workflows/docpublish.yml#L48. It's getting the password from the environment variable with `-e` which is what you want, so if you have access to the CI you can do it.

utzig:
> I could remove the plain text option and only use the environment variable possibility

That would be less bad for sure.

d3zd3z:
Probably some relationship to #599 here, which may or may not help with this particular case. Using something like a software hsm could allow an hsm-aware imgtool to not need to prompt for passwords.

PR #957, opened 2021-03-05
Fix mbedtls ecdsa signatures for Cypress

romanjoe:
This PR fixes recent changes, regarding Mbed TLS ECDSA for signatures introduced with this commit https://github.com/mcu-tools/mcuboot/commit/641af4530cf0de1990a539d76590b97fd0e272f2 for Cypress.


romanjoe:
@d3zd3z please review, tnx

boot/bootutil/src/image_ec256.c
```diff
@@ -163,7 +210,12 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
     pubkey = (uint8_t *)bootutil_keys[key_id].key;
     end = pubkey + *bootutil_keys[key_id].len;
 
+#ifdef CY_MBEDTLS_HW_ACCELERATION
+    mbedtls_ecdsa_init(&ctx);
+    rc = bootutil_parse_eckey(&ctx, &pubkey, end);
+#else
     rc = bootutil_import_key(&pubkey, end);
+#endif
```
d3zd3z, line 218: The context is defined by the backend-specific code. We should be able to define a context, so that this doesn't need an ifdef, and two different calls, but just two different functions.

boot/bootutil/src/image_ec256.c
```diff
@@ -178,6 +230,13 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
     /*
      * This is simplified, as the hash length is also 32 bytes.
      */
+#ifdef CY_MBEDTLS_HW_ACCELERATION
```
d3zd3z, line 233: Can we do this without the ifdefs? Or at least can we do it so that the ifdefs are in the header file instead of here.

Although, I am kind of realizing that the API surface isn't quite right to make this clean, and it just gets a bit messier each time we add one. Right now, though, having most of the ifdef blocks in `ecdsa_p256.h` and a few here isn't ideal.
The difference seems to be in terms of how a given backend wants the keys formatted. I suspect the key decoding needs to be what is conditionalized, but these are large enough that it doesn't really make sense to put it in the header file. In fact, there isn't really any benefit to putting the ifdefs in the header file (or even having it), since it is only included by a single C file.

d3zd3z (changes requested):
Other than having two functions that otherwise should really have the same signature, this is OK. I do think, in a larger picture view, we need to refactor this so that the distinction for the backends is done differently. Perhaps it would be worth looking at the PSA crypto API (as long as that isn't too complex), we could either use newer Mbed TLS, or have our own wrapper for it. Wrappers for ECC algorithms isn't going to hurt as the algorithms themselves take long enough that the extra time isn't really a factor. In fact, the whole 'inline' thing is kind of silly here.

d3zd3z:
Thinking about this, I think I'd like this to go in as is. This "abstraction" is kind of breaking down, and we need to rethink how it works, but the burden to do that shouldn't be just on someone trying to push a change through for their target.

I'll think about this and try to come up with a cleaner way to separate, based on the use cases we have now.

d3zd3z (approved):
LGTM

romanjoe:
@d3zd3z Thank you for merging this PR. Ready to support further refactoring for our target, if you come up with better separation approach.

PR #961, opened 2021-03-10
Clippy cleanups of the simulator

d3zd3z:
Clippy is a Rust tool that offers suggestions on coding style. Apply these suggestions. Most of the changes don't change the code at all. A few do make changes, but should have the same end result.


utzig (approved):
LGTM

Issue #526, opened 2019-08-10
Encrypted images use '0' for a nonce.


d3zd3z:
The method for encrypting images always uses a nonce of zero. Although we should be generating a fresh key for each encryption, there still should be a nonce generated uniquely for each image. All of the crypoanalysis of the cipher methods assume the nonce will be pseudorandomly generated, and therefore the current use is beyond this analysis.

In order to provide backward compatibility with existing images, I propose adding a TLV that can provide a nonce. If the encryptor of an image provides this, then this nonce should be given in this TLV, and used (likely added, via 64-bit add to the offset) as part of operating in counter mode.

PR #962, opened 2021-03-11
Introduce random IV and salt for encrypted image

romanjoe:
This PR introduces randomised IV (nonce) and salt as input to form HKDF data for encrypted images.

Commits:
**imgtool:** adds flag `--use-random-iv` which allows randomised data embedding to HKDF data - turned off by default to keep compatibility.
**bootutil:** changes to support update in mcuboot
**cypress:** updates to cypress code to support changes

This PR is dependent on: https://github.com/mcu-tools/mcuboot/pull/957 and https://github.com/mcu-tools/mcuboot/pull/884

### DIFF
--- a/boot/bootutil/src/encrypted.c
+++ b/boot/bootutil/src/encrypted.c
@@ -417,12 +417,16 @@ boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
 {
     int rc;
 
-    rc = bootutil_aes_ctr_set_key(&enc_state[slot].aes_ctr, bs->enckey[slot]);
+    rc = bootutil_aes_ctr_set_key(&enc_state[slot].aes_ctr, bs->encinfo[slot]);
     if (rc != 0) {
         boot_enc_drop(enc_state, slot);
         enc_state[slot].valid = 0;
         return -1;
     }
+
+    (void)memcpy(enc_state[slot].aes_iv,
+            &bs->encinfo[slot][BOOTUTIL_CRYPTO_AES_CTR_KEY_SIZE],
+            BOOTUTIL_CRYPTO_AES_CTR_BLOCK_SIZE);
 
     enc_state[slot].valid = 1;
 
@@ -430,6 +434,7 @@ boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
 }
 
 #define EXPECTED_ENC_LEN        BOOT_ENC_TLV_SIZE
+#define EXPECTED_ENC_EXT_LEN    BOOT_ENC_TLV_SIZE + BOOT_ENC_TLV_EXT_SIZE
 
 #if defined(MCUBOOT_ENCRYPT_RSA)
 #    define EXPECTED_ENC_TLV    IMAGE_TLV_ENC_RSA2048
@@ -437,23 +442,20 @@ boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
 #    define EXPECTED_ENC_TLV    IMAGE_TLV_ENC_KW
 #elif defined(MCUBOOT_ENCRYPT_EC256)
 #    define EXPECTED_ENC_TLV     IMAGE_TLV_ENC_EC256
-#    define EXPECTED_ENC_EXT_LEN (EXPECTED_ENC_LEN + BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE)
 #    define EC_PUBK_INDEX       (0)
 #    define EC_TAG_INDEX        (65)
 #    define EC_CIPHERKEY_INDEX  (65 + 32)
-_Static_assert(EC_CIPHERKEY_INDEX + BOOT_ENC_KEY_SIZE == EXPECTED_ENC_LEN,
+#    define EC_SALT_INDEX       (65 + 32 + 16)
+_Static_assert(EC_SALT_INDEX == EXPECTED_ENC_LEN,
         "Please fix ECIES-P256 component indexes");
 #elif defined(MCUBOOT_ENCRYPT_X25519)
 #    define EXPECTED_ENC_TLV    IMAGE_TLV_ENC_X25519
 #    define EC_PUBK_INDEX       (0)
 #    define EC_TAG_INDEX        (32)
 #    define EC_CIPHERKEY_INDEX  (32 + 32)
-_Static_assert(EC_CIPHERKEY_INDEX + BOOT_ENC_KEY_SIZE == EXPECTED_ENC_LEN,
+#    define EC_SALT_INDEX       (32 + 32 + 16)
+_Static_assert(EC_SALT_INDEX == EXPECTED_ENC_LEN,
         "Please fix ECIES-X25519 component indexes");
-#endif
-
-#ifndef EXPECTED_ENC_EXT_LEN
-#define EXPECTED_ENC_EXT_LEN        EXPECTED_ENC_LEN
 #endif
 
 /*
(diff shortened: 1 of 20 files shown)


utzig:
Are changes related to `CY_MBEDTLS_HW_ACCELERATION` supposed to be here as well? Is there a rationale for the change?

d3zd3z:
This will need to be rebased, now that #957 has merged.

romanjoe:
@utzig - actually, no this PR is related to #957, now when it is merged, i have rebased this one

romanjoe:
@d3zd3z thank you David, rebased this one, to master. No it is related to only one another #884. I'm taking a look on CI fails. Can you please comment on implementation as a whole, thank you!

utzig:
@romanjoe Rebase!

scripts/imgtool/image.py
```diff
@@ -288,9 +296,13 @@ def ecies_hkdf(self, enckey, plainkey):
         return cipherkey, ciphermac, pubk
 
     def create(self, key, public_key_format, enckey, dependencies=None,
-               sw_type=None, custom_tlvs=None):
+               sw_type=None, custom_tlvs=None, use_random_iv=False):
         self.enckey = enckey
 
+        if use_random_iv:
+            self.hkdf_salt = os.urandom(32)
+            self.hkdf_len  += 16 * 2   # 48 for basic scheme + 16 * 2 for random IVs
```
utzig, line 304: Remove the extra space before `+=`

utzig (changes requested):
The crypto changes look sane to me, but a few doubts:

* What is to be gained from adding a random nonce to every image generation given that the key is already randomized. If randomizing the key is not secure enough, why is adding an extra randomized parameter more secure? Also isn't this somewhat equivalent to saying AES is not secure?
* The HKDF receives as input the diffie-hellman of a "just" randomized private key, what extra security is to be gained by adding a salt in this case?
* Have you considered making `bootutil` build-time configurable? (tbh, I am not sure there's anything to be gained!)
* I don't see Simulator changes to test with the random IV, ain't it suppose to be added?

While these are not really `IV`s, I am OK with the terminology...

d3zd3z:
> What is to be gained from adding a random nonce to every image generation given that the key is already randomized. If randomizing the key is not secure enough, why is adding an extra randomized parameter more secure? Also isn't this somewhat equivalent to saying AES is not secure?

It is _probably_ secure as is. However, every analysis of the security of AES has been done with the assumption of a random IV. By not using a random IV, we're kind of venturing into new territory. Realistically, it is something we're probably going to keep running into when people try to do a security analysis of MCUboot, and adding in IVs will make that analysis easier to do.

d3zd3z:
> The HKDF receives as input the diffie-hellman of a "just" randomized private key, what extra security is to be gained by adding a salt in this case?

I think this is similar, just in that the analysis was done with salt, and people will be more comfortable with it that way.

I have received similar feedback from other sources about these issues.

boot/bootutil/src/encrypted.c
```diff
@@ -475,14 +481,22 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
 #if defined(MCUBOOT_ENCRYPT_EC256) || defined(MCUBOOT_ENCRYPT_X25519)
     bootutil_hmac_sha256_context hmac;
     bootutil_aes_ctr_context aes_ctr;
+    uint8_t salt[BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE];
     uint8_t tag[BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE];
     uint8_t shared[SHARED_KEY_LEN];
-    uint8_t derived_key[BOOTUTIL_CRYPTO_AES_CTR_KEY_SIZE + BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE];
+    uint8_t derived_key[BOOTUTIL_CRYPTO_AES_CTR_KEY_SIZE + BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE + BOOTUTIL_CRYPTO_AES_CTR_BLOCK_SIZE * 2];
```
d3zd3z, line 487: This line is getting fairly long, perhaps split after one of the `+` signs?

boot/bootutil/src/encrypted.c
```diff
@@ -586,13 +622,13 @@ boot_enc_decrypt(const uint8_t *buf, uint8_t *enckey)
 
     bootutil_hmac_sha256_init(&hmac);
 
-    rc = bootutil_hmac_sha256_set_key(&hmac, &derived_key[16], 32);
+    rc = bootutil_hmac_sha256_set_key(&hmac, &derived_key[BOOTUTIL_CRYPTO_AES_CTR_BLOCK_SIZE], BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE);
```
d3zd3z, line 625: Likewise here on the longer lines.

d3zd3z (commented):
Just some otherwise minor formatting things. I'm ok with the main concept here. I don't think this actually adds any security, but I suspect we'll keep getting pestered about it because we're not doing things the "standard" way.

Once Fabio's concerns are addressed, I can add my approval.

boot/bootutil/src/encrypted.c
romanjoe: Fixed

boot/bootutil/src/encrypted.c
romanjoe: Fixed

scripts/imgtool/image.py
romanjoe: fixed

romanjoe: @utzig Well basically all the changes came from NIST requirements. Our security experts have reviewed mcuboot code in scope of secure code review of CypressBootloader, put our attention to this implementation and requested changes. We came with same arguments as yours - what is an added value of 2 randomised datas instead of one randomised data, but it is a standard recommendation/requirement.

There are no simulator changes. Unfortunately did not have much time for implementation. We made changes backwards compatible, and it would only have effect when image signed with `--use-random-iv` flag is programmed.

romanjoe: There also an issue opened for this same feature https://github.com/mcu-tools/mcuboot/issues/526 as @d3zd3z pointed out, when we have discussed this.

nvlsianpu:
>> What is to be gained from adding a random nonce to every image generation given that the key is already randomized. If randomizing the key is not secure enough, why is adding an extra randomized parameter more secure? Also isn't this somewhat equivalent to saying AES is not secure?

>It is probably secure as is. However, every analysis of the security of AES has been done with the assumption of a random IV. By not using a random IV, we're kind of venturing into new territory. Realistically, it is something we're probably going to keep running into when people try to do a security analysis of MCUboot, and adding in IVs will make that analysis easier to do.

So purpose of this feature should be described in documentation along with note that it gives no gain as key is randomized anyway.

scripts/imgtool/image.py
```diff
@@ -267,13 +270,18 @@ def ecies_hkdf(self, enckey, plainkey):
             newpk = X25519PrivateKey.generate()
             shared = newpk.exchange(enckey._get_public())
         derived_key = HKDF(
-            algorithm=hashes.SHA256(), length=48, salt=None,
+            algorithm=hashes.SHA256(), length=self.hkdf_len, salt=self.hkdf_salt,
             info=b'MCUBoot_ECIES_v1', backend=default_backend()).derive(shared)
+        if self.hkdf_salt is not None:
+            key_nonce = derived_key[48:64]
+            self.enc_nonce = derived_key[64:76] + bytes([0] * 4)
```
utzig, line 277: While you are deriving 48 + 32 bytes of key material, here you only use 12 of the 16 remaining bytes (plus 4 zero-pad bytes), so 4 bytes are never consumed from the hkdf, why is that?

romanjoe:
@romanjoe @utzig - i have updated docs, as we agreed on a meeting

scripts/imgtool/image.py
kotkcy: We use additional 32 bytes from key material as:
- 16 bytes for key encryption IV
- 16 bytes for CTR image encryption IV (where 4 bytes are initially zero)

kotkcy: This IVs generation was made to be consistently with AES data block size (16 bytes)

PR #963, opened 2021-03-12
bootutil: Fix boot_write_trailer write alignment

de-nordic:
The commit fixes alignment issue in boot_write_trailer, that would
occur in case when input buffer length would be greater than
alignment parameter of flash.

Possible scenario:
```
let align = 4
let inlen = 7
```
ccording to the previous logic `align < inlen => align = inlen` (line 316-318). which means that
```
320: memset(&buf[inlen], erased_val, align - inlen);
```
does not write any erase value to fill in buffer because `align - inlen == 0`. and line
```
322: rc = flash_area_write(fap, off, buf, align);
```
writes unaligned buffer due to `align == inlen == 7`

Test PR for zephyr: https://github.com/zephyrproject-rtos/zephyr/pull/33292

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -294,47 +294,6 @@ boot_read_enc_key(int image_index, uint8_t slot, struct boot_status *bs)
 }
 #endif
 
-/**
- * Write trailer data; status bytes, swap_size, etc
- *
- * @returns 0 on success, != 0 on error.
- */
-static int
-boot_write_trailer(const struct flash_area *fap, uint32_t off,
-        const uint8_t *inbuf, uint8_t inlen)
-{
-    uint8_t buf[BOOT_MAX_ALIGN];
-    uint8_t align;
-    uint8_t erased_val;
-    int rc;
-
-    align = flash_area_align(fap);
-    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN) {
-        return -1;
-    }
-    erased_val = flash_area_erased_val(fap);
-    if (align < inlen) {
-        align = inlen;
-    }
-    memcpy(buf, inbuf, inlen);
-    memset(&buf[inlen], erased_val, align - inlen);
-
-    rc = flash_area_write(fap, off, buf, align);
-    if (rc != 0) {
-        return BOOT_EFLASH;
-    }
-
-    return 0;
-}
-
-static int
-boot_write_trailer_flag(const struct flash_area *fap, uint32_t off,
-        uint8_t flag_val)
-{
-    const uint8_t buf[1] = { flag_val };
-    return boot_write_trailer(fap, off, buf, 1);
-}
-
 int
 boot_write_copy_done(const struct flash_area *fap)
 {
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -251,6 +251,10 @@ int boot_write_image_ok(const struct flash_area *fap);
 int boot_write_swap_info(const struct flash_area *fap, uint8_t swap_type,
                          uint8_t image_num);
 int boot_write_swap_size(const struct flash_area *fap, uint32_t swap_size);
+int boot_write_trailer(const struct flash_area *fap, uint32_t off,
+                       const uint8_t *inbuf, uint8_t inlen);
+int boot_write_trailer_flag(const struct flash_area *fap, uint32_t off,
+                            uint8_t flag_val);
 int boot_read_swap_size(int image_index, uint32_t *swap_size);
 int boot_slots_compatible(struct boot_loader_state *state);
 uint32_t boot_status_internal_off(const struct boot_status *bs, int elem_sz);
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -334,7 +334,7 @@ boot_write_magic(const struct flash_area *fap)
  *
  * @returns 0 on success, != 0 on error.
  */
-static int
+int
 boot_write_trailer(const struct flash_area *fap, uint32_t off,
         const uint8_t *inbuf, uint8_t inlen)
 {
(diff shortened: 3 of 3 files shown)


boot/bootutil/src/bootutil_misc.c
```diff
@@ -309,13 +309,13 @@ boot_write_trailer(const struct flash_area *fap, uint32_t off,
...
-    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN) {
+    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN ||
+	BOOT_MAX_ALIGN % align) {
         return -1;
     }
     erased_val = flash_area_erased_val(fap);
-    if (align < inlen) {
-        align = inlen;
-    }
+    align = ((inlen + align - 1) / align) * align;
```
d3zd3z, line 317: The division seems kind of heavy here, but I suppose this code doesn't run that frequently. We could ensure somewhere that the alignment is a power of two (it doesn't really make sense for it to not be), and then change this to `((inline + align - 1) & ~(align - 1)`, but in this instance, it is probably not worth doing that.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -309,13 +309,13 @@ boot_write_trailer(const struct flash_area *fap, uint32_t off,
     int rc;
 
     align = flash_area_align(fap);
-    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN) {
+    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN ||
+	BOOT_MAX_ALIGN % align) {
```
d3zd3z, line 313: I'd be more comfortable with: `.. || BOOT_MAX_ALIGN % align != 0` just to make it clearer. But this is ok as well.

utzig:
This function is `static` and was refactored from a code that only ever writes up to the alignment size, so in practice there is no chance it would ever write more than it should, or do you think it would @de-nordic? Btw, I did not know there was a copypaste in `bootutil_public.c`, maybe should do some refactoring to merge the copypaste from those two files into some new file...

boot/bootutil/src/bootutil_misc.c
utzig: Probably more efficient in runtime using `BOOT_MAX_ALIGN & (align - 1)`

boot/bootutil/src/bootutil_misc.c
de-nordic: Will change.

boot/bootutil/src/bootutil_misc.c
de-nordic: Will change

de-nordic:
> 
> This function is `static` and was refactored from a code that only ever writes up to the alignment size, so in practice there is no chance it would ever write more than it should, or do you think it would @de-nordic?

There is parameter check for `inlen` vs `BOOT_MAX_ALIGN ` so I guess someone has predicted possibility of getting data longer than alignment buffer by accident. The function is there so there is probability that someone would finally give it `inlen` that would be accepted by first check, but would be unaligned write... and would probably fix it the same way finally.
The extra arithmetic does not add that much weight to the function anyway: there is syscall to take flash parameters and perform write, and with the function not being that much popular the time spent on extra check can be neglected.

In other perspective: if we are sure that `inlen` would never be bigger than the `BOOT_MAX_ALIGN ` and if `BOOT_MAX_ALIGN` could be declared high enough to cover every possible write alignment for a time, we could probably neglect all the checks except for the `align` vs `inlen` (to get the proper write len). Similar code is in `boot_write_status` and nobody bothers to check if `BOOT_MAX_ALIGN` can take align

We should probably check all the flash parameters before the first write is even done so that the problem could be reported early,
The param check if in `boot_write_trailer` gives you -1 for combination of 3 possible reasons.

> Btw, I did not know there was a copypaste in `bootutil_public.c`, maybe should do some refactoring to merge the copypaste from those two files into some new file...

Yes they probably should. It seems that there is a lot of shared code between bootutil_public.c and bootutil_misc.c.
I do not know why, so I have not taken upon myself to make the code common in this PR.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -309,13 +309,13 @@ boot_write_trailer(const struct flash_area *fap, uint32_t off,
     int rc;
 
     align = flash_area_align(fap);
-    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN) {
+    if (inlen > BOOT_MAX_ALIGN || align > BOOT_MAX_ALIGN ||
+	(BOOT_MAX_ALIGN & (align - 1)) == 0) {
```
utzig, line 313: It is `!=`, not `==`, also there is tab here, MCUboot uses spaces, please fix, same applies on the other file.

utzig (changes requested):

utzig (changes requested):
Please remove `boot/bootutil/cscope.out`

de-nordic:
> 
> 
> Please remove `boot/bootutil/cscope.out`

Ain't my day.

boot/bootutil/src/bootutil_misc.c
utzig: Sorry about this, but it was a wrong suggestion on my part; that does not work, the correct way to do the same as the modulo that you done originally is:

```
... || (align <= BOOT_MAX_ALIGN && (align & (align - 1)) == 0) {
```

I am not sure that is really something that needs protection against though...

utzig: Not my day as well I guess, just need to add this extra || test `(align & (align - 1)) != 0`...

nvlsianpu (commented):
> Btw, I did not know there was a copypaste in bootutil_public.c, maybe should do some refactoring to merge the copypaste from those two files into some new file...

I recommend to keep bootutil_public.c implementation, make it global and use in bootutil_misc.c

nvlsianpu (approved):

utzig (approved):
LGTM

PR #964, opened 2021-03-22
boot/zephyr: cleanup GPIO pin setup

nvlsianpu:
Removed configuration of gpio pin using flags which were
removed for zephyr-project long ago. Also old method of read
the pin value was removed.
No sense to keep that as nowadays MCUboot is also not compatible
with zephyr version which supports these flags and methods.


nvlsianpu:
@de-nordic

utzig (approved):
LGTM

de-nordic (approved):

PR #968, opened 2021-03-24
cypress: Add support for PSoC 62 1M and 512K chips, refactor makefiles

romanjoe:
This PR introduces support for 1M and 512K chips of PSoC 62 families.

* makefiles build system refactored
* psoc6pdl is substituted to mtb-pdl-cat1 repo, core-lib updated
* added support in build system and code for 1M and 512K chips
* added support for external memory for 1M and 512K
* only OVERWRITE mode supported for all families
* documentation updated with 512K and 1M details

@d3zd3z please review, thanks!


utzig (approved):
LGTM

PR #979, opened 2021-04-13
cmake: fix evaluation with empty signature key file

niclas-ats:
When CONFIG_BOOT_SIGNATURE_KEY_FILE is left empty, the cmake find command gets
called with a wrong number of arguments. This patch skips the search for the
signature key file within config files in that case.

Signed-off-by: Nico Lastzka <Nico.Lastzka@ATSonline.de>


nvlsianpu:
cc @MarekPieta

MarekPieta (approved):

d3zd3z (approved):

PR #982, opened 2021-04-23
RFC: Replace direct access to flash_area/flash_sector fields with getter functions

de-nordic:
Intention of this PR is to present an idea that should allow the mcuboot to become system agnostic regarding
the internal structure of flash_area and flash_sector objects.
The idea is to provide getter/setter macros (here) or functions that should only be used to access the fields of objects
and forbid direct access to fields.
Such approach should allow greater flexibility regarding management and processing of such objects as it should allow additional, system processing of returned values, which, after obtaining an object (for example flash_area) is not possible,

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -282,11 +282,11 @@ bs_upload(char *buf, int len)
 
     if (off == 0) {
         curr_off = 0;
-        if (data_len > fap->fa_size) {
+        if (data_len > flash_area_get_size(fap)) {
             goto out_invalid_data;
         }
 #ifndef CONFIG_BOOT_ERASE_PROGRESSIVELY
-        rc = flash_area_erase(fap, 0, fap->fa_size);
+        rc = flash_area_erase(fap, 0, flash_area_get_size(fap));
         if (rc) {
             goto out_invalid_data;
         }
@@ -316,10 +316,11 @@ bs_upload(char *buf, int len)
         BOOT_LOG_ERR("Unable to determine flash sector size");
         goto out;
     }
-    if (off_last != sector.fs_off) {
-        off_last = sector.fs_off;
-        BOOT_LOG_INF("Erasing sector at offset 0x%x", sector.fs_off);
-        rc = flash_area_erase(fap, sector.fs_off, sector.fs_size);
+    if (off_last != flash_sector_get_off(&sector)) {
+        off_last = flash_sector_get_off(&sector);
+        BOOT_LOG_INF("Erasing sector at offset 0x%x", flash_sector_get_off(&sector));
+        rc = flash_area_erase(fap, flash_sector_get_off(&sector),
+                              flash_sector_get_size(&sector));
         if (rc) {
             BOOT_LOG_ERR("Error %d while erasing sector", rc);
             goto out;
@@ -367,9 +368,11 @@ bs_upload(char *buf, int len)
             }
             /* Assure that sector for image trailer was erased. */
             /* Check whether it was erased during previous upload. */
-            if (off_last < sector.fs_off) {
-                BOOT_LOG_INF("Erasing sector at offset 0x%x", sector.fs_off);
-                rc = flash_area_erase(fap, sector.fs_off, sector.fs_size);
+            if (off_last < flash_sector_get_off(&sector)) {
+                BOOT_LOG_INF("Erasing sector at offset 0x%x",
+                             flash_sector_get_off(&sector));
+                rc = flash_area_erase(fap, flash_sector_get_off(&sector),
+                                      flash_sector_get_size(&sector));
                 if (rc) {
                     BOOT_LOG_ERR("Error %d while erasing sector", rc);
                     goto out;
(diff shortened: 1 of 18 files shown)


de-nordic:
@davidvincze I guess from the test logs that I need to work out the RFC pr for the changes to the https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/ too (I have already cloned the repo and seen that there is an independent flash_area definitions), although I have a question how do I redirect this PR to use my other PR or fork wit changes for the testing with Travis?

nvlsianpu:
> I guess from the test logs that I need to work out the RFC pr for the changes to the https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/ too (I have already cloned the repo and seen that there is an independent flash_area definitions), although I have a question how do I redirect this PR to use my other PR or fork wit changes for the testing with Travis?

@RcColes Can you help?

nvlsianpu (approved):

utzig (approved):

RcColes:
> > I guess from the test logs that I need to work out the RFC pr for the changes to the https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/ too (I have already cloned the repo and seen that there is an independent flash_area definitions), although I have a question how do I redirect this PR to use my other PR or fork wit changes for the testing with Travis?
> 
> @RcColes Can you help?

Hmm, it seems like there's a chicken and egg scenario going on here. There needs to be some changes to TF-M in order for CI to pass, but then again we can't make changes to TF-M related to this patch until this patch is merged. I've seen this in one other issue before, but thankfully we were able to resolve by tweaking the patch to make it not an interface change.

I'm not sure what the long-term solution to this is. Is it possible that we could make the FIH testsuite indicative rather than required? We'd still need to put the patch into TF-M shortly afterwards (because then the FIH testsuite would be failing on mcuboot master), but this would break the deadlock around which patch needs to come first.

So the proposed order would be:
* Commit to mcuboot despite CI failing
* Create patch to TF-M
* Merge patch to TF-M
* Update FIH testsuite to use new version of TF-M

Though of course the testsuite failing for any period of time isn't ideal still. Thoughts?

utzig:
> Though of course the testsuite failing for any period of time isn't ideal still. Thoughts?

My counter-proposal:

1) Switch to https://github.com/mcu-tools/trusted-firmware-m which I will set up to mirror the upstream repo from ARM
2) Create a new branch called "mcuboot", or something like that, that starts on the sha that was is currently used in ARM's tfm.
3) Now local work. Submit PR to be merged on this tfm branch.
4) As part of the PR update local script to either use the branch or fetch and run on a PR in a process similar to what is used with west on Zephyr.
5) Run the local PR with the PR in our own tfm fork.
6) Once it passes, merge the PR on the local fork of tfm in the "mcuboot" branch; update the mcuboot PR back to the tip of the mcuboot branch
7) Merge here.

Out of scope but relevant would be that those PRs are upstreamed to ARM's tfm. And eventually we force push to a new sha when upstream tfm patches are merged. So this basically avoids the locks. Thoughts?

RcColes:
> > Though of course the testsuite failing for any period of time isn't ideal still. Thoughts?
> 
> My counter-proposal:
> 
>     1. Switch to https://github.com/mcu-tools/trusted-firmware-m which I will set up to mirror the upstream repo from ARM
> 
>     2. Create a new branch called "mcuboot", or something like that, that starts on the sha that was is currently used in ARM's tfm.
> 
>     3. Now local work. Submit PR to be merged on this tfm branch.
> 
>     4. As part of the PR update local script to either use the branch or fetch and run on a PR in a process similar to what is used with west on Zephyr.
> 
>     5. Run the local PR with the PR in our own tfm fork.
> 
>     6. Once it passes, merge the PR on the local fork of tfm in the "mcuboot" branch; update the mcuboot PR back to the tip of the mcuboot branch
> 
>     7. Merge here.
> 
> 
> Out of scope but relevant would be that those PRs are upstreamed to ARM's tfm. And eventually we force push to a new sha when upstream tfm patches are merged. So this basically avoids the locks. Thoughts?

Yes that sounds like a better idea. I'm wary of forks, but I guess in this case it makes the most sense of the options. We've just got to be careful that we make changes to the fork that they get upstreamed in a sensible timeframe.

de-nordic:
@utzig @RcColes 
Thanks for your input on the case. It seems that my change causes a lot of headache.
I could not figure out how the labeling in tfm goes and how it is tied to mcoboot master but I guess that is some kind of higher configuration management magic.

Anyway, since this is a lot of headache, to make it worth I would kindl  ask everyone that can to look at the idea and put it under merciless scrutiny. Thanks in advance.

d3zd3z (commented):
One minor comment, the 3 commits that modify each platform only differ in the first word of the patch subject. It would probably be nice to add a paragraph to the commit text that says something like: "This commit modifies the Zephyr files." or something like that.

de-nordic:
> 
> One minor comment, the 3 commits that modify each platform only differ in the first word of the patch subject. It would probably be nice to add a paragraph to the commit text that says something like: "This commit modifies the Zephyr files." or something like that.

@d3zd3z Sure. I was considering squashing them together into single commit (there are 4 of them actually), but I do not know whether it is ok to place changes to different platforms into single commit.

d3zd3z:
> I do not know whether it is ok to place changes to different platforms into single commit.

Overall, I don't think it matters all that much. We haven't focused on that kind of separation before. So, feel free to squash them if you'd like. It is good, though, to keep logically distinct commits separate.

de-nordic:
> 
> 
> > I do not know whether it is ok to place changes to different platforms into single commit.
> 
> Overall, I don't think it matters all that much. We haven't focused on that kind of separation before. So, feel free to squash them if you'd like. It is good, though, to keep logically distinct commits separate.

OK, I have squashed the changes, for mcuboot ports, into single commit.

d3zd3z (approved):

PR #984, opened 2021-04-27
Mbed: Make scratch BlockDevice customizable

boraozgen:
This change makes it possible to provide a custom scratch region BlockDevice, such as an external flash.

@AGlass0fMilk

### DIFF
--- a/boot/mbed/src/secondary_bd.cpp
+++ b/boot/mbed/src/secondary_bd.cpp
@@ -17,17 +17,13 @@
  * limitations under the License
  */
 
-#if MCUBOOT_DIRECT_XIP
-
 #include "flash_map_backend/secondary_bd.h"
 #include "platform/mbed_toolchain.h"
 #include "FlashIAPBlockDevice.h"
+#include "bootutil_priv.h"
 
-/**
- * For an XIP build, the secondary BD is provided by mcuboot by default.
- *
- * This is a weak symbol so the user can override it.
- */
+#if MCUBOOT_DIRECT_XIP
+
 MBED_WEAK mbed::BlockDevice* get_secondary_bd(void) {
     static FlashIAPBlockDevice secondary_bd(MBED_CONF_MCUBOOT_XIP_SECONDARY_SLOT_ADDRESS,
             MCUBOOT_SLOT_SIZE);
@@ -37,4 +33,12 @@ MBED_WEAK mbed::BlockDevice* get_secondary_bd(void) {
 
 #endif
 
+#if MCUBOOT_SWAP_USING_SCRATCH
 
+MBED_WEAK mbed::BlockDevice* get_scratch_bd(void) {
+    static FlashIAPBlockDevice scratch_bd(MCUBOOT_SCRATCH_START_ADDR, MCUBOOT_SCRATCH_SIZE);
+
+    return &scratch_bd;
+}
+
+#endif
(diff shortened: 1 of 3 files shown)


boraozgen:
I actually wanted to ask if external scratch could cause any issues, as I never stumbled upon such an example. Thanks for the remark about encrypted images. Maybe this could be documented somewhere as a limitation.

AGlass0fMilk:
> Thanks for the remark about encrypted images. Maybe this could be documented somewhere as a limitation.

The best thing to do would be to make it a mutually-exclusive configuration option. This prevents developers from unknowingly enabling a potential security loophole. Unfortunately you can't rely on a developer reading a documentation footnote for something like this :sweat_smile: 

Perhaps we could introduce a new configuration option in the `mbed_lib.json` that allows you to retarget the scratch block device by overriding a function, similar to how the update/secondary block device is provided to Mcuboot now. Enabling this new option _and_ encrypted firmware updates at the same time can cause a compile-time error.

boraozgen:
I believe this should be documented in mcuboot core level, since this is not a port related consideration.

AGlass0fMilk:
> I believe this should be documented in mcuboot core level, since this is not a port specific consideration.
> 
> > Unfortunately you can't rely on a developer reading a documentation footnote for something like this
> 
> They better read the documentation if they are implementing an advanced security feature like this :)

While I agree with both points, I still think the compiler should be made to warn users at least. The hope is that mcuboot will become easier to use in Mbed-OS as integration is refined. Ease-of-use means more developers will just assume things work...

Maybe I'm a bit cynical but I think we should add a compile-time warning/error if the scratch BlockDevice overwrite option is enabled and encrypted images is enabled at the same time.

Do you have any thoughts on this @LDong-Arm?

AGlass0fMilk:
> @AGlass0fMilk How would you implement this check? Currently it is not a configuration option, and I don't know of a way to determine if the weak function has been overridden or not. Adding a configuration option just for warning purposes seems unnecessary to me.

I see your point. More configuration (especially just for warnings) will surely lead to configuration hell.

How about this: Add a disclaimer to the docstring comment of the weak `get_scratch_bd` function that says putting the scratch region on an external memory may cause security loopholes. Maybe include the information that @d3zd3z posted above.

Docstring comments are much likelier to be read/glanced at than documentation...

boot/mbed/src/secondary_bd.cpp
```diff
@@ -37,4 +38,17 @@ MBED_WEAK mbed::BlockDevice* get_secondary_bd(void) {
 
 #endif
 
+#if MCUBOOT_SWAP_USING_SCRATCH
+
+/**
+ * By default, scratch space is at the end of internal flash, after the main application.
+ *
+ * This is a weak symbol so the user can override it.
```
AGlass0fMilk, line 46: Add security warning here about putting scratch region on external flash.

AGlass0fMilk (changes requested):

boot/mbed/src/secondary_bd.cpp
boraozgen: Done

AGlass0fMilk:
Nit: can you combine the docstring comments and just put all the info above the declaration in the header? Typically I just put docstrings in the header.

Sorry for the small changes 😅

boraozgen:
You're right, how about this?

PR #985, opened 2021-04-27
Mbed: Remove unused header size config

boraozgen:
Removing the header size configuration which apparently does not do anything.

What was the reason behind this parameter? @AGlass0fMilk

### DIFF
--- a/boot/mbed/mbed_lib.json
+++ b/boot/mbed/mbed_lib.json
@@ -25,12 +25,6 @@
             "help": "Size of the scratch area, in bytes. If needed, please set on a per-target basis.",
             "macro_name": "MCUBOOT_SCRATCH_SIZE"
         },
-        "header-size": {
-            "help": "Header size, in bytes, prepended to the bootable application image. Should be one or multiple times the sector size.",
-            "macro_name": "MCUBOOT_HEADER_SIZE",
-            "required": true,
-            "value": 4096
-        },
         "validate-primary-slot": {
             "help": "Always check the signature of the image in the primary slot before booting, even if no upgrade was performed. This is recommended if the boot time penalty is acceptable.",
             "macro_name": "MCUBOOT_VALIDATE_PRIMARY_SLOT",


d3zd3z:
Do you know how the header size is getting set on the images? In Zephyr at least, this needs to be known by Zephyr, as it needs to be able to be run with the header, and imgtool needs it, in order to properly pad the image (and write the header).

boraozgen:
In Mbed OS, we set the application start address, which must be (header address + header size). The OS does not know about the bootloader or the header, and cares only about the start address of the application. The responsibility to match the addresses are left to the developer.

utzig:
@boraozgen You have to sign off your commit to pass the CI.

boraozgen:
> @boraozgen You have to sign off your commit to pass the CI.

Will do.

AGlass0fMilk:
I don't think this should be merged. The header size is used in the bootloader.

Early on in the Mbed-OS port of Mcuboot, I had minimal configuration options to try to prevent invalid configuration headaches. When @LDong-Arm joined in the effort, he added more configuration options so that developers could explicitly specify their memory configuration.

This is more flexible in that advanced users can accommodate other parts of internal flash that may be used for things like non-volatile settings storage, etc.

I think the header size declaration is necessary to keep, especially for the bootloader build.

boraozgen:
> The header size is used in the bootloader.

Where is it used exactly?

LDong-Arm:
I think it's safe to remove this config:
* When signing an image, the header size is specified with `--header-size` so MCUboot knows it. Presumably the header size is tied to the image instead of the bootloader?
* The application inside an image has `target.mbed_app_start` set so the code starts after the image header.

boraozgen:
@LDong-Arm exactly. My comment [here](https://github.com/AGlass0fMilk/mbed-mcuboot-demo/issues/10#issuecomment-827570349) is also relevant to this topic.

utzig:
> * When signing an image, the header size is specified with `--header-size` so MCUboot knows it. Presumably the header size is tied to the image instead of the bootloader?

Yes, the header size is part of the header of the image itself; the bootloader does not have an image size setting per se.

AGlass0fMilk:
> > The header size is used in the bootloader.
> 
> Where is it used exactly?

A quick search of this repo for "MCUBOOT_HEADER_SIZE" shows only the `mbed_lib.json` file so I would say this is safe to remove.

I can't recall exactly why this option was added when I first did the Mbed-OS port. I vaguely recall mcuboot at the time requiring it?

Otherwise, the Mbed-OS port used to try and take care of all the configuration math for you. Instead of allowing you to configure things like the application start and size and all that, it would calculate it from these configurations.

Regardless, I agree this can be removed. Thanks @boraozgen!

utzig (approved):

PR #987, opened 2021-04-29
boot: Add AES256 support for image encryption

salomethirot-arm:
Support only works when using mbedtls as the cryptographic library.

Signed-off-by: Salome Thirot <salome.thirot@arm.com>


utzig:
@salomethirot-arm Please also add the simulator changes and enable it on GH actions.

boot/bootutil/include/bootutil/image.h
```diff
@@ -89,7 +89,7 @@ struct flash_area;
 #define IMAGE_TLV_RSA3072_PSS       0x23   /* RSA3072 of hash output */
 #define IMAGE_TLV_ED25519           0x24   /* ed25519 of hash output */
 #define IMAGE_TLV_ENC_RSA2048       0x30   /* Key encrypted with RSA-OAEP-2048 */
-#define IMAGE_TLV_ENC_KW128         0x31   /* Key encrypted with AES-KW-128 */
+#define IMAGE_TLV_ENC_KW            0x31   /* Key encrypted with AES-KW 128 or 256*/
```
d3zd3z, line 92: As mentioned by @utzig these changes should be matched in the simulator code, along with additional tests for the new algorithm.

d3zd3z (approved):
Excellent work. Thank you for the simulator support.

salomethirot-arm:
Thanks. Just want to point out that @utzig just asked me to add the following test to the workflow: sig-ecdsa-mbedtls + enc-aes256- ec256 + primary-slot-validate, and after testing it locally I found some failures, you might want them to be resolved before merging this.

utzig:
> Thanks. Just want to point out that @utzig just asked me to add the following test to the workflow: sig-ecdsa-mbedtls + enc-aes256- ec256 + primary-slot-validate, and after testing it locally I found some failures, you might want them to be resolved before merging this.

Hmmm, try changing your `build.rs` with `MCUBOOT_SWAP_SAVE_ENCTLV` disabled and see if that works.

salomethirot-arm:
> > Thanks. Just want to point out that @utzig just asked me to add the following test to the workflow: sig-ecdsa-mbedtls + enc-aes256- ec256 + primary-slot-validate, and after testing it locally I found some failures, you might want them to be resolved before merging this.
> 
> Hmmm, try changing your `build.rs` with `MCUBOOT_SWAP_SAVE_ENCTLV` disabled and see if that works.

No I just forgot to add the aes_key_size argument to some functions in tlv.rs. I'm about to update the PR with the correction

sim/src/tlv.rs
```diff
@@ -535,15 +589,21 @@ impl ManifestGen for TlvGen {
...
-                    let okm = match prk.expand(&[b"MCUBoot_ECIES_v1"], OkmLen(48)) {
+                    let okm = if aes256 { match prk.expand(&[b"MCUBoot_ECIES_v1"], OkmLen(64)) {
                         Ok(okm) => okm,
                         Err(_) => panic!("Failed building HKDF OKM"),
-                    };
-                    let mut buf = [0u8; 48];
+                    } } else { match prk.expand(&[b"MCUBoot_ECIES_v1"], OkmLen(48)) {
+                        Ok(okm) => okm,
+                        Err(_) => panic!("Failed building HKDF OKM"),
+                    } };
```
utzig, line 605: I think if you added something like `let okm_len = if aes256 { 64 } else { 48 };` in the line above `let okm = ...`, you could just use that value and this block would not need to be duplicated.

sim/src/tlv.rs
```diff
@@ -535,15 +589,21 @@ impl ManifestGen for TlvGen {
...
+                    let mut buf = if aes256 {vec![0u8; 64] } else { vec![0u8; 48] };
```
utzig, line 606: nit, missing space `{ vec![0u...`

sim/src/tlv.rs
```diff
@@ -569,13 +636,15 @@ impl ManifestGen for TlvGen {
             buf.append(&mut cipherkey);
 
             if self.kinds.contains(&TlvKinds::ENCEC256) {
-                assert!(buf.len() == 113);
+                let size = if aes256 { 129 } else { 113 };
+                assert!(buf.len() == 129 || buf.len() == 113);
```
utzig, line 640: Could not this be `assert!(buf.len() == size)` since you are already know which size it's supposed to be?

sim/src/image.rs
```diff
@@ -1294,17 +1296,26 @@ fn install_image(flash: &mut SimMultiFlash, slot: &SlotInfo, len: usize,
     tlv.add_bytes(&b_img);
 
     // Generate encrypted images
-    let flag = TlvFlags::ENCRYPTED as u32;
-    let is_encrypted = (tlv.get_flags() & flag) == flag;
+    let flag = TlvFlags::ENCRYPTED_AES128 as u32 | TlvFlags::ENCRYPTED_AES256 as u32;
+    let is_encrypted = ((tlv.get_flags() & flag) == TlvFlags::ENCRYPTED_AES128 as u32)
+                    || ((tlv.get_flags() & flag) == TlvFlags::ENCRYPTED_AES256 as u32);
```
utzig, line 1301: A nit since the code is already there, but this would be easier to do by just `(tlv.get_flags() & flag) != 0`, which would work for any number of flags, or a single one as well...

sim/mcuboot-sys/build.rs
```diff
@@ -280,18 +293,42 @@ fn main() {
...
         conf.define("MBEDTLS_CONFIG_FILE", Some("<config-ec.h>"));
     } else if (sig_ecdsa || enc_ec256) && !enc_kw {
         conf.define("MBEDTLS_CONFIG_FILE", Some("<config-asn1.h>"));
     } else if sig_ed25519 || enc_x25519 {
         conf.define("MBEDTLS_CONFIG_FILE", Some("<config-asn1.h>"));
-    } else if enc_kw {
+    } else if enc_kw || enc_aes256_kw {
         conf.define("MBEDTLS_CONFIG_FILE", Some("<config-kw.h>"));
+    } else if enc_aes256_x25519 {
+        conf.define("MBEDTLS_CONFIG_FILE", Some("<config-ed25519.h>"));
```
utzig, line 331: A bit funny that `sig_ed25519` uses `config-asn1.h` and `enc_aes256_x25519` uses `config-ed25519.h`. No need to change anything here, but this should be reviewed in the future...

utzig (changes requested):
This is pretty good work! I left a few comments which are basically nits. I'd suggest that future contributions break the PR in commits where there is one for changes to bootutil+docs, one for imgtool, one for the simulator, one for the CI, etc, but please don't do it now since it will mess the review!

docs/encrypted_images.md
```diff
@@ -94,17 +95,17 @@ libraries. The whole key encryption can be summarized as:
 * Derive the new keys from the secret using HKDF (built on HMAC-SHA256). We
   are not using a `salt` and using an `info` of `MCUBoot_ECIES_v1`, generating
   48 bytes of key material.
-* A new random encryption key of 16 bytes is generated (for AES-128). This is
+* A new random encryption key is generated (for AES). This is
   the AES key used to encrypt the images.
-* The key is encrypted with AES-128-CTR and a `nonce` of 0 using the first
-  16 bytes of key material generated previously by the HKDF.
+* The key is encrypted with AES-128-CTR or AES-256-CTR and a `nonce` of 0 using
+  the first 16 bytes of key material generated previously by the HKDF.
```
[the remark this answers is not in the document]
salomethirot-arm: 16, the size of the nonce doesn't change.

sim/mcuboot-sys/build.rs
salomethirot-arm: Yes I agree, enc_aes256_x25519 uses mbedtls contrary to sig_x25519, ideally it should be mentioned in the name of the feature to reduce confusion but I didn't want to end-up with a very long name so I left it like this

utzig (approved):
LGTM, well done!

PR #989, opened 2021-05-03
Serial recovery features

wcappelle:
I recently made some local changes to the mcuboot-zephyr sources for allowing me to use the mcuboot on a STM32 Cortex-M0 where the serial recovery mode will be used to enter the mcboot and to transfer encrypted images. 
Therefore I made some patches which are split up as following:

* fix on unaligned memory access on M0 targets wrt serial mcumgr commands
* Adding echo support for the echo command of mcumgr
* Adding a feature where the echo command (within a timeout) can be used as a trigger to stay in the mcuboot, as for now this is only possible with a GPIO pin.
* Allow a cached validation result of the main partition (now this is either always, or never). on a M0 this takes some significant time in the boot process.
* Adding support to allow encrypted images to be send & decrypted over the serial recovery methods in a single application context. This has been done by decryption in place, and is not power failsafe, but can always be redone when failed. Anyhow serial recovery isn't power failsafe, so don't think this is a bad thing. On a STM32L0@32MHz with 100kb of imagesize, this process takes about 2 minutes to complete (transfer included 9600baud)

It's my first upload to the community, so please be kind for any newbies mistakes :-)


utzig:
@wcappelle The autogeneration of the encryption key into code was already discussed many times before, so I doubt this will be OK. Regarding style, we use spaces outside parenthesis e not inside so `if( x )` is wrong. and `if (x) ` is correct, please fix style.

wcappelle:
@utzig , thanks for the feedback. If these changes for encryption key generation will not be ok, i can always remove this if asked. 
For the code guidelines, is there a .clang-format file available for these to be applied immediately? I've done the changes on the code style as suggested.

mateo-ligne-netatmo:
Hello everyone, 
I will be interested by the "decryption in place" part of this pull request. What is the current status of this point ? Are you waiting for suggested modifications following may 11th meeting ? Thanks for answers !

wcappelle:
@mateo-ligne-netatmo, I have no idea what the status is, I was expecting a review, but I have no idea how this need to proceed. 
@utzig , any feedback would be welcome?
I've seen on the PR #1088, some similar changes will be done on the key generation (more extended, and not limited to only EC256), do if these are merged, I will merge these changes in.

de-nordic:
@wcappelle can you maybe separate "Fix unaligned buffer access in serial boot mode" to other PR; I do not think that it is good idea to mix fixes and features, and the fix seem quite straightforward so it will probably be accepted quicker.

You also seem to mix several features in one PR which makes it harder to review.

Regarding the addition of echo command: you need to add some configuration option to turn the feature off at compile time, not all uses of mcuboot require this feature but it adds code.

I am ok with adding "Regarding the "Add feature for optional mcumgr ping command within a timeout to enter" but why just not rely on already existing SMP processing in boot_serial.c? The code is already there, you just need to make some mods to make it work for your purpose. 

The "Add support for signed images in single loader mode" should be given own PR because it is more security related and will require closer look at do to having a lot of code and being quite complicated, at first glance.

boot/boot_serial/src/boot_serial.c
```diff
@@ -89,7 +89,8 @@ MCUBOOT_LOG_MODULE_DECLARE(mcuboot);
 #endif
 
 static char in_buf[BOOT_SERIAL_INPUT_MAX + 1];
-static char dec_buf[BOOT_SERIAL_INPUT_MAX + 1];
+static char dec_buf[BOOT_SERIAL_INPUT_MAX + 1] __attribute__((aligned));
+static char img_buf[BOOT_SERIAL_INPUT_MAX + 1] __attribute__((aligned));
```
nvlsianpu: I would see this intermediate buffer here only for architectures which requires it. Perfectly this can be added in #ifdef xx #endif. xx should be defined if an arch requires it.

Alternatiovelly flash_area_write() might be added in 3 steps: 1st for unaligned buffer beginning, 2nd for aligned middle buffer part and 3rd for unaligned ending part.

boot/boot_serial/src/boot_serial.c
```diff
@@ -421,6 +421,34 @@ bs_upload(char *buf, int len)
     flash_area_close(fap);
 }
 
+static void
```
nvlsianpu, line 450: This should be added as optional. Also worth to know that the protocol will respond to unknown command with error code - so is echo really needed?

boot/boot_serial/src/boot_serial.c
```diff
@@ -446,6 +447,7 @@ bs_echo(char *buf, int len)
     cbor_encode_text_string(&bs_rsp, echo_buf, echo_buflen);
     cbor_encoder_close_container(&bs_root, &bs_rsp);
     boot_serial_output();
+    bs_echoreceived = 1;
```
nvlsianpu, line 475: maybe you can use for that any command we already have, for instance the `image list` ( bs_list()) command. I think no obstacles to do that.

wcappelle:
Thanks for the review/comments @nvlsianpu  & @de-nordic .
I will re-arrange the PR & split it in multiple pieces.
It also seems I can remove the unaligned access fix since it was only fixing the lack of support of unaligned access in the flash_area_write function, which is in the meantime fixed in the place it should have been fixed in the first place (stm32 flash driver in zephyr).

Give me some time to finish it, so i'll see if I still find some time to split them all before the holidays.

nvlsianpu:
> I will re-arrange the PR & split it in multiple pieces.
Great!

> It also seems I can remove the unaligned access fix since it was only fixing the lack of support of unaligned access in the flash_area_write function, which is in the meantime fixed in the place it should have been fixed in the first place (stm32 flash driver in zephyr)

I was also thinking that this might be better solution as things like unaligned buffer access are everywhere.

wcappelle:
first 2 commits are split here:
#1253  
#1254 
#1255
The DFU timeout depends on 1253, so will be added when that one is completed

boot/boot_serial/src/boot_serial.c
d3zd3z: At least with the Zephyr build system, values that aren't referenced won't be included in the build. I'm not sure it is worth adding conditional compilation, if the linker is just going to remove the symbols and data anyway.

wcappelle:
closing since all sub PR's have been merged in.

PR #995, opened 2021-05-07
bootutil_public: allow to confirm padded image without copy-done flag

nvlsianpu:
This patch is for fixing:
https://github.com/zephyrproject-rtos/zephyr/issues/34683

copy-done flag is used to design completely copied image.
It is used to recognize whether need to complete image upgrade.
If image is running then:
1) it was already verified and copied by MCUboot
   copy-done flag is set for sure.
2) or it was programmed directly using programming interface

For case 2) when image contains magic but not copy-done flag it was
impossible to confirm it. This patch removes needs of having copy-done
flag set while confirming. This make API behavior similar to situation
when the not-padded image was programmed.

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -583,11 +583,10 @@ boot_set_confirmed(void)
         goto done;
     }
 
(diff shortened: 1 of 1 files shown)


nvlsianpu:
cc @nandojve

utzig (approved):

PR #996, opened 2021-05-07
boot: bootutil: loader: Populate header when read sector fails

sigvartmh:
The primary image won't boot if read sector fails in the loader.
This is due to the header not being populated so that the TLV
values cannot be read out correctly.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1560,6 +1560,9 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
          * if there is one.
          */
         BOOT_SWAP_TYPE(state) = BOOT_SWAP_TYPE_NONE;
+        (void) boot_read_image_header(state, BOOT_PRIMARY_SLOT,
+				      boot_img_hdr(state, BOOT_PRIMARY_SLOT),
+				      bs);
         return;
     }
 


sigvartmh:
@nvlsianpu Please review and maybe add suggestions on how this can be resolved better.

boot/bootutil/src/loader.c
```diff
@@ -1560,6 +1560,9 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
          * if there is one.
          */
         BOOT_SWAP_TYPE(state) = BOOT_SWAP_TYPE_NONE;
+        (void) boot_read_image_header(state, BOOT_PRIMARY_SLOT,
```
sigvartmh, line 1563: Don't know if there is any point in handling the return value of this. If it fails it fails and there is nothing we can do.

tamasban: Hmm, as I see the `boot_read_image_headers` should return with success if only the reading of the secondary slot has failed, if the `require_all` paramter set to `false`, but it seems to be okay when calling it in line 1570.

So, I would expect that execution is continued if only the reading of secondary slot has failed.

nvlsianpu:
@sigvartmh Can you resolve the conflict?

PR #1002, opened 2021-05-08
imgtool: Add support for write_size > 8

arvinf:
Increase write_size paramter to 16 to support SoCs such as Atmel SAM4S or
SAMV70 with 128-bit default flash alignments.

Signed-off-by: Arvin Farahmand <arvinf@ip-logix.com>

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -510,7 +510,7 @@ class Image():
         if overwrite_only:
             return MAX_ALIGN * 2 + magic_size
         else:
-            if write_size not in set([1, 2, 4, 8]):
+            if write_size not in set([1, 2, 4, 8, 16]):
                 raise click.BadParameter("Invalid alignment: {}".format(
                     write_size))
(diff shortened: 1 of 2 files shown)


d3zd3z:
If we plan to support these large alignments in the bootloader code, we should also add the 16-byte alignment to the end of image.rs.  For large-writes, we test 128 and 512, but it looks like 16 will be common. A quick test with 'master' shows that setting the write alignment to 16 fails most tests at this point.

d3zd3z (approved):
I'm ok with these changes, although '> 8' might be a little misleading, since this only supports a write size of 16 in addition to currently supported sizes.

arvinf:
I'll edit the commit message.

utzig (changes requested):
I believe this is misleading, since it won't really add any support for 16-byte alignments, only make a user believe it does by having that option in `imgtool`. Before bootutil supports alignments larger than 8-byte there is no reason to actually have this option. Also this duplicates effort which was already done in #949. Maybe you should work on improving that PR?

PR #1005, opened 2021-05-11
Fix version comparison result causing boot failure

### LINKED ISSUE #976: boot_go() Returns boot_version_cmp() Value
`boot_go()` and consequently `context_boot_go()` will return a `boot_version_cmp()` return value if no pre-compile features are defined (MCUBOOT_DIRECT_XIP_REVERT, MCUBOOT_RAM_LOAD, MCUBOOT_HW_ROLLBACK_PROT, MCUBOOT_MEASURED_BOOT, MCUBOOT_DATA_SHARING). 

`boot_version_cmp()` returns -1, 0 or 1 to imply version comparison results. These values can be incorrectly interpreted as a non-zero error code when returned from `boot_go()`. 

https://github.com/mcu-tools/mcuboot/blob/986c212ea902e2c58bce5860195db605684edf7f/boot/bootutil/src/loader.c#L2347-L2348

AGlass0fMilk:
The variable `rc` is used in `context_boot_go` for a number of operations, mostly checking return codes for an error. It is also used to store the result of comparing the installed application version numbers (using `boot_version_cmp`).

At the end of `context_boot_go`, `rc` is returned as the result of the operation.

In some configurations, namely direct XIP (without revert or any other extra boot checks), it is possible for the comparison result to be the last value of `rc`. In most cases, this will cause `context_boot_go` to report a failure (ie: non-zero return code), as the value of `rc` will be 1 or -1 if the installed applications aren't the same version.

This commit resets the value of `rc` to 0 after the version comparison has taken place.

Fixes #976

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2365,6 +2365,9 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
                 }
             }
 
+            /* Reset rc since we use it above for boot_version_cmp */
+            rc = 0;
+
 #ifdef MCUBOOT_DIRECT_XIP_REVERT
             rc = boot_select_or_erase(&slot_state, selected_slot);
             if (rc != 0) {


d3zd3z (approved):
I have also added #1008 to add a regression test for this case. We can't add it yet, because there isn't yet sim test support for direct-xip.

Otherwise, this seems fine.

AGlass0fMilk:
Let's get that merge!

PR #1006, opened 2021-05-11
Enable no signature verification configuration for Mbed-OS.

AGlass0fMilk:
This commit introduces changes to allow the Mbed-OS port to disable use of signature verification. Previously this was not possible even though it is a valid mcuboot configuration.

### DIFF
--- a/boot/mbed/app_enc_keys.c
+++ b/boot/mbed/app_enc_keys.c
@@ -30,8 +30,6 @@ extern unsigned int ecdsa_pub_key_len;
 #define HAVE_KEYS
 extern const unsigned char ed25519_pub_key[];
 extern unsigned int ed25519_pub_key_len;
-#else
-#error "No public key available for given signing algorithm."
 #endif
 
 /*
--- a/boot/mbed/include/mcuboot_config/mcuboot_config.h
+++ b/boot/mbed/include/mcuboot_config/mcuboot_config.h
@@ -18,6 +18,7 @@
 #define SIGNATURE_TYPE_RSA      0
 #define SIGNATURE_TYPE_EC256    1
 #define SIGNATURE_TYPE_ED25519  2
+#define SIGNATURE_TYPE_NONE     3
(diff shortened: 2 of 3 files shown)


boot/mbed/app_enc_keys.c
```diff
@@ -30,8 +30,6 @@ extern unsigned int ecdsa_pub_key_len;
 #define HAVE_KEYS
 extern const unsigned char ed25519_pub_key[];
 extern unsigned int ed25519_pub_key_len;
-#else
-#error "No public key available for given signing algorithm."
```
boraozgen, line 34: Is this file supposed to be excluded from compilation for no signature configuration?

LDong-Arm: As far as I know, Mbed CLI1 cannot conditionally ignore files. Assuming files will be compiled and disabling unused code with macros is the only way.

d3zd3z: The linker also might leave out symbols that aren't used. It might not be necessary to conditionalize the entire file.

d3zd3z (approved):
> Previously this was not possible even though it is a valid mcuboot configuration.

It may be a valid configuration, but it isn't really a recommended configuration, as images can easily be forged.

Otherwise, this is ok.

AGlass0fMilk:
> > Previously this was not possible even though it is a valid mcuboot configuration.
> 
> 
> 
> It may be a valid configuration, but it isn't really a recommended configuration, as images can easily be forged.
> 
> 
> 
> Otherwise, this is ok.

I wanted to use mcuboot in my open source projects where a user can update the firmware of a device in the field. In this case, it's not as much of a concern. But it is nice to be able to use mcuboot for _just_ verifying the hash i.e. integrity of the firmware when the security is not a requirement.

More like an arduino-style bootloader. Mcuboot may be overkill for this but it's way easier than rolling my own.

PR #1007, opened 2021-05-11
bootutil_public.c: modify boot_set_pending() and boot_set_confirmed()

Sherryzhang2:
bootutil_public.c: add two user APIs

    Add boot_set_pending_multi_image() and
    boot_set_confirmed_multi_image() APIs so that the user can set the
    image with given index as pending, confirmed.

    Change-Id: Ifca04d396b38c6c64581703794071f6b42e9dfbf

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -3,7 +3,7 @@
  *
  * Copyright (c) 2017-2019 Linaro LTD
  * Copyright (c) 2016-2019 JUUL Labs
- * Copyright (c) 2019-2020 Arm Limited
+ * Copyright (c) 2019-2021 Arm Limited
  * Copyright (c) 2020 Nordic Semiconductor ASA
  *
  * Original license:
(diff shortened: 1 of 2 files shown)


utzig (changes requested):
This is not OK, these APIs are used by external code, libraries and whatever else, so no...

d3zd3z:
What about something like creating a `boot_set_pending_n` and likewise for the other that takes an additional argument. The regular `boot_set_pending` would just call `boot_set_pending_n` with a zer0, but it would allow aware code to begin using the new call.

Sherryzhang2:
> This is not OK, these APIs are used by external code, libraries and whatever else, so no...

Yes, I used this in the Firmware Update partition in the Trusted Firmware M. Why this is not OK?

nvlsianpu (changes requested):
> What about something like creating a boot_set_pending_n and likewise for the other that takes an additional argument. The regular boot_set_pending would just call boot_set_pending_n with a zer0, but it would allow aware code to begin using the new call.

This is exactly how this can be done.
For instance this API is used by zephyr-project, so need keep compatibility.

Sherryzhang2:
> > What about something like creating a boot_set_pending_n and likewise for the other that takes an additional argument. The regular boot_set_pending would just call boot_set_pending_n with a zer0, but it would allow aware code to begin using the new call.
> 
> This is exactly how this can be done.
> For instance this API is used by zephyr-project, so need to keep compatibility.

Updated the PR accordingly.

boot/bootutil/src/bootutil_public.c
```diff
@@ -471,25 +471,28 @@ boot_swap_type(void)
...
- *                                  1=run image forever.
+ *                          only tested once:
+ *                               0=run image once, then confirm or revert.
+ *                               1=run image forever.
  *
  * @return                  0 on success; nonzero on failure.
  */
 int
-boot_set_pending(int permanent)
+boot_set_pending_multi_image(int image_index, int permanent)
```
utzig, line 488: The changes look good to me, but how about calling them `boot_set_pending_multi` and `boot_set_confirmed_multi`  (no image postfix) so they both match the naming scheme used for `boot_swap_type_multi`?

Sherryzhang2: Done

utzig (approved):
LGTM, thanks!

utzig:
@Sherryzhang2 Ah, the current commit message does not match the new prototypes, please fix it.

nvlsianpu (approved):

Sherryzhang2:
> @Sherryzhang2 Ah, the current commit message does not match the new prototypes, please fix it.

Done

PR #1015, opened 2021-05-24
bootutil: Reduce boot_initialize_area variants

de-nordic:
The commit removes implementation of boot_initialize_area
specific for flash_area_to_sectors, and applies changes to
the flash_area_get_sectors using variant, to make it
alternatively work with flash_area_to_sectors.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -189,43 +189,12 @@ boot_write_sz(struct boot_loader_state *state)
     return elem_sz;
 }
 
-#ifndef MCUBOOT_USE_FLASH_AREA_GET_SECTORS
 static int
 boot_initialize_area(struct boot_loader_state *state, int flash_area)
 {
-    int num_sectors = BOOT_MAX_IMG_SECTORS;
-    int rc;
-
-    if (flash_area == FLASH_AREA_IMAGE_PRIMARY(BOOT_CURR_IMG(state))) {
-        rc = flash_area_to_sectors(flash_area, &num_sectors,
-                                   BOOT_IMG(state, BOOT_PRIMARY_SLOT).sectors);
-        BOOT_IMG(state, BOOT_PRIMARY_SLOT).num_sectors = (size_t)num_sectors;
-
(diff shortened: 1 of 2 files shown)


boot/bootutil/src/loader.c
```diff
@@ -245,14 +215,18 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
         return BOOT_EFLASH;
     }
 
+#ifdef MCUBOOT_USE_FLASH_AREA_GET_SECTORS
     rc = flash_area_get_sectors(flash_area, &num_sectors, out_sectors);
+#else
+    _Static_assert(sizeof(int) <= sizeof(uint32_t), "Fix needed");
```
tamasban, line 221: This was officially introduced only in C11, but some earlier compiler versions already support it. Since it already appears in the code and nobody has complained, it is fine. I have not found any statement in the repo that which C version is supported officially.

tamasban (approved):
Thanks, nice enhancement!

nvlsianpu (approved):

boot/bootutil/src/loader.c
d3zd3z: It kind of depends on the environment used to build MCUboot, which is fairly diverse. Given the instances of C11 features, perhaps that is adequate. We also have other habits, such as declaring all variables at the top of functions that hasn't been necessary since C89, so who knows what we are targeting.

boot/bootutil/src/loader.c
```diff
@@ -189,44 +189,14 @@ boot_write_sz(struct boot_loader_state *state)
...
-    return rc;
-}
-#else  /* defined(MCUBOOT_USE_FLASH_AREA_GET_SECTORS) */
-static int
-boot_initialize_area(struct boot_loader_state *state, int flash_area)
-{
-    uint32_t num_sectors;
-    struct flash_sector *out_sectors;
-    size_t *out_num_sectors;
-    int rc;
```
utzig, line 200: There are two empty lines here, one can be removed.

utzig (approved):

boot/bootutil/src/loader.c
de-nordic: Fixed.

PR #1016, opened 2021-05-25
Make boot_read_swap_state declaration public

de-nordic:
The PR contains series of commits that:
 - make `boot_read_swap_state` public and available from bootutil_public.h;
 - change `fixup_revert` to use provided flash area object pointer to obtain swap state by call to `boot_read_swap_state`, instead of previously used `boot_read_swap_state_by_id`; this makes ID param (`sec_id`) no longer needed;
 - makes `boot_swap_sector` call `boot_read_swap_state` on already opened scratch flash area, instead of using `boot_read_swap_state_by_id`, with the area ID, which caused double open on the flash area;
 - changes logic of `boot_set_pending_multi` by moving `flash_area_open` to the top of the function, making that the only place where flash area object is obtained in the function; switches the function to use `boot_read_swap_state`;
 - switches `boot_set_confirmed_multi` to use `boot_read_swap_state`;

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil_public.h
+++ b/boot/bootutil/include/bootutil/bootutil_public.h
@@ -4,7 +4,7 @@
  * Copyright (c) 2017-2019 Linaro LTD
  * Copyright (c) 2016-2019 JUUL Labs
  * Copyright (c) 2019-2021 Arm Limited
- * Copyright (c) 2020 Nordic Semiconductor ASA
+ * Copyright (c) 2020-2021 Nordic Semiconductor ASA
  *
  * Original license:
  *
@@ -223,13 +223,25 @@ int boot_read_image_ok(const struct flash_area *fap, uint8_t *image_ok);
 /**
  * @brief Read the image swap state
  *
- * @param flash_area_id Id of flash parttition from which trailer will be read.
- * @param state Struture for holding swap state.
+ * @param flash_area_id id of flash partition from which state will be read;
+ * @param state pointer to structure for storing swap state.
  *
- * @return 0 on success, nonzero errno code on fail.
+ * @return 0 on success; non-zero error code on failure;
  */
 int
 boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state);
+
+/**
+ * @brief Read the image swap state
+ *
+ * @param fa pointer to flash_area object;
+ * @param state pointer to structure for storing swap state.
+ *
+ * @return 0 on success; non-zero error code on failure.
+ */
+int
+boot_read_swap_state(const struct flash_area *fa,
+                     struct boot_swap_state *state);
 
 #define BOOT_MAGIC_ARR_SZ \
     (sizeof boot_img_magic / sizeof boot_img_magic[0])
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -492,24 +492,23 @@ boot_set_pending_multi(int image_index, int permanent)
     uint8_t swap_type;
     int rc;
 
-    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_SECONDARY(image_index),
-                                    &state_secondary_slot);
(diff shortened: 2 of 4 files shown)


de-nordic:
@Sherryzhang2 can you look at the changes too?

boot/bootutil/src/bootutil_public.c
```diff
@@ -524,26 +523,24 @@ boot_set_pending_multi(int image_index, int permanent)
...
 
     default:
         assert(0);
-        return BOOT_EBADIMAGE;
+        rc = BOOT_EBADIMAGE;
     }
+
+done:
+    flash_area_close(fap);
+    return 0;
```
utzig, line 543: Should probably return `rc`, no?

de-nordic: (after lots of swearing to himself, he has finally admitted) Yes, you are right @utzig.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -231,6 +231,18 @@ int boot_read_image_ok(const struct flash_area *fap, uint8_t *image_ok);
 int
 boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state);
 
+/**
+ * @brief Read the image swap state
+ *
+ * @param flash_area fa pointer to flash_area object.
```
Sherryzhang2, line 237: Remove "flash_area"?

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -231,6 +231,18 @@ int boot_read_image_ok(const struct flash_area *fap, uint8_t *image_ok);
...
+ * @param state Struture for holding swap state.
+ *
+ * @return 0 on success, nonzero errno code on fail.
```
Sherryzhang2, line 240: "errno" -> "error"?
"fail" -> "failure"?

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Done

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Done

boot/bootutil/src/bootutil_public.c
```diff
@@ -524,26 +523,24 @@ boot_set_pending_multi(int image_index, int permanent)
...
+        break;
 
     default:
         assert(0);
-        return BOOT_EBADIMAGE;
+        rc = BOOT_EBADIMAGE;
     }
+
+done:
+    flash_area_close(fap);
```
d3zd3z, line 542: Admittedly, `flash_area_close` doesn't return an error in any current configuration. But we could do something like:
```c
    int rc2 = flash_area_close(fap);
    if (rc == 0) {
        rc = rc2;
    }
```

although, I'm not sure this is really necessary.

d3zd3z (approved):
I'm ok with this, my comment is only a suggestion and doesn't have to be changed.

utzig:
@de-nordic Please squash all commits, reading one by one does not make a lot of sense.

boot/bootutil/src/bootutil_public.c
de-nordic: Unfortunately the original author, of the flash_area_close, did not predict any return value from the close function, so it has void type. I was considering changing that in Zephyr, but dropped the idea.

de-nordic:
> 
> 
> @de-nordic Please squash all commits, reading one by one does not make a lot of sense.

OK, I have squashed to 3:
 - the one that published boot_read_swap_state;
 - the one that changes revert_fixup signature;
 - the one that applies boot_read_swap_state where possible;

utzig (approved):
LGTM. The last commit looks weird with two subjects and signed-off-by lines; otherwise nice refactoring.

de-nordic:
> 
> 
> LGTM. The last commit looks weird with two subjects and signed-off-by lines; otherwise nice refactoring.

Sorry for that, fixed.

PR #974, opened 2021-04-06
boot: Multi-image boot for direct-xip and ram-load

mark-horvath-arm:
Multiple image boot support is implemented for direct-xip and ram-load
strategies. Dependency checking is also added, and for ram-load more
executable ram regions can be set.

Signed-off-by: Mark Horvath <mark.horvath@arm.com>
Change-Id: I59b346277dcb2c17cef4800f1e556b4ea9144845

PR #1019, opened 2021-05-26
Add support in sim for direct-xip and ramload

d3zd3z:
Fix enough in the simulator so that the direct-xip and ramload configurations can be run. All of the existing tests assume that MCUboot modifies flash to make use of the new image. This PR does not add additional tests for this functionality, but fixes enough so that they can be compiled, and that inappropriate tests will be avoided.

### DIFF
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -403,10 +403,14 @@ boot_img_sector_off(const struct boot_loader_state *state, size_t slot,
 #endif  /* !defined(MCUBOOT_USE_FLASH_AREA_GET_SECTORS) */
 
 #ifdef MCUBOOT_RAM_LOAD
+#define IMAGE_RAM_BASE ((uintptr_t)0)
+
 #define LOAD_IMAGE_DATA(hdr, fap, start, output, size)       \
-    (memcpy((output),(void*)((hdr)->ih_load_addr + (start)), \
+    (memcpy((output),(void*)(IMAGE_RAM_BASE + (hdr)->ih_load_addr + (start)), \
     (size)), 0)
 #else
+#define IMAGE_RAM_BASE ((uintptr_t)0)
+
 #define LOAD_IMAGE_DATA(hdr, fap, start, output, size)       \
     (flash_area_read((fap), (start), (output), (size)))
 #endif /* MCUBOOT_RAM_LOAD */
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -403,7 +403,26 @@ boot_img_sector_off(const struct boot_loader_state *state, size_t slot,
 #endif  /* !defined(MCUBOOT_USE_FLASH_AREA_GET_SECTORS) */
 
 #ifdef MCUBOOT_RAM_LOAD
-#define IMAGE_RAM_BASE ((uintptr_t)0)
+#   ifdef __BOOTSIM__
+
+/* Query for the layout of a RAM buffer appropriate for holding the
+ * image.  This will be per-test-thread, and therefore must be queried
+ * through this call. */
+struct bootsim_ram_info {
+    uint32_t start;
+    uint32_t size;
+    uintptr_t base;
+};
+struct bootsim_ram_info *bootsim_get_ram_info(void);
+
+#define IMAGE_GET_FIELD(field) (bootsim_get_ram_info()->field)
+#define IMAGE_RAM_BASE IMAGE_GET_FIELD(base)
+#define IMAGE_EXECUTABLE_RAM_START IMAGE_GET_FIELD(start)
+#define IMAGE_EXECUTABLE_RAM_SIZE IMAGE_GET_FIELD(size)
+
+#   else
+#       define IMAGE_RAM_BASE ((uintptr_t)0)
+#   endif
 
 #define LOAD_IMAGE_DATA(hdr, fap, start, output, size)       \
     (memcpy((output),(void*)(IMAGE_RAM_BASE + (hdr)->ih_load_addr + (start)), \
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2242,6 +2242,8 @@ print_loaded_images(struct boot_loader_state *state,
 {
     uint32_t active_slot;
 
+    (void)state;
+
     IMAGES_ITER(BOOT_CURR_IMG(state)) {
         active_slot = slot_usage[BOOT_CURR_IMG(state)].active_slot;
 
@@ -2354,6 +2356,9 @@ boot_verify_ram_load_address(struct boot_loader_state *state,
     uint32_t img_end_addr;
     uint32_t exec_ram_start;
     uint32_t exec_ram_size;
+
+    (void)state;
+
 #ifdef MULTIPLE_EXECUTABLE_RAM_REGIONS
     int      rc;
 
@@ -2582,11 +2587,13 @@ static inline int
 boot_remove_image_from_sram(struct boot_loader_state *state,
                             struct slot_usage_t slot_usage[])
 {
+    (void)state;
+
     BOOT_LOG_INF("Removing image from SRAM at address 0x%x",
                  slot_usage[BOOT_CURR_IMG(state)].img_dst);
 
-    memset((void*)slot_usage[BOOT_CURR_IMG(state)].img_dst, 0,
-           slot_usage[BOOT_CURR_IMG(state)].img_sz);
+    memset((void*)(IMAGE_RAM_BASE + slot_usage[BOOT_CURR_IMG(state)].img_dst),
+           0, slot_usage[BOOT_CURR_IMG(state)].img_sz);
 
     slot_usage[BOOT_CURR_IMG(state)].img_dst = 0;
     slot_usage[BOOT_CURR_IMG(state)].img_sz = 0;
@@ -2608,6 +2615,8 @@ boot_remove_image_from_flash(struct boot_loader_state *state, uint32_t slot)
     int area_id;
     int rc;
     const struct flash_area *fap;
+
+    (void)state;
 
     BOOT_LOG_INF("Removing image %d slot %d from flash", BOOT_CURR_IMG(state),
                                                          slot);
@@ -2917,7 +2926,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 {
     struct slot_usage_t slot_usage[BOOT_IMAGE_NUMBER];
     int rc;
-    fih_int fih_rc;
+    fih_int fih_rc = fih_int_encode(0);
 
     memset(state, 0, sizeof(struct boot_loader_state));
     memset(slot_usage, 0, sizeof(struct slot_usage_t) * BOOT_IMAGE_NUMBER);
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -857,6 +857,7 @@ boot_validated_swap_type(struct boot_loader_state *state,
 
     return swap_type;
 }
+#endif
 
 /**
  * Erases a region of flash.
@@ -874,6 +875,7 @@ boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
     return flash_area_erase(fap, off, sz);
 }
 
+#if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
 /**
  * Copies the contents of one flash region to another.  You must erase the
  * destination region prior to calling this function.
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2416,7 +2416,7 @@ boot_copy_image_to_sram(struct boot_loader_state *state, int slot,
     }
 
     /* Direct copy from flash to its new location in SRAM. */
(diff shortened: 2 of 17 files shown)


utzig (dismissed):
A few nitpicks.

sim/src/caps.rs
```diff
@@ -39,6 +41,12 @@ impl Caps {
     pub fn get_num_images() -> usize {
         (unsafe { bootutil_get_num_images() }) as usize
     }
+
+    /// Query if this configuration performs some kind of upgrade by writing to flash.
+    pub fn modifies_flash() -> bool {
+        // All other configurations perform upgrades by writing to flash.
+        !(Self::RamLoad.present() || Self::DirectXip.present())
```
tamasban, line 48: In case of ram-load if image is faulty (could happen in negative test) the flash is erased:
https://github.com/mcu-tools/mcuboot/pull/974/commits/b005cde0550ec3fe25da6777e5335cae72a11918#diff-a004fdf5907914a360ce1e7da94221c0165d1c4c8ba7b8c4eaf67bd63af395c1R2863

boot/bootutil/src/bootutil_priv.h
```diff
@@ -406,10 +406,14 @@ boot_img_sector_off(const struct boot_loader_state *state, size_t slot,
 #endif  /* !defined(MCUBOOT_USE_FLASH_AREA_GET_SECTORS) */
 
 #ifdef MCUBOOT_RAM_LOAD
+#define IMAGE_RAM_BASE ((uintptr_t)3)
```
tamasban, line 409: Might not have got it fully, but this seems to have an impact not only on the simulator environment rather the real targets as well.

sim/Cargo.toml
```diff
@@ -26,6 +26,8 @@ enc-x25519 = ["mcuboot-sys/enc-x25519"]
 enc-aes256-x25519 = ["mcuboot-sys/enc-aes256-x25519"]
 bootstrap = ["mcuboot-sys/bootstrap"]
 multiimage = ["mcuboot-sys/multiimage"]
+ramload = ["mcuboot-sys/ramload"]
```
tamasban, line 29: Nit: Might ram-load better. In the design.md it is refernced in this way.

tamasban (changes requested):
Thanks, David for the initial steps in the simulator environment!

boot/bootutil/src/bootutil_priv.h
nvlsianpu: Look like I see the same. ^^

boot/bootutil/src/loader.c
```diff
@@ -780,6 +781,7 @@ boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
     return flash_area_erase(fap, off, sz);
 }
 
+#if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
```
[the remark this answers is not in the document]
d3zd3z: This one is a little less clear, since the two places it is used have different ifdef blocks around them. I think it works out OK to just always have the function, and let the linker eliminate it if it isn't used.

sim/src/caps.rs
d3zd3z: Maybe "modifies_flash" isn't the best name. I'm trying to capture the case of moving images around. Maybe 'moves_images' or something like that?

boot/bootutil/src/bootutil_priv.h
d3zd3z: Yeah, this isn't done yet, just a placeholder to get it to compile in the simulator. I'm working on actually testing these features now, which will involve putting something meaningful here.

d3zd3z:
I've revised this I believe to account for review feedback. This tests both ram loading and direct xip with single and multiple images. I have not addressed #1028, as there is another patch that fixes that. Once that is merged, we can also try enabling this feature in sim tests.

d3zd3z:
I've also removed the "(trivial)" from the title, since this is no longer trivial, and actually does test, at least basic tests, of these features. It might make sense to write additional tests for the revert feature.

d3zd3z:
Replaced `ramload` with either `ram_load` or `ram-load` as appropriate.

d3zd3z:
@utzig Any chance to be able to re-review? I think I'd addressed everything, and have added support for testing both of these features.

sim/src/caps.rs
d3zd3z: @tamasban aside from that, I'm not certain that this behavior is actually desirable. If the external system does a partial download, and power were lost in the midst, it would have to start over. If power is unreliable, it might never complete. Or does it only do the erase if the image is marked for upgrade.

The only benefit I see to clearing a bad image would be to make subsequent boots faster.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -403,7 +403,26 @@ boot_img_sector_off(const struct boot_loader_state *state, size_t slot,
 #endif  /* !defined(MCUBOOT_USE_FLASH_AREA_GET_SECTORS) */
 
 #ifdef MCUBOOT_RAM_LOAD
-#define IMAGE_RAM_BASE ((uintptr_t)3)
+#   ifdef __BOOTSIM__
```
tamasban, line 406: extra spaces, also in line 423-425

boot/bootutil/src/bootutil_priv.h
tamasban: Could you set the macro value to 0? Then it won't break the functionality and its value is changed anyway to 0 in a later commit.

sim/src/image.rs
```diff
@@ -1374,25 +1365,37 @@ impl Images {
 }
 
 impl RamData {
+    // TODO: This is not correct. The second slot of each image should be at the same address as
+    // the primary.
     fn new(slots: &[[SlotInfo; 2]]) -> RamData {
         let mut addr = RAM_LOAD_ADDR;
         let mut places = BTreeMap::new();
+        // println!("Setup:-------------");
```
tamasban, line 1373: some debug  print left in

sim/mcuboot-sys/src/lib.rs
```diff
@@ -32,6 +32,11 @@ impl RamBlock {
         &self.ram
     }
 
+    /// Borrow a piece of the ram, with 'offset' being the beginning of the buffer.
```
tamasban, line 35: The subject line of this commit might not correct? How this change is related to DIRECT_XIP?

tamasban (changes requested):
Thanks for the simulator support for ram-load and direct-xip features!

Just minor comments, Rust code is not reviewed deeply.

boot/bootutil/src/bootutil_priv.h
d3zd3z: The spaces are intentional? I'm not sure why people insist on not indenting preprocessor directives.

sim/mcuboot-sys/src/lib.rs
d3zd3z: Github doesn't make it clear which commit this comment goes with. I'll try go figure out from this line, though.

sim/src/image.rs
d3zd3z: There are a lot of commented out debug print messages in the simulator. I propose changing these into low level log messages, but since this isn't the only place, maybe making that a separate PR.

sim/mcuboot-sys/src/lib.rs
d3zd3z: But, related to this, this adds an accessor needed by the DIRECT_XIP testing code.

boot/bootutil/src/bootutil_priv.h
tamasban: ACK.

sim/src/image.rs
tamasban: ACK

utzig (changes requested):
* b70b6f4673d2ee65d5a0627c46614b2edf324444, ccfd7d6d40cab7a7364e133a61a5bbbbbf2e0f15 and e21a37b0354218f2d317ddcc301f8505b57fbd04 should probably be just a single commit since all deals with the results from boot_go;
* 9340ab994a600eb110c624068e81248d3b15d7d3 is doing a change that 5686da1e7b13f7069b3dafaf257122dcf20a213f is undoing, so they can be fixed up;
* 08cb79d9b9e5bc6b6632dcc94f3b77757b89d0f9 can be squashed into e14169bc040c77ad68fc8025df39664fba976be0 as far as I understand;
* Are the `TODO`s in efea321487e404f5dad0b4c15d37da342a28016c, 5ae28338074eb34fc786d7c4b997c3ee01820419 and cb5b843e011fc18139246344e8cf41b7a4af6e38 supposed to be left there?
* efea321487e404f5dad0b4c15d37da342a28016c adds a bunch of code that is later removed by cb5b843e011fc18139246344e8cf41b7a4af6e38. Wouldn't it be better to either just add or just remove something instead of adding, removing, adding again whatever? It's hard to follow...

d3zd3z:
> b70b6f4, ccfd7d6 and e21a37b should probably be just a single commit since all deals with the results from boot_go

I'll squash the first two, but the second one really does do something new, and I really think it should still be a separate commit.

> 9340ab9 is doing a change that 5686da1 is undoing, so they can be fixed up;

I combined these to avoid the modification and reversion of the code.

> 08cb79d can be squashed into e14169b as far as I understand;

Combined

> Are the TODOs in efea321, 5ae2833 and cb5b843 supposed to be left there?

They are intentional, if that is what you are asking. All are suggestions on possible future ways to improve the simulator. The ones in 5ae2833 are all suggesting looking into a finding a way to remove a possibly expensive `clone()`. The rest are suggesting things that aren't yet being tested but could be.

> efea321 adds a bunch of code that is later removed by cb5b843. Wouldn't it be better to either just add or just remove something instead of adding, removing, adding again whatever? It's hard to follow...

I guess that depends on how we want to view the git history. The testing is implemented in two stages, and the `run_ram_load` code present in efea321 is needed in order to make that commit meaningful. cb5b843 adds support for testing multiple images, which makes `run_ram_load` need to be more complex. I could squash these, but I think it makes the resulting commits harder to follow. They are definitely doing different things.

d3zd3z:
Add be146bef, to fix a problem with the simulator implementation of fa_device_id caused by c304a7f8.

d3zd3z:
Any chance for a review: @de-nordic, @tamasban, or @nvlsianpu ? Thanks.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -403,10 +403,33 @@ boot_img_sector_off(const struct boot_loader_state *state, size_t slot,
...
+struct bootsim_ram_info *bootsim_get_ram_info(void);
+
+#define IMAGE_GET_FIELD(field) (bootsim_get_ram_info()->field)
+#define IMAGE_RAM_BASE IMAGE_GET_FIELD(base)
+#define IMAGE_EXECUTABLE_RAM_START IMAGE_GET_FIELD(start)
+#define IMAGE_EXECUTABLE_RAM_SIZE IMAGE_GET_FIELD(size)
+
+#   else
+#       define IMAGE_RAM_BASE ((uintptr_t)0)
+#   endif
```
de-nordic, line 425: I am not an expert in code style, but I think that if you indent the preprocessor directives, the blocks within conditional compilation directives should also be indented.
So the struct above should be also shifted right and the `#define` should be `#    define`, otherwise it looks like bad merge.
I think that when you indent the directives, then they should be treated as part of code, not as preprocessor directives, and the code should be formatted with them.

de-nordic (approved):
Only reviewed the C code; my Rust is not good enough to have viable input.

Left comment regarding indenting of preprocessor directives, otherwise looks OK.
Thanks.

d3zd3z:
@de-nordic it might be worth discussing how we want to handle indenting of preprocessor directives. It is kind of messy how they are separate, syntactically, from the code, which usually means editors will fight against trying to indent them together. I can just unindent them all for now, or what I might propose is a separate PR to fix up the indentation based on whatever we might propose from discussions.

de-nordic:
> 
> 
> @de-nordic it might be worth discussing how we want to handle indenting of preprocessor directives(...)

I would happily discuss it, I am actually using indentations although I do not find it to be widely accepted style.

>I can just unindent them all for now, or what I might propose is a separate PR to fix up the indentation based on whatever we might propose from discussions.

I think that this would be good move, for the PR, because as I have said, when you look at the overall file it looks like bad merge at this point.

d3zd3z:
@de-nordic writes:
> I would happily discuss it, I am actually using indentations although I do not find it to be widely accepted style.

I've noticed a trend amongst embedded C programmers, to treat the C language kind of like it was in the 70s. Features newer than that seem to be ignored or rejected, especially features that came from C++.

Although, indenting by leaving the '#' in the left column and adding spaces after it has _always_ worked.

de-nordic:
> 
> I've noticed a trend amongst embedded C programmers, to treat the C language kind of like it was in the 70s. Features newer than that seem to be ignored or rejected, especially features that came from C++.

It is not unique to embedded; I think that this is caused by some past misunderstandings.

> 
> Although, indenting by leaving the '#' in the left column and adding spaces after it has _always_ worked.

The first C manual I have read, which I thought was C language specification actually, is in K&R notation, and within it the # in preprocessor directives are followed by space (https://www.bell-labs.com/usr/dmr/www/cman.pdf).

utzig (approved):
LGTM

PR #1026, opened 2021-06-11
boot: introduce MCUBOOT_CPU_IDLE() for support low power single thread 

nvlsianpu:
If a port supports single thread or is bare-metal then might be need
to switch to idle mode explicitly form MCUboot code.
The call allows to enable lower power consumption while waiting for
incoming transmission in serial recovery etc.

Especially:
Sine a while there is no silent idle thread created automatically while
CONFIG_MULTITHREADING=n in zephyr. Since that any single thread application
needs to call k_cpu_idle() by itself for entering idle mode, which
allows for reduction power consumption.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -613,6 +613,7 @@ boot_serial_start(const struct boot_uart_funcs *f)
 
     off = 0;
     while (1) {
+        MCUBOOT_CPU_IDLE();
         rc = f->read(in_buf + off, sizeof(in_buf) - off, &full_line);
         if (rc <= 0 && !full_line) {
             continue;
--- a/samples/mcuboot_config/mcuboot_config.template.h
+++ b/samples/mcuboot_config/mcuboot_config.template.h
@@ -146,4 +146,14 @@
  *    do { do watchdog feeding here! } while (0)
  */
 
+/* If a OS ports support single thread mode or is bare-metal then:
+ * This macro implements call that switches CPU to an idle state, from which
+ * the CPU may be woken up by, for example, UART transmission event.
+ * 
+ * Otherwise this macro should be no-op.
+ */
+#define MCUBOOT_CPU_IDLE() \
+    do {                   \
+    } while (0)
+
 #endif /* __MCUBOOT_CONFIG_H__ */
--- a/boot/cypress/MCUBootApp/config/mcuboot_config/mcuboot_config.h
+++ b/boot/cypress/MCUBootApp/config/mcuboot_config/mcuboot_config.h
@@ -152,4 +152,11 @@
 #define NUM_ECC_BYTES (256 / 8)
 #endif /* ENC_IMG */
 
+/*
+ * No direct idle call implemented
+ */
+#define MCUBOOT_CPU_IDLE() \
+    do {                   \
+    } while (0)
+
 #endif /* MCUBOOT_CONFIG_H */
--- a/boot/mbed/include/mcuboot_config/mcuboot_config.h
+++ b/boot/mbed/include/mcuboot_config/mcuboot_config.h
@@ -79,5 +79,11 @@
     do {                                        \
     } while (0)
 
+/*
+ * No direct idle call implemented
+ */
+#define MCUBOOT_CPU_IDLE() \
+    do {                   \
+    } while (0)
 
 #endif /* __MCUBOOT_CONFIG_H__ */
--- a/boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h
+++ b/boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h
@@ -101,4 +101,11 @@
 #define MCUBOOT_WATCHDOG_FEED()    do {} while (0)
 #endif
 
+/*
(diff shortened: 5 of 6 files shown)


nvlsianpu:
cc @nordic-krch

nvlsianpu (commented):
Have theory (especially after looking into mynewt os.h file taht it is actually legacy and better to move new call/macro somewhere else.

utzig:
I think this would be better using a macro, as is done with `MCUBOOT_WATCHDOG_FEED`, eg `MCUBOOT_IDLE_WFE`, adding it to `mcuboot_config.h` since this file already exists for all ports. Also you need to at least add the empty macro for other ports.

samples/mcuboot_config/mcuboot_config.template.h
```diff
@@ -146,4 +146,15 @@
  *    do { do watchdog feeding here! } while (0)
  */
 
+/*
+ * If a OS ports support single thread mode or is bare-metal then:
+ * This macro shall implements call which switches CPU to idle states
+ * which wakes on an event like expected uart transmission.
```
de-nordic, line 152: this macro implements call that switches CPU to an idle state, from which the CPU may be woken up by, for example, UART transmission event.

nvlsianpu: thanks!

de-nordic (approved):
Looks OK, thanks!

boot/zephyr/include/os/os.h
```diff
@@ -0,0 +1,8 @@
+
+
+static inline void os_cpu_idle_wfe(void)
+{
+  if (!IS_ENABLED(CONFIG_MULTITHREADING)) {
+    k_cpu_idle();
+  }
+}
```
utzig, line 8: Shouldn't this file be removed?

nvlsianpu: ach, I forgot

utzig (approved):

ioannisg:
> I think this would be better using a macro, as is done with `MCUBOOT_WATCHDOG_FEED`, eg `MCUBOOT_IDLE_WFE`, adding it to `mcuboot_config.h` since this file already exists for all ports. Also you need to at least add the empty macro for other ports.

Isn't WFE, though an ARM-only terminology? Would we want a more generic name to support other ARCHes?

nvlsianpu:
@ioannisg WFE mean Wait For Event. ARM just has named the instruction well.
I can change name to something else (?) or remove WFE part.
Let me know what you prefer.

ioannisg:
> @ioannisg WFE mean Wait For Event. ARM just has named the instruction well.
> I can change name to something else (?) or remove WFE part.
> Let me know what you prefer.

I think you need to change this, because this is confusing with the ARM terminology, and won't be good, e.g. for RiscV. But also for ARM is confusing a bit, since here, this macro calls k_cpu_idle, and _that_ functions does not use WFE instruction :p

nvlsianpu:
@ioannisg I renamed the macro to `MCUBOOT_CPU_IDLE()`.

ioannisg (approved):
Looks fine to me.

samples/mcuboot_config/mcuboot_config.template.h
```diff
@@ -146,4 +146,14 @@
  *    do { do watchdog feeding here! } while (0)
  */
 
+/* If a OS ports support single thread mode or is bare-metal then:
+ * This macro implements call that switches CPU to an idle state, from which
+ * the CPU may be woken up by, for example, UART transmission event.
+ *
```
d3zd3z, line 152: There is a trailing space on this line.

boot/boot_serial/src/boot_serial.c
```diff
@@ -613,6 +613,7 @@ boot_serial_start(const struct boot_uart_funcs *f)
 
     off = 0;
     while (1) {
+        MCUBOOT_CPU_IDLE();
```
d3zd3z, line 616: Is there a race here? What if data comes in, and an interrupt has already processed it. This would then block, and not wake up until perhaps something else woke the system up.

I'm wondering if we wouldn't be better off making the `read` call below implemented with blocking if that makes sense on the platform instead of hoping that this IDLE call will work.

nvlsianpu: >  the read call below implemented with blocking 

I don't understand what you mean exactly.

I'm not 100% sure - this is working well on zephyr.
Wouldn't be just needed to check whether incoming buffer is empty before call the idle() (in case serial backend serve use line-by-line).

nvlsianpu: @nordic-krch Any opinion?

d3zd3z: The race is probably rare, especially because uarts are fairly slow, but if data comes in before the idle line is reached, the idle line could potentially block forever. Really, it would also depend on whether the device gives another interrupt if more characters come in, since there are multiple characters.

So, this might not show up either very often or even at all, but it is still racy.

nvlsianpu: Zephyr is using k_cpu_idle() as well, so if this is real issue - it is zephyr-rtos issue.
This is our conclusion after talk with @nordic-krch

PR #1035, opened 2021-06-23
bootutil: Fix some misspelling variables

### LINKED ISSUE #1028: Compile error when `MCUBOOT_DIRECT_XIP_REVERT` is defined
`boot/bootutil/loader.c:2292` and `boot/bootutil/loader.c:2213` refer to the variable `slot`, that was renamed in ccaf7f88 to `active_slot`. Looks like just a rename to fix it, though I guess this means we don't have test coverage for `MCUBOOT_DIRECT_XIP_REVERT`.

FarK:
There ir a couple of misspelled variables (slot -> active_slot).

fix #1028

Signed-off-by: Carlos Falgueras García <carlos.falgueras@wslw.es>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2289,7 +2289,7 @@ boot_select_or_erase(struct boot_loader_state *state,
          * to prevent it from being selected again on the next reboot.
          */
         BOOT_LOG_DBG("Erasing faulty image in the %s slot.",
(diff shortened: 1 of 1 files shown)


d3zd3z (changes requested):
The change itself is fine. I suggest fixing the "ir" to perhaps "are" in the commit text, though :-)

FarK:
Of course, thank you for the feedback :)

nvlsianpu (approved):

d3zd3z (approved):

PR #1037, opened 2021-06-24
bootutil: Fix memory leak on loader.c::split_image_check()

FarK:
The tmpbuf must be freed before exit.

Signed-off-by: Carlos Falgueras García <carlos.falgueras@wslw.es>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -515,7 +515,7 @@ split_image_check(struct image_header *app_hdr,
                   struct image_header *loader_hdr,
                   const struct flash_area *loader_fap)
 {
-    static void *tmpbuf;
+    static void *tmpbuf = NULL;
     uint8_t loader_hash[32];
     fih_int fih_rc = FIH_FAILURE;
 
@@ -529,13 +529,14 @@ split_image_check(struct image_header *app_hdr,
     FIH_CALL(bootutil_img_validate, fih_rc, NULL, 0, loader_hdr, loader_fap,
              tmpbuf, BOOT_TMPBUF_SZ, NULL, 0, loader_hash);
     if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
-        FIH_RET(fih_rc);
+        goto out;
     }
 
     FIH_CALL(bootutil_img_validate, fih_rc, NULL, 0, app_hdr, app_fap,
              tmpbuf, BOOT_TMPBUF_SZ, loader_hash, 32, NULL);
 
 out:
+    free(tmpbuf);
     FIH_RET(fih_rc);
 }
 #endif /* !MCUBOOT_DIRECT_XIP && !MCUBOOT_RAM_LOAD */


de-nordic:
If I may suggest something: the `free` will do nothing if it is given `NULL` pointer, maybe initialize the `tmpbuf` pointer to `NULL`, it will remain such if `malloc` fails and have single `out:` label to cover both paths?

de-nordic:
@FarK I have missed the `static` qualifier of the `tmpbuf`, that is why there is double check for `!tmpbuff` because the buffer may have been allocated in previous run. It is leak, but it seems to be designed that way, so the buffer gets allocated once and then it is just references.
I think we should go with the `free` you have implemented, in which case the definition of `tmpbuf` should be altered by removal of `static` and the check (!tmpbuf) is redundant.
@utzig do you know why this allocation is done such way?

FarK:
Is `split_image_check()` used somewhere? I couldn't find any place where it's used.

de-nordic:
> `split_image_check`

`split_go`

FarK:
> > `split_image_check`
> 
> `split_go`

Yes of course, but `split_go()` is called from anywhere I think.

de-nordic:
I am not familiar with the usage of this function, at least in possible Zephyr boot paths.

boot/bootutil/src/loader.c
```diff
@@ -515,7 +515,7 @@ split_image_check(struct image_header *app_hdr,
                   struct image_header *loader_hdr,
                   const struct flash_area *loader_fap)
 {
-    static void *tmpbuf;
+    static void *tmpbuf = NULL;
```
nvlsianpu, line 518: static local variable is automatically initialized to 0 (so NULL).

nvlsianpu (changes requested):
This is code used by TFM.
This is not a memory leak in reality.

Let's wait for TFM team comments.
@tamasban @RcColes ^^

boot/bootutil/src/loader.c
FarK: Indeed. Thank you! I'll wait before push more changes though, just in case this finally isn't a leak.

d3zd3z:
Two issues here. First, I'm not sure I'd really call this a memory leak, as in the only thing that will happen after this code completes is a jump to the new image, where the memory will be completely reused by the running image.

Second, given that this is a static, it looks like the intent was to allow the same buffer to be reused on subsequent calls. By freeing it (and especially, but not re-NULL-ing the pointer), this will result in a memory use error if the function is called again.

boot/bootutil/src/loader.c
```diff
@@ -529,13 +529,14 @@ split_image_check(struct image_header *app_hdr,
...
     if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
-        FIH_RET(fih_rc);
+        goto out;
     }
 
     FIH_CALL(bootutil_img_validate, fih_rc, NULL, 0, app_hdr, app_fap,
              tmpbuf, BOOT_TMPBUF_SZ, loader_hash, 32, NULL);
 
 out:
+    free(tmpbuf);
```
d3zd3z, line 539: This is a static. You need to have a `tmpbuf = NULL;`. Otherwise a subsequent call to this function will result in the freed buffer being used.

d3zd3z (changes requested):

d3zd3z:
My main suggestion, however, would be to not apply this patch. It fixes something that isn't really a problem.

FarK:
There is another `malloc()`, in `split_go()`, and this does have its `free()`. It seems like the situation is the same as in `split_image_check()`, this heap will be lost when the image boots. Anyway, I think a comment in-code like "Why this is not a leak" will be great.

Maybe could this buffer be static? Like in https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L479 or in https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/single_loader.c#L35

boot/bootutil/src/loader.c
tamasban: Actually the `split_go()` functionality is not used by TF-M. We have just modified the code (adding FIH_CALL macros) to not break the build.

Do we know is there any user of this `split_go`  feature?

nvlsianpu: What I see is that split_go() was something used in the mynewt bootloader https://github.com/apache/mynewt-core/blob/2673c6c414f05028abf12039033d3ec27f87d98b/boot/split/src/split.c#L66
Looks like a fossil here.
@utzig I think it can be removed form mcuboot. Can you confirm?

utzig: No, I don't think it can. But I can ask on the Mynewt ML and deprecate it for next version (if there are no opposition). I would suggest not changing that code in your PRs; if it is broken and nobody fixes it, then nobody is really using it, otherwise users can send fixes.

d3zd3z:
I'd be ok making the buffer just a static. The only benefit I see to allocating here is if this function doesn't get called. I don't see any point in trying to free this buffer, however.

FarK:
I don't know how to proceed with this. I think we all agreed this is not a leak, so maybe this PR should be closed? I could make a new PR adding an explanatory commentary or making the buffer static, but I think it will be pointless if this function has no use and it will be deprecated. Therefore, I think I should wait for @utzig to bring some info from Mynewt about the use of `split_go()`.

d3zd3z:
I would suggest just closing the PR. I do believe `split_go()` is used by MyNewt, but it still happens right before a boot, that will discard the current heap.

PR #1038, opened 2021-06-24
Allow to boot primary image if secondary one is unreachable

nvlsianpu:
In case secondary image device has a malfunction,
It might be still possible to boot the primary image.

This make sense especially if the secondary image resides in an
external flash which might be damaged while SoC is still working.

I also added configuration which helps to test this configuration:

For building mcuboot use following comand-line:
west build -d build/mcuboot_y -b nrf52840dk_nrf52840 \
bootloader/mcuboot/boot/zephyr/ -- \
-DDTC_OVERLAY_FILE=./boards/nrf52840dk_ram.overlay \
-DOVERLAY_CONFIG=./boards/flash_sim_driver.conf

Need additionally to modify the zephyr/drivers/flash/flash_simulator.c
driver, so flash_sim_read() always returns an error e.g: -EINVAL.


boot/bootutil/src/loader.c
```diff
@@ -1654,7 +1655,11 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
          * if there is one.
          */
         BOOT_SWAP_TYPE(state) = BOOT_SWAP_TYPE_NONE;
-        return;
+        if (BOOT_EFLASH == rc)
```
d3zd3z, line 1658: Prefer `if (rc == BOOT_EFLASH)`. Gcc will warn (which will stop the build at least in the sim) if an assignment were used instead of equality. Otherwise, this is more readable with the variable on the left.

d3zd3z (changes requested):

de-nordic (approved):
Looks OK. Thanks!

d3zd3z (approved):

PR #1041, opened 2021-06-28
boot : Adding encrypted ram-load support

salomethirot-arm:
This patch adds the possibility to boot using the ram-loading strategy with an
encrypted image. This is enabled when both the flags MCUBOOT_RAM_LOAD
and MCUBOOT_ENC_IMAGES are defined.

Signed-off-by: Hugo L'Hostis <hugo.lhostis@arm.com>
Signed-off-by: Salome Thirot <salome.thirot@arm.com>

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -149,8 +149,8 @@ struct image_tlv {
     uint16_t it_len;    /* Data length (not including TLV header). */
 };
 
-#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags && IMAGE_F_ENCRYPTED_AES128) \
-                        || ((hdr)->ih_flags && IMAGE_F_ENCRYPTED_AES256))
+#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags & IMAGE_F_ENCRYPTED_AES128) \
+                        || ((hdr)->ih_flags & IMAGE_F_ENCRYPTED_AES256))
 #define MUST_DECRYPT(fap, idx, hdr) \
     ((fap)->fa_id == FLASH_AREA_IMAGE_SECONDARY(idx) && IS_ENCRYPTED(hdr))
 
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -143,8 +143,10 @@ _Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
 #else
 #define ARE_SLOTS_EQUIVALENT()    1
 
+#ifdef MCUBOOT_DIRECT_XIP
 #ifdef MCUBOOT_ENC_IMAGES
-#error "Image encryption (MCUBOOT_ENC_IMAGES) is not supported when MCUBOOT_DIRECT_XIP or MCUBOOT_RAM_LOAD mode is selected."
+#error "Image encryption (MCUBOOT_ENC_IMAGES) is not supported when MCUBOOT_DIRECT_XIP is selected."
+#endif
 #endif
 #endif /* MCUBOOT_DIRECT_XIP || MCUBOOT_RAM_LOAD */
 
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2379,6 +2379,109 @@ boot_verify_ram_load_address(struct boot_loader_state *state,
     return 0;
 }
 
+#ifdef MCUBOOT_ENC_IMAGES
+
+/**
+ * Copies and decrypts an image from a slot in the flash to an SRAM address.
+ *
+ * @param  state    Boot loader status information.
+ * @param  slot     The flash slot of the image to be copied to SRAM.
+ * @param  hdr      The image header.
+ * @param  src_sz   Size of the image.
+ * @param  img_dst  Pointer to the address at which the image needs to be
+ *                  copied to SRAM.
+ *
+ * @return          0 on success; nonzero on failure.
+ */
+static int
+boot_decrypt_and_copy_image_to_sram(struct boot_loader_state *state,
+                                    uint32_t slot, struct image_header *hdr,
+                                    uint32_t src_sz, uint32_t img_dst )
+{
+    /* The flow for the decryption and copy of the image is as follows :
+     * 1. The whole image is copied to the RAM (header + payload + TLV).
+     * 2. The encryption key is loaded from the TLV in flash.
+     * 3. The image is then decrypted chunk by chunk in RAM (1 chunk
+     * is 1024 bytes). Only the payload section is decrypted.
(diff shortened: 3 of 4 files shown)


d3zd3z (commented):
We should wait for #1019 to merge, and add some test support for this change to the sim.

boot/bootutil/include/bootutil/image.h
```diff
@@ -149,8 +149,8 @@ struct image_tlv {
     uint16_t it_len;    /* Data length (not including TLV header). */
 };
 
-#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags && IMAGE_F_ENCRYPTED_AES128) \
-                        || ((hdr)->ih_flags && IMAGE_F_ENCRYPTED_AES256))
+#define IS_ENCRYPTED(hdr) (((hdr)->ih_flags & IMAGE_F_ENCRYPTED_AES128) \
```
d3zd3z, line 152: This is just a bug, and should probably be a separate commit that just fixes it (along with an explanation).

boot/bootutil/src/bootutil_priv.h
```diff
@@ -143,8 +143,10 @@ _Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
 #else
 #define ARE_SLOTS_EQUIVALENT()    1
 
+#ifdef MCUBOOT_DIRECT_XIP
```
d3zd3z, line 146: I wonder if `#if defined(MCUBOOT_DIRECT_XIP) && defined(MCUBOOT_ENC_IMAGES)` might make this a little clearer.

boot/bootutil/src/loader.c
```diff
@@ -2379,6 +2379,109 @@ boot_verify_ram_load_address(struct boot_loader_state *state,
...
+ * @param  src_sz   Size of the image.
+ * @param  img_dst  Pointer to the address at which the image needs to be
+ *                  copied to SRAM.
+ *
+ * @return          0 on success; nonzero on failure.
+ */
+static int
+boot_decrypt_and_copy_image_to_sram(struct boot_loader_state *state,
+                                    uint32_t slot, struct image_header *hdr,
+                                    uint32_t src_sz, uint32_t img_dst )
```
d3zd3z, line 2399: Stray space before the ')'.

boot/bootutil/src/loader.c
```diff
@@ -2379,6 +2379,109 @@ boot_verify_ram_load_address(struct boot_loader_state *state,
...
+    area_id = flash_area_id_from_multi_image_slot(BOOT_CURR_IMG(state), slot);
+    rc = flash_area_open(area_id, &fap_src);
+    if (rc != 0){
+        return BOOT_EFLASH;
+    }
+
+    tlv_off = BOOT_TLV_OFF(hdr);
+
+    /* Copying the whole image in RAM */
+    rc = flash_area_read(fap_src, 0, (void *)img_dst, src_sz);
```
d3zd3z, line 2432: It'd probably be clearer to have another variable that stores the destination address, as a pointer. If you search the code for IMAGE_RAM_BASE, you can see what is needed to make this work under the simulator. Basically, instead of `(void *)img_dst, you'll want `(void *)(IMAGE_RAM_BASE + img_dst)`. This will then work on target as well as the simulator.

You could either change both instances, or as I mentioned, create a new variable to hold that value.

boot/bootutil/src/loader.c
```diff
@@ -2539,11 +2642,19 @@ boot_load_image_to_sram(struct boot_loader_state *state,
             return rc;
         }
 #endif
-
+#ifdef MCUBOOT_ENC_IMAGES
+        /* decrypt image if encrypted and copy it to RAM */
+        if (IS_ENCRYPTED(hdr)){
+            rc = boot_decrypt_and_copy_image_to_sram(state, active_slot, hdr, img_sz, img_dst);
+        }else {
```
d3zd3z, line 2649: Missing space between `}` and the `else`.

boot/bootutil/src/loader.c
```diff
@@ -2539,11 +2642,19 @@ boot_load_image_to_sram(struct boot_loader_state *state,
             return rc;
         }
 #endif
-
+#ifdef MCUBOOT_ENC_IMAGES
+        /* decrypt image if encrypted and copy it to RAM */
+        if (IS_ENCRYPTED(hdr)){
```
d3zd3z, line 2647: Missing space before the `{`.

d3zd3z (changes requested):
This looks good. With some changes I've mentioned regarding `IMAGE_RAM_BASE`, this does work in the simulator, and running something like:
```sh
% cd sim
% cargo run --features sig-rssa validate-primary-slot ram-load enc-rsa
```
Will show what is failing when that happens. I will continue to investigate the test failure this causes (it is likely the simulator needs to be taught about encrypted ram loading).

boot/bootutil/include/bootutil/image.h
salomethirot-arm: Just created https://github.com/mcu-tools/mcuboot/pull/1072 to fix this

d3zd3z:
My initial testing is showing that it is unable to validate the image slow when ram loading and one of the signatures is enabled, for example:
```sh
cargo test --features 'sig-rsa ram-load enc-rsa'
```
gives me a failure at:
```
test ram_load_first ... [2021-07-28T18:25:18Z ERROR bootsim::image] Failed to execute ram-load
thread 'main' panicked at 'assertion failed: !image.run_ram_load()', sim/tests/core.rs:63:1
```
and stepping through the debugger shows that `boot_image_check` is trying to verify the signature on the encrypted image instead of the decrypted one in warm. It also fails when `validate-primary-slot` is not present.

salomethirot-arm:
Thanks @d3zd3z for reporting this. I will look into it.

salomethirot-arm:
Hi @d3zd3z. Thank you for testing this. I fixed a couple of issues that affected the image validation. However there is one last that I'm not sure whether it comes from my code or the simulator. At the moment the encrypted ram_load test fails because the size used to compare the images in `run_ram_load()` is 4 bytes larger than the size used by mcuboot to decrypt the image (loaded with `boot_read_image_size()`). Those 4 bytes are the only ones that differ between both images and they seem to have been added by some padding code in the simulator in `install_image()`. I'm not sure why the problem would come from my code here, and I'm not hugely familiar with the simulator, do you have any idea on this?

d3zd3z:
I will investigate the padding. The padding is done in order to make the image a multiple of the write alignment, but it shouldn't really affect the image sizes seen.

d3zd3z:
I'll see if I can fix this, as I think it is a simulator issue.

salomethirot-arm:
Thank you!

d3zd3z:
Just to let you know, I'm still investigating this. I have run into an issue of other tests failing within the simulator that I need to track down as well.

salomethirot-arm:
Hi @d3zd3z, any update on this?

PR #1047, opened 2021-07-02
bootutil: Include missing header and prevent ASSERT redefinition

gustavonihei:
This PR intends to provide two fixes for minor issues encountered during the creation of a new port:
- The `mcuboot_config.h` header is missing from `bootutil_public.h` includes for the evaluation of the `MCUBOOT_HAVE_ASSERT_H` definition.
- `ASSERT` macro may be already defined by the underlying OS, so we need to avoid redefining it.


gustavonihei:
@nvlsianpu @utzig @d3zd3z @davidvincze, could you please take a look at this small PR?

d3zd3z (approved):

PR #1050, opened 2021-07-07
Initialize aes-ctr during swap recovery

### LINKED ISSUE #1046: bootutil_aes_ctr_init is never called during a swap recovery
Hello,

So I'm using MCUboot for a project of mine, and I noticed a weird behavior and I'm not sure whether that is intentional.

With `MCUBOOT_ENC_IMAGES` and `MCUBOOT_SWAP_USING_SCRATCH`, during a "normal" update, `bootutil_aes_ctr_init` is called as expected before any use of `bootutil_aes_ctr_set_key` or `bootutil_aes_ctr_en(/de)crypt`, but during a swap recovery it is not the case.

If a swap is interrupted (e.g., by resetting the board at any time after the first swap status was persisted, for example during the "Secondary Slot => Scratch Area" phase of the swap), the swap will be resumed as expected, but putting a breakpoint in the `bootutil_aes_ctr_init` (and/or logging) shows it is never called on this execution path; `bootutil_aes_ctr_set_key` and `bootutil_aes_ctr_encrypt` are both called before any call to `bootutil_aes_ctr_init`.

This causes no problems when no hardware crypto is used with `MCUBOOT_USE_TINYCRYPT` (since the init function is empty), but once I implemented a hardware aes_ctr function, the init function is actually required and thus the encryption fails during a swap recovery (but works as expected is the swap is never interrupted). Is this an oversight or did I miss something?

The quick fix I did on my side is in [loader.c](https://github.com/mcu-tools/mcuboot/blob/1c4562473e4a82e8e6b641b223a935c48e0b5b74/boot/bootutil/src/loader.c#L1247), in `boot_swap_images` (the link should be a the correct line), I added a call to `boot_enc_init`, so the code looks like this instead:

```
        for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
            boot_enc_init(BOOT_CURR_ENC(state), slot); // !
            rc = boot_read_enc_key(image_index, slot, bs);
            assert(rc == 0);
```

Adding this line fixes my issue and the swap can be recovered as expected.

Can someone confirm this?
I thought I should let you know, just in case it is really an issue.

pokemaniak:
Fixes #1046 

`bootutil_aes_ctr_init` is never called when MCUboot resumes a swap after a reset, which is a problem in case of hw implementation of AES-CTR.

This pull request simply fixes this by calling `boot_enc_init` before setting keys when `!boot_status_is_reset()`

Signed-off-by: Alex Ferragni <alex.ferragni@csem.ch>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1245,6 +1245,7 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
 
 #ifdef MCUBOOT_ENC_IMAGES
         for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
+            boot_enc_init(BOOT_CURR_ENC(state), slot);
             rc = boot_read_enc_key(image_index, slot, bs);
             assert(rc == 0);
 


d3zd3z:
I want to think a little bit on how we might be able to test this. The simulator doesn't test this, and maybe we want some extra work to happen to make sure this is properly initialized, perhaps even wrapping the structure with one that has an extra field in it to detect if the initialization didn't happen (and only on sim builds).

PR #1058, opened 2021-07-12
[DNM ] Introduce image access hooks 

nvlsianpu:
Introduced hooks interface which allows to override MCUboot's native rutines in order to serve cases when MCUboot doesn't have direct access to an image slot (especially a primary image slot).

Hooks allows developers to customize mcuboot behavior without hacking the mcuboot code itself.


boot/bootutil/include/bootutil/boot_hooks.h
```diff
@@ -0,0 +1,75 @@
...
+
+/** Hook for provide image header data.
+ *
+ * @param img_index the index of the images pair
+ * @param slot slot number
+ * @param img_hed image header structure to be populated
+ *
+ * @retval 0: header was read/populated
+ *         FIH_FAILURE: image is invalid,
+ *         1 if hook not implemented for the image-slot,
```
de-nordic, line 45: The 1 may also mean `follow the normal path`; at least it looks like that in the code, which is useful: for example user may trigger something over GPIO or UART, with the hooks, but relay on the "original" path to do the true update.

nvlsianpu: applied your comment. Instead of `1` now I'm using `BOOT_HOOK_REGULAR` for transfer that status

sigvartmh (changes requested):

nvlsianpu:
@sigvartmh I removed files you mentioned. They are kept here https://github.com/mcu-tools/mcuboot/pull/1084 for help purpose.

sigvartmh (changes requested):
I have some changes to suggest I just need to clean them up a bit.

boot/bootutil/src/loader.c
```diff
@@ -756,8 +763,15 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         }
     }
 #endif
-
-    FIH_CALL(boot_image_check, fih_rc, state, hdr, fap, bs);
+#ifdef MCUBOOT_IMAGE_ACCESS_HOOKS
+    if (slot == 0) {
```
sigvartmh, line 767: instead of having this if slot check could we just pass the slot to the function like this
`boot_image_check_hook(int slot, int img_index);`

sigvartmh (changes requested):
See the above change^

boot/bootutil/src/loader.c
nvlsianpu: sure, will change this as you are suggesting.

sigvartmh: Also remove `if (slot == 0) {` now.

nvlsianpu: Done

sigvartmh:
I'll try out these changes and up-date  https://github.com/nrfconnect/sdk-mcuboot/pull/165

boot/boot_serial/src/boot_serial.c
```diff
@@ -186,17 +187,34 @@ bs_list(char *buf, int len)
             if (flash_area_open(area_id, &fap)) {
                 continue;
             }
+#ifdef MCUBOOT_IMAGE_ACCESS_HOOKS
```
d3zd3z, line 190: Rather than make this conditional, can we just always have hooks, even if there is hook that just accesses the data directly?

d3zd3z (changes requested):
My primary comment is that I'd like to see this done without ifdefs. Just add the hook stuff to the main boot files, always. We can either conditionalize the implementations in the header (to generate no code for the hooks), or we can include a stub implementation that does nothing (a function that does nothing will be 1 instruction on arm, and the call 1-2).

But, we really need to be moving to the direction of fewer ifdefs in the boot code, not more.

Otherwise, the idea seems fine, I'm just worried about the code getting more cluttered.

boot/boot_serial/src/boot_serial.c
de-nordic: I think it takes more operations, depending on number of parameters the function takes.
Maybe inlined stubs could be optimize by compiler.

I kinda agree on that no-ifdefs policy as this annoys me in C a lot; the solution is to either use the `IS_ENABLED` type of macro
```
if (IS_ENABLED(MCUBOOT_IMAGE_ACCESS_HOOKS) && boot_read_image_header_hook(image_index, slot, &hdr) == BOOT_HOOK_REGULAR) {
...
}
```
like in Zephyr, but this also requires code to be specifically written as you need to have all definitions used within such `if` available till compiler is able to optimize them out.
Another option would be to define macros, when `MCUBOOT_IMAGE_ACCESS_HOOKS` is not defined, that pretend to be hook functions and return specific literal, but this has to be carefully done, for example:
```
#ifndef MCUBOOT_IMAGE_ACCESS_HOOKS
#define boot_read_image_header_hook(just, dump, it) SPECIFIC_LITERAL
#endif

int rc = boot_read_image_header_hook(image_index, slot, &hdr);
if (rc == NOT_THAT_SPECIFIC_LITERAL) {
  ... // should be optimized out by compiler when it figures out that rc is set to literal
}
```
and also requires all definitions used within the block to be defined during compilation.

boot/boot_serial/src/boot_serial.c
```diff
@@ -397,8 +415,12 @@ bs_upload(char *buf, int len)
                     goto out;
                 }
             }
-        }
 #endif
+#ifdef MCUBOOT_IMAGE_ACCESS_HOOKS
+            rc = boot_serial_uploaded_hook(img_num, fap, img_size);
```
de-nordic, line 420: The `rc` is not validated anywhere, the static analysis tool will be poking us for that.

de-nordic (approved):
Looks OK.
Left one comment over non-checked RC.
I agree with @d3zd3z that we should limit number of #ifdefs (https://github.com/mcu-tools/mcuboot/pull/1058/files#r694135467), but I am not sure what approach should we follow; I think that it can be left as it is for this PR.

We @d3zd3z should figure common approach to ifdefs, and probably set some milestone date or that, and walk through the code and refactor the ifdefs in some consistent way.

boot/boot_serial/src/boot_serial.c
nvlsianpu: What about construction with macro like this.
``` C
#ifdef MCUBOOT_IMAGE_ACCESS_HOOKS
#define HOOK_CALL(f, ret, ret_defult, ...) \
    do { \
            ret = f(__VA_ARGS__); \
    } while (0)
#else
#define HOOK_CALL(f, ret, ret_defult, ...) \
    do { \
            ret = ret_default; \
    } while (0)
#endif
```

```suggestion
        HOOK_CALL(boot_read_image_header_hook, rc, BOOT_HOOK_REGULAR, BOOT_CURR_IMG(state), i, boot_img_hdr(state, i));
        if (rc == BOOT_HOOK_REGULAR)
        {
            flash_area_read(fap, 0, &hdr, sizeof(hdr));
        }
```

` if (rc == BOOT_HOOK_REGULAR)` can be embedded into HOOK_CALL macro if you are thinking that it is better.

boot/boot_serial/src/boot_serial.c
```diff
@@ -186,17 +187,34 @@ bs_list(char *buf, int len)
...
-              bootutil_img_validate(NULL, 0, &hdr, fap, tmpbuf, sizeof(tmpbuf),
-                                    NULL, 0, NULL)) {
-                flash_area_close(fap);
-                continue;
+            if (hdr.ih_magic == IMAGE_MAGIC)
+            {
+#ifdef MCUBOOT_IMAGE_ACCESS_HOOKS
+                FIH_CALL(boot_image_check_hook, fih_rc, image_index, slot);
+                if (fih_eq(fih_rc, BOOT_HOOK_REGULAR))
+#endif
```
d3zd3z: I think having ifdefs that don't match up with the C structure bother a lot more than having all of the ifdefs. Can we at least put this ifdef in a header, and define these as empty macros (well, ones like @de-nordic showed that don't generate warnings) when the feature isn't defined. That would at least move the clutter to a header instead of all through out the source code.

d3zd3z (changes requested):

boot/boot_serial/src/boot_serial.c
nvlsianpu: now is treated as operation error

boot/boot_serial/src/boot_serial.c
nvlsianpu: Introduced macro for call the hook which become no-op in case hooks are not enabled.

nvlsianpu:
I messed with hooks which uses FIH_CALL.

PR #1065, opened 2021-07-15
boot: bootutil: Fix memory alignment of RAM buffer

MarekPieta:
Change fixes memory alignment of the RAM buffer that is used to temprorarily store data during swap. Some FLASH drivers require word-aligned input data buffer. Using unaligned buffer results in FLASH write error.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -813,7 +813,7 @@ boot_copy_region(struct boot_loader_state *state,
     uint8_t image_index;
 #endif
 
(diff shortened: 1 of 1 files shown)


nvlsianpu (approved):

d3zd3z (approved):

d3zd3z:
This change is specific to GCC (and probably will work with clang). The code only currently works with these compilers anyway, so it really isn't a problem to add this, but we are, at some point, probably going to have to abstract some of this at least a little so that we can make this work on other compilers.

PR #1067, opened 2021-07-19
boot: Revert "boot: Check shared area more carefully before init"

Sherryzhang2:
The commit "boot: Check shared area more carefully before init" adds
the support that the shared area can be used by other boot stages.
Before this commit, MCUboot initialize the shared memory based on a
global variable which indicates whether the shared memory has been
initialized by MCUboot. After this commit, MCUboot also checks the magic
value as well as the sanity check before initialization. So if the data
in shared memory retains after a reset, MCUboot does not initialize the
shared memory thus 'SHARED_MEMORY_OVERWRITE' error happens in
'boot_add_data_to_shared_area'.

So reverted this commit temporarily.

Change-Id: I1fc390bc17f90c2624024bc101ba8b4d5a75fe23

### DIFF
--- a/boot/bootutil/src/boot_record.c
+++ b/boot/bootutil/src/boot_record.c
@@ -62,18 +62,13 @@ boot_add_data_to_shared_area(uint8_t        major_type,
 
     boot_data = (struct shared_boot_data *)MCUBOOT_SHARED_DATA_BASE;
 
-    /* Check whether first time to call this function. If it is, then check
-     * whether the shared data area needs to be initialised.
+    /* Check whether first time to call this function. If does then initialise
+     * shared data area.
      */
     if (!shared_memory_init_done) {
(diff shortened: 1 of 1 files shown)


tamasban (approved):
I approve this revert:
-  with regard to the coming TF-M code freeze deadline due to the upcoming release
-  furthermore, its counterpart in TF-M has not been reviewed and  merged

d3zd3z (commented):
I think this is ok, but I think the commit text could use a bit more explanation as to why this is needed. What are the scenarios where the magic value is present but the data is not valid?

Sherryzhang2:
> I think this is ok, but I think the commit text could use a bit more explanation as to why this is needed. What are the scenarios where the magic value is present but the data is not valid?

Done.
In the updated commit message, I added the description of "why" this revert is needed. But I did not write a description about when this commit can be added back as I do not want to give that commitment.

d3zd3z (approved):

PR #1070, opened 2021-07-20
Introduce random IV and salt - REBASED to latest changes

romanjoe:
Creating new PR, since previous one (https://github.com/mcu-tools/mcuboot/pull/962) became incompatible with latest changes in mcuboot.

This one does not include @utzig changes (https://github.com/mcu-tools/mcuboot/pull/980), because now they cause failures in sim run. Do we need to investigate on that?

Commits:
imgtool: adds flag --use-random-iv which allows randomised data embedding to HKDF data - turned off by default to keep compatibility.
bootutil: changes to support update in mcuboot
cypress: updates to cypress code to support changes

### DIFF
--- a/boot/bootutil/include/bootutil/enc_key.h
+++ b/boot/bootutil/include/bootutil/enc_key.h
@@ -45,6 +45,7 @@ extern "C" {
 
 struct enc_key_data {
     uint8_t valid;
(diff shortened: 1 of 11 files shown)


utzig:
> This one does not include @utzig changes (#980), because now they cause failures in sim run. Do we need to investigate on that?

Yes, otherwise there's not sim testing support.

PR #1074, opened 2021-07-25
zephyr: do not set defaults for LOG_IMMEDIATE Kconfig

mnkp:
LOG_IMMEDIATE Kconfig option has been repurposed and is now a Zephyr's internal, non-visible symbol that should not be modified by the end user.

The option's default value has been added in the past to configure synchronous logging if multi-threading is disabled. At present it prevents the user from configuring synchronous logging mode when multi-threading is enabled.

Remove the defaults altogether as the logging mode used by MCUBoot is defined in prj.conf.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -663,10 +663,6 @@ config MULTITHREADING
 	default n if SOC_FAMILY_NRF
 	default y
 
-config LOG_IMMEDIATE
-	default n if MULTITHREADING
-	default y
-
 config LOG_PROCESS_THREAD
 	default n # mcuboot has its own log processing thread
 


boot/zephyr/Kconfig
```diff
@@ -663,10 +663,6 @@ config MULTITHREADING
 	default n if SOC_FAMILY_NRF
 	default y
 
-config LOG_IMMEDIATE
```
nvlsianpu, line 666: why not use LOG_MODE_IMMEDIATE instead?

This is commit with was introduced this https://github.com/mcu-tools/mcuboot/commit/9a4946ce9a6d20e1710096cac5a7b476f85de65f for some expalanation.

mnkp: I have seen the above commit, however a later commit e75e33d reverted the change by defining logging mode in prj.conf. It felt like a good time to clean it up.

Having the default and setting the Kconfig option in prj.conf is of course not the same but
- logging in Zephyr is undergoing refactoring so the settings are not really stable at the moment.
- LOG_MODE_IMMEDIATE is now a choice option, cannot be assigned a default value, the logic would need to be reworked.
- if we are not choosing the synchronous logging mode then the default in Zephyr is the deferred mode. It is however quite expensive, requires extra ~4K of flash.

mnkp, line 666: There is one more occurrence of `LOG_IMMEDIATE` in this file on line 445, however I believe it's usage is consistent with how the symbol is supposed to be used / is used by Zephyr. Please double check.

nvlsianpu: Yes, it is consistent.

nvlsianpu: Ok. Will merge this then.
I prefer to configure as much as possible automatically for the developer (prj.conf is not automatic thing), but in that state of things it doesn't make sense as you elaborated.

nvlsianpu (approved):

PR #1077, opened 2021-07-27
boot: zephyr: add integration platforms to the allow list

ioannisg:
The platforms that are listed as integration platforms in
the Zephyr boot sample need to be included in the platform
allow list, otherwise zephyr CI will through an error.

Signed-off-by: Ioannis Glaropoulos <Ioannis.Glaropoulos@nordicsemi.no>

### DIFF
--- a/boot/zephyr/sample.yaml
+++ b/boot/zephyr/sample.yaml
@@ -1,34 +1,42 @@
 sample:
   description: mcuboot test build sample
   name: mcuboot
-common:
-  integration_platforms:
-    - nrf52840dk_nrf52840
-    - nrf52840dongle_nrf52840
 
 tests:
   sample.bootloader.mcuboot:
     tags: bootloader_mcuboot
     platform_allow: nrf52840dk_nrf52840 frdm_k64f
+    integration_platforms:
+      - nrf52840dk_nrf52840
   sample.bootloader.mcuboot.serial_recovery:
     extra_args: OVERLAY_CONFIG=serial_recovery.conf
     platform_allow:  nrf52840dk_nrf52840
+    integration_platforms:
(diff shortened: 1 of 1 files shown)


nvlsianpu (commented):
Why you added nrf52840dongle_nrf52840 to cases where it had been not listed before?

ioannisg:
> d been not lis

@nvlsianpu 
The reason is that it was added as integration platform. So as far as I understood, if a platform is listed as an integration platform, but it is not in the allow-list, twister throws an error. So I added the 52840 dongle wherever it was possible. Apparently, this was not possible for the _qspi_ variant of the sample, so i had to remove the "common" definition of the integration platforms and move the integration platform definition inside each test variant.

nvlsianpu:
@ioannisg allow_platform list can be wider than integration_platform. I put nrf52840dongle target only where it is needed. Rationale is it is based on the same SoC as the DK, and DK is much more common. I also want to reduce CI infrastructure load.
The bug you are fixing is "quite new"

What you can do is to not extend load as follow: don't add dongle to cases where it was not available on allow platform before.

ioannisg:
> @ioannisg allow_platform list can be wider than integration_platform. I put nrf52840dongle target only where it is needed. Rationale is it is based on the same SoC as the DK, and DK is much more common. I also want to reduce CI infrastructure load.
> The bug you are fixing is "quite new"
> 
> What you can do is to not extend load as follow: don't add dongle to cases where it was not available on allow platform before.

Fixed that now, @nvlsianpu . Of course, I had to remove the common definition of the integration platform list and keep the case-by-case definition of the int platforms in each variant.

nvlsianpu (approved):

PR #1080, opened 2021-07-28
Custom image property list comand

nvlsianpu:
Added service for custom command for fetch non-standard image status.

cddl-gen header also fixed: https://github.com/NordicSemiconductor/cddl-gen/pull/77

### DIFF
--- a/boot/zephyr/boot_serial_extensions.c
+++ b/boot/zephyr/boot_serial_extensions.c
@@ -50,6 +50,7 @@ static int bs_custom_storage_erase(cbor_state_t *cs)
     return rc;
 }
 
+#ifdef MCUBOOT_MGMT_CUSTOM_IMG_LIST
 static int custom_img_status(int image_index, uint32_t slot,char *buffer,
                         ssize_t len)
 {
@@ -117,6 +118,8 @@ static int bs_custom_img_list(cbor_state_t *cs)
 #ifndef ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST
     #define ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST 1
 #endif
+#endif /*MCUBOOT_MGMT_CUSTOM_IMG_LIST*/
+
 int bs_peruser_system_specific(const struct nmgr_hdr *hdr, const char *buffer,
                                int len, cbor_state_t *cs)
 {
@@ -128,9 +131,11 @@ int bs_peruser_system_specific(const struct nmgr_hdr *hdr, const char *buffer,
                 mgmt_rc = bs_custom_storage_erase(cs);
             }
         } else if (hdr->nh_op == NMGR_OP_READ) {
+#ifdef MCUBOOT_MGMT_CUSTOM_IMG_LIST
             if (hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST) {
                 mgmt_rc = bs_custom_img_list(cs);
             }
+#endif
         }
     }
 
--- a/boot/zephyr/boot_serial_extensions.c
+++ b/boot/zephyr/boot_serial_extensions.c
@@ -15,42 +15,131 @@
 #include "../boot_serial/src/boot_serial_priv.h"
 #include "../boot_serial/src/cbor_encode.h"
 
+#include "bootutil/image.h"
+
 MCUBOOT_LOG_MODULE_DECLARE(mcuboot);
 
-int bs_peruser_system_specific(const struct nmgr_hdr *hdr, const char *buffer,
-                               int len, cbor_state_t *cs)
+static int bs_custom_storage_erase(cbor_state_t *cs)
 {
-    int mgmt_rc = MGMT_ERR_ENOTSUP;
-    int rc = -ENOTSUP;
+    int rc;
 
-    if (hdr->nh_group == ZEPHYR_MGMT_GRP_BASE &&
-        hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_ERASE_STORAGE &&
-        hdr->nh_op == NMGR_OP_WRITE) {
+    const struct flash_area *fa;
 
-        const struct flash_area *fa;
+    rc = flash_area_open(FLASH_AREA_ID(storage), &fa);
 
-        rc = flash_area_open(FLASH_AREA_ID(storage), &fa);
-
+    if (rc < 0) {
+        LOG_ERR("failed to open flash area");
+    } else {
+        rc = flash_area_erase(fa, 0, FLASH_AREA_SIZE(storage));
         if (rc < 0) {
-            LOG_ERR("failed to open flash area");
-        } else {
-            rc = flash_area_erase(fa, 0, FLASH_AREA_SIZE(storage));
-            if (rc < 0) {
-                LOG_ERR("failed to erase flash area");
-            }
-            flash_area_close(fa);
+            LOG_ERR("failed to erase flash area");
         }
-        if (rc == 0) {
-            mgmt_rc = MGMT_ERR_OK;
-        } else {
-            mgmt_rc = MGMT_ERR_EUNKNOWN;
-        }
+        flash_area_close(fa);
+    }
+    if (rc == 0) {
+        rc = MGMT_ERR_OK;
+    } else {
+        rc = MGMT_ERR_EUNKNOWN;
     }
 
     map_start_encode(cs, 10);
     tstrx_put(cs, "rc");
-    uintx32_put(cs, mgmt_rc);
+    uintx32_put(cs, rc);
     map_end_encode(cs, 10);
 
     return rc;
+}
+
+static int custom_img_status(int image_index, uint32_t slot,char *buffer,
+                        ssize_t len)
+{
+    uint32_t area_id;
+    struct flash_area const *fap;
+    struct image_header hdr;
+    int rc;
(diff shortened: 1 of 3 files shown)


boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
-    uintx32_put(cs, mgmt_rc);
+    uintx32_put(cs, rc);
     map_end_encode(cs, 10);
 
     return rc;
 }
+
+#ifndef ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST
+    #define ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST 1
+#endif
```
de-nordic, line 125: OK, so this will be gone when the mcumgr header in Zephyr gets updated?

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
+            break;
+        }
+    }
+
+    if (mgmt_rc == MGMT_ERR_ENOTSUP) {
+        map_start_encode(cs, 10);
+        tstrx_put(cs, MGMT_ERR_ENOTSUP);
+        uintx32_put(cs, mgmt_rc);
+        map_end_encode(cs, 10);
+    }
```
de-nordic, line 148: This is great, I should have done that in original implementation of the command.

boot/zephyr/boot_serial_extensions.c
nvlsianpu: Will remove this as soon as zephyr-main will be updated.

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
+    struct flash_area const *fap;
+    struct image_header hdr;
+    int rc;
+    int img_install_stat = 0;
+
+    area_id = flash_area_id_from_multi_image_slot(image_index, slot);
+
+    rc = flash_area_open(area_id, &fap);
+    if (rc) {
+        return rc;
```
de-nordic, line 66: Space at the end of line.

boot/boot_serial/src/cbor_encode.h
```diff
@@ -55,10 +55,10 @@ bool bstrx_encode(cbor_state_t *state, const cbor_string_type_t *result);
 bool tstrx_encode(cbor_state_t *state, const cbor_string_type_t *result);
 
 #define tstrx_put(state, string) \
-	tstrx_encode(state, &(cbor_string_type_t){.value = (const uint8_t *)string, len = (sizeof(string) - 1)})
+	tstrx_encode(state, &(cbor_string_type_t){.value = (const uint8_t *)string, .len = (sizeof(string) - 1)})
```
de-nordic, line 58: The file at the top contains sha of the commit in the cddl repo that the file comes from. I think that this should be updated, and corrected in all other cddl files.

boot/zephyr/boot_serial_extensions.c
de-nordic: Oh I did, just to tired to remember...

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
+
+#ifndef ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST
+    #define ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST 1
+#endif
+int bs_peruser_system_specific(const struct nmgr_hdr *hdr, const char *buffer,
+                               int len, cbor_state_t *cs)
+{
+    int mgmt_rc = MGMT_ERR_ENOTSUP;
+
+    if (hdr->nh_group == ZEPHYR_MGMT_GRP_BASE && hdr->nh_op == NMGR_OP_WRITE) {
```
de-nordic, line 131: This has to be divided because for the `ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST` the `nh_op` is `NMGR_OP_READ` as it reads state, does not modify board.

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
+    char tmpbuf[64];	/* Buffer should fit version and flags */
+
+    map_start_encode(cs, 10);
+
+    for (int img = 0; img < MCUBOOT_IMAGE_NUMBER; img++) {
+        for (int slot = 0; rc == 0 && slot < 2; slot++) {
+            rc = custom_img_status(img, slot, tmpbuf, sizeof(tmpbuf));
+            if (rc == 0) {
+                intx32_put(cs, img * 2 +slot);
+                tstrx_put(cs, tmpbuf);
```
de-nordic, line 102: The `tstrx_put` should be replaced with `tsrx_put_term`, otherwise the cbor will encode entire tmpbuf, instead of the string that is within it.

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,137 @@
...
+{
+    int rc = 0;
+    char tmpbuf[64];	/* Buffer should fit version and flags */
+
+    map_start_encode(cs, 10);
+
+    for (int img = 0; img < MCUBOOT_IMAGE_NUMBER; img++) {
+        for (int slot = 0; rc == 0 && slot < 2; slot++) {
+            rc = custom_img_status(img, slot, tmpbuf, sizeof(tmpbuf));
+            if (rc == 0) {
```
de-nordic, line 100: I think it would be best to report every target image that exists instead of just slot.

de-nordic:
Got error 3 at the and but with slightly modified test script and removed one if got:
```
{0: 'ver=0.0.0.0,install_stat=0', 1: '', 'rc': 0}
```

boot/boot_serial/src/cbor_encode.h
nvlsianpu: ^^ I sent patch to cddl-gen as well, but i don't have any influence on when it will be merged.

boot/zephyr/boot_serial_extensions.c
nvlsianpu: ok

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,140 @@
...
+    int mgmt_rc = MGMT_ERR_ENOTSUP;
+
+    if (hdr->nh_group == ZEPHYR_MGMT_GRP_BASE) {
+        if (hdr->nh_op == NMGR_OP_WRITE) {
+            if (hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_ERASE_STORAGE) {
+                mgmt_rc = bs_custom_storage_erase(cs);
+            }
+        } else if (hdr->nh_op == NMGR_OP_READ) {
+            if (hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST) {
+                mgmt_rc = bs_custom_img_list(cs);
```
de-nordic, line 141: If you return non-0 here, no response will be sent out, this is due to how a result code of the `bs_peruser_sustem_specific` calls is processed here
https://github.com/mcu-tools/mcuboot/blob/fc1ad8d080e6b6252dc5337273c6195500d7caf0/boot/boot_serial/src/boot_serial.c#L509-L511

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,131 @@
...
+    int rc = 0;
+    char tmpbuf[64];	/* Buffer should fit version and flags */
+
+    map_start_encode(cs, 10);
+
+    for (int img = 0; img < MCUBOOT_IMAGE_NUMBER; img++) {
+        for (int slot = 0; slot < 2; slot++) {
+            rc = custom_img_status(img, slot, tmpbuf, sizeof(tmpbuf));
+
+            intx32_put(cs, img * 2 +slot);
```
de-nordic, line 101: Add 1 to image numbers, so that enumeration would start from 1, this makes it coherent with -n to mcumgr

nvlsianpu: done

de-nordic:
I think that the command should be only enabled when `MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD` is enabled; it only then makes sense. In such case, rather than iterating through images/slots the `flash_area_id_from_direct_image` should be considered for obtaining area_id for an image to verify.
The direct image upload is independent from `MCUBOOT_IMAGE_NUMBER`, you may have it set to 1 or 2 and yet be able to directly upload to any image you select.

boot/zephyr/boot_serial_extensions.c
```diff
@@ -15,42 +15,136 @@
...
+            if (hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST) {
+                mgmt_rc = bs_custom_img_list(cs);
+            }
+#endif
+        }
+    }
+
+    if (mgmt_rc == MGMT_ERR_ENOTSUP) {
+        map_start_encode(cs, 10);
+        tstrx_put(cs, MGMT_ERR_ENOTSUP);
```
de-nordic, line 144: This should be `tstrx_put(cs, "rc");`

nvlsianpu: done

de-nordic (approved):
Looks OK. I have been able to test it with custom script that allows to trigger the command and it works fine. All added Kconfig paths have been tested.

PR #1087, opened 2021-08-03
boot: zephyr: Configure mimxrt685_evk board

MaureenHelm:
The mimxrt685_evk board has large slots so we need to increase
CONFIG_BOOT_MAX_IMG_SECTORS from the default.

Signed-off-by: Maureen Helm <maureen.helm@nxp.com>

### DIFF
--- a/boot/zephyr/boards/mimxrt685_evk_cm33.conf
+++ b/boot/zephyr/boards/mimxrt685_evk_cm33.conf
@@ -1,0 +1,4 @@
+# Copyright 2021 NXP
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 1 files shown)


d3zd3z:
Do you know if this has been tested with this value? There are some structures at least within MCUboot that are sized based on this value. It still should be possible to make it work, though.

MaureenHelm:
> Do you know if this has been tested with this value? There are some structures at least within MCUboot that are sized based on this value. It still should be possible to make it work, though.

I tested with https://docs.zephyrproject.org/latest/samples/subsys/mgmt/mcumgr/smp_svr/README.html

@mmahadevan108 FYI

d3zd3z (approved):

PR #1088, opened 2021-08-04
boot: zephyr: User defined encryption key

bn-norbit:
Currently the image encryption keys are hard coded. This pull request provides configuration options so a custom encryption key file can be provided, similar to how custom signature key files are solved.

RSA  encryption has been built and tested using Zephyr 2.4 and nRF52840.

Other pull requests have also provided solutions for replacing the hard coded encryption key:
  * https://github.com/mcu-tools/mcuboot/pull/596
  * https://github.com/mcu-tools/mcuboot/pull/989

These pull requests have not yet been accepted.


boot/zephyr/Kconfig
```diff
@@ -289,6 +289,26 @@ config BOOT_ENCRYPT_X25519
 	  described under "ECIES-X25519 encryption" in docs/encrypted_images.md.
 endif # !SINGLE_APPLICATION_SLOT
 
+config BOOT_ENCRYPT_KEY_FILE
```
str4t0m, line 292: I believe BOOT_USE_ENCRYPT_KEY_FILE should be moved inside `#if !SINGLE_APPLICATION_SLOT`.

boot/zephyr/Kconfig
```diff
@@ -289,6 +289,26 @@ config BOOT_ENCRYPT_X25519
...
+	string "PEM key file"
+	default ""
```
str4t0m, line 294: When the keys are auto generated, ideally there would be no need for default keys in boot/zephyr/keys.c.
Those could be removed like it was done for the signature keys. Specifying enc-ec256-priv.pem, enc-rsa2048-priv.pem, enc-x25519-priv.pem as default keys for the related encryption mechanisms would be a first step.
For X25519 you would additionally have to take care that the automatically generated name does not match.
As result the BOOT_USE_ENCRYPT_KEY_FILE would no longer be needed.

boot/zephyr/CMakeLists.txt
```diff
@@ -290,6 +290,42 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
...
+
+    set(GENERATED_PRIVKEY ${ZEPHYR_BINARY_DIR}/autogen-privkey.c)
+    add_custom_command(
+        OUTPUT ${GENERATED_PRIVKEY}
+      COMMAND
+      ${PYTHON_EXECUTABLE}
+      ${MCUBOOT_DIR}/scripts/imgtool.py
+      getpriv
+      -k
+      ${ENCRYPTION_KEY_FILE}
```
str4t0m, line 321: You could consider adding the --minimal option.
Which was used for the default ec256 and x25519 keys.

str4t0m (commented):
I'd welcome an option to easily pass the encryption key, and have left a few comments.

boot/zephyr/Kconfig
bn-norbit: You are correct. Moved inside `#if !SINGLE_APPLICATION_SLOT`

boot/zephyr/Kconfig
bn-norbit: Adopted the same strategy as for generating public signature keys.

boot/zephyr/CMakeLists.txt
bn-norbit: Added `--minimal` for ec256 and x25519 keys only, and not for rsa key. The generated keys are equal to the keys removed from `keys.c`.

bn-norbit:
@str4t0m The Sim and Mynewt workflows failed due to commit signature error. I have force-pushed a fix. Can you approve the workflows one more time?

bn-norbit:
@str4t0m Added tests for encryption. Think this is ready for re-review. BTW: the "Re-request review" button doesn't work for me.

bn-norbit:
@str4t0m Can you run workflows again? Sorry for messing up...

maximevince:
I am interested in this. I have raised an issue with similar concerns here: https://github.com/mcu-tools/mcuboot/issues/1268
The question on a practical way to have unique per-device private keys remains. (E.g. something similar to what BOOT_HW_KEY does for signing keys)

wcappelle:
@d3zd3z , As mentioned in my PR, I can rework mine when this gets merged in. Similar approach has been taken in this PR, so it would not be that hard to rework for me. I prefer this PR work over mine since it's more extensive than my work in that area.
Note that #989 has been split into #1255 wrt keys & encrypted images.

samples/zephyr/run-tests.sh
```diff
@@ -191,6 +191,115 @@ pyocd commander -c reset
 echo "Expected result: hello1 runs"
 ok_yn
 
+echo '--------------------------------------------------------'
```
d3zd3z, line 194: Can you make comparable updates to samples/zephyr/mcutests/mcutests.go. The run-tests.sh is kind of deprecated, and at least all of my testing is done with the Go version.

d3zd3z (changes requested):
Looks good, just need to update the Go test runner.

samples/zephyr/run-tests.sh
bn-norbit: Looks like the same feature has been merged in #1255 on 13 Jan (!)

d3zd3z:
And we seem to have still ended up with conflicts, sorry. I hope this isn't too difficult to resolve.

PR #1089, opened 2021-08-06
boot: zephyr: Do not use `irq_lock()` if using arm cleanup

sigvartmh:
`irq_lock()` sets `BASEPRI_MAX` aka. the `BASEPRI` mask to whatever
zephyr has configured it to be by the value of `_EXC_IRQ_DEFAULT_PRIO`.

However by calling arm_cleanup() we also do the call to
`__disable_irq()` setting the PRIMASK to 1. Meaning the only exceptions
we can recive is fault exceptions. Masking out more exceptions does not
really make sense.

The rational behind this change is that sometimes applications booted by MCUBoot will not expect the `BASEPRI`
to be set to something other than 0(No effect). Meaning if they depend on
using some exception which now is masked out by `BASEPRI` they will
fail. 

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -183,7 +183,6 @@ static void do_boot(struct boot_rsp *rsp)
                                      rsp->br_image_off +
                                      rsp->br_hdr->ih_hdr_size);
 
-    irq_lock();
 #ifdef CONFIG_SYS_CLOCK_EXISTS
     sys_clock_disable();
 #endif
@@ -202,6 +201,8 @@ static void do_boot(struct boot_rsp *rsp)
 
 #if CONFIG_CPU_HAS_ARM_MPU || CONFIG_CPU_HAS_NXP_MPU
     z_arm_clear_arm_mpu_config();
+#else
+    irq_lock();
 #endif
 
 #if defined(CONFIG_BUILTIN_STACK_GUARD) && \


sigvartmh:
Another solution could be to `__set_BASEPRI(0)` before jumping to the next image or
```C
unsigned int key = irq_lock()
...
irq_unlock(key);
<jump>
```
Both these should be safe as the only exceptions which are still enabled are fault exceptions.
We could do `__disable_fault_irq ` also if we want to make sure all interrupts except `NMI` does not go through.

sigvartmh:
@ioannisg @d3zd3z @nvlsianpu could someone add someone to review this or review this if you are able to.

ioannisg:
I will take  a look, thanks.

d3zd3z (approved):
I guess the only real concern I have is whether we will get interrupts before they are expected in the application. Typically, the interrupts we would see in the bootloader would be about the UART, which tend to be relatively slow (and most setups aren't sending characters to the device at this point anyway).

But, I'm ok changing this, and we can deal with any issues that come up.

boot/zephyr/main.c
```diff
@@ -202,6 +201,8 @@ static void do_boot(struct boot_rsp *rsp)
 
 #if CONFIG_CPU_HAS_ARM_MPU || CONFIG_CPU_HAS_NXP_MPU
     z_arm_clear_arm_mpu_config();
+#else
+    irq_lock();
```
ioannisg, line 205: So the irq_lock() is not invoked when the CPU has MPU. Why? What is the relation between cleaning the IRQ state and whether the SoC has the MPU peripheral? I think these two are unrelated.

ioannisg:
> The rational behind this change is that sometimes applications booted by MCUBoot will not expect the BASEPRI
to be set to something other than 0(No effect). Meaning if they depend on
using some exception which now is masked out by BASEPRI they will
fail.

Zephyr applications will enable IRQs before jumping to main. So I do not see why this is a problem.

boot/zephyr/main.c
sigvartmh: Miss aligned the #else will fix it.

PR #1095, opened 2021-08-10
Move Cargo.toml to top level

d3zd3z:
Instead of requiring cargo to be run in the sim directory, create a
top-level "workspace" Cargo file, that references the others.  The main
result of this is that cargo can now be run in the top of the workspace,
and the 'target' directory will be placed there, rather than in the sim
directory.

This is primarily of benefit to tools such as RLS, Rust-analyzer and
various IDEs that expect a crate at the top level of the tree.


nvlsianpu (approved):
CI is passing which proves that this patch is OK.

nvlsianpu:
@d3zd3z looks like merging this cause many CI failures on master.

PR #1096, opened 2021-08-12
Prevent compilation of Espressif files on Mbed build

lambda-shuttle:
The old mbed CLI 1 tool, currently used to compile the MCUboot port for Mbed™ OS, compiles all files unless otherwise specified in `.mbedignore`.   The espressif directory in the boot folder was added to the ignore file to prevent failed builds. 

@pan-

### DIFF
--- a/.mbedignore
+++ b/.mbedignore
@@ -2,6 +2,7 @@ boot/boot_serial/*
 boot/mynewt/*
 boot/zephyr/*
 boot/cypress/*
+boot/espressif/*
 ci/*
 docs/*
 ptest/*


pan- (approved):

d3zd3z (approved):

d3zd3z (changes requested):
As mentioned by the failing test, you'll need to add a Signed-off-by line to the commit text. It is a little difficult getting that to happen with commits made by github. Maybe we should change our checking script. If you (Sudarshan) would like one of the maintainers to rebase the commit so that isn't needed, let us know. But, you will need to add the SoB to this commit. If you want us to do it, just state permission in a comment and we can also add a SoB for you.

lambda-shuttle:
Thanks for the information! I'll keep this in mind when making pull requests in the future. Yes, I'm fine with the repository's maintainers adding a SoB line for me.

d3zd3z:
I also added the additional text from the PR to the commit.

d3zd3z (approved):

PR #1102, opened 2021-08-17
ci: Run sim tests at top level

d3zd3z:
Implement the sim tests at top level change that was reverted earlier, after fixing the issues that were breaking it back then.  This converts the mcuboot tree into a Cargo workspace, with projects for the existing parts of the code. The cargo commands can now be run at the top-level in the tree, and when doing so, will run tests on all of the projects in the workspace, rather than just one.


utzig (approved):
LGTM

sim/mcuboot-sys/build.rs
```diff
@@ -110,14 +111,14 @@ fn main() {
         conf.file("../../ext/tinycrypt/lib/source/ecc.c");
         conf.file("../../ext/tinycrypt/lib/source/ecc_dsa.c");
         conf.file("../../ext/tinycrypt/lib/source/ecc_platform_specific.c");
-        conf.include("../../ext/mbedtls/library");
-        conf.file("../../ext/mbedtls-asn1/src/platform_util.c");
-        conf.file("../../ext/mbedtls-asn1/src/asn1parse.c");
+        conf.conf.include("../../ext/mbedtls/library");
+        conf.file("../../ext/mbedtls/library/platform_util.c");
+        conf.file("../../ext/mbedtls/library/asn1parse.c");
```
utzig, line 116: Mixing `ext/mbedtls-asn1/include` and adding the Mbed TLS source here could end up in API breakage, wouldn't it be better to just change this in another PR, fully removing the bundled ASN1 parser it that's the case? Also I am not sure I see the reason for doing this, is there a ASN1 parser security bug or something that is not working correctly?

d3zd3z: The problem is that we're, in some configurations, pulling in these two files from both libraries. When used as a library, the linker picks one, I'm not even sure which one. I can pull this change out, but it does means that a bare `cargo test` at the top of the tree will fail.

I'm only proposing not using the bundled code in the sim.

utzig: > I'm only proposing not using the bundled code in the sim.

Sure, move on then.

PR #1103, opened 2021-08-17
boot/espressif: Clone ESP-IDF only when required

shubhamkulkarni97:
Added `update = none` in .gitmodules to avoid unnecessary clone of ESP-IDF submodule.

The submodule will be cloned only when `--checkout` flag is passed when updating submodules.

This change will avoid maintaining two copies of IDF in Zephyr environment.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,3 +26,4 @@
 	path = boot/espressif/hal/esp-idf
 	url = https://github.com/espressif/esp-idf.git
 	branch = release/v4.3
+	update = none


d3zd3z:
How large does this end up being? I'm wondering if this will be confusing to any users. Right now, we have other SDKs that are always cloned. I'm not sure of a better way to handle this, though.

shubhamkulkarni97:
@d3zd3z The ESP-IDF submodule is around 400 MB in size.

The documentation PR https://github.com/mcu-tools/mcuboot/pull/1118 includes the `--checkout` flag for updating the submodules. So documentation will help users to update submodule correctly.

The main motive behind this PR is that adding this change will eliminate duplicate copy of ESP-IDF in Zephyr development environment.

Please let me know if it sounds good.

shubhamkulkarni97:
Closing this PR as none of the mcuboot submodules are cloned in Zephyr

PR #1105, opened 2021-08-18
boot_serial: cbor_encode: Fix encoding data length

### LINKED ISSUE #1104: serial recovery: can't upload any image bigger than 64 KB
We have cbor encode issue which cause bad integer encoding so some values can be encoded on byte length which is not power of 2.
for instance 0x010144 was encoded as `19 01 01 44`, but should be as `1a 00 01 01 44`
bug was Introduced with https://github.com/mcu-tools/mcuboot/pull/997

MarekPieta:
Change fixes encoding data length and adds support for big endian byte ordering. According to specification, data length can be encoded either on 1, 2, 4 or 8 bytes.

fixes #1104

### DIFF
--- a/boot/boot_serial/src/cbor_encode.c
+++ b/boot/boot_serial/src/cbor_encode.c
@@ -86,12 +86,35 @@ static bool value_encode_len(cbor_state_t *state, cbor_major_type_t major_type,
 static uint32_t get_result_len(const void *const input, uint32_t max_result_len)
 {
 	uint8_t *u8_result  = (uint8_t *)input;
+	size_t i;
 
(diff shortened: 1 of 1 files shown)


nvlsianpu:
cc  @oyvindronningstad

nvlsianpu (approved):

de-nordic (approved):
Looks OK.

nvlsianpu:
@MarekPieta Can you rebase on the main HEAD? Currently this PR inherits some CI issues from the previous main HEAD.

MarekPieta:
Rebased

PR #1109, opened 2021-08-19
boot/nuttx: Enable crypto backend according to configuration from NuttX

gustavonihei:
This PR intends to enable the NuttX port to select between Mbed TLS and TinyCrypt as the cryptographic backends.

The selection is achieved by means of NuttX `make menuconfig` command.


gustavonihei:
@utzig This will enable MCUboot to be built on NuttX without Mbed TLS. This commit will also allow using Mbed TLS as the crypto backend when Mbed TLS support is eventually added to NuttX.

utzig (approved):

PR #1112, opened 2021-08-20
boot: zephyr: Fix misaligned #else for `irq_lock`

sigvartmh:
So in PR #1089 I misaligned the `#else` used for the `irq_lock` as it
should be aligned with `CONFIG_MCUBOOT_CLEANUP_ARM_CORE`. So that
`irq_lock`is called when we don't call `__disable_irq` from arm core
clean up code.


sigvartmh:
ref: https://github.com/mcu-tools/mcuboot/pull/1089#discussion_r692963442
@ioannisg @d3zd3z

nvlsianpu (approved):

nvlsianpu:
cc @ioannisg

d3zd3z (approved):

PR #1114, opened 2021-08-23
Fix unknown type name "size_t" error in "boot_record.h"

AGlass0fMilk:
This commit adds an include clause to "boot_record.h" to include the stddef.h header that contains the declaration of the "size_t" type.

Attempting to include "boot_record.h" in a build for Mbed-OS would throw an error complaining "size_t" was unknown, adding a note to inculde "stddef.h"


AGlass0fMilk:
Before anyone asks:

https://stackoverflow.com/questions/50893099/why-is-size-t-in-stddef-and-not-stdint

d3zd3z (approved):
LGTM

PR #1115, opened 2021-08-23
Mbed add shared data config

AGlass0fMilk:
This PR adds configuration parameters for Mbed-OS to use the shared data feature of mcuboot.

It also introduces an iterator-style utility class for Mbed-OS applications booted by mcuboot to iterate through the TLV-encoded values in the shared data region.

### DIFF
--- a/boot/bootutil/include/bootutil/boot_record.h
+++ b/boot/bootutil/include/bootutil/boot_record.h
@@ -18,6 +18,7 @@
 #define __BOOT_RECORD_H__
 
 #include <stdint.h>
+#include <stddef.h>
 #include "bootutil/image.h"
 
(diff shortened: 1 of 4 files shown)


AGlass0fMilk:
@d3zd3z @utzig Any thoughts on this? The only thing I wasn't certain about was including the `DataShare` class in this repo -- I figured the core mcuboot team wouldn't want this to be added to this repository :)

It doesn't make sense to me to separate the data share iteration utilities from this code however.

Do you think it would be better to make this class a C API that is provided by mcuboot for all operating systems?

AGlass0fMilk:
@d3zd3z Any update on this?

d3zd3z (approved):

Issue #983, opened 2021-04-26
Swap copy size optimization does not work if scratch area is larger than the image


boraozgen:
In the design document it is mentioned that the number of swapped sectors is optimized based on the image sizes in case the slot size is larger than the images:

> Note2: The bootloader tries to copy only used sectors (based on largest image installed on any of the slots), minimizing the amount of sectors copied and reducing the amount of time required for a swap operation.

This optimization does not work if the scratch area size is larger than the image size. `boot_copy_sz` function uses the scratch area size as the lower limit of its output. This output value is used without checking if the image size is smaller than the scratch area, resulting in a unnecessarily large copy operation.

As I was working on a patch, I realized that there is a larger logic error in the `swap_run` function. The last sector ID variables are incremented together, which means in case sector sizes of the primary and secondary slots are different, the slot with the larger sector size may get incremented to a point where the ID is more than the available sectors. Afterwards this ID is used to calculate the copy size in `boot_copy_sz`. Since the output is capped by the scratch area size, it does not matter but it is still a logic error.

My proposal would be to compare the output of `boot_copy_sz` with the largest of `primary_slot_size` and `secondary_slot_size`, and use the smallest number of this comparison. I could work on a PR if this issue is acknowledged.

hjuul:
I've got a problem that is likely to be related to the "logic error" you're mentioning.

- My firmware image is larger than the swap area size (swap is 256KiB).
- My internal flash has 256KiB sectors, and slot1 is in a QSPI flash with 4KiB sectors.

What I'm seeing is that MCUboot always swaps all sectors regardless of my image size. The swap is functional, but takes several minutes more than it should.

Slack discussion:
https://mcuboot.slack.com/archives/C6CB5BXGC/p1627994014013400

nvlsianpu:
> This optimization does not work if the scratch area size is larger than the image size. boot_copy_sz function uses the scratch area size as the lower limit of its output. This output value is used without checking if the image size is smaller than the scratch area, resulting in a unnecessarily large copy operation.

I beat that for-lop counting of the copy size will end execution if `last_sector_idx` decrements below zero. So ther is another limit which implies from image size to be copied.

PR #1131, opened 2021-09-07
Fix bug causing long swap times in certain cases

hjuul:
When the secondary slot is stored in a flash with smaller sector size than
the primary slot, MCUboot would swap more sectors than necessary.
This is an attempt to fix that.

This PR is related to #983.

More details in this Slack thread: https://mcuboot.slack.com/archives/C6CB5BXGC/p1627994014013400

### DIFF
--- a/boot/bootutil/src/swap_scratch.c
+++ b/boot/bootutil/src/swap_scratch.c
@@ -700,20 +700,20 @@ swap_run(struct boot_loader_state *state, struct boot_status *bs,
            primary_slot_size += boot_img_sector_size(state,
                                                      BOOT_PRIMARY_SLOT,
                                                      last_sector_idx);
+            last_sector_idx++;
         }
         if ((secondary_slot_size < copy_size) ||
             (secondary_slot_size < primary_slot_size)) {
            secondary_slot_size += boot_img_sector_size(state,
                                                        BOOT_SECONDARY_SLOT,
                                                        last_idx_secondary_slot);
+            last_idx_secondary_slot++;
         }
         if (primary_slot_size >= copy_size &&
                 secondary_slot_size >= copy_size &&
                 primary_slot_size == secondary_slot_size) {
             break;
         }
-        last_sector_idx++;
-        last_idx_secondary_slot++;
     }
 
     swap_idx = 0;


nvlsianpu (approved):
fix looks sane.
Thanks!

pennam:
@hjuul tests are complainig about your commit email address... maybe you shoud replace <helge@fastmail.com> with <helge.juul@nortekgroup.com>

d3zd3z:
This is failing ci with `../../boot/bootutil/src/swap_scratch.c:575: boot_swap_sectors: Assertion `rc == 0' failed.`

hjuul:
Ok, so I was encouraged to create this PR based on a naïve workaround I had for [this issue](https://github.com/mcu-tools/mcuboot/issues/983#issuecomment-898311096).
Now that it fails, I don't know the code base enough to make it work.
Are there any maintainers that are willing to take this PR further?

pennam:
since i'm interested in using this i'll will be happy to help to found out why is failing. Any suggestion from the maintainers is really appreciated. Is there a way to reproduce and debug the issues on the simulator?

utzig (commented):
I believe this is just plain wrong, the purpose of that loop is to match the copy sizes based on both sectors from primary and secondary, so sizes in multiples will work. The update proposed in this PR would try to write a 4K sector in internal flash into a 2K flash on SPI, for example, which obviously does not work. How many sectors are we talking about (in your case)? I believe it should be only enough to match the internal flash, so say you have 4K internal and 2K external, that would be a single sector, is this really that bad?

hjuul:
@utzig , in my case I have 256KiB sector size on internal flash (primary slot) and only 4KiB sector size on external flash (secondary slot). So yeah, it's really bad.

utzig:
> @utzig , in my case I have 256KiB sector size on internal flash (primary slot) and only 4KiB sector size on external flash (secondary slot). So yeah, it's really bad.

I think you will basically have to add special handling for the last sector, since the swap code would like to erase 256KiB on the last sector fitting up to the image size, but you only want to erase up to the size enough for fitting the image, which could be much smaller than that, and also try to synchronize this between both slots, erasing one size in the primary and another size for the secondary. I don't think it will be remotely close to this simplistic approach, but feel free to prove me wrong! ;-)

utzig:
> @utzig , in my case I have 256KiB sector size on internal flash (primary slot) and only 4KiB sector size on external flash (secondary slot). So yeah, it's really bad.

How long does it take to erase 4KiB on that flash?

utzig:
> @utzig , in my case I have 256KiB sector size on internal flash (primary slot) and only 4KiB sector size on external flash (secondary slot). So yeah, it's really bad.

Just read your Zephyr thread here: https://github.com/zephyrproject-rtos/zephyr/issues/38548. Your QSPI flash, MT25QU256, supports 4KB, 32KB, and 64KB sector erases, and the driver should choose automatically, for example, I implemented this for Mynewt, and I assume Zephyr's driver can also do it, otherwise it could be implemented in a few lines of code like this: https://github.com/apache/mynewt-core/blob/master/hw/mcu/nxp/kinetis/src/hal_qspi.c#L352.

PR #1132, opened 2021-09-08
boot: zephyr: Add multi-image DFU for Thingy:53

MarekPieta:
Change enables multi-image DFU in Thingy:53 configuration.


MarekPieta:
The required features are currently not supported in MCUboot repository. The configuration change may be introduced later.

PR #1137, opened 2021-09-13
zephyr: support the hook file by MCUBOOT_BOOTUTIL library

nvlsianpu:
This patch make possible MCUBOOT_BOOTUTIL to integrate the
hook file on their own. This is intended to support hook while
the library is just part of the application.
The hooks file for the MCUboot build is now included in the same manner.

### DIFF
--- a/boot/bootutil/zephyr/CMakeLists.txt
+++ b/boot/bootutil/zephyr/CMakeLists.txt
@@ -16,6 +16,30 @@ zephyr_library_named(mcuboot_util)
 zephyr_library_sources(
   ../src/bootutil_public.c
     )
+
+if(CONFIG_BOOT_IMAGE_ACCESS_HOOKS)
+  if(NOT CONFIG_BOOT_IMAGE_ACCESS_HOOKS_FILE STREQUAL "")
+    if(IS_ABSOLUTE ${CONFIG_BOOT_IMAGE_ACCESS_HOOKS_FILE})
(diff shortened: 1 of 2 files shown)


d3zd3z (approved):

d3zd3z:
Merging, as another change has disabled the failing Travis test.

boot/bootutil/zephyr/CMakeLists.txt
```diff
@@ -16,6 +16,30 @@ zephyr_library_named(mcuboot_util)
...
+
+  if(DEFINED HOOKS_FILE)
+      zephyr_library_sources(
+        ${HOOKS_FILE}
+      )
+  else()
+    message(STATUS "No hooks implementation file.")
+  endif()
+endif()
+
```
tejlmand, line 42: Why all this logic ?

A downstream user can simply do:
```
# <downstream-path>/CMakeLists.txt
target_sources(mcuboot_util PRIVATE <hook-file.c>)
```

No need for Kconfig or anything like it.

tejlmand (commented):
Little late to the game, but I don't understand this overly complex setup, when the same can be accomplished with one line i a downstream CMake file.

boot/bootutil/zephyr/CMakeLists.txt
nvlsianpu: ^^ You still might do like you mention in the downstream/local fork. Actually it will be much more suitable. Her I have been aiming to provide something to make this configurable, in case any user want to use mainline MCUboot without touching it source code (and CMake is not a configuration interface).

PR #1138, opened 2021-09-13
ci: Disable travis test until they can be fixed

d3zd3z:
The Travis tests build MCUboot against a specific version of TF-M, which
it no longer works with.  Newer versions of TF-M seem to not support the
mps2/an521 target that the tests don't build that way either.

Until these tests can be fixed, just disable all of them.  Future
possibilities include fixing the issues between TF-M and MCUboot, or
making these tests able to run within Zephyr.


utzig (approved):
LGTM

PR #1142, opened 2021-09-15
Fix bug: Potential call to `BlockDevice::size` when uninitialized

AGlass0fMilk:
This commit fixes a bug that can cause an uninitialized BlockDevice to be queried for its size. In some cases, this can return unexpected results (eg: 0).

The check for initializing the BlockDevice is now made before any operations are made on it.

@LDong-Arm @boraozgen 

### DIFF
--- a/boot/mbed/src/flash_map_backend.cpp
+++ b/boot/mbed/src/flash_map_backend.cpp
@@ -97,6 +97,9 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
     if (open_count[id] == 1) {
         MCUBOOT_LOG_DBG("initializing flash area %d...", id);
         result = bd->init();
+        if(result) {
+            return result;
+        }
     }
 
     fap->fa_size = (uint32_t) bd->size();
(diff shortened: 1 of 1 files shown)


boot/mbed/src/flash_map_backend.cpp
```diff
@@ -91,15 +91,17 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
...
-    fap->fa_size = (uint32_t) bd->size();
 
+    int result = 0;
     /* Only initialize if this isn't a nested call to open the flash area */
     if (open_count[id] == 1) {
         MCUBOOT_LOG_DBG("initializing flash area %d...", id);
-        return bd->init();
-    } else {
-        return 0;
+        result = bd->init();
```
LDong-Arm, line 99: Not critical, but shall we consider returning early here if init() fails?

AGlass0fMilk: I suppose that would make sense.

boot/mbed/src/flash_map_backend.cpp
```diff
@@ -91,15 +91,20 @@ int flash_area_open(uint8_t id, const struct flash_area** fapp) {
...
+        if(result) {
```
utzig, line 100: Not sure how lax the Mbed OS coding style is, but this could use a space to match other code: `if (`. Also please rebase on latest master.

d3zd3z:
Is there still an interest in getting this in? Other suggestion would be to put "mbed: " at the start of the commit text, to make it clear what this affects.

PR #1143, opened 2021-09-15
Mbed-OS: Prevent use of BlockDevice with unreliable erase value

AGlass0fMilk:
The default return value of `BlockDevice::get_erase_value` is -1. When implicitly cast from an `int` to a `uint8_t`, this gets truncated to `0xFF`, which is normally a reasonable erase value.

However, the intent of the Mbed-OS BlockDevice API is that an **integer** value of -1 is to be interpreted as an "unreliable" erase value. ie: the underlying BlockDevice does not have a known erase value.

If a user attempts to use such a BlockDevice (eg: SDBlockDevice at the moment, see ARMmbed/mbed-os#15070), they may encounter hard to debug errors.

Since MCUboot expects some fields to be the "erased value" under certain boot/swap states, a BlockDevice with an unreliable erase value is not compatible with MCUboot. It is recommended that the user wrap the incompatible BlockDevice in a [`FlashSimBlockDevice`](https://github.com/ARMmbed/mbed-os/blob/64a34193b126aed25497021b7617f9dadf7ae7a8/storage/blockdevice/include/blockdevice/FlashSimBlockDevice.h) to provide a known erase value.

This commit adds an assert that will (hopefully) catch such an error during development.

### DIFF
--- a/boot/mbed/src/flash_map_backend.cpp
+++ b/boot/mbed/src/flash_map_backend.cpp
@@ -198,7 +198,20 @@ uint8_t flash_area_align(const struct flash_area* fap) {
 
 uint8_t flash_area_erased_val(const struct flash_area* fap) {
     mbed::BlockDevice* bd = flash_map_bd[fap->fa_id];
-    return bd->get_erase_value();
+
+    int erased_val = bd->get_erase_value();
+
+    /**
+     * If the BlockDevice returns -1 for the get_erase_value (IMPORTANT: distinct from byte value 0xFF),
+     * this means the erase value is unreliable and cannot be used. MCUboot does not support
+     * such BlockDevices. Some internal checks rely on comparison to a known erase value.
+     *
+     * Wrap this BlockDevice in a FlashSimBlockDevice to provide a known erase value and ensure
+     * compatibility with MCUboot.
+     */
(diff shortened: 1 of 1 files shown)


AGlass0fMilk:
@LDong-Arm @boraozgen

AGlass0fMilk:
Updated the comments to recommend use of the [`FlashSimBlockDevice`](https://github.com/ARMmbed/mbed-os/blob/64a34193b126aed25497021b7617f9dadf7ae7a8/storage/blockdevice/include/blockdevice/FlashSimBlockDevice.h) to ensure a known erase value and enable compatibility with MCUboot.

See ARMmbed/mbed-os#15071

d3zd3z (changes requested):
The change looks fine to me. Do you mind wrapping the lines, though, in the commit text, so that they don't exceed about 75 columns.

d3zd3z:
The failed mynewt test should pass after re-pushing. The certificate issue has been resolved. Rebasing will bring in a change that ignores the certificate, though.

d3zd3z:
Is there still interest in getting this merged?

AGlass0fMilk:
Unfortunately I no longer contribute to Mbed-OS, so someone else will have to push this along.

PR #1145, opened 2021-09-16
ci: Update the TF-M version in fih test

Sherryzhang2:
Fix build time error in fih test script

Change-Id: I357ca9266629310deddf8431aa912f3fdbe9f34c


Sherryzhang2:
The CI failure is caused by
ERROR: cannot verify developer.arm.com's certificate, issued by ‘CN=GlobalSign RSA OV SSL CA 2018,O=GlobalSign nv-sa,C=BE’:

utzig:
> The CI failure is caused by
> ERROR: cannot verify developer.arm.com's certificate, issued by ‘CN=GlobalSign RSA OV SSL CA 2018,O=GlobalSign nv-sa,C=BE’:

#1147 should fix it, I hope.

utzig:
@Sherryzhang2 Please rebase.

Sherryzhang2:
> @Sherryzhang2 Please rebase.

Done

utzig:
@tamasban @Sherryzhang2 I had to add the following commit to fix the change in format on the YAML: https://github.com/mcu-tools/mcuboot/pull/1145/commits/36d640d4fd214b9afef9afdfecd8ef942a3b2c92. I have not tracked down where it started happening, so if you want to go ahead, I am fine with this working as is now!

tamasban (changes requested):
@utzig: Thanks for the fix.

@Sherryzhang2:
I have noticed in the test logs that the test is executed but in the logs it is never seen that the system is booted successfully (this is very unlikely):

`    System booted (undesired behaviour) (0)`

This check is done based on parsing log messages from TF-M. This message is looked for in the log:
  `Secure image initializing`

But for some reason this was not  found by the scripts, instead of this was the last line:
`Description: 'ECDSA signature test of attest token' (8) `

@matetothpal tries to help out. 

Would be better to parse for a string that is not project-specific, instead comes from MCUboot and mark the end of a successful boot.

matetothpal:
@Sherryzhang2 I had a look at this, and it seems that Tamas is right: The log message expected by the test environment is not present in the TF-M log. Fortunately the issue is quite easy to fix: see this commit: https://github.com/matetothpal/mcuboot/commit/bac9fb93cbe633c0e27b917bc0feec51af0cc4f3

Currently these situations, where the CI environment is unable to detect a booting TF-M, remain undetected, unless someone takes a close look at the logs. So I suggest to add an extra step to the test, when a TF-M run is done with intact payload, and the environment attempts to detect booted TF-M (and fail if not). I created a PoC for this here: https://github.com/matetothpal/mcuboot/commit/de4f5ed05d5140d6a28790337bb1f04dffca6966 Feel free to improve it (for example it probably would be a good idea to define the desired log from TF-M in a single place). 

@tamasban Using a printout from MCUBoot would probably solve the problem of being dependent on an external project. It actually would make it possible to get rid entirely from TF-M in the CI system, the payload content could be a signed chunk of 0 bytes. 
However I see value in having an actual working payload in the test, as it is more accurate in a sense that it can measure code executing after that last MCUBoot log printed.

tamasban:
@matetothpal: Thanks, to look into it! Still worries me that print-out changes in TF-M project can break FI test in MCUboot and it might remain unnoticed untilTF-M version is gets updated. Would be good to decouple them.

I see your points about having a real payload in the test. But this payload could be a fixed image (prints a single line and loop), which is stored in the repository. In this way, we can be sure that the end of the test is always recognised.

@utzig:
 I suggest merging this PR including the fix in [bac9fb9](https://github.com/matetothpal/mcuboot/commit/bac9fb93cbe633c0e27b917bc0feec51af0cc4f3), this could unblock the activity to move the script to GitHub workflow. Are you okay with this? 

@Sherryzhang2, 
Could you enhance the FI test in the following way in a distinct PR:
  - Create a minimalistic Secure payload, which just prints a single line and FI test greps for this.
  - Store this binary in the MCUboot repo and always load this to QEMU by the script.
  - Apply the detection of a successful boot before test execution. To notice if for some reason the static image cannot boot-up properly. E.g,: memory layout of an521 was modified in the TF-M repo and the static image memory address does not align.

utzig:
> @utzig:
> I suggest merging this PR including the fix in [bac9fb9](https://github.com/matetothpal/mcuboot/commit/bac9fb93cbe633c0e27b917bc0feec51af0cc4f3), this could unblock the activity to move the script to GitHub workflow. Are you okay with this?

Yes, from the discussion I gather it is not a "correct" solution, but as long as someone is working on fixing the real issue I am fine with this as is. Maybe open an issue on GH to leave it documented?

tamasban:
> Maybe open an issue on GH to leave it documented?

I have opened #1154.

Sherryzhang2:
@tamasban  Please see my comment on https://github.com/mcu-tools/mcuboot/issues/1154#issuecomment-924620460

tamasban (approved):

PR #1149, opened 2021-09-17
Eliminate duplicate C files in mcuboot-sys

d3zd3z:
Depending on configuration, the simulator's mcuboot-sys crate either pulls files from `ext/mbedtls` or from `ext/mbedtls-asn1`, which is a few files extracted from the full Mbed TLS tree. However, there are now some configurations where the simulator ends up bringing in both sets of these files. In the current tests, the linker ends up arbitrarily choosing one of these. This also causes the cargo tests to fail when run directly in the `sim/mcuboot-sys` directory, as they are linked directly instead of through a library, and the compiler complains of duplicates.

Fix this by not using the asn1 copy in the simulator.

In addition, add a dummy feature for `large-write` to the mcuboot-sys crate so that all of this can be tested using ptest and having `ci/sim_run.sh` run in the mcuboot-sys crate.

### DIFF
--- a/sim/mcuboot-sys/Cargo.toml
+++ b/sim/mcuboot-sys/Cargo.toml
@@ -80,6 +80,9 @@ direct-xip = []
 # Check (in software) against version downgrades.
 downgrade-prevention = []
 
+# Large write.  Not meaningful, but present here so that the
+# full-suite tests will work for this configuration.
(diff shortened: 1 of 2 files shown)


utzig (approved):
LGTM

PR #1153, opened 2021-09-20
Add CMake Support for Mbed port

lambda-shuttle:
Currently, MbedOS applications using this port require that they are built with the old mbed-cli. However, `mbed-tools` is the new, future command-line tool for MbedOS; `mbed-tools` is also known as Mbed CLI 2. This new tool makes use of "Ninja as a build system, and CMake to generate the build environment and manage the build process in a compiler-independent manner." More information can be found [here](https://os.mbed.com/docs/mbed-os/v6.12/build-tools/mbed-cli-2.html). 

A CMakeLists.txt was added to the port (`mcuboot/boot/mbed`). This pull request also includes changes to avoid including the private header (`bootutil_priv.h`) in bootutil's src directory - the required macro `MCUBOOT_SWAP_USING_SCRATCH` is now setup in mbed's `flash_map_backend.cpp`. 

@LDong-Arm


d3zd3z:
Is there a desire to get this merged?

If so, a few issues to address:
- Rebase the changes, this will be needed to fix the merge conflicts, but also to eliminate the extra merge commit in the history.
- There are multiple lines with stray whitespace at the end of the line.

PR #1155, opened 2021-09-22
zephyr: Check zephyr,uart-mcumgr as candidate for serial recovery

de-nordic:
The commit modifies selection of boot serial UART by first checking
the Zephyr chosen zephyr,uart-mcumgr and then reverting to the
zephyr,console, as a secondary candidate.
In case when both nodes are present and point to the same device,
error will be reported.

### DIFF
--- a/boot/zephyr/serial_adapter.c
+++ b/boot/zephyr/serial_adapter.c
@@ -22,8 +22,16 @@
 #include "bootutil/bootutil_log.h"
 #include <zephyr/usb/usb_device.h>
 
-#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE)
+#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE) && \
+    (!DT_HAS_CHOSEN(zephyr_uart_mcumgr) ||                              \
+     DT_SAME_NODE(DT_CHOSEN(zephyr_uart_mcumgr), DT_CHOSEN(zephyr_console)))
 #error Zephyr UART console must been disabled if serial_adapter module is used.
+#endif
+
+#if defined(CONFIG_BOOT_SERIAL_CDC_ACM) && \
+    defined(CONFIG_UART_CONSOLE) && !DT_HAS_CHOSEN(zephyr_uart_mcumgr)
+#error Zephyr UART console must been disabled if CDC ACM is enabled and MCUmgr \
+       has not been redirected to other UART with DTS chosen zephyr,uart-mcumgr.
 #endif
 
 BOOT_LOG_MODULE_REGISTER(serial_adapter);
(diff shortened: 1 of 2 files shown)


boot/zephyr/serial_adapter.c
```diff
@@ -192,7 +195,11 @@ static int
 boot_uart_fifo_init(void)
 {
 #ifdef CONFIG_BOOT_SERIAL_UART
+#if DT_HAS_CHOSEN(zephyr_uart_mcumgr)
+	uart_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_uart_mcumgr));
+#else
 	uart_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_console));
+#endif
 #elif CONFIG_BOOT_SERIAL_CDC_ACM
 	uart_dev = DEVICE_DT_GET_ONE(zephyr_cdc_acm_uart);
```
mlsvrts, line 204: Currently, it is not really feasible to use a composite cdc acm usb device, such that one interface is used for logging and the other is used for serial recovery. This is because [`DEVICE_DT_GET_ONE`](https://docs.zephyrproject.org/2.7.0/reference/drivers/index.html?highlight=device_dt_get_one#c.DEVICE_DT_GET_ONE) returns an arbitrary valid node, which can of course collide with the cdc acm uart being used as the logging backend. 

So, maybe it would be useful to also try to use `zephyr,uart-mcumgr` as the default cdc acm uart device selection, and fall back to grabbing any valid node if it is not present?

```suggestion
#if DT_HAS_CHOSEN(zephyr_uart_mcumgr)
        uart_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_uart_mcumgr));
#else
        uart_dev = DEVICE_DT_GET_ONE(zephyr_cdc_acm_uart);
#endif
```

d3zd3z (approved):

d3zd3z:
Unfortunately, this seems to have developed a minor conflict that need to be resolved.

nvlsianpu:
@de-nordic Can you refresh this PR - I think it make sense.

boot/zephyr/serial_adapter.c
nvlsianpu: +1 for binding virtual port using zephyr,uart-mcumgr chosen node as well

nordicjm (approved):

boot/zephyr/serial_adapter.c
de-nordic: I would rather have that as another PR, because I am not that familiar with CDC ACM devices and reluctant to bring change I do not completely understand.

nordicjm:
Needing to use this for a zephyr test in qemu, is it possible to fix the failing issue and merge?

d3zd3z:
Looks like we just need to fix the CI failure. Do we need to add mynewt support for this?

de-nordic:
> Looks like we just need to fix the CI failure. Do we need to add mynewt support for this?

Will rebase to fix CI. These are Zephyr only sources affected here. mynewt may have different mechanism for this.

de-nordic:
@mlsvrts Can you take a look?

mlsvrts:
I'm not able to easily pull and test this anymore, but these changes should support the composite CDC-ACM use-case, so it looks good to me, thanks!

PR #1157, opened 2021-09-24
Add FIH hardening tests to workflows

utzig:
Add new GH workflow to run the FIH hardening tests and disable Travis.


tamasban (approved):
Thanks for porting it to GH Workfow!

d3zd3z (approved):
looks good

Issue #1094, opened 2021-08-09
Consider not storing dev/test keys in repo


d3zd3z:
Currently, development keys are stored in the repo. As long as the instructions are followed, this should not result in these keys being used in production. However it is possible for this to happen. One way to fix this is to remove these keys, and instead, have all of the systems that used the development keys instead create a set if they are needed. Ideally, this would be created once in a developer's workspace, and the same keys would be reused. This needs to be coordinated across all of the systems that use MCUboot, including the simulator, Zephyr, MyNewt, and the other supported platforms.

bn-norbit:
#1088 can also address this issue for Zephyr. If the default keys are removed, the user is forced to generate an encryption key and set  CONFIG_BOOT_ENCRYPT_KEY_FILE accordingly. The encryption tests can generate a keys on the fly.

d3zd3z:
Thanks @bn-norbit, This does help this for Zephyr. A challenge with this issue is addressing it for every platform before the keys can be removed. But, maybe when we address it for a given platform, that platform needs to be changed so that it never uses the in-tree keys.

PR #1162, opened 2021-09-29
espressif: Enable image signature verification

almir-okato:
This PR adds MbedTLS and Tinycrypt security lib options for signature verification to Espressif's port.
Sample keys for each type of signature (RSA, EC256 and ED25519) were also added.


almir-okato:
@d3zd3z and @utzig could you please take a look?

boot/espressif/keys_sample/ec256-sample.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
d3zd3z, line 1: Any reason for this target to have it's own set of sample keys, instead of using the ones at the top of the tree?

almir-okato: There isn't any special reason, I may did something wrong and when I tested it didn't work. I tested it again and works fine.
Should I change to use the keys in the root directory?

d3zd3z: I think it would be better to use the root directory keys. We have an issue #1094 about not storing these keys, but generating them when needed. It'll be easier to track down all the situations that do this if they are all at least using the same set of keys.

almir-okato: Hi @d3zd3z, I've applied your suggestion and also removed those key samples. Please take a look.

almir-okato:
> @almir-okato Please rebase on latest `main`.

@utzig, done. Thanks!

PR #1166, opened 2021-10-07
sim: update ram load to use image size comparison

utzig:
When building images for testing they are aligned to a flash write size, which works fine for flash based comparisons, but since RAM is always reset to 0, if the flash is erased to 0xff, the final bytes of the alignment will differ even though they are not relevant (not part of the image itself).

This commit adds a real image size parameter to the ImageData, so it can be used by the RAM load to compare only the relevant bits of the image (ignore the padding), and also updates the RAM test routine to use the correct image size.

### DIFF
--- a/ci/fih-tests_install.sh
+++ b/ci/fih-tests_install.sh
@@ -18,7 +18,7 @@ set -e
 
 DOCKER_DIR=docker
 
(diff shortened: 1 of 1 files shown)


utzig:
@tamasban @d3zd3z Is someone looking at this tfm repo issue?

PR #1169, opened 2021-10-11
boot: zephyr: allow custom handler for indication LED

kvaellning:
Allows to provide a custom implementation for the indication LED.
This can be used enable more in-depth information about the boot state,
or use custom handlers for LEDs which are shared with the app firmware
(e.g. different blink states, PWM)

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -114,6 +114,17 @@ static inline bool boot_skip_serial_recovery()
 BOOT_LOG_MODULE_REGISTER(mcuboot);
 
 #ifdef CONFIG_MCUBOOT_INDICATION_LED
+
+#ifdef CONFIG_MCUBOOT_INDICATION_LED_CUSTOM
+/**
+ * Define these functions in your custom library
+ */
+extern void mcuboot_led_init(void);
+extern void mcuboot_led_active(void);
+extern void mcuboot_led_error(void);
+extern void mcuboot_led_finish(void);
+
+#else
 /*
  * Devicetree helper macro which gets the 'flags' cell from a 'gpios'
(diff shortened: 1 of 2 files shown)


d3zd3z (approved):

d3zd3z:
Do you mind rebasing this and resolving the conflict. It shouldn't be too difficult to resolve.

kvaellning:
@d3zd3z It's updated

boot/zephyr/main.c
```diff
@@ -541,12 +572,20 @@ void main(void)
...
+#ifdef CONFIG_MCUBOOT_INDICATION_LED
+        mcuboot_led_error();
+#endif
         FIH_PANIC;
     }
 
     BOOT_LOG_INF("Bootloader chainload address offset: 0x%x",
                  rsp.br_image_off);
 
+#ifdef CONFIG_MCUBOOT_INDICATION_LED
```
d3zd3z, line 584: Do these need ifdefs? Aren't they defined as empty when `CONFIG_MCUBOOT_INDICATION_LED` is not defined?

kvaellning: Interesting, this appears to have entered the code on the first rebase a while ago. Thought I actually catched and fixed all.

nvlsianpu (commented):
I think https://github.com/mcu-tools/mcuboot/pull/1332 provides API which is substitute of this PR content. Maybe even more generic.
@kvaellning  Shall we close this PR?

kvaellning:
@nvlsianpu Yes it can be closed.
But tbh, using both CONFIG_USE_INDICATION_LED and #1332 may first activate the LED, than you are overwriting the state again. This may not be intended. But if necessary, I would build upon the new API and see if there is a good way to get it run.

PR #1184, opened 2021-10-22
Enhance simulator to generate maximally sized images

d3zd3z:
This allows boundary condition testing within the simulator, to make sure we handle the cases when the image is a large as it can possibly be.

This enables maximal images in those configurations where we are able to accurately determine what that image size is. For overwrite-only+large-write, the calculation is incorrect (and results in a negative image size), and for swap-move, it doesn't account for the padding sector needed for the move. For those, we just leave the current image sizes.


utzig (approved):
LGTM

PR #1186, opened 2021-10-27
Add support for partial RAM_LOADING

RcColes:
Add new public API function to perform RAM_LOADING on a specific image. Designed to be used when only a subset of the images need to be loaded, or images need to be loaded in multiple stages.

In order to do this, change how `context_boot_go` handles state. It now doesn't clear the context when called. Add a new API function to clear this state so that external callers can still use `context_boot_go`. The semantics of `boot_go` are unchanged.

Also moves slot_usage into the bootloader state.


d3zd3z (approved):
Looks good to me.

Issue #1191, opened 2021-11-02
Possible size overflow for header block size in image encryption


kvaellning:
**Issue:**

I am using MCUboot with Zephyr, and set a header size of 0x1000. In `loader:boot_copy_region`, this leads to a chunk size of 0x400 (buffer size). If on start of the image (`loader.c:907`), this leads to a silent overflow in `blk_sz` on line 910.

Interesting code:
`blk_sz = chunk_sz - hdr->ih_hdr_size; // chunk_sz = 0x400; hdr->ih_hdr_size = 0x1000`.

**Suggested fix:**

1. Check for this overflow, return `BOOT_EBADARGS`
2. imgtool should check for meaningful values of the header size.

nvlsianpu:
@kvaellning Is so big header something that you needed? Current code support header up to 1024 B size.
If this is just a bug hunting then I would go with fixes you suggested.

kvaellning:
This was basically a misconfiguration on my side, and I was curious why my image was corrupted without seeing an error. For my application, I can live with smaller header sizes (which I now use). I‘ll provide a pull request with the suggested fixes by tomorrow.

d3zd3z:
Adding a check is a good first step. The code could be reworked to not require that the header fit into the first block, though.

The header size is determined by the number of IRQs on the SoC (at least for Arm). The number of IRQs in the vector table determines the required alignment of the image (hence the size of the header). 256 IRQs isn't that uncommon, and I could see needing more.

nvlsianpu:
^^ I'm looking at better solution.

PR #1192, opened 2021-11-03
boot: Avoid overflow if header size is too big and encryption is used

kvaellning:
Fixes issue #1191.

This tries to prevent misconfiguration of the header size in the imgtool,
and also adds code which prevents the overflow if the (not encrypted) header
is skipped.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -920,6 +920,7 @@ boot_copy_region(struct boot_loader_state *state,
     uint8_t image_index;
 #endif
 
+    /* buffer size limits maximum header size. This affects IMAGE_HEADER_SIZE_MAX in imgtool.py */
     TARGET_STATIC uint8_t buf[1024] __attribute__((aligned(4)));
 
 #if !defined(MCUBOOT_ENC_IMAGES)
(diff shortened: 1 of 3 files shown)


boot/bootutil/src/loader.c
```diff
@@ -966,6 +966,10 @@ boot_copy_region(struct boot_loader_state *state,
                 if (off + bytes_copied < hdr->ih_hdr_size) {
                     /* do not decrypt header */
                     blk_off = 0;
+                    /* check for overflow, return BOOT_EBADIMAGE */
```
d3zd3z, line 969: `/* Check for overflow. */` is sufficient. Stating what the return does is redundant and just increases the likelihood that the code gets changed in the future and the comment missed.

scripts/imgtool/image.py
```diff
@@ -40,6 +40,7 @@
 
 IMAGE_MAGIC = 0x96f3b83d
 IMAGE_HEADER_SIZE = 32
+IMAGE_HEADER_SIZE_MAX = 0x400 # limited by buffer size in loader.c:boot_copy_region
```
d3zd3z, line 43: It might also be helpful to put a comment in loader.c where the buffer is defined that it should match this declaration here.

d3zd3z (changes requested):

nvlsianpu:
@kvaellning I opened alternative to your patch, see #1199

d3zd3z (changes requested):
Updates look good. Can you please squash the commits, and repush (with -f)?

kvaellning:
@d3zd3z I am a little bit unsure here - as mentioned by @nvlsianpu, there is a better fix already provided. Anyway, I will just bundle it together, but if it's not requiered anymore then I would simply close the pull-request.

kvaellning:
Closed since fixed.

PR #1194, opened 2021-11-03
espressif: Allow use of imgtool other than the one from repository

gustavonihei:
This PR intends to make the build system prioritize the usage of `imgtool` installed on the system.

If `imgtool` is not found, then CMake will fallback to the current behavior of using the one from the repository.

The motivation for this change is due to NuttX build system relying on the `imgtool` installed via `pip`.


utzig (approved):
LGTM

nvlsianpu (approved):

PR #1196, opened 2021-11-03
espressif: Fix flash_area_[read/write] support for unaligned accesses

gustavonihei:
## Summary ##
This PR intends to refactor both the `flash_area_read` and `flash_area_write` functions for properly supporting unaligned read and write operations on SPI Flash.

This refactor also handles read and write operations whose length in bytes are not multiple of 4.

## Testing ##
Created a custom testing infra for validating the added changes:
https://godbolt.org/z/834rjjhMq

### DIFF
--- a/boot/espressif/port/esp_mcuboot.c
+++ b/boot/espressif/port/esp_mcuboot.c
@@ -16,7 +16,33 @@
 #include "bootloader_flash.h"
 #include "bootloader_flash_priv.h"
 
-#define ARRAY_SIZE(arr) sizeof(arr)/sizeof(arr[0])
+#ifndef ARRAY_SIZE
+#  define ARRAY_SIZE(arr)           (sizeof(arr) / sizeof((arr)[0]))
+#endif
+
+#ifndef MIN
+#  define MIN(a, b)                 (((a) < (b)) ? (a) : (b))
+#endif
+
+#ifndef ALIGN_UP
+#  define ALIGN_UP(num, align)      (((num) + ((align) - 1)) & ~((align) - 1))
+#endif
+
+#ifndef ALIGN_DOWN
+#  define ALIGN_DOWN(num, align)    ((num) & ~((align) - 1))
+#endif
+
+#ifndef ALIGN_OFFSET
+#  define ALIGN_OFFSET(num, align)  ((num) & ((align) - 1))
+#endif
+
+#ifndef IS_ALIGNED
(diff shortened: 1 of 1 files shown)


gustavonihei:
@utzig The CI failed during the checkout of some submodules. Could you please restart the jobs?

gustavonihei:
> @utzig The CI failed during the checkout of some submodules. Could you please restart the jobs?

Thanks for restarting it.
The CI actually passed, but I force pushed to make the static assertion cleaner:
```patch
--- a/boot/espressif/port/esp_mcuboot.c
+++ b/boot/espressif/port/esp_mcuboot.c
@@ -42,7 +42,7 @@
 
 #define FLASH_BUFFER_SIZE           256 /* SPI Flash block size */
 
-_Static_assert((FLASH_BUFFER_SIZE & 0x3) == 0, "Buffer size for SPI Flash operations must be 4-byte aligned.");
+_Static_assert(IS_ALIGNED(FLASH_BUFFER_SIZE, 4), "Buffer size for SPI Flash operations must be 4-byte aligned.");
 
 #define BOOTLOADER_START_ADDRESS 0x1000
 #define BOOTLOADER_SIZE CONFIG_ESP_BOOTLOADER_SIZE
```

gustavonihei:
I've implemented a simple testing infra for validating the added changes: https://godbolt.org/z/834rjjhMq

boot/espressif/port/esp_mcuboot.c
```diff
@@ -86,68 +112,148 @@ void flash_area_close(const struct flash_area *area)
...
         return -1;
     }
-    if (len < 4) {
-        memcpy(dst, read_ptr, len);
-    }
+
     return 0;
 }
 
+static bool aligned_flash_write(size_t dest_addr, const void *src, size_t size)
```
d3zd3z, line 185: This shouldn't be needed. In fact, it is important that it not be present. The simulator verifies that all writes are on alignment boundaries, and of an alignment length, so that this is needed indicates something, somewhere else is wrong.

d3zd3z (changes requested):

boot/espressif/port/esp_mcuboot.c
gustavonihei: Thanks for the feedback! I've updated the PR, now the write operation no longer handles any alignment.

d3zd3z (approved):

PR #1197, opened 2021-11-03
ci: Enable build of espressif port for supported chips

gustavonihei:
This PR intends to enable the CI build of the `espressif` port of MCUboot for the currently supported chips:
- ESP32
- ESP32-S2
- ESP32-C3


utzig (approved):
LGTM

PR #1206, opened 2021-11-12
loader: add checking of reset address

hakonfam:
In a multi image context it is possible for a user to upload an image
to the wrong secondary slot. As the same key is used for both images
MCUboot will overwrite image 0 with a variant of image 1.

If direct overwrite is enabled it is not trivial to recover from this.

To mitigate this issue we introduce a check of the reset address within
the vector table.

If the reset address in the new image is not contained within the
primary slot the image is deemed incorrect and is deleted from the
secondary slot

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -773,6 +773,41 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         goto out;
     }
 
+#if MCUBOOT_IMAGE_NUMBER > 1 && !defined(MCUBOOT_ENC_IMAGES) && defined(MCUBOOT_VERIFY_IMG_ADDRESS)
+    /* Verify that the image in the secondary slot has a reset address
+     * located in the primary slot. This is done to avoid users incorrectly
+     * overwriting an application written to the incorrect slot.
+     * This feature is only supported by ARM platforms.
+     */
+    if (area_id == FLASH_AREA_IMAGE_SECONDARY(BOOT_CURR_IMG(state))) {
+        const struct flash_area *pri_fa = BOOT_IMG_AREA(state, BOOT_PRIMARY_SLOT);
+        struct image_header *secondary_hdr = boot_img_hdr(state, slot);
+        uint32_t reset_value = 0;
+        uint32_t reset_addr = secondary_hdr->ih_hdr_size + sizeof(reset_value);
+
+        rc = flash_area_read(fap, reset_addr, &reset_value, sizeof(reset_value));
+        if (rc != 0) {
+            fih_rc = fih_int_encode(1);
+            goto out;
+        }
+
+        if (reset_value < pri_fa->fa_off || reset_value> (pri_fa->fa_off + pri_fa->fa_size)) {
+            BOOT_LOG_ERR("Reset address of image in secondary slot is not in the primary slot");
+            BOOT_LOG_ERR("Erasing image from secondary slot");
+
+            /* The vector table in the image located in the secondary
+             * slot does not target the primary slot. This might
+             * indicate that the image was loaded to the wrong slot.
+             *
+             * Erase the image and continue booting from the primary slot.
+             */
+            flash_area_erase(fap, 0, fap->fa_size);
+            fih_rc = fih_int_encode(1);
+            goto out;
+        }
+    }
+#endif
+
 out:
     flash_area_close(fap);
 
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -257,9 +257,14 @@ config BOOT_SWAP_SAVE_ENCTLV
 	  JTAG/SWD or primary slot in external flash).
 	  If unsure, leave at the default value.
 
+config BOOT_ENCRYPT_IMAGE
+	bool
+	help
+	  Hidden option used to check if any image encryption is enabled.
+
 config BOOT_ENCRYPT_RSA
 	bool "Support for encrypted upgrade images using RSA"
-	default n
+	select BOOT_ENCRYPT_IMAGE
 	help
 	  If y, images in the secondary slot can be encrypted and are decrypted
 	  on the fly when upgrading to the primary slot, as well as encrypted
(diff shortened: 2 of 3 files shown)


boot/bootutil/src/loader.c
```diff
@@ -773,6 +773,39 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         goto out;
     }
 
+#if BOOT_IMAGE_NUMBER > 1 && !defined(MCUBOOT_ENC_IMAGES) && defined(MCUBOOT_VERIFY_IMG_ADDRESS)
```
d3zd3z, line 776: The only thing I suggest is putting at least a comment in here specifying that this is Arm specific. The Kconfig catches that on Zephyr, but it could be confusing for someone on another platform to enable this.

This also catches situations, such as dual XIP, where the image linked at the wrong address gets loaded.

If this is a test build (with swap), the situation should actually recover after a reboot, and the images are reverted.

boot/bootutil/src/loader.c
```diff
@@ -773,6 +773,39 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
+    /* Verify that the image in the secondary slot has a reset address
+     * located in the primary slot. This is done to avoid users incorrectly
+     * overwriting an application written to the incorrect slot.
+     */
+    if (area_id == FLASH_AREA_IMAGE_SECONDARY(BOOT_CURR_IMG(state))) {
```
nvlsianpu, line 781: I would extract this patch to one function which will be no-operation if configuration is irrelevant, (it might be done like was done for hooks callbacks, using an appropriate macro). This will help with legibility of this quite complicated code.
sth like:
```C
static int ensure_startup_address_or_delete(...)
```

boot/bootutil/src/loader.c
```diff
@@ -773,6 +773,39 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
+        const struct flash_area *pri_fa = BOOT_IMG_AREA(state, BOOT_PRIMARY_SLOT);
+        struct image_header *secondary_hdr = boot_img_hdr(state, slot);
+        uint32_t reset_value = 0;
+        uint32_t reset_addr = secondary_hdr->ih_hdr_size + sizeof(reset_value);
+
+        rc = flash_area_read(fap, reset_addr, &reset_value, sizeof(reset_value));
+        if (rc != 0) {
+            return BOOT_EFLASH;
```
nvlsianpu, line 789: flash area should be closed before returning to the caller (by mcuboot internal policy).
Would by not i case if this bock will be converted to the function.

boot/bootutil/src/loader.c
hakonfam: > The only thing I suggest is putting at least a comment in here specifying that this is Arm specific. The Kconfig catches that on Zephyr, but it could be confusing for someone on another platform to enable this.

Yeah, I will add a comment.

> This also catches situations, such as dual XIP, where the image linked at the wrong address gets loaded.

I will see if I can add a note on this (at least in the kconfig help text)

> If this is a test build (with swap), the situation should actually recover after a reboot, and the images are reverted.

True. This is why I only `default y` if you don't perform swapping.

boot/zephyr/Kconfig
```diff
@@ -257,9 +257,14 @@ config BOOT_SWAP_SAVE_ENCTLV
 	  JTAG/SWD or primary slot in external flash).
 	  If unsure, leave at the default value.
 
+config BOOT_ENCRYPT_IMAGE
```
nvlsianpu, line 260: This property is used n the source code.

nvlsianpu (commented):
I think this is valuable patch will be working well for ARM cores.

Probably you are aware of existing TLV dedicated to store of the application startup address
(see IMAGE_F_ROM_FIXED). Right now this is used by direct-xip for check the image, but it can be extended to be used along with swap modes as well. Of course a check using this TLV type requires that image must be signed with this, which might not satisfy your requirement.

hakonfam:
> Probably you are aware of existing TLV dedicated to store of the application startup address (see IMAGE_F_ROM_FIXED). Right now this is used by direct-xip for check the image, but it can be extended to be used along with swap modes as well. Of course a check using this TLV type requires that image must be signed with this, which might not satisfy your requirement.

I'm aware of this TLV, but IMO this solution is more robust since it (as you say) does not require additional metadata which a user might not add. The code size for checking the TLV would be similar to this.

I guess in the future we could have support for both types of checks, and those who know that the TLV is in place can opt out of the vector table checking.

boot/zephyr/Kconfig
hakonfam: I don't use it in the source code, but I use it as a helper kconfig to ensure that `MCUBOOT_VERIFY_IMG_ADDRESS` cannot be enabled if any form of image encryption is in place.

boot/bootutil/src/loader.c
hakonfam: I added a `goto` which should ensure that the area is closed.

I will try to come up with a helper function to see if that will makes things more clear.

boot/zephyr/Kconfig
nvlsianpu: ok, missed that

boot/bootutil/src/loader.c
gustavonihei: This patch hurts the MCUboot hardware-agnostic architecture. Even if we eventually wrapped this in an `#ifdef ARM`, this opens the door to further shortcuts throughout the core library, which I believe would be greatly undesirable.

There should be a way for achieving the same result by properly creating a new generic interface in the abstraction layer.
All we need here is a method for identifying whether the image should run from slot 0 or 1, it should not be explicitly coupled to an architecture-specific detail.

nvlsianpu: ^^
Right. Then proper approach would be to define an interface which will be implemented by architecture dedicated source.

d3zd3z (approved):

nvlsianpu:
I think this PR was merged to fast (https://github.com/mcu-tools/mcuboot/pull/1206#discussion_r752282942)
No need to revert it as it doesn't brake anything, but would be good to define proper interface and move architecture depended check to an architecture specific file. I will register the issue for track this.

PR #1208, opened 2021-11-12
espressif: Use BOOT_LOG_* macros instead of the MCUBOOT_LOG_*

gustavonihei:
## Summary

This PR intends to make the logging infra on the Espressif port uniform to the bootutil library.

This PR also refines the include directives, by removing unused headers and making the usage of brackets and quotes a bit more coherent.

## Impact
This PR brings no functional changes, so it shouldn't bring any impact.

## Testing

CI build pass.
Also tested in every board supported by the Espressif port:
- ESP32
- ESP32-S2
- ESP32-C3

### DIFF
--- a/boot/espressif/main.c
+++ b/boot/espressif/main.c
@@ -5,18 +5,19 @@
  */
 
 #include <bootutil/bootutil.h>
+#include <bootutil/bootutil_log.h>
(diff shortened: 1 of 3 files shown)


utzig (approved):
LGTM

PR #1210, opened 2021-11-14
nuttx: solve referenciation error in memset call

tito97sp:
**Solve:**
For GCC 4.9: 
error: argument to 'sizeof' in 'memset' call is the same expression as the destination; did you mean to dereference it? [-Werror=sizeof-pointer-memaccess]

### DIFF
--- a/boot/nuttx/src/flash_map_backend/flash_map_backend.c
+++ b/boot/nuttx/src/flash_map_backend/flash_map_backend.c
@@ -552,7 +552,7 @@ int flash_area_erase(const struct flash_area *fa, uint32_t off, uint32_t len)
       return ERROR;
     }
 
-  memset(buffer, erase_val, sizeof(buffer));
+  memset(buffer, erase_val, sector_size);
 
   i = 0;
 


boot/nuttx/src/flash_map_backend/flash_map_backend.c
```diff
@@ -552,7 +552,7 @@ int flash_area_erase(const struct flash_area *fa, uint32_t off, uint32_t len)
       return ERROR;
     }
 
-  memset(buffer, erase_val, sizeof(buffer));
+  memset(buffer, erase_val, sizeof(*buffer));
```
utzig, line 555: Not OK using `sizeof(*buffer)`, should probably be `memset(buffer, erase_val, sector_size);`

tito97sp: **Right**. With `memset(buffer, erase_val, sector_size)` compiles correctly.

gustavonihei: Indeed, the solution proposed by @utzig should be the correct one.

gustavonihei (approved):
LGTM

utzig:
@tito97sp Please squash both commits into a single commit; the original commit message was OK.

utzig:
> @tito97sp Please squash both commits into a single commit; the original commit message was OK.

Ah, and also sign off the commit.

PR #1213, opened 2021-11-16
nuttx: solve referenciation error in memset call

tito97sp:
Signed-off-by: Andres Sanchez <tito97_sp@hotmail.com>

Solve:
For GCC 4.9:
error: argument to 'sizeof' in 'memset' call is the same expression as the destination; did you mean to dereference it? [-Werror=sizeof-pointer-memaccess]


gustavonihei (approved):
LGTM

pkarashchenko (approved):

utzig (approved):

PR #1215, opened 2021-11-19
boot_serial: cleanup max output size calculation

nvlsianpu:
Calculation of BOOT_SERIAL_OUT_MAX was based on MCUBOOT_IMAGE_NUMBER
while in other places BOOT_IMAGE_NUMBER is taken int account as
images number. Let's align the calculation to use same literal value
as others.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -63,12 +63,8 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
-#ifndef BOOT_IMAGE_NUMBER
-#define BOOT_IMAGE_NUMBER MCUBOOT_IMAGE_NUMBER
-#endif
-
 #define BOOT_SERIAL_INPUT_MAX   512
-#define BOOT_SERIAL_OUT_MAX     (128 * MCUBOOT_IMAGE_NUMBER)
+#define BOOT_SERIAL_OUT_MAX     (128 * BOOT_IMAGE_NUMBER)
 
 #ifdef __ZEPHYR__
 /* base64 lib encodes data to null-terminated string */
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -130,14 +130,6 @@ struct boot_status {
 
 extern const uint32_t boot_img_magic[4];
 
-#ifdef MCUBOOT_IMAGE_NUMBER
-#define BOOT_IMAGE_NUMBER          MCUBOOT_IMAGE_NUMBER
-#else
-#define BOOT_IMAGE_NUMBER          1
-#endif
-
-_Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
-
 #if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
 #define ARE_SLOTS_EQUIVALENT()    0
 #else
--- a/boot/bootutil/include/bootutil/bootutil.h
+++ b/boot/bootutil/include/bootutil/bootutil.h
@@ -36,6 +36,14 @@
 extern "C" {
 #endif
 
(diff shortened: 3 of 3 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -68,7 +68,7 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #endif
 
 #define BOOT_SERIAL_INPUT_MAX   512
-#define BOOT_SERIAL_OUT_MAX     (128 * MCUBOOT_IMAGE_NUMBER)
+#define BOOT_SERIAL_OUT_MAX     (128 * BOOT_IMAGE_NUMBER)
```
gustavonihei, line 71: `bootutil_priv.h` also defines BOOT_IMAGE_NUMBER` and performs some validation on it.
Please, consider reusing that definition and removing lines 66-68.

gustavonihei (approved):
LGTM

boot/boot_serial/src/boot_serial.c
nvlsianpu: right

gustavonihei: It seems `mcuboot/boot/bootutil/src/` is not part of the include paths for the build of `boot_serial` library, which makes sense since it is not part of `bootutil`.

Does it make sense to move those defines to the public interface of `booutil`?

utzig: As an alternative you can add `@mcuboot/boot/bootutil` as a dependency here: https://github.com/mcu-tools/mcuboot/blob/main/boot/boot_serial/pkg.yml#L28

de-nordic (approved):
In commit message `taken int` -> `taken into` typo, and `images number` to `image's number`, if possessive, or `number of images`, if it means "many images"

Otherwise looks OK

boot/boot_serial/src/boot_serial.c
utzig: Oh, I got this wrong, `bootutil_priv.h` is not in an `include/` directory, hmmm, not sure but it does not seem too much like a "private" file after it is shared...

utzig: Taking a look now my suggestion would be to move:

```
#ifdef MCUBOOT_IMAGE_NUMBER
#define BOOT_IMAGE_NUMBER          MCUBOOT_IMAGE_NUMBER
#else
#define BOOT_IMAGE_NUMBER          1
#endif

_Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
```

to `boot/bootutil/include/bootutil/bootutil.h` so the definitions are not "private" anymore.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -130,12 +130,6 @@ struct boot_status {
 
 extern const uint32_t boot_img_magic[4];
 
-#ifdef MCUBOOT_IMAGE_NUMBER
-#define BOOT_IMAGE_NUMBER          MCUBOOT_IMAGE_NUMBER
-#else
-#define BOOT_IMAGE_NUMBER          1
-#endif
-
 _Static_assert(BOOT_IMAGE_NUMBER > 0, "Invalid value for BOOT_IMAGE_NUMBER");
```
gustavonihei, line 133: This assertion should be moved as well to the public header.

nvlsianpu:
@utzig ?

utzig:
> @utzig ?

?

nvlsianpu:
Can you approve/merge?

gustavonihei:
> Can you approve/merge?

Please, take a look at this comment https://github.com/mcu-tools/mcuboot/pull/1215#discussion_r760088128, it should be addressed before merging.

utzig:
> Can you approve/merge?

I was wondering why you don't want to address: https://github.com/mcu-tools/mcuboot/pull/1215#discussion_r760088128

nvlsianpu:
I just forgot.

utzig (approved):
LGTM

PR #1217, opened 2021-11-22
bootutil: Increase minimum flash write

gustavonihei:
This PR intends to resume development on #949 by addressing raised comments and performing further clean ups.

## Pending tasks:
- [x] ~`imgtool`: Customize **magic** according to the `--align` field.~
- [x] ~Update documentation regarding new **magic** value for images whose maximum write alignment are different than 8 bytes.~


gustavonihei:
Comments from #949 are addressed in [this commit](ef44bcb). I'll squash it later.

talih0:
@gustavonihei In zephyr, flash_area_align() returns uint8_t.
This will limit alignment to 128 as I understand.

talih0:
Also in
https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/bootutil_public.c#L344

align is uint8_t

gustavonihei:
> flash_area_align

Well noted. **Apache NuttX* and **Mbed OS** ports also define the return type of `flash_area_align` as `uint8_t`, which is the correct API according to the Porting documentation: https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md#flash-map

@d3zd3z @utzig Should we update the API and all ports to `uint16_t` in accordance to what is used by the simulator?
https://github.com/mcu-tools/mcuboot/blob/main/sim/mcuboot-sys/csupport/storage/flash_map.h#L126

gustavonihei:
@talih0 I've changed the return type of `flash_area_align` to `uint32_t` and fixed the implementation of every port maintained inside the MCUboot repo. Other ports (Zephyr included), may require a PR to their specific repositories once this one gets merged.

talih0:
> @talih0 I've changed the return type of `flash_area_align` to `uint32_t` and fixed the implementation of every port maintained inside the MCUboot repo. Other ports (Zephyr included), may require a PR to their specific repositories once this one gets merged.

thanks, I'll add the changes to zephyr, and test on xmc4700 where the write size is 256.

talih0:
I did a quick test with zephyr on xmc4700, and the test - revert pattern is working fine for me. I'll look over the code in the next few days again.

We should probably update boot/zephyr/include/mcuboot_config/mcuboot_config.h to set MCUBOOT_BOOT_MAX_ALIGN via CONFIG_xx option.

utzig:
> I did a quick test with zephyr on xmc4700, and the test - revert pattern is working fine for me. I'll look over the code in the next few days again.
> 
> We should probably update boot/zephyr/include/mcuboot_config/mcuboot_config.h to set MCUBOOT_BOOT_MAX_ALIGN via CONFIG_xx option.

You might want to take a look at #841 which is a proposal for an alternative algorithm that was designed to support large sectors, "real large". Using the standard status algorithm is very wasteful for any flash using larger than 32 bytes of write size. IMO the main reason for supporting 16/32 bytes here is to allow for AES-128/256 HW encrypted flashes.

sim/src/image.rs
```diff
@@ -1796,12 +1796,13 @@ fn verify_trailer(flash: &SimMultiFlash, slot: &SlotInfo,
 
     failed |= match magic {
         Some(v) => {
-            if v == 1 && &copy[24..] != MAGIC {
+            let magic_off = (c::boot_max_align() * 3) + (c::boot_magic_sz() - MAGIC.len());
+            if v == 1 && &copy[magic_off..] != MAGIC {
```
utzig, line 1800: @gustavonihei Those changes to remove hard-coded values look good, please submit them in a separate PR, and I mean all changes to `image.rs` but the adding of `16, 32` alignments to testing.

gustavonihei: Sure, created PR https://github.com/mcu-tools/mcuboot/pull/1226.

sim/src/image.rs
```diff
@@ -1528,7 +1528,11 @@ fn install_image(flash: &mut SimMultiFlash, slot: &SlotInfo, len: ImageSize,
                 // c::boot_status_sz(dev.align() as u32) as usize
                 16 + 4 * dev.align()
             } else {
-                c::boot_trailer_sz(dev.align() as u32) as usize
+                fn align_up(num: u32, align: u32) -> u32 {
```
d3zd3z, line 1531: Probably should move this function to the top level. There are a handful of utility functions later in this file.
Also, probably best to have an assertion to make sure that 'align' is a power of two `assert!(align.is_power_of_two())`.

d3zd3z (changes requested):
BTW, it is the case that for swap-move, the end of the TLV/data must not be in the same sector as the trailer.

sim/src/image.rs
gustavonihei: Moved it into a dedicated file `utils.rs`.

d3zd3z:
The only thing I'm seeing missing from earlier discussion was to change the magic value when the maximum alignment was changed. Ideally, the layout would be the same when the max align was set to 8, and then different when different values were chosen. I think we could encode the max alignment into the magic number so that the different configurations would have different magic. Otherwise, the code will confusingly misinterpret the trailer if compiled with a different value than the generated image.

gustavonihei:
> The only thing I'm seeing missing from earlier discussion was to change the magic value when the maximum alignment was changed. Ideally, the layout would be the same when the max align was set to 8, and then different when different values were chosen. I think we could encode the max alignment into the magic number so that the different configurations would have different magic. Otherwise, the code will confusingly misinterpret the trailer if compiled with a different value than the generated image.

PTAL at https://github.com/gustavonihei/mcuboot/pull/1217/commits/97d847d6619c3a7bded516005848e89083c12cc2 which implements the new magic value, which embeds the maximum write alignment information.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -47,6 +47,14 @@
 extern "C" {
 #endif
 
+#ifndef ALIGN_UP
+#define ALIGN_UP(num, align)    (((num) + ((align) - 1)) & ~((align) - 1))
```
nvlsianpu, line 51: Iooks like we need something even more common header for both xxx_public.h files

nvlsianpu (commented):
I'm start looking into this PR.

boot/bootutil/include/bootutil/bootutil_public.h
gustavonihei: Indeed. `bootutil_public.c` and `bootutil_misc.c` also duplicate several identical helpers functions.
I believe this refactor may deserve a dedicated PR due to the amount of changes, as it would make the current one even more extensive.

nvlsianpu: ^^ sounds good to me. Let's not complicate this PR more.

boot/bootutil/src/swap_scratch.c
```diff
@@ -514,6 +515,17 @@ boot_swap_sectors(int idx, uint32_t sz, struct boot_loader_state *state,
      * controls if special handling is needed (swapping last sector).
      */
     last_sector = boot_img_num_sectors(state, BOOT_PRIMARY_SLOT) - 1;
+    sector_sz = boot_img_sector_size(state, BOOT_PRIMARY_SLOT, 0);
```
nvlsianpu, line 518: sectors might be of different size each other,

nvlsianpu (changes requested):
I mainly concern on the commit of `bootutil: Support trailers larger than sector size for swap scratch`
so far we allows irregular slot memory layout. Until we quits from this requirement, code should support it.

boot/bootutil/src/bootutil_public.c
```diff
@@ -51,22 +51,34 @@
...
 #else
 BOOT_LOG_MODULE_REGISTER(mcuboot_util);
 #endif
 
-const uint32_t boot_img_magic[] = {
-    0xf395c277,
-    0x7fefd260,
-    0x0f505235,
-    0x8079b62c,
+#if BOOT_MAX_ALIGN == 8
```
nvlsianpu, line 62: So far MCUboot defined same magic for all allowed BOOT_MAX_ALIGN values.
What is the reason behind changing that?

boot/bootutil/src/swap_scratch.c
gustavonihei: But does MCUboot support different sectors sizes within a single slot?

boot/bootutil/src/bootutil_public.c
```diff
@@ -316,14 +315,32 @@ int
...
 {
     uint32_t off;
+    uint32_t pad_off;
     int rc;
+    uint8_t magic[BOOT_MAGIC_ALIGN_SIZE];
+    uint8_t erased_val;
 
     off = boot_magic_off(fap);
 
+    /* image_trailer structure was modified with additional padding such that
```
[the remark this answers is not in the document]
gustavonihei: It is important that the MAGIC field stays at the very last position of the trailer, this way the bootloader is able to parse it regardless of the maximum write alignment of the application image.

boot/bootutil/src/swap_scratch.c
utzig: Yes it does: https://github.com/mcu-tools/mcuboot/blob/main/sim/src/image.rs#L342

boot/bootutil/src/bootutil_priv.h
```diff
@@ -140,7 +140,26 @@ struct boot_status {
  *      (`MCUBOOT_ENC_IMAGES`).
  */
 
-extern const uint32_t boot_img_magic[4];
+union boot_img_magic_t
+{
+    struct {
```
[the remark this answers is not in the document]
gustavonihei: It wouldn't hurt to add the packed attribute here, but it is not needed in this case since the alignment requirement for the inner struct is **2** imposed by the `uint16_t` member. So the compiler won't add padding to this region.
Besides, the static assertion on line **162** will ensure we won't get caught in any misbehavior of the compiler.

boot/bootutil/src/swap_scratch.c
gustavonihei: Correct me if I am wrong, but `FlashId::Image0` starts af offset **0x20000** which already falls into the sector with 128KB size.

boot/bootutil/src/bootutil_public.c
gustavonihei: The `BOOT_MAX_ALIGN` was not configurable, so it always assumed maximum write alignment of 8 bytes.
This PR now allows it to be configurable with a range of values.

Selecting any value other than 8 will change the layout of the image trailer, so a new magic value is required in order for the bootloader to be able to identify the new layout.

boot/bootutil/src/swap_scratch.c
utzig: > Correct me if I am wrong, but `FlashId::Image0` starts af offset **0x20000** which already falls into the sector with 128KB size.

Yeah, you're right.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -177,10 +181,9 @@ static inline uint32_t
 boot_enc_key_off(const struct flash_area *fap, uint8_t slot)
 {
 #if MCUBOOT_SWAP_SAVE_ENCTLV
-    return boot_swap_size_off(fap) - ((slot + 1) *
-            ((((BOOT_ENC_TLV_SIZE - 1) / BOOT_MAX_ALIGN) + 1) * BOOT_MAX_ALIGN));
+    return boot_swap_size_off(fap) - ((slot + 1) * BOOT_ENC_TLV_ALIGN_SIZE);
 #else
-    return boot_swap_size_off(fap) - ((slot + 1) * BOOT_ENC_KEY_SIZE);
+    return boot_swap_size_off(fap) - ((slot + 1) * BOOT_ENC_KEY_ALIGN_SIZE);
 #endif
```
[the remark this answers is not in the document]
gustavonihei: Sorry, could you please be more specific on this?
`BOOT_ENC_TLV_ALIGN_SIZE` and `BOOT_ENC_KEY_ALIGN_SIZE` may still have different sizes.

gustavonihei:
> I mainly concern on the commit of `bootutil: Support trailers larger than sector size for swap scratch` so far we allows irregular slot memory layout. Until we quits from this requirement, code should support it.

Could you please point where this supported is implemented? To the extent of my research, MCUboot seems to assume that sectors within a given image slot are equally sized. Different image slots, though, may be allocated into memory regions of distinct sectors sizes.

nvlsianpu:
See swap_scratch.c file:

https://github.com/mcu-tools/mcuboot/blob/3ee626f36ed9c2ad0d259fda164aa2d880f2c436/boot/bootutil/src/swap_scratch.c#L171
```C
/*
 * Slots are compatible when all sectors that store up to to size of the image
 * round up to sector size, in both slot's are able to fit in the scratch

 * area, and have sizes that are a multiple of each other (powers of two
 * presumably!).
 */
int
boot_slots_compatible(struct boot_loader_state *state)
```

boot/bootutil/src/swap_scratch.c
utzig: Just tried here and it works as expected:

```diff
diff --git a/sim/src/image.rs b/sim/src/image.rs
index b45791e3..6e771871 100644
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -341,7 +341,9 @@ impl ImagesBuilder {
                 // STM style flash.  Large sectors, with a large scratch area.
                 let dev = SimFlash::new(vec![16 * 1024, 16 * 1024, 16 * 1024, 16 * 1024,
                                         64 * 1024,
-                                        128 * 1024, 128 * 1024, 128 * 1024],
+                                        32 * 1024, 32 * 1024, 64 * 1024, // primary
+                                        32 * 1024, 32 * 1024, 64 * 1024, // secondary
+                                        128 * 1024],                     // scratch
                                         align as usize, erased_val);
                 let dev_id = 0;
                 let mut areadesc = AreaDesc::new();
```

utzig: Now thinking better about it, it was good point you made, I think it's better to have this memory model above in the tests, so I will submit this patch once I am finished with "professional" work for the day.

gustavonihei:
> See swap_scratch.c file:
> 
> https://github.com/mcu-tools/mcuboot/blob/3ee626f36ed9c2ad0d259fda164aa2d880f2c436/boot/bootutil/src/swap_scratch.c#L171
> 
> ```c
> /*
>  * Slots are compatible when all sectors that store up to to size of the image
>  * round up to sector size, in both slot's are able to fit in the scratch
> 
>  * area, and have sizes that are a multiple of each other (powers of two
>  * presumably!).
>  */
> int
> boot_slots_compatible(struct boot_loader_state *state)
> ```

@nvlsianpu PTAL, take a look at [gustavonihei/mcuboot@`15f0aa9` (#1217)](https://github.com/gustavonihei/mcuboot/pull/1217/commits/15f0aa9176330c4159727c6d98a49126135a68d8), now the calculation of the trailer size should consider distinct sized sectors within the image slot.

boot/bootutil/src/bootutil_misc.c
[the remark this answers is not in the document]
alexdolzhenkovbdl: Sorry, it is my fail, I didn't realized that they are no the same.

scripts/imgtool/main.py
```diff
@@ -288,8 +288,10 @@ def convert(self, value, param, ctx):
               help='Specify the value of security counter. Use the `auto` '
               'keyword to automatically generate it from the image version.')
 @click.option('-v', '--version', callback=validate_version,  required=True)
-@click.option('--align', type=click.Choice(['1', '2', '4', '8']),
+@click.option('--align', type=click.Choice(['1', '2', '4', '8', '16', '32']),
               required=True)
+@click.option('--max-align', type=click.Choice(['1', '2', '4', '8', '16', '32']),
```
utzig, line 293: Eventually once other tools are updated this extra parameter could be removed and the new MAGIC with max-align could be used all the time...

utzig (commented):
LGTM, just left a few doubts.

sim/src/image.rs
```diff
@@ -2020,12 +2040,12 @@ pub fn show_sizes() {
...
+#[cfg(not(feature = "max-align-32"))]
 fn test_alignments() -> &'static [usize] {
     &[1, 2, 4, 8]
 }
 
-#[cfg(feature = "large-write")]
+#[cfg(feature = "max-align-32")]
 fn test_alignments() -> &'static [usize] {
-    &[1, 2, 4, 8, 128, 512]
+    &[32]
```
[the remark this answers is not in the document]
gustavonihei: The current MAGIC pattern dedicates 2 bytes to the maximum write alignment value. In order to keep the MAGIC constant data in the simulator, it assumes a fixed maximum write alignment of 32 bytes when the test feature `max-align-32` is enabled.

sim/src/image.rs
```diff
@@ -135,6 +136,16 @@ impl ImagesBuilder {
...
         }
 
+        if !unsupported_caps.contains(&Caps::SwapUsingScratch) {
+            let (_, scratch_len, _) = match areadesc.find(FlashId::ImageScratch) {
+                Some(info) => info,
+                None => return Err("insufficient partitions".to_string()),
+            };
+            if scratch_len < c::boot_trailer_sz(align as u32) as usize {
+                return Err("insufficient size for scratch partition".to_string());
+            }
```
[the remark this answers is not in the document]
gustavonihei: I've just attempted another run on `enc-aes256-rsa` without this filtering and the execution is halted by this assertion:
https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/bootutil_misc.c#L147

The assertion ensures that the scratch area is larger than the trailer.
If I remove the assertion, the `enc-aes256-rsa` test still finishes successfully.
Should I remove the assertion?

jamike:
I tested sucessfully the patch set in a TF-M Build on stm/b_u585i_iot02a (max-align 16)

sim/src/image.rs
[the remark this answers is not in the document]
utzig: The assertion is correct for the trailer in the primary slot, probably have to update that function with a calculation of trailer size for scratch and use `fap` to determine which to use. You can leave as is, or fix it if you feel like it.

sim/src/image.rs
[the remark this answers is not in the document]
gustavonihei: In order to keep `MAGIC` a `const` array of bytes, I believe we would need to create a new `max-align-16` feature to configure the first two bytes accordingly. This scales poorly and will add several new jobs to the CI. Unfortunately  `features` cannot carry over integer values, it is just a flag.
To the best of my currently limited Rust knowledge we would need to refactor the way we manipulate `MAGIC` to make it a global mutable resource. Having it as a `static mut` does not seem as an elegant solution either.
How does that sound to you?

sim/src/image.rs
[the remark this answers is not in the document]
gustavonihei: I've tried the following patch to adjust the status offset to the size of the scratch area, but it didn't work and the `perm_with_fails` and `revert_with_fails` tests failed:
```patch
diff --git a/boot/bootutil/src/bootutil_misc.c b/boot/bootutil/src/bootutil_misc.c
index 67f0fe3..3bac13d 100644
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -139,6 +139,11 @@ boot_status_off(const struct flash_area *fap)
 
     off_from_end = boot_trailer_sz(elem_sz);
 
+    if (flash_area_get_id(fap) == FLASH_AREA_IMAGE_SCRATCH) {
+        /* Image trailer may be larger than the scratch area */
+        off_from_end %= flash_area_get_size(fap);
+    }
+
     assert(off_from_end <= flash_area_get_size(fap));
     return flash_area_get_size(fap) - off_from_end;
 }
```

sim/src/image.rs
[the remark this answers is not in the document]
utzig: That is a bit too much and not very scalable, I will pursue a different path but in the interim I run all tests also using 16 bytes and it worked fine.

sim/src/image.rs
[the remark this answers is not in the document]
utzig: I pushed two commits, one which removes that section of code above, that you can squash into the main commit it changes, and another with a fix to the boot_status_off function, which should correct the behavior and allow it to pass CI. Otherwise this looks pretty good and I will approve afterwards (assuming it passes CI!).

boot/bootutil/src/bootutil_misc.c
```diff
@@ -139,6 +147,12 @@ boot_status_off(const struct flash_area *fap)
 
     off_from_end = boot_trailer_sz(elem_sz);
 
+#if MCUBOOT_SWAP_USING_SCRATCH
```
nvlsianpu, line 150: See my dislike for this how this get fixed, rather another call boot_trailer_xyz_sz() should be provided instead

utzig: As far as I could see the only place that needs to know the size of the trailer inside scratch was here, so another call would just be overkill that should not be used anywhere else. Do you have any plans on using it?

nvlsianpu: I don't have plan to reuse new call. It is just about legibility of the code - how the trailer size is calculated.
Contribution to the size from the status is much smaller in case of scratch are, and this is what is subtracted here.

Don't treat my comments as a no-go.

utzig: It's a fair assessment, I will do it as suggested later today once I'm done with work for the day. I was also curious to understand a bit better why this did not impact the swap with scratch operation the way it was. Sorry that this was a bit rushed.

scripts/imgtool/main.py
nvlsianpu: I wonder whether not better to only extend --align allowed values range.
New magic will then apply only for 16 and 32, so answer is no,
i think we need to explain what is the difference between --align and - max-align to the user

And: Should `--align value` >= `--max_align value` ?

boot/bootutil/src/bootutil_misc.c
utzig: #1246 should address your concerns. I took a look at the effect of the bug and the swap operation was indeed writing at the wrong offset inside scratch, but the right amount of data, which is possibly still caught by the calculations that are done before starting a swap, but anyway it's still weird we didn't catch that before...

utzig:
I just tried testing the max-align of 8, 4 and 2 in the simulator. For 8 it does not work because it assumes the old magic, which is fine. For 4 it works. For <=2 it does not work because it fails in assertions inside bootutil. It should be easy to fix but I would suggest in the scope of this PR to restrict the make max-align to only accept 16 and 32 (as well as maintaining the previous format with 8). Later on smaller sizes could be added if anyone feels it's a valuable addition.

utzig:
@gustavonihei Please rebase and drop the previous status offset commit.

utzig (approved):
LGTM, thanks!

nvlsianpu (approved):

d3zd3z (approved):

PR #1221, opened 2021-11-25
mynewt: Add call to custom pre-boot function

kasjer:
Custom boot_preboot() function can be used to start
DFU in similar way as is already done in zephyr.

Proposed function name can be discussed if someone thinks it's inappropriate.

### DIFF
--- a/boot/mynewt/syscfg.yml
+++ b/boot/mynewt/syscfg.yml
@@ -28,6 +28,9 @@ syscfg.defs:
     BOOT_CUSTOM_START:
         description: 'Override hal_system_start with a custom start routine'
         value: 0
+    BOOT_PREBOOT:
+        description: 'Call boot_preboot() function before booting application'
+        value:
 
 syscfg.vals:
     SYSINIT_CONSTRAIN_INIT: 0
--- a/boot/mynewt/src/main.c
+++ b/boot/mynewt/src/main.c
@@ -51,6 +51,10 @@
 void boot_custom_start(uintptr_t flash_base, struct boot_rsp *rsp);
 #endif
 
+#if MYNEWT_VAL(BOOT_PREBOOT)
+void boot_preboot(void);
+#endif
+
 #if defined(MCUBOOT_SERIAL)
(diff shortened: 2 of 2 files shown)


boot/mynewt/syscfg.yml
```diff
@@ -28,6 +28,9 @@ syscfg.defs:
     BOOT_CUSTOM_START:
         description: 'Override hal_system_start with a custom start routine'
         value: 0
+    BOOT_PREBOOT:
+        description: 'Call boot_preboot() function before booting application'
+        value:
```
utzig, line 33: What is the default supposed to be?

kasjer: For now by default function does not exist.
If package dfu from mynewt core will be included. It will provide this function, then function will check selected pin to see if DFU should be enabled and if so it will start TinyUSB/DFU functionality.
This will be very helpful for NRF53840 dongle where there is no debugger on board and flashing is requires special cable. Once this is enabled inserting dongle with BUTTON_1 pressed will start DFU and dfu-util or dfu-tool can be used to flash image into SLOT0.

But maybe someone will have other purpose for this so that is why I did not propose function name that implies DFU.

utzig: No, I just meant, do you want to make the value `0` or `1`? Make it explicit.

kasjer: Right now I have dfu package that sets this to 1. This package is not BSP/APP/TARGET so it would not be able to set this value if it was already set.
If you want I can set this here and remove it there and add restriction that will force user to specify it in target.
Or I could add specific syscfg value that will include specific package from core and it will also set BOOT_PREBOOT on mcuboot side.

utzig: > If you want I can set this here and remove it there and add restriction that will force user to specify it in target.
Or I could add specific syscfg value that will include specific package from core and it will also set BOOT_PREBOOT on mcuboot side.

Too much, thanks.

utzig (approved):
LGTM

PR #1222, opened 2021-11-25
nuttx: fix compilation warnings

pkarashchenko:
Signed-off-by: Petro Karashchenko <petro.karashchenko@gmail.com>

Solve:
For GCC 10.3:
```
warning: data definition has no type or storage class
     #define BOOT_LOG_MODULE_DECLARE(module)  MCUBOOT_LOG_MODULE_DECLARE(module)
warning: format '%d' expects argument of type 'int', but argument 5 has type 'int32_t' {aka 'long int'} [-Wformat=]
     syslog(LOG_INFO, "%s: " format "\n", __FUNCTION__, ##__VA_ARGS__)
warning: format '%zu' expects argument of type 'size_t', but argument 4 has type 'long unsigned int' [-Wformat=]
     syslog(LOG_DEBUG, "%s: " format "\n", __FUNCTION__, ##__VA_ARGS__)
```

### DIFF
--- a/boot/nuttx/src/flash_map_backend/flash_map_backend.c
+++ b/boot/nuttx/src/flash_map_backend/flash_map_backend.c
@@ -64,9 +64,9 @@ struct flash_device_s
 
   struct partition_info_s partinfo;
 
-  int     fd;          /* File descriptor for an open flash area */
-  int32_t refs;        /* Reference counter */
-  uint8_t erase_state; /* Byte value of the flash erased state */
+  int      fd;          /* File descriptor for an open flash area */
+  uint32_t refs;        /* Reference counter */
+  uint8_t  erase_state; /* Byte value of the flash erased state */
 };
 
 /****************************************************************************
@@ -232,11 +232,11 @@ static struct flash_device_s *lookup_flash_device_by_offset(uint32_t offset)
  *   Retrieve flash area from the flash map for a given ID.
  *
  * Input Parameters:
- *   fa_id - ID of the flash area.
+ *   id - ID of the flash area.
  *
  * Output Parameters:
- *   fa    - Pointer which will contain the reference to flash_area.
- *           If ID is unknown, it will be NULL on output.
+ *   fa - Pointer which will contain the reference to flash_area.
+ *        If ID is unknown, it will be NULL on output.
  *
  * Returned Value:
  *   Zero on success, or negative value in case of error.
@@ -254,7 +254,7 @@ int flash_area_open(uint8_t id, const struct flash_area **fa)
   dev = lookup_flash_device_by_id(id);
   if (dev == NULL)
     {
-      BOOT_LOG_ERR("Undefined flash area: %d", id);
+      BOOT_LOG_ERR("Undefined flash area: %" PRIu8, id);
 
       return ERROR;
     }
@@ -263,7 +263,7 @@ int flash_area_open(uint8_t id, const struct flash_area **fa)
 
   if (dev->refs++ > 0)
     {
-      BOOT_LOG_INF("Flash area ID %d already open, count: %d (+)",
+      BOOT_LOG_INF("Flash area ID %" PRIu8 " already open, count: %" PRIu32 " (+)",
                    id, dev->refs);
 
       return OK;
@@ -319,7 +319,7 @@ int flash_area_open(uint8_t id, const struct flash_area **fa)
 
   dev->fd = fd;
 
-  BOOT_LOG_INF("Flash area %d open, count: %d (+)", id, dev->refs);
+  BOOT_LOG_INF("Flash area %" PRIu8 " open, count: %" PRIu32 " (+)", id, dev->refs);
 
   return OK;
 
@@ -363,7 +363,7 @@ void flash_area_close(const struct flash_area *fa)
       return;
     }
 
-  BOOT_LOG_INF("Close request for flash area %" PRIu8 ", count: %d (-)",
+  BOOT_LOG_INF("Close request for flash area %" PRIu8 ", count: %" PRIu32 " (-)",
                fa->fa_id, dev->refs);
 
   if (--dev->refs == 0)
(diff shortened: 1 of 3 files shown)


boot/nuttx/src/flash_map_backend/flash_map_backend.c
```diff
@@ -65,7 +65,7 @@ struct flash_device_s
   struct partition_info_s partinfo;
 
   int     fd;          /* File descriptor for an open flash area */
-  int32_t refs;        /* Reference counter */
+  int     refs;        /* Reference counter */
```
gustavonihei, line 68: I believe the proper fix lies on the log format rather than changing the data type here.
Please, consider maintaining the fixed-width type and fixing the formats where applicable.

pkarashchenko: I really do not think that it matters. The MCUboot is "Secure boot for 32-bit Microcontrollers", so `int` should be capable of capable of carrying same values as `int32_t`. I chose a single change head instead of dozen of changes in log calls because fixes seems equivalent to me.

gustavonihei: > I really do not think that it matters. The MCUboot is "Secure boot for 32-bit Microcontrollers", so int should be capable of capable of carrying same values as int32_t.

This is a matter of consistency. In general it is more recommended to rely on fixed-width data types, which makes the overall code clearer and less prone to portability issues.

> I chose a single change head instead of dozen of changes in log calls because fixes seems equivalent to me.

I didn't find those dozen of log calls, there only 3 references that you'd need to change:
https://github.com/mcu-tools/mcuboot/blob/main/boot/nuttx/src/flash_map_backend/flash_map_backend.c#L266-L267
https://github.com/mcu-tools/mcuboot/blob/main/boot/nuttx/src/flash_map_backend/flash_map_backend.c#L322
https://github.com/mcu-tools/mcuboot/blob/main/boot/nuttx/src/flash_map_backend/flash_map_backend.c#L366-L367

pkarashchenko: > This is a matter of consistency. In general it is more recommended to rely on fixed-width data types, which makes the overall code clearer and less prone to portability issues.
I understand your point, but taking into account the definition of MCUboot itself as a "32-bit Microcontrollers" variant I do not see portability issue as a strong argument.
> I didn't find those dozen of log calls, there only 3 references that you'd need to change:
Yeah, probably the warning log was too long so I didn't count all places, but rather saw that a single line change will fix all.

Anyway I will proceed and apply 3 changes instead of one.

gustavonihei: > I understand your point, but taking into account the definition of MCUboot itself as a "32-bit Microcontrollers" variant I do not see portability issue as a strong argument.

Just because MCUboot only supports 32-bit microcontrollers today does not mean it will always remain like that. That's why I said "**less prone to** portability issues".

> Anyway I will proceed and apply 3 changes instead of one.

Amount of changed lines should not matter as long as the solution is the most appropriate.

pkarashchenko: Done. Please check code consistency improvements

gustavonihei (changes requested):
The declaration of all `flash_area_*` functions follow the Flash Map API documentation on https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md#flash-map

Changing just on the NuttX port will make it inconsistent.

pkarashchenko:
> The declaration of all `flash_area_*` functions follow the Flash Map API documentation on https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md#flash-map
> 
> Changing just on the NuttX port will make it inconsistent.

What do you suggest? Revert back changes or fix docs + other ports?

gustavonihei:
> > The declaration of all `flash_area_*` functions follow the Flash Map API documentation on [`main`/docs/PORTING.md#flash-map](https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md?rgh-link-date=2021-11-25T13%3A11%3A59Z#flash-map)
> > Changing just on the NuttX port will make it inconsistent.
> 
> What do you suggest? Revert back changes or fix docs + other ports?

I'd suggest to restrict the scope of this PR to the initial changes required for fixing the compilation warnings.
If you really want to improve the Flash Map API, I'd suggest to create another PR and modifying all ports at once.

pkarashchenko:
> > > The declaration of all `flash_area_*` functions follow the Flash Map API documentation on [`main`/docs/PORTING.md#flash-map](https://github.com/mcu-tools/mcuboot/blob/main/docs/PORTING.md?rgh-link-date=2021-11-25T13%3A11%3A59Z#flash-map)
> > > Changing just on the NuttX port will make it inconsistent.
> > 
> > 
> > What do you suggest? Revert back changes or fix docs + other ports?
> 
> I'd suggest to restrict the scope of this PR to the initial changes required for fixing the compilation warnings. If you really want to improve the Flash Map API, I'd suggest to create another PR and modifying all ports at once.

Done

boot/nuttx/src/flash_map_backend/flash_map_backend.c
```diff
@@ -801,5 +801,12 @@ int flash_area_id_from_image_offset(uint32_t offset)
 
   BOOT_LOG_INF("offset:%" PRIu32, offset);
 
-  return dev->fa_cfg->fa_id;
+  if (dev != NULL)
+    {
+	  return dev->fa_cfg->fa_id;
```
gustavonihei, line 806:
```suggestion
        return dev->fa_cfg->fa_id;
```
Whitespaces instead of tabs.

pkarashchenko: Done

gustavonihei (approved):
LGTM

utzig (approved):
LGTM

PR #1226, opened 2021-11-30
sim: Calculate offsets from boot_max_align instead of hardcoding them

gustavonihei:
This PR intends to prepare the simulator to execute images with a max align value other than 8 bytes.

**Note**: change extracted from patch series proposed on PR https://github.com/mcu-tools/mcuboot/pull/1217

### DIFF
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -1796,12 +1796,13 @@ fn verify_trailer(flash: &SimMultiFlash, slot: &SlotInfo,
 
     failed |= match magic {
         Some(v) => {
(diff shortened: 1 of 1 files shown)


utzig:
There was another bit of code which also was not related to changing the alignments that could also be here...

utzig:
> There was another bit of code which also was not related to changing the alignments that could also be here...

`mark_permanent_upgrade`

gustavonihei:
> There was another bit of code which also was not related to changing the alignments that could also be here...

True, I've updated the PR. Thanks for the heads up!

utzig (approved):
LGTM, thanks!

PR #1238, opened 2021-12-10
sim: define stm32f4 with non-linear sectors slots

utzig:
Change the flash layout of the stm32f4 simulated device to include sectors of different sizes inside both slots; both slots have been updated to still share the same layout. This layout changes allow for testing and validation that a swap with scratch is possible even if not all sectors inside a slot are the same size, as long as they all fit in scratch.

### DIFF
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -339,9 +339,13 @@ impl ImagesBuilder {
         match device {
             DeviceName::Stm32f4 => {
                 // STM style flash.  Large sectors, with a large scratch area.
-                let dev = SimFlash::new(vec![16 * 1024, 16 * 1024, 16 * 1024, 16 * 1024,
-                                        64 * 1024,
-                                        128 * 1024, 128 * 1024, 128 * 1024],
+                // The flash layout as described is not present in any real STM32F4 device, but it
+                // serves to exercise support for sectors of varying sizes inside a single slot,
+                // as long as they are compatible in both slots and all fit in the scratch.
+                let dev = SimFlash::new(vec![16 * 1024, 16 * 1024, 16 * 1024, 16 * 1024, 64 * 1024,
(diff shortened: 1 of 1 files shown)


gustavonihei (approved):

sim/src/image.rs
```diff
@@ -339,9 +339,13 @@ impl ImagesBuilder {
...
                 // STM style flash.  Large sectors, with a large scratch area.
-                let dev = SimFlash::new(vec![16 * 1024, 16 * 1024, 16 * 1024, 16 * 1024,
-                                        64 * 1024,
-                                        128 * 1024, 128 * 1024, 128 * 1024],
+                // The flash layout as described is not present in any real STM32F4 device, but it
+                // serves to exercise support for sectors of varying sizes inside a single slot,
+                // as long as they are compatible in both slots and all fit in the scratch.
+                let dev = SimFlash::new(vec![16 * 1024, 16 * 1024, 16 * 1024, 16 * 1024, 64 * 1024,
+                                        32 * 1024, 32 * 1024, 64 * 1024,
+                                        32 * 1024, 32 * 1024, 64 * 1024,
```
gustavonihei, line 347: How about 
```suggestion
                                        32 * 1024, 64 * 1024, 32 * 1024,
                                        64 * 1024, 16 * 1024, 16 * 1024, 32 * 1024,
```
How about going a bit further and making primary and secondary slots have different number of sectors. Then we would achieve greater coverage of the `boot_slots_compatible` function.

utzig: This is a bit unrealistic, and not supported; either both slots have to share the same layout with whatever sector sizes they have, or a number of sectors of one has to fit into the other in a linear, from start to finish fashion, which is broken at the beginning by `32 + 64` does not fit into `64`. The main problem solved, apart from the sectors match, is one of having two flashes with different sector sizes.

Issue #1211, opened 2021-11-15
Software based downgrade prevention with single application slot


mateo-ligne-netatmo:
I am working on a project using MCUBoot (v1.7.99-ncs3) where my microcontroller will receive updates through UART from another chip using mcumgr library. 

I would like to use both following configuration: 
    • CONFIG_SINGLE_APPLICATION_SLOT
    • CONFIG_MCUBOOT_DOWNGRADE_PREVENTION

Currently the downgrade prevention option compare the version of the two stored firmware, but in my case I would like to compare the version of the new firmware with stored one before overwriting it.
   
Has this use case been considered? Is it possible to read the firmware header before loading it? 

Thanks for your answers, 
Matéo

d3zd3z:
When MCUboot is configured for single application slot, the downgrade prevention code wouldn't ever be run, as it is part of the upgrade process, which MCUboot isn't really involved in. This check you are requesting would need to be added to the serial recovery code.

However, the request is reasonable, and I don't see any reason this couldn't be implemented there.

mateo-ligne-netatmo:
Hi, 
I'm implementing the comparison of software version between stored firmware and firmware to be send over recovery UART from MCUmgr. Currently, I am able to allow / disallow flash write considering version but the issue I am facing is that in case of a blocked update, MCUmgr keep trying to send the firmware unless I fake the reception by incrementing offset. 

Do you have any return code or other solution that should stop an ongoing MCUmgr update ? 

Thanks for your answers,
Matéo

d3zd3z:
It may have just been an unexpected situation from the perspective of MCUmgr. Perhaps we do need to define an additional error code to indicate that the image has been rejected (I could see images being rejected for other reasons, say it is able to tell it is for the wrong device). MCUmgr (and probably newtmgr) might have to be modified to support this, though.

mateo-ligne-netatmo:
In linked PR #1239, I added this new error code that I called MGMT_ERR_REJECTED_UPD, it might be renamed and defined in several other files as I only used it in a Serial recovery configuration. Otherwise I made a development on newtmgr (https://github.com/apache/mynewt-newtmgr/pull/187) stopping the firmware sending on the reception of this newly defined code.

PR #1239, opened 2021-12-13
Sw version based downgrade prevention

mateo-ligne-netatmo:
Hello everyone, 

This PR follow a [previous issue](https://github.com/mcu-tools/mcuboot/issues/1211)  where I express my need of having a software version comparison before loading a new firmware through **Recovery uart** in a **Single application slot** configuration. 

The firmware update will be monitored by **MCUmgr**, and in order to stop the update from it, I defined a new error type that will stop update process. This MCUmgr feature is the [following open PR](https://github.com/apache/mynewt-newtmgr/pull/187)

I also added a commit allowing to use 2 UART peripheral at the same time one for Logs and the other for MCUmgr communication. 
 
It's my first MCUBoot upload so don't hesitate to make feedbacks or suggestions, I will take them into account !

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -487,7 +487,7 @@ out:
     flash_area_close(fap);
 
 #ifdef MCUBOOT_ENC_IMAGES
-    if (curr_off == img_size) {
+    if (curr_off == img_size && rc == MGMT_ERR_OK) {
         /* Last sector received, now start a decryption on the image if it is encrypted*/
         rc = boot_handle_enc_fw();
     }
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -58,7 +58,7 @@
 #include "bootutil_priv.h"
 #endif
 
-#ifdef MCUBOOT_ENC_IMAGES
+#if defined MCUBOOT_ENC_IMAGES || defined MCUBOOT_DOWNGRADE_PREVENTION
 #include "single_loader.h"
 #endif
 
@@ -102,6 +102,9 @@ static cbor_state_backups_t dummy_backups;
 static cbor_state_t cbor_state = {
     .backups = &dummy_backups
 };
+#ifdef MCUBOOT_DOWNGRADE_PREVENTION
+static bool updatable;
+#endif /* MCUBOOT_DOWNGRADE_PREVENTION */
 
 /**
  * Function that processes MGMT_GROUP_ID_PERUSER mcumgr group and may be
@@ -266,6 +269,10 @@ bs_upload(char *buf, int len)
     static off_t off_last = -1;
     struct flash_sector sector;
 #endif
+#ifdef MCUBOOT_DOWNGRADE_PREVENTION
+    struct image_header* inc_hdr = NULL;
+    struct image_header st_hdr;
+#endif /* MCUBOOT_DOWNGRADE_PREVENTION */
 
     img_num = 0;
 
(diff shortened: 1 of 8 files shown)


utzig (changes requested):
Please sign-off your commits.

boot/boot_serial/src/boot_serial.c
```diff
@@ -317,6 +328,36 @@ bs_upload(char *buf, int len)
         goto out;
     }
 
+#ifdef MCUBOOT_DOWNGRADE_PREVENTION
+    /* Comparison should happen once, on the reception of header */
+    if (off == 0)
+    {
```
d3zd3z, line 334: Brace should be at the end of the line with the if.

boot/boot_serial/src/boot_serial.c
```diff
@@ -317,6 +328,36 @@ bs_upload(char *buf, int len)
...
+        inc_hdr = (struct image_header*)img_data;
+        rc = BOOT_HOOK_CALL(boot_read_image_header_hook,
+                            BOOT_HOOK_REGULAR, 0, 0, &st_hdr);
+        if (rc == BOOT_HOOK_REGULAR)
+        {
```
d3zd3z, line 339: likewise

boot/boot_serial/src/boot_serial.c
```diff
@@ -317,6 +328,36 @@ bs_upload(char *buf, int len)
...
+            flash_area_read(fap, 0, &st_hdr, sizeof(st_hdr));
+        }
+        updatable = true;
+        /* Check if a firmware is already stored, if yes, compare versions */
+        if (st_hdr.ih_img_size < flash_area_get_size(fap))
+        {
```
d3zd3z, line 345: here as well

boot/boot_serial/src/boot_serial.c
```diff
@@ -317,6 +328,36 @@ bs_upload(char *buf, int len)
...
+            rc = boot_version_cmp(&st_hdr.ih_ver,
+                                  &inc_hdr->ih_ver);
+            if (rc > 0) {
+                BOOT_LOG_INF("Update aborted due to strictly lower firmware version");
+                updatable = false;
+            }
+        }
+    }
+    if (!updatable)
+    {
```
d3zd3z, line 355: and here

d3zd3z (changes requested):
Can you also write a bit more in the commit text. The summary lines are good, but the body (above the signed off by line) should have a bit of an explanation as to why a change is being made. (probably not needed for the typo fix).

boot/boot_serial/src/boot_serial.c
```diff
@@ -255,6 +262,10 @@ bs_upload(char *buf, int len)
     static off_t off_last = -1;
     struct flash_sector sector;
 #endif
+#ifdef MCUBOOT_DOWNGRADE_PREVENTION
+    struct image_header* inc_hdr = NULL;
+    struct image_header st_hdr;
+#endif /*MCUBOOT_DOWNGRADE_PREVENTION*/
```
d3zd3z, line 268: Also, sorry to nit, but please put spaces between the comment characters and the symbol. I notice there are a few others that have snuck in without the spaces, but the original ones do have the spaces.

boot/zephyr/serial_adapter.c
```diff
@@ -22,8 +22,8 @@
 #include "bootutil/bootutil_log.h"
 #include <usb/usb_device.h>
 
-#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE)
-#error Zephyr UART console must been disabled if serial_adapter module is used.
+#if DT_SAME_NODE(DT_CHOSEN(zephyr_uart_mcumgr), DT_CHOSEN(zephyr_console))
```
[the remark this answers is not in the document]
mateo-ligne-netatmo: Yes, but what about the case where we want to have UART console and DFU over UART on two separate UART device ? As far as I understood this check disallow this possibility, whereas my changes enable it if console UART and MCUmgr UART are 2 differents devices.

d3zd3z (approved):
For future reference, please format the text of the commit text to 75-ish columns. If there are no other changes, I can fix this up before this goes in, however.

mateo-ligne-netatmo:
Hello, I have rebased my PR with current MCUboot main especially following https://github.com/mcu-tools/mcuboot/pull/1255. It turns out that on a failed firmware update (for too low SW version for instance), stored binary won't be overwritten but it will be decrypted a second time which will break the applicative. To counter that, I've just had a return code check before decryption.

d3zd3z:
I believe @utzig concerns have been addressed.

boot/zephyr/serial_adapter.c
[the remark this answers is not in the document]
nvlsianpu: @de-nordic Can you take a look on this and the next line?

PR #1246, opened 2021-12-16
bootutil: fix scratch status offset calculation

utzig:
Update `boot_status_off` to take into consideration that scratch only has to store status for one swap of sectors, thus requiring much less flash space than what is required in the primary slot.

This commits does a bit of refactoring on the functions that return parts of trailer size to make the code clearer.

This should fix the bug found while testing #1217


gustavonihei (approved):
LGTM

nvlsianpu (approved):

PR #1248, opened 2021-12-16
sim: Homogenize max-align-32 feature check

d3zd3z:
This feature check was done statically instead of dynamically in the
build.rs file.  Change this to match the rest.  Perhaps we might want to
change all of them to be done at compile time, but that should be done
with all of the features.


utzig (approved):
LGTM

gustavonihei (approved):
LGTM!

nvlsianpu (approved):

PR #1249, opened 2021-12-17
espressif: Fix support for signing algorithms

gustavonihei:
This PR intends to complete the support for signing algorithms for the current supported chips. Some algorithms were disabled due to build issues.

This PR also adds more CI jobs for building with support for Secure Boot V2 and the following signing algorithms:
- RSA-2048
- RSA-3072
- ECDSA-P256
- ED25519

## Note
**ESP32 ECO 3** still does not support **ED25519** when Secure Boot V2 is enabled due to the bootloader image exceeding the absolute binary size limit of 48KB imposed by Secure Boot V2, as documented here:\
https://docs.espressif.com/projects/esp-idf/en/v4.3/esp32/security/secure-boot-v2.html#bootloader-size

The chip's ROM bootloader will emit the following error before even loading the MCUboot bootloader image:
```
rst:0x10 (RTCWDT_RTC_RESET),boot:0x12 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:2, clock div:2
secure boot v2 enabled
Bootloader is too large for 48KB staging area.
ets_secure_boot_v2.c 74
```
An issue will be created for later addressing this limitation.


utzig (approved):

gustavonihei:
@utzig I've changed the way the new jobs are created. Instead of creating a new matrix item for each signature algorithm, they are now done sequentially in a single job to avoid the time-consuming environment preparation.

PR #1252, opened 2021-12-18
boot/nuttx: Switch from printf to syslog

nandojve:
Use syslog in favor of printf function for print messages. The syslog ensure that all characters are corrected flushed. Syslog is already used when log is enabled and this only switch to same log interface.

Signed-off-by: Gerson Fernando Budke <nandojve@gmail.com>

### DIFF
--- a/boot/nuttx/main.c
+++ b/boot/nuttx/main.c
@@ -23,7 +23,7 @@
 
 #include <nuttx/config.h>
 
-#include <stdio.h>
+#include <syslog.h>
 
 #include <sys/boardctl.h>
 
@@ -52,7 +52,7 @@ static void do_boot(struct boot_rsp *rsp)
   ret = flash_area_open(area_id, &flash_area);
   assert(ret == OK);
 
-  printf("Booting from %s...\n", flash_area->fa_mtd_path);
+  syslog(LOG_INFO, "Booting from %s...\n", flash_area->fa_mtd_path);
 
   info.path        = flash_area->fa_mtd_path;
   info.header_size = rsp->br_hdr->ih_hdr_size;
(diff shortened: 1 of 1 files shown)


boot/nuttx/main.c
```diff
@@ -22,6 +22,7 @@
  ****************************************************************************/
 
 #include <nuttx/config.h>
+#include <syslog.h>
 
 #include <stdio.h>
```
gustavonihei, line 28:
```suggestion

#include <stdio.h>
#include <syslog.h>
```
nit: `syslog.h` header should part of the same group of `stdio.h`.

boot/nuttx/main.c
```diff
@@ -79,17 +80,19 @@ int main(int argc, FAR char *argv[])
...
   if (fih_not_eq(fih_rc, FIH_SUCCESS))
     {
-      fprintf(stderr, "Unable to find bootable image\n");
+      syslog(LOG_ERR, "Unable to find bootable image\n");
       FIH_PANIC;
     }
 
   do_boot(&rsp);
 
+  syslog(LOG_ERR, "Something went wrong, should not get here!");
```
gustavonihei, line 95: This log line adds little to no information to the Bootloader user.
I suggest to either rephrase it by adding a proper error message with context information, or simply remove it.

gustavonihei: If the intent is to catch a missing (or bad) implementation of `BOARDIOC_BOOT_IMAGE` boardctl call, then catch this error on the `do_boot` function.

gustavonihei (approved):
LGTM

pkarashchenko (approved):

boot/nuttx/main.c
```diff
@@ -24,6 +24,7 @@
 #include <nuttx/config.h>
 
 #include <stdio.h>
```
pkarashchenko, line 26: Do we need to consider if `#include <stdio.h>` can be removed?

nandojve: Indeed, it is not necessary.

nandojve:
Updated commit credentials.

PR #1253, opened 2021-12-20
boot_serial: Add support for the echo command

wcappelle:
This PR adds support for the echo command in mcumgr.


boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,41 @@ bs_upload(char *buf, int len)
...
 }
 
+#ifdef MCUBOOT_BOOT_MGMT_ECHO
+static bool 
+decode_echo(cbor_state_t *state, cbor_string_type_t *result)
+{
+	bool int_res;
+	size_t bsstrdecoded;
+
+	bool tmp_result = (((map_start_decode(state) && (int_res = (multi_decode(2, 2, &bsstrdecoded, (void *)tstrx_decode, state, result, sizeof(cbor_string_type_t))), ((map_end_decode(state)) && int_res)))));
```
de-nordic, line 449: This is not readable.

wcappelle: 100% right :-). I was already wondering how this passed code review in the decode_Upload method.

utzig: This looks like it was auto-generated by `cddl-gen`.

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,45 @@ bs_upload(char *buf, int len)
...
+    map_end_decode(state);
+    return true;
+}
+
+
+static void
+bs_echo(char *buf, int len)
+{
+    const char *echo_buf="";
+    size_t echo_buflen = strlen(echo_buf);
```
de-nordic, line 463: No point to initialize `echo_buf` to "' and `strlen` is wasted call.

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,45 @@ bs_upload(char *buf, int len)
...
+    size_t bsstrdecoded;
+
```
de-nordic, line 465: stray whiteline

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,45 @@ bs_upload(char *buf, int len)
...
+    cbor_string_type_t str[2];
+
+    if (entry_function((const uint8_t *)buf, len, (const void *)str, &bsstrdecoded, (void *)decode_echo, 1, 2)) {
```
de-nordic, line 468: what is the point of so many casts?

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,45 @@ bs_upload(char *buf, int len)
...
+        echo_buf = (char*)str[1].value;
+        echo_buflen = str[1].len;
+    }
+
+    map_start_encode(&cbor_state, 10);
+    tstrx_put(&cbor_state, "r");
+    tstrx_encode(&cbor_state, &(cbor_string_type_t){.value = (const uint8_t *)echo_buf, .len = echo_buflen});
```
de-nordic, line 475: Anonymous definition of `cbor_string_type_t` structure as function parameter makes the code hard to read and nothing more. Don't you actually want to return what has been decoded into `str`? Why not use the `str` directly here?

boot/zephyr/Kconfig
```diff
@@ -556,6 +556,12 @@ config BOOT_ERASE_PROGRESSIVELY
 	 on some hardware that has long erase times, to prevent long wait
 	 times at the beginning of the DFU process.
 
+config BOOT_MGMT_ECHO
+	bool "Enable echo command"
+	default "n"
```
nvlsianpu, line 561:
```suggestion
```
For any Kconfig property we always get implicate `default n` if not gives it y explicitly.

boot/boot_serial/src/boot_serial.c
wcappelle: true, fixed

boot/boot_serial/src/boot_serial.c
wcappelle: I took the cbor_decode_Upload  as example. I could remove the str cast, but all others are needed, but harmless. I'm not really into cbor & all of it's functionality, so I just took what I could find in the current (generated) code. Suggestions or pointers to look into are welcome to optimize of course.

boot/boot_serial/src/boot_serial.c
wcappelle: good catch, I've updated the code to just use the found str in the buffer, which skips this anonymous struct afterall.

boot/zephyr/Kconfig
wcappelle: Thanks for clarifying. I've updated the file to your comment.

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,39 @@ bs_upload(char *buf, int len)
...
+    size_t bsstrdecoded;
+
+    if (!map_start_decode(state)) {
+        return false;
+    }
+    if (!multi_decode(2, 2, &bsstrdecoded, (void *)tstrx_decode, state, result, sizeof(cbor_string_type_t))) {
+        return false;
+    }
+    map_end_decode(state);
+    return true;
```
de-nordic, line 455: Shouldn't the map_end_decode happen regardless whether multi_decode is successful or not?
I have checked decode_Upload and it seems to call on map_end_decode independently from `multi_decode` returned value.

de-nordic (commented):
I left one comment regarding decode_echo.

Before anyone else notices: yes there is no check whether incoming cbor payload is "d":<echo string> map entry, so no check whether key is "d", but I think this is ok. The entire echo command is characteristic enough and it should be enough to just respond with the <echo string>, regardless of key value.

boot/boot_serial/src/boot_serial.c
wcappelle: It's a pain to read it, but you're right, it seems like doing so. I've added the end decode now

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,40 @@ bs_upload(char *buf, int len)
...
+
+    if (!map_start_decode(state)) {
+        return false;
+    }
+    if (!multi_decode(2, 2, &bsstrdecoded, (void *)tstrx_decode, state, result, sizeof(cbor_string_type_t))) {
+        map_end_decode(state);
+        return false;
+    }
+    map_end_decode(state);
+    return true;
```
de-nordic, line 456:
```
int ret;
...
ret = multi_decode(...);
map_end_decode(state);
return ret;
```

wcappelle: or like that :-)
Thanks!

de-nordic (approved):
Looks OK. I have tested that on rf52840dk_nrf52840 and it works.

wcappelle:
> Looks OK. I have tested that on rf52840dk_nrf52840 and it works.

Thanks @de-nordic for the review & test.

wcappelle:
@nvlsianpu @utzig , could you find some time to finish this review? Thanks!

d3zd3z:
The stray end of line space I mentioned earlier is still present.

d3zd3z (changes requested):
Oops, apparently this comment is still pending.

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,45 @@ bs_upload(char *buf, int len)
     flash_area_close(fap);
 }
 
+#ifdef MCUBOOT_BOOT_MGMT_ECHO
+static bool
```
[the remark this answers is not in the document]
wcappelle: Oops! fixed!

wcappelle:
@d3zd3z , I think I've tackled all of your comments. If not let me know, and will fix them asap.

wcappelle:
@d3zd3z , could you check if all is fine for this PR?

d3zd3z (approved):

boot/boot_serial/src/boot_serial.c
```diff
@@ -439,6 +439,38 @@ bs_upload(char *buf, int len)
...
+    size_t bsstrdecoded;
+    cbor_string_type_t str[2];
+
+    if (entry_function((const uint8_t *)buf, len, str, &bsstrdecoded, (void *)decode_echo, 1, 2)) {
+        map_start_encode(&cbor_state, 10);
+        tstrx_put(&cbor_state, "r");
+        tstrx_encode(&cbor_state, &str[1]);
+        map_end_encode(&cbor_state, 10);
+        boot_serial_output();
+    }
```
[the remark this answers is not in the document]
de-nordic: With the default settings 112 long echo string is enough to show the problem.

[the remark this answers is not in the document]
d3zd3z: As long as this doesn't introduce something like a buffer overflow, perhaps this should be fixed in a separate PR, since it applies to more than just the echo command.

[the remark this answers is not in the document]
de-nordic: OK, so this doesn't. But I have found problem within boot_serial_output, that I am fixing.

de-nordic (approved):

PR #1254, opened 2021-12-20
single loader: Add caching of validation state of an image 

wcappelle:
Allow a cached validation result of the main slot (now this is either always, or never). on a M0 this takes some significant time in the boot process. The validation result is stored in the swap area, which is checked before being boot in serial recovery mode.


boot/zephyr/single_loader.c
```diff
@@ -49,6 +49,37 @@ boot_image_validate(const struct flash_area *fa_p,
...
+    fih_int fih_rc = FIH_FAILURE;
+
+    memset(&state, 0, sizeof(struct boot_swap_state));
+    rc = boot_read_swap_state(fa_p, &state);
+    if( rc != 0)
+        FIH_RET(FIH_FAILURE);
+    if( state.magic != BOOT_MAGIC_GOOD
+            || state.image_ok != BOOT_FLAG_SET ) {
+        /* At least validate the image once */
+        FIH_CALL(boot_image_validate, fih_rc, _fa_p, hdr);
```
wcappelle, line 67: note to self, use fa_p instead of _fa_p

d3zd3z:
How does this differ from just not setting VALIDATE_PRIMARY_SLOT?

wcappelle:
@d3zd3z This PR differs from the VALIDATE_PRIMARY_SLOT in the way that the result of the validation is cached in the main flash. So when a new image is loaded into the main slot, only once the validation is being done of that image, and cached at the last bytes of the main partition (stored in the swap area at the end of the main partition - 'image ok' is being reused as cache flag). At next boot, the flag will be checked, and validation will be skipped if it was a good image.
This has as a consequence that it's less secure, but has a benefit that the validation step does not happen at every boot, which is not fast on a slow mcu like a ARM-M0@32MHz. It's a tradeoff between speed & security in my case.

d3zd3z:
It is still unclear to me how this behavior differs from having VALIDATE_PRIMARY_SLOT disabled. The image will be validated before it is installed in the primary slot, but on subsequent boots it will be assumed to have been validated.

wcappelle:
In serial recovery, you just flash an image directly into the main slot without verification of signature. It's only after it's flashed you can check if it's a valid image. Serial recovery doesn't check thihs, but it's the normal boot process which handles this. If you don't have the option VALIDATE_PRIMARY_SLOT enabled, there is no validation happening of the contents received in serial recovery, and you can boot any image loaded by this mode. 
This PR is therefore making sure that whatever image you flash via serial recovery, you still have this validation step done once.

d3zd3z:
> This PR is therefore making sure that whatever image you flash via serial recovery, you still have this validation step done once.

What happens if the image isn't valid? Do we erase the image. The thing is, with validate-primary-slot turned off, we aren't really a secure bootloader. If we're relying on secure boot, why not add the image validation as part of that, although, admittedly, that is challenging to do. Trying to "cache" the verification seems a bit misleading, since it doesn't give us any security.

For this case, this implementation does seem to make sense. I don't know if we just want to make it clearer that this isn't adding security to the image, just preventing invalid images from easily being loaded.

d3zd3z (approved):

wcappelle:
> 

@d3zd3z , no the image is not erased for now, but could be added if wanted? You're right, it does lower the security, but it's at least better than disabling the validation at all I think? If you have any suggestions to make it a bit more secure, i'm happy to get feedback. For sure I could add this warning in the KConfig that it is lowering the security if one could tamper with the flash?

d3zd3z:
I think I'd be ok with this as long as we are clear in the Kconfig (and probably the docs) that this feature isn't secure, as it trusts. Maybe it'd be good to explain where this is useful compared to just disabling the verification of slot 0.

wcappelle:
@d3zd3z , i've updated the KConfig & docs with this information. Could you have a look if this is the information you've wanted? Thanks!

PR #1255, opened 2021-12-20
single_loader: boot_serial: Add support for encrypted images 

wcappelle:
Adding support to allow encrypted images to be send & decrypted over the serial recovery methods in a single application context. This has been done by decryption in place, and is not power failsafe, but can always be redone when failed. Anyhow serial recovery isn't power failsafe, so don't think this is a bad thing. On a STM32L0@32MHz with 100kb of imagesize, this process takes about 2 minutes to complete (transfer included 9600baud)

The decryption is triggered after the last chunk of an upload via serial recovery is received.
The decryption will read a flash eraseblock out of flash, decrypt it, erase the flash block, and write the decrypted block back to the flash. 

There is also some work done for the keys in the Kconfig. I'm aware there is a duplicate PR for this. I can rework this other PR is merged in.


boot/boot_serial/src/boot_serial.c
```diff
@@ -200,6 +204,17 @@ bs_list(char *buf, int len)
                                    fih_rc, image_index, slot);
                 if (fih_eq(fih_rc, BOOT_HOOK_REGULAR))
                 {
+#ifdef MCUBOOT_ENC_IMAGES
+                    if (slot == 0 && IS_ENCRYPTED(&hdr)) {
+                        /* Clear the encrypted flag we didn't supply a key
+                        * This flag could be set if there was a decryption in place
+                        * performed before. We will try to validate the image without
+                        * decryption by clearing the flag in the heder. If
```
d3zd3z, line 212: The is a stray space at the end of this line.

boot/zephyr/CMakeLists.txt
```diff
@@ -304,6 +305,47 @@ if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
   zephyr_library_sources(${GENERATED_PUBKEY})
 endif()
 
+# CONF_FILE points to the KConfig configuration files of the bootloader.
+unset(CONF_DIR)
+foreach (filepath ${CONF_FILE})
```
d3zd3z, line 310: Cmake convention never has spaces between the keyword and the argument.

d3zd3z (changes requested):
Other than some minor whitespace issues, this looks ok to me.

wcappelle:
@d3zd3z , thanks for your review. I've updated the requested changes.

d3zd3z (approved):

PR #1259, opened 2021-12-22
espressif: Use "TAG" field from SOC_LOG* macros from IDF libraries

gustavonihei:
This PR intends to make use of the TAG field from `SOC_LOG*` macros used by some IDF components.

Besides making the log messages more intuitive, it also brings the benefit of removing the need for passing `-Wno-unused-variable` compiler flag, since now TAG is being used as part of the log message.


utzig (approved):
LGTM

PR #1260, opened 2021-12-22
imgtool: Update version 1.8.1 after PR#1217 been merged.

jamike:
Since imgtool supports larger align value, and align_max. Updating
https://pypi.org/project/imgtool is a must.

### DIFF
--- a/scripts/imgtool/__init__.py
+++ b/scripts/imgtool/__init__.py
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-imgtool_version = "1.8.0"
+imgtool_version = "1.8.1"


utzig:
@d3zd3z I'm not sure if this would also require a full minor release of MCUboot or if `imgtool` can just be release separately with its own version. The main issue we had with version numbers before was that people don't like alpha/beta so we have resort to proper stable numbering.

gustavonihei:
Does MCUboot comply to [semantic versioning rules](https://semver.org/)? If so, a Minor release seems to be more appropriate (e.g. 1.9.0):

> MINOR version when you add functionality in a backwards compatible manner [...]

Maybe `imgtool` needs to have its own versioning, so that it becomes free from the stricter release process from MCUboot.

nvlsianpu:
Can we design the version 1.8.0.1? @utzig Is that possible?

utzig:
> Can we design the version 1.8.0.1? @utzig Is that possible?

That does not follow Semantic Versioning. But pypi.org is OK with this number afair and it should work correctly, upgrade from 1.8.0, and so on. For the purpose of doing this "pre-release of new features" I think it's one reasonable idea.

> Does MCUboot comply to [semantic versioning rules](https://semver.org/)? If so, a Minor release seems to be more appropriate (e.g. 1.9.0):
> 
> > MINOR version when you add functionality in a backwards compatible manner [...]

Yes, but MCUboot has always been 1.x.y and was mostly compatible, if not 100%, with previous versions. I guess we just don't like incrementing the MAJOR! What seems more reasonable to me is to make it 1.8.1 because next MCUboot release (for reals!) will probably be 1.9.0. So we could either do a minor full MCUboot release 1.8.1, or just `imgtool`, that was my main doubt.

d3zd3z:
I'm ok with a 1.8.1, but I think we should release mcuboot and imgtool together.

utzig:
> I'm ok with a 1.8.1, but I think we should release mcuboot and imgtool together.

@jamike #1170 can be used as a reference of what files need to be updated to do a proper 1.8.1 release. We can tag it afterwards.

jamike:
As we are targeting a mcuboot release, can we try to merge https://github.com/mcu-tools/mcuboot/pull/900, Before I create the PR for mcuboot 1.8.1

paulwuertz:
Without this PR one cannot sign images for the sam_e70b_xplained with imgtool, just installed the forks imgtool manually because of this...

d3zd3z:
As discussed at the TSC call, we will release a 1.9.0 for all of MCUboot+imgtool.

PR #1263, opened 2021-12-28
bootutil: miscellaneous improvements to code

romanjoe:
Recently we have had a code review of our bootloader project based on mcuboot.
Here are some fixed we made to common files as result of code review.

- add additional input params check
- introduce defines instead of hardcoded values
- refactor some excessive goto statements

### DIFF
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -82,8 +82,10 @@ struct boot_status {
     uint8_t swap_type;    /* The type of swap in effect */
     uint32_t swap_size;   /* Total size of swapped image */
 #ifdef MCUBOOT_ENC_IMAGES
+#define BOOT_UNINITIALIZED_KEY_FILL 0xFF
     uint8_t enckey[BOOT_NUM_SLOTS][BOOT_ENC_KEY_ALIGN_SIZE];
 #if MCUBOOT_SWAP_SAVE_ENCTLV
+#define BOOT_UNINITIALIZED_TLV_FILL 0xFF
     uint8_t enctlv[BOOT_NUM_SLOTS][BOOT_ENC_TLV_ALIGN_SIZE];
 #endif
 #endif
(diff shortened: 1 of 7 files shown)


boot/bootutil/src/bootutil_priv.h
```diff
@@ -82,8 +82,10 @@ struct boot_status {
     uint8_t swap_type;    /* The type of swap in effect */
     uint32_t swap_size;   /* Total size of swapped image */
 #ifdef MCUBOOT_ENC_IMAGES
+#define BOOT_UNINITIALIZED_KEY_FILL 0xFF
```
gustavonihei, line 85: The convenience for avoid duplicating the `ifdef` comes at the price of hurting the code organization which, for me, makes it not worth it.
Please, evaluate moving these definitions before the struct declaration.

romanjoe: You mean `#define BOOT_UNINITIALIZED_KEY_FILL 0xFF` and `#define BOOT_UNINITIALIZED_TLV_FILL 0xFF`? Because ifdefs itself controls wether `uint8_t enckey[BOOT_NUM_SLOTS][BOOT_ENC_KEY_ALIGN_SIZE];` and `uint8_t enctlv[BOOT_NUM_SLOTS][BOOT_ENC_TLV_ALIGN_SIZE];` are included in struct declaration.

gustavonihei: > You mean `#define BOOT_UNINITIALIZED_KEY_FILL 0xFF` and `#define BOOT_UNINITIALIZED_TLV_FILL 0xFF`?

Yes, exactly those definitions.

boot/bootutil/src/encrypted.c
```diff
@@ -79,15 +79,11 @@ key_unwrap(const uint8_t *wrapped, uint8_t *enckey)
 
     bootutil_aes_kw_init(&aes_kw);
     rc = bootutil_aes_kw_set_unwrap_key(&aes_kw, bootutil_enc_key.key, *bootutil_enc_key.len);
-    if (rc != 0) {
-        goto done;
-    }
-    rc = bootutil_aes_kw_unwrap(&aes_kw, wrapped, TLV_ENC_KW_SZ, enckey, BOOT_ENC_KEY_SIZE);
-    if (rc != 0) {
-        goto done;
+
```
d3zd3z, line 82: I'm not sure I'm convinced this is an improvement. It makes the code a little more fragile if code ends up being added after this block, it could be missed how the error handling is done. The change wouldn't affect the generated code, regardless.

d3zd3z: Also, there is some spurious whitespace on the blank line here.

d3zd3z (changes requested):
My main suggestion would be to split this into a couple of patches that fix specific things. Changing the hash size checks is technically a behavior change (an improvement), and is probably best as its own patch.

PR #1269, opened 2022-01-07
allow leading zeros in version number

Olstyle:
I could not find a reason why leading zeros in version numbers should be filtered out, so I changed that part of the checking regex. As a result I can now use version numbers like `01.01`, which lead to the same image header as `1.1` .

### DIFF
--- a/scripts/imgtool/version.py
+++ b/scripts/imgtool/version.py
@@ -28,7 +28,7 @@ SemiSemVersion = namedtuple('SemiSemVersion', ['major', 'minor', 'revision',
                                                'build'])
 
 version_re = re.compile(
-    r"""^([1-9]\d*|0)(\.([1-9]\d*|0)(\.([1-9]\d*|0)(\+([1-9]\d*|0))?)?)?$""")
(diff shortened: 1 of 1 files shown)


d3zd3z:
Leading zeros are explicitly forbidden by https://semver.org/, which we try to follow.

d3zd3z:
Do you have a use case that requires this? The problem I see with this is that 01.01 will be interpreted the same by imgtool, but might be interpreted differently in other tools.

Olstyle:
> Do you have a use case that requires this? The problem I see with this is that 01.01 will be interpreted the same by imgtool, but might be interpreted differently in other tools.
In our old environment I try to modernize without antagonizing to many people the version numbering was 4 digit BCD. So 01.01 was 0x0101 whereas 1.1 would become 0x0011 if you don't look out in the tooling. So leading zeros, especially after the point are the standard notation for all "old" developers.

Olstyle:
> Leading zeros are explicitly forbidden by https://semver.org/, which we try to follow.

I can accept that, but again not find any reference to this in the docs (though the code in the version.py does say that it is supposed to be "Semi Semantic Versioning"). 
Whether it makes sense to enforce a versioning standard on the users of mcuboot on top of the actual limits of mcuboot itself is a different question. In my mind it should not be up to mucboot to decide how applications are versioned as long as the version can be represented.

PR #1270, opened 2022-01-07
imgtool: Add clear image generation with encryption capability

jamike:
Create an option to generate a clear image with encryption capability that can be installed on a primary slot. Since image has encryption capability image can be swap and encrypted in secondary slot

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -305,7 +305,7 @@ class Image():
         return cipherkey, ciphermac, pubk
 
     def create(self, key, public_key_format, enckey, dependencies=None,
-               sw_type=None, custom_tlvs=None, encrypt_keylen=128):
+               sw_type=None, custom_tlvs=None, encrypt_keylen=128, clear=False):
         self.enckey = enckey
 
         # Calculate the hash of the public key
@@ -472,13 +472,14 @@ class Image():
                 else:
                     tlv.add('ENCX25519', enctlv)
 
-            nonce = bytes([0] * 16)
-            cipher = Cipher(algorithms.AES(plainkey), modes.CTR(nonce),
-                            backend=default_backend())
-            encryptor = cipher.encryptor()
-            img = bytes(self.payload[self.header_size:])
-            self.payload[self.header_size:] = \
-                encryptor.update(img) + encryptor.finalize()
+            if not clear:
+                nonce = bytes([0] * 16)
+                cipher = Cipher(algorithms.AES(plainkey), modes.CTR(nonce),
+                                backend=default_backend())
+                encryptor = cipher.encryptor()
+                img = bytes(self.payload[self.header_size:])
+                self.payload[self.header_size:] = \
+                    encryptor.update(img) + encryptor.finalize()
 
         self.payload += prot_tlv.get()
         self.payload += tlv.get()
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -254,6 +254,10 @@ class BasedIntParamType(click.ParamType):
               type=click.Choice(['128','256']),
               help='When encrypting the image using AES, select a 128 bit or '
                    '256 bit key len.')
+@click.option('-c', '--clear', required=False, is_flag=True, default=False,
+              help='Output a non-encrypted image with encryption capabilities,'
+                   'so it can be installed in the primary slot, and encrypted '
+                   'when swapped to the secondary.')
 @click.option('-e', '--endian', type=click.Choice(['little', 'big']),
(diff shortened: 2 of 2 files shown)


scripts/imgtool/image.py
```diff
@@ -477,8 +477,9 @@ def create(self, key, public_key_format, enckey, dependencies=None,
                             backend=default_backend())
             encryptor = cipher.encryptor()
             img = bytes(self.payload[self.header_size:])
-            self.payload[self.header_size:] = \
-                encryptor.update(img) + encryptor.finalize()
+            if clear == False:
```
utzig, line 480: Pythonic and correct way to do this: `if not clear:`

Olstyle:
That's exactly the feature I was missing while discussing how an MCUs production image could be assembled here: https://github.com/mcu-tools/mcuboot/discussions/1258

scripts/imgtool/main.py
```diff
@@ -254,6 +254,10 @@ def convert(self, value, param, ctx):
               type=click.Choice(['128','256']),
               help='When encrypting the image using AES, select a 128 bit or '
                    '256 bit key len.')
+@click.option('-c', '--clear', required=False, is_flag=True, default=False,
+              help='Output an encrypted image file that can be installed on '
+                   'primary slot. (clear image code with encrypt capability '
+                   'in secondary slot during swap.)')
```
Olstyle, line 260: This is incomplete. You also need to define this as an input to sign and forward it to create
```
def sign(key, public_key_format, align, version, pad_sig, header_size,
         pad_header, slot_size, pad, confirm, max_sectors, overwrite_only,
         endian, encrypt_keylen, encrypt, clear, infile, outfile, dependencies,
         load_addr, hex_addr, erased_val, save_enctlv, security_counter,
         boot_record, custom_tlv, rom_fixed, max_align):
```
in line 306
and
```
    img.create(key, public_key_format, enckey, dependencies, boot_record,
               custom_tlvs, int(encrypt_keylen), clear)
```   
in line 356
are needed to have this option accepted and forwarded. (I tested the tool with those additions and then got the expected binary which was encrypted on swap on my target).

jamike: Correct, thanks.

scripts/imgtool/image.py
```diff
@@ -477,8 +477,9 @@ def create(self, key, public_key_format, enckey, dependencies=None,
                             backend=default_backend())
             encryptor = cipher.encryptor()
             img = bytes(self.payload[self.header_size:])
-            self.payload[self.header_size:] = \
-                encryptor.update(img) + encryptor.finalize()
+            if not clear:
+                self.payload[self.header_size:] = \
+                    encryptor.update(img) + encryptor.finalize()
```
utzig, line 482: If you're not going to use the `encryptor` when `clear` is set, is there a reason to leave the above cryptographic primitive instantiation enabled?

jamike: No reason,  I ll fix that soon

scripts/imgtool/image.py
```diff
@@ -471,14 +471,15 @@ def create(self, key, public_key_format, enckey, dependencies=None,
                     tlv.add('ENCEC256', enctlv)
                 else:
                     tlv.add('ENCX25519', enctlv)
-
-            nonce = bytes([0] * 16)
-            cipher = Cipher(algorithms.AES(plainkey), modes.CTR(nonce),
+
```
utzig, line 474: You left stray spaces on this line.

scripts/imgtool/image.py
```diff
@@ -471,14 +471,15 @@ def create(self, key, public_key_format, enckey, dependencies=None,
...
                 else:
                     tlv.add('ENCX25519', enctlv)
-
-            nonce = bytes([0] * 16)
-            cipher = Cipher(algorithms.AES(plainkey), modes.CTR(nonce),
+            
+            if not clear:
+                nonce = bytes([0] * 16)
+                cipher = Cipher(algorithms.AES(plainkey), modes.CTR(nonce),
                             backend=default_backend())
```
utzig, line 478: Please align this with the above arguments (4 more spaces I think).

scripts/imgtool/main.py
utzig, line 260: Technically speaking it's not an "encrypted image", I would suggest something like: "Output a non-encrypted image with encryption capabilities, so it can be installed in the primary slot, and encrypted when swapped to the secondary."

utzig (changes requested):
A few more nits and it should be fine...

utzig (approved):
LGTM

PR #1274, opened 2022-01-11
doc: fixed release-notes.md

FrancescoSer:
Fixed typos and grammar mistakes.
Rephrased slightly unclear sentences.
Fixed punctuation.
Unified spaces after periods.
Unified tenses used in listed changes.

Signed-off-by: Francesco Domenico Servidio <francesco.servidio@nordicsemi.no>

NOTE: No changes in the meaning of sentences were introduced.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,85 +1,84 @@
-# MCUboot release notes
+# Release notes
 
 - Table of Contents
 {:toc}
 
 ## Version 1.8.0
 
-The 1.8.0 release of MCUboot contains numerous fixes, and adds support
-for the NuttX RTOS, and the Espressif ESP32 SDK.
+The 1.8.0 release of MCUboot contains numerous fixes and adds support
+for the NuttX RTOS and the Espressif ESP32 SDK.
 
 ### About this release
 
-- Add support for the NuttX RTOS.
-- Add support for the Espressif ESP32 SDK.
+- Added support for the NuttX RTOS.
+- Added support for the Espressif ESP32 SDK.
 - `boot_serial` changed to use cddl-gen, which removes the dependency
   on tinycbor.
-- Add various hooks to be able to change how image data is accessed.
-- Cypress supports Mbed TLS for encryption.
-- Support using Mbed TLS for ECDSA.  This can be useful if Mbed TLS is
-  brought in for another reason.
-- Add simulator support for testing direct-XIP and ramload.
-- Support Mbed TLS 3.0.  Updates the submodule for Mbed TLS to 3.0.
-- Enable direct-xip mode in Mbed-OS port.
-- extract `bootutil_public` library, a common interface for MCUboot
+- Added various hooks to be able to change how the image data is accessed.
+- Cypress now supports Mbed TLS for encryption.
+- Added support for using Mbed TLS for ECDSA. This is useful if Mbed TLS is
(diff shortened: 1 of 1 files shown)


nvlsianpu (approved):

docs/release-notes.md
```diff
@@ -1,150 +1,149 @@
...
-The 1.8.0 release of MCUboot contains numerous fixes, and adds support
-for the NuttX RTOS, and the Espressif ESP32 SDK.
+The 1.8.0 release of MCUboot contains numerous fixes and adds support
+for the NuttX RTOS and the Espressif ESP32 SDK.
 
 ### About this release
 
-- Add support for the NuttX RTOS.
-- Add support for the Espressif ESP32 SDK.
+- Added support for the NuttX RTOS.
```
d3zd3z, line 13: I'm going to leave a bit -1 in this grammar change. The use of the present tense is very intentional, and overall makes the release notes much easier to read. The same goes for commit text, which should also be written in the present tense.

d3zd3z (changes requested):
I realize the release notes aren't consistent as to tense, mostly due to different people doing the releases. I do feel, fairly strongly, though that the release notes should be written in the simpler present tense.

docs/release-notes.md
FrancescoSer: @d3zd3z  While I personally prefer the past tense, I don't have any issue with switching them all to the present tense, as long as they are consistent if that's the issue. I just think that the current mix of tenses does not make much sense.

gustavonihei: Actually, the discussion is not between *present* or *past* tense, because that would assume that the verb goes in the Indicative mood, which should not be the case.
A Release Notes (as well as Git commit messages) usually contains verbs in the Imperative mood, which are more suitable to point that *things will change from this point forward*.

d3zd3z: Kind of drifting off topic, but there are languages that have a past imperative tense.

More seriously, do you think we should add comments to the release notes process docs stating the release notes should be in imperative mood? I see that a lot more mixed that the text of git commits, but I find it easier to go from git commits in imperative mood to release notes based on those.

BTW, I have seen the indicative present in commit text. It feels weird to me. "Adding support for ...", although I'm not sure if that is actually present indicative "I am adding support".

FrancescoSer: @gustavonihei The imperative mood is used usually as a convention for the first line of commit messages because also commands like git merge or git revert automatically create messages in imperative. I guess one could use the argument that it makes sense to give "orders" to the codebase in imperative mood.

But we are talking about release notes here. Do you really want your release notes to be just a copy-paste of commit messages? In the context of a release note, it doesn't really make much sense to me. When the release note is published, the "order" to the codebase is something that happened in the past, so using the imperative mood in this context looks plain wrong to me. To whom are you ordering the verb's action? To the reader? It doesn't make sense to me.

But, assuming this does actually makes sense, it doesn't look like the release notes are actually consistent in following your own rule.

Anyway, even if I think this is plain wrong, I'm just an external contributor here, so if you guys think my edit of the tenses here is "controversial", I can revert it and keep the rest of the changes.

d3zd3z:
I'm still a bit reluctant to make grammar changes to previous release notes, especially without an agreed-upon guideline for how they should be.

PR #1279, opened 2022-01-14
samples/zephyr: Remove the body of the shell test runner

d3zd3z:
Instead of trying to main two test runners for the Zephyr sample
directory, have the old test runner just print out a message that users
should use the new test runner.  Eventually, we can just remove old
runner entirely.

Instead of asking the user for each test if it worked or not, the new
test runner looks at the serial log and makes this conclusion itself.
It is able to run the whole test suite completely automatically.


d3zd3z:
I don't have good visibility into who uses run-tests.sh, and there who would be impacted by this. Please feel free to mention people you'd like to review this.

nvlsianpu:
@MarekPieta We don't use that. Right?

MarekPieta:
I looked at the code and it seems like the tests are supported. I never ran it locally, but I guess that they should work with Zephyr.

MarekPieta:
I think we may need to completely remove files relates to tests in case they are not maintained to avoid misleading users.

d3zd3z:
> I think we may need to completely remove files relates to tests in case they are not maintained to avoid misleading users.

Which files are you referring to? That's kind of the point of this change. Right now, we have both a shell script runner, and a small Go program that invoke the tests. The Go version has significantly more functionality, as it is able to automatically verify that the tests work. I run this regularly, especially as we are preparing for a release. I was hoping to remove the shell on so that they don't drift apart. That has the consequence of making people who are using the shell one migrate to using the other one.

MarekPieta:
If the Go program works fine with Zephyr's test we can remove the shell script. I was just not sure if we maintain the tests for Zephyr. To be honest I never ran them locally.

nvlsianpu (approved):

PR #1280, opened 2022-01-19
boot_serial: zephyr: Add optional timeout to enter serial recovery

wcappelle:
This PR adds the possibility to only enter the bootloader's
serial recovery mode when a mcumgr command is received within a
given timeout.

This PR is the last part of the split off PR #989. The work has been done for Zephyr only (although should be easy to integrate in others).

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -93,6 +93,7 @@ const struct boot_uart_funcs *boot_uf;
 static uint32_t curr_off;
 static uint32_t img_size;
 static struct nmgr_hdr *bs_hdr;
+static bool bs_entry;
 
 static char bs_obuf[BOOT_SERIAL_OUT_MAX];
 
@@ -555,6 +556,9 @@ boot_serial_input(char *buf, int len)
     } else {
         bs_rc_rsp(MGMT_ERR_ENOTSUP);
     }
+#ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
+    bs_entry = true;
+#endif
 }
 
 static void
@@ -670,25 +674,29 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
  * Task which waits reading console, expecting to get image over
  * serial port.
  */
-void
-boot_serial_start(const struct boot_uart_funcs *f)
+static void
+boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
 {
     int rc;
     int off;
     int dec_off = 0;
     int full_line;
     int max_input;
+    int elapsed_in_ms = 0;
 
     boot_uf = f;
     max_input = sizeof(in_buf);
 
     off = 0;
-    while (1) {
+    while (timeout_in_ms > 0 || bs_entry) {
         MCUBOOT_CPU_IDLE();
         MCUBOOT_WATCHDOG_FEED();
+#ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
+        uint32_t start = k_uptime_get_32();
+#endif
         rc = f->read(in_buf + off, sizeof(in_buf) - off, &full_line);
         if (rc <= 0 && !full_line) {
-            continue;
+            goto check_timeout;
         }
         off += rc;
         if (!full_line) {
@@ -698,7 +706,7 @@ boot_serial_start(const struct boot_uart_funcs *f)
                  */
                 off = 0;
             }
-            continue;
+            goto check_timeout;
         }
         if (in_buf[0] == SHELL_NLIP_PKT_START1 &&
           in_buf[1] == SHELL_NLIP_PKT_START2) {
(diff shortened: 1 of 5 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -93,6 +93,7 @@ const struct boot_uart_funcs *boot_uf;
 static uint32_t curr_off;
 static uint32_t img_size;
 static struct nmgr_hdr *bs_hdr;
+static int bs_entry;
```
de-nordic, line 96: Can this be just bool?

boot/boot_serial/src/boot_serial.c
```diff
@@ -683,12 +687,17 @@ boot_serial_start(const struct boot_uart_funcs *f)
...
     off = 0;
-    while (1) {
+    while (timeout_in_ms > 0 || bs_entry) {
         MCUBOOT_CPU_IDLE();
         MCUBOOT_WATCHDOG_FEED();
+#ifdef __ZEPHYR__
+        uint32_t start = k_uptime_get_32();
+#else
+        uint32_t start = os_get_uptime_usec()/1000;
+#endif
```
de-nordic, line 697: I am little against adding more __ZEPHYR__/else blocks in common code: if number of supported systems would increase then we would have a lot of these.
Can we solve that by adding some function declaration that different systems could provide own implementations for?

boot/boot_serial/src/boot_serial.c
wcappelle: yes, makes sense. done

boot/boot_serial/src/boot_serial.c
wcappelle: I understand what you mean, but not sure where to put the platform dependant header in that case.. i'm only familiar with zephyr, so for any other build systems/platforms I would need some advice to proceed. Could you agree if I would make it a define at top to start with, so the code is not double ifdef'ed? 
```
#ifdef __ZEPHYR__
#define MCUBOOT_UPTIME_MS_U32 (k_uptime_get_32())
#else
#define MCUBOOT_UPTIME_MS_U32 (os_get_uptime_usec()/1000)
#endif
```

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -185,6 +185,10 @@
 #define MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD
 #endif
 
+#ifdef CONFIG_BOOT_SERIAL_WAIT_FOR_DFU
+#define MCUBOOT_SERIAL_WAIT_FOR_DFU
+#endif
+
```
de-nordic, line 191: Make this something like:
```
#ifdef CONFIG_MCUBOOT_SERIAL_WAIT_FOR_DFU
#define MCUBOOT_SERIAL_WAIT_FOR_DFU CONFIG_MCUBOOT_SERIAL_WAIT_FOR_DFU
#else
#define MACUBOOT_SERIAL_WAIT_FOR_DFU 0
#endif
```
Add the definition, of `MCUBOOT_SERIAL_WAIT_FOR_DFU`, or each file system supported by Zephyr (just search mcuboot tree for them); for example (shameless plug) https://github.com/mcu-tools/mcuboot/blob/main/boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h#L89-L93

This way the code will not be available for systems that do not support it without `#ifdef __ZEPHYR__`, you just can add `#if MCUBOOT_SERIAL_WAIT_FOR_DFU != 0`.

boot/boot_serial/src/boot_serial.c
de-nordic: You do not have to do this for every system (look https://github.com/mcu-tools/mcuboot/pull/1280/files#r791572629), when you clearly state that that system does not support it.
You just have to propose some common function name like `os_time_uptime_ms` (or something) and provide implementation for systems that you know and make sure that these systems link that function.
Only systems that set `MCUBOOT_SERIAL_WAIT_FOR_DFU != 0` will have to implement the function, so if somebody would look into the mcuboot_config.h of own system then the person will know that this is hardcoded to 0 and some work is expected to happen before the timeout works.
Look into for example how each system code within mcuboot provides own `flash_area_id_to_multi_image_slot`.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
wcappelle: Thanks for the hint. however i'm looking into the different mcuboot_config.h, and doesn't look like it's common at all to define something to 0 if the feature isn't availlable like for the mynewt you've shown. I will stick with the ifdef instead of assigning a value since that would require all configs to be touched/polluted if you ask me. 
I'll do the changes, and you can argue with this comment if you don't like it, and we can see for alternative options then maybe?

wcappelle:
@de-nordic , thanks for your review. I've reworked the changes you've requested, just look at my comment on the ifdef part. I've done it slightly different to avoid polluting/touching all of the configs. If you don't like it let me know, or suggest another approach, and i'm happy to fix this.

boot/boot_serial/src/boot_serial.c
```diff
@@ -714,5 +721,35 @@ boot_serial_start(const struct boot_uart_funcs *f)
             boot_serial_input(&dec_buf[2], dec_off - 2);
         }
         off = 0;
+check_timeout:
+        /* Subtract elapsed time */
+#ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
+        timeout_in_ms -= (k_uptime_get_32() - start);
+#endif
```
de-nordic, line 728: I guess it is OK, although usage of Zephyr specific k_ call here, in common code, drops system agnostic solution on the next dev that enables `MCUBOOT_SERIAL_WAIT_FOR_DFU` for other system.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
de-nordic: I do not think that this would pollute anything, you have to have separate commits for other sys though, this rather informs users that configure mcuboot for their systems that such option exists in common code but is not supported by their system yet.
We do not have common, generic, config file that would gather all common code defines and then a system specific code would override them, so this is way to inform users about possibilities.

de-nordic: @d3zd3z Sorry for dragging you here, but I would like your opinion on the approach as I do not want to talk @wcappelle into doing something that mcuboot does not care about.

d3zd3z (approved):

wcappelle:
@de-nordic , I see @d3zd3z approved this PR , does this mean nothing else need to happen regarding your comment/request for feedback?
Could you let me know how to proceed if possible with your open review comment? Thanks.

de-nordic (approved):

PR #1282, opened 2022-01-19
V1.9.0 rel

d3zd3z:
Preparations for releasing v1.9.0-rc1.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -3,6 +3,33 @@
 - Table of Contents
 {:toc}
 
+## Version 1.9.0
+
+The 1.9.0 release of MCUboot contains various bug fixes, improves
+support on some recent targets, and adds support for devices with a
+write alignment larger than 8.
+
+This change introduces a potentially incompatible change to the format
+of the image trailer.  If `BOOT_MAX_ALIGN` is kept at 8, the trailer
+format does not change.  However, to support larger write alignments,
+this value can be increased, which will result in a different magic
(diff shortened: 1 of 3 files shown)


docs/release-notes.md
```diff
@@ -3,6 +3,32 @@
 - Table of Contents
 {:toc}
 
+## Version 1.9.0
+
+The 1.9.0 release of MCUboot contains various bug fixes, improves
+support on some recent targets, and adds support for devices with a
+write alignment larger than 8.
```
utzig, line 10: But `imgtool` allows only up to 32, not sure larger than this was tested...

utzig (approved):
LGTM

docs/release-notes.md
nvlsianpu: up to 32 B: https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/bootutil_public.h#L86

docs/release-notes.md
```diff
@@ -3,6 +3,32 @@
...
+
+This change introduces a potentially incompatible change to the format
```
nvlsianpu, line 12: I would mention that new feature is accessible over `MCUBOOT_BOOT_MAX_ALIGN`, if this is undefined or defined to 8 the mcuboot is backward compatible.

docs/release-notes.md
utzig: Thanks, I didn't remember that, we should cite this in the release notes.

PR #1283, opened 2022-01-19
zephyr: rsa: Use a smaller sha256 implementation

SebastianBoe:
Reduce the size of the sha256 implementation by 1300 bytes by using an
implementation that runs 30% slower.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>


boot/zephyr/include/config-rsa.h
```diff
@@ -64,6 +64,7 @@
 #define MBEDTLS_MD_C
 #define MBEDTLS_OID_C
 #define MBEDTLS_SHA256_C
+#define MBEDTLS_SHA256_SMALLER
```
utzig, line 67: Why do it only for RSA configuration?

SebastianBoe: It's the only config I had size problems with.

I can add it to ec, ed25519, kw, and rsa-kw as well if you wish.

sebo@sebo:~/ncs/bootloader/mcuboot$ ack SHA256 boot/zephyr/include/config-*
boot/zephyr/include/config-asn1.h
42:// #define MBEDTLS_SHA256_C

boot/zephyr/include/config-ec.h
67:#define MBEDTLS_SHA256_C

boot/zephyr/include/config-ed25519.h
62:#define MBEDTLS_SHA256_C

boot/zephyr/include/config-kw.h
59:#define MBEDTLS_SHA256_C

boot/zephyr/include/config-rsa.h
66:#define MBEDTLS_SHA256_C
67:#define MBEDTLS_SHA256_SMALLER

boot/zephyr/include/config-rsa-kw.h
65:#define MBEDTLS_SHA256_C

utzig: > It's the only config I had size problems with.

Then you have not tried `ed25519`! :-)

> I can add it to ec, ed25519, kw, and rsa-kw as well if you wish.

Yeah, I don't see why leave them inconsistent. In Mynewt I left this option enabled by default so all configurations already use it.

nvlsianpu (approved):

boot/zephyr/include/config-rsa.h
SebastianBoe: Done

utzig (approved):
LGTM

Issue #1146, opened 2021-09-16
boot/bootutil: image header size check is not precise enough


nvlsianpu:
related to #1140 

looks like `boot_is_header_valid()` is a very rough check regard sizes: it checks whether the header and image binaries fit in the slot.
Slot must be able to accommodate other entities like TLVs, the image trailer and the swap status.

I'm not sure whether this check shouldn't be more precise.

nvlsianpu:
@d3zd3z Are you answer to the questions I'm asking above?

d3zd3z:
Provoked by #1140, I've started trying to add size limit checking to the simulator. What I want to do is be able to make an image that is exactly the largest size possible, and make sure it works.

But, yes, the maximum size does need to include the header, the image itself, the TLV, status, and the trailer. At best, they can all be packed together for a solid image.

The TLV is the trickiest, as it changes size depend on the things that are included within it.

utzig:
> Actually, it might fail pretty badly if swap ends up blowing away the end of the TLV in the process of writing the trailer.

This will probably happen.

nvlsianpu:
looks like bootutil_img_validate() is a place where precise check can be added, alternatively the check can be added somewhere around.

utzig:
> looks like bootutil_img_validate() is a place where precise check can be added, alternatively the check can be added somewhere around.

I don't disagree (have not looked at the code!), but if there is a problem with the checks, maybe add some simulator test that exercises the incorrect behavior?

nvlsianpu:
I'm going to provide a patch for this ticket, will try to add test if I manage to.

utzig:
Fixed by #1286 (hopefully the bot won't be closing this again)

nvlsianpu:
> Fixed by https://github.com/mcu-tools/mcuboot/pull/1286 (hopefully the bot won't be closing this again)

Yes, It was fixed.

PR #1286, opened 2022-01-21
boot: add precise check of the image size

nvlsianpu:
It is possible that image in the slot is so big
that MCUboot swap metadata will interfere with
its content during the swap operation.

This patch introduces additional check to the image
validation procedure.

added `oversized_bootstrap` test for ensure that too big image will be rejected while bootsraping
added `oversized_secondary_slot` test for ensure that too big image upgrade attempt will be rejected. I doesn't work for `max-align-32` feature due to not precise operations on flash image in sim code. Fixing of this problem is out of scope for this PR.

Objective is to fix issue signaled by #1146
this patch is also successor of #1140

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -387,3 +387,27 @@ boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
     return 0;
 }
 #endif
+
+uint32_t bootutil_max_image_size(const struct flash_area *fap)
+{
+#if defined(MCUBOOT_SWAP_USING_SCRATCH)
+    return boot_status_off(fap);
+#elif defined(MCUBOOT_SWAP_USING_MOVE)
+    struct flash_sector sector;
+    /* get the last sector offset */
+    int rc = flash_area_sector_from_off(boot_status_off(fap), &sector);
+    if (rc) {
+        BOOT_LOG_ERR("Unable to determine flash sector of the image trailer");
+        return 0; /* Returning of zero here should cause any check which uses
+                   * this value to fail.
+                   */
+    }
+    return flash_sector_get_off(&sector);
+#elif defined(MCUBOOT_OVERWRITE_ONLY)
+    return boot_swap_info_off(fap);
+#elif defined(MCUBOOT_DIRECT_XIP)
+    return boot_swap_info_off(fap);
+#elif defined(MCUBOOT_RAM_LOAD)
+    return boot_swap_info_off(fap);
+#endif
+}
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -335,6 +335,28 @@ impl ImagesBuilder {
         }
     }
 
+    pub fn make_oversized_bootstrap_image(self) -> Images {
+        let mut flash = self.flash;
+        let ram = self.ram.clone(); // TODO: Avoid this clone.
+        let images = self.slots.into_iter().enumerate().map(|(image_num, slots)| {
+            let dep = BoringDep::new(image_num, &NO_DEPS);
+            let primaries = install_no_image();
+            let upgrades = install_image(&mut flash, &slots[1],
+                ImageSize::Oversized, &ram, &dep, false);
+            OneImage {
+                slots,
+                primaries,
+                upgrades,
+            }}).collect();
+        Images {
+            flash,
+            areadesc: self.areadesc,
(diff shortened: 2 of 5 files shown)


nvlsianpu:
cc @MarekPieta

boot/bootutil/src/bootutil_misc.c
```diff
@@ -387,3 +387,31 @@ boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
     return 0;
 }
 #endif
+
```
d3zd3z, line 390: It'd be nice to make sure we've got a good match between this code and what is in https://github.com/mcu-tools/mcuboot/blob/main/sim/src/image.rs#L1516-L1534. Whether they should even be separate is another consideration.

I did do a little testing where I added '4' to the max size to make sure it was causing failures.

sim/src/image.rs
```diff
@@ -1522,14 +1522,19 @@ fn install_image(flash: &mut SimMultiFlash, slot: &SlotInfo, len: ImageSize,
                 // This computation is incorrect, and we need to figure out the correct size.
                 // c::boot_status_sz(dev.align() as u32) as usize
                 16 + 4 * dev.align()
-            } else {
+            } else if Caps::SwapUsingMove.present() {
                 let sector_size = dev.sector_iter().next().unwrap().size as u32;
                 align_up(c::boot_trailer_sz(dev.align() as u32), sector_size) as usize
+            } else if Caps::SwapUsingScratch.present() {
+                c::boot_trailer_sz(dev.align() as u32) as usize
```
nvlsianpu, line 1529: I wonder why introduction of this line imposes badly on so many tests.

d3zd3z: The move swap will require an additional sector (beyond the trailer) in slot 0, as it needs to slide slot 0 down 1 sector. The original computation in the code wasn't correct.

nvlsianpu: ^^ That patch should affect only swap-using-scratch builds. I need to check whether SwapUsinMove is taken into account properly.

nvlsianpu: I found out that this code path changes behavior only for the swap using scratch. Rest goes with ImageSize::Given(size).

nvlsianpu:
@d3zd3z I added new test for prove that issue existed and is fixed by the patch.
This path should be considered as important on.

utzig (commented):
Have not looked at the Rust changes yet...

boot/bootutil/src/bootutil_misc.c
```diff
@@ -387,3 +387,31 @@ boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
...
+                   * this value to fail.
+                   */
+    }
+    return flash_sector_get_off(&sector);
+#elif defined(MCUBOOT_OVERWRITE_ONLY)
+#ifdef MCUBOOT_ENC_IMAGES
+    return boot_enc_key_off(fap, 1);
+#else
+    return boot_swap_size_off(fap);
+#endif
```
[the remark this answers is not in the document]
nvlsianpu: This was obviously my mistake. For bootstrapping the limit should be the same as for selected regular update mode.
As for subsequent modes f. boot_read_swap_state() is parsing swap state fields - which gives the limit.

nvlsianpu:
@utzig @d3zd3z Ping

sim/src/image.rs
```diff
@@ -1516,21 +1675,22 @@ fn install_image(flash: &mut SimMultiFlash, slot: &SlotInfo, len: ImageSize,
...
             slot_len - HDR_SIZE - trailer - tlv_len
+        },
+        ImageSize::Oversized => {
+            let trailer = image_largest_trailer(dev);
+            let tlv_len = tlv.estimate_size();
+            info!("slot: 0x{:x}, HDR: 0x{:x}, trailer: 0x{:x}",
+                slot_len, HDR_SIZE, trailer);
+            // the overflow size is rougly estimated to work for all
+            // configurations. It might be precise if tlv_len will be maked precise.
+            slot_len - HDR_SIZE - trailer - tlv_len + dev.align()*4
```
d3zd3z, line 1691: There is an extra space at the end of this line. I'll probably run rustfmt on this code after this merges, so it really isn't that big of a deal, though.

d3zd3z (approved):
Just a space issue, and I need to rustfmt anyway.

Issue #1288, opened 2022-01-25
imgtool: erased values should be omitted in trailer


wcappelle:
When generating an image with trailer (--pad or --confirm), the imgtool also adds the erased values into that area:
```
19841 :020000040807EB
19842 :10FFE000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF21
19843 :10FFF00077C295F360D2EF7F3552500F2CB67980DF
19844 :00000001FF

```
 This is not wanted because on MCU's which use ECC into their flash, will write the erased values, and compute an ECC on these values. The flash driver even does this on 0XFF (erased) values. Since the ECC of the 0XFF... value is not 0XFF, this will be written to the ECC location in flash, which will prevent the next write to fail since ECC cannot be overwritten.

This leads to various problems in mcuboot/mcuboot lib which uses the image trailer, like the writing of the image_ok or any other flag in the trailer fails because the values can not be (re-)written.

The STM32L5 MCU is affected by this problem.

PR #1289, opened 2022-01-25
imgtool: skip erased values in trailer generation

wcappelle:
When generating an image with trailer (--pad or --confirm), the
imgtool also adds the erased values into that area. This is not
wanted for flash drivers which use ECC and treat writing erased
values as a real write action, which cannot be overwritten anymore.
See issue #1288

For now the change in implementation is always to skip these in the trailer, but i'm not sure if this suits another purpose to still add these? I always can make an option in the tool not to change the current image generation. 

image trailer before change:
```
$ tail -n 4 zephyr/zephyr.encrypted.confirmed.signed.hex.original 
:020000040807EB
:10FFE000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF21
:10FFF00077C295F360D2EF7F3552500F2CB67980DF
:00000001FF
```
image trailer after change:
```
$ tail -n 4 zephyr/zephyr.encrypted.confirmed.signed.hex 
:0897800062AD74BDB5788A6387
:020000040807EB
:10FFF00077C295F360D2EF7F3552500F2CB67980DF
:00000001FF
```

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -242,15 +242,15 @@ class Image():
                                                   self.save_enctlv,
                                                   self.enctlv_len)
                 trailer_addr = (self.base_addr + self.slot_size) - trailer_size
-                padding = bytearray([self.erased_val] * 
-                                    (trailer_size - len(self.boot_magic)))
-                padding += self.boot_magic
                 if self.confirm and not self.overwrite_only:
                     magic_size = 16
                     magic_align_size = align_up(magic_size, self.max_align)
                     image_ok_idx = -(magic_align_size + self.max_align)
-                    padding[image_ok_idx] = 0x01  # image_ok = 0x01
-                h.puts(trailer_addr, bytes(padding))
+                    flag = bytearray([self.erased_val] * magic_align_size)
+                    flag[0] = 0x01 # image_ok = 0x01
+                    h.puts(trailer_addr + image_ok_idx, bytes(flag))
+                h.puts(trailer_addr + (trailer_size - len(self.boot_magic)),
+                       bytes(self.boot_magic))
             h.tofile(path, 'hex')
         else:
             if self.pad:


scripts/imgtool/image.py
```diff
@@ -242,15 +242,15 @@ def save(self, path, hex_addr=None):
...
                     magic_size = 16
                     magic_align_size = align_up(magic_size, self.max_align)
                     image_ok_idx = -(magic_align_size + self.max_align)
-                    padding[image_ok_idx] = 0x01  # image_ok = 0x01
-                h.puts(trailer_addr, bytes(padding))
+                    flag = bytearray([self.erased_val] * magic_align_size)
+                    flag[0] = 0x01 # image_ok = 0x01
+                    h.puts(trailer_addr + image_ok_idx, bytes(flag))
+                h.puts(trailer_addr + (trailer_size - len(self.boot_magic))
+                       , bytes(self.boot_magic))
```
d3zd3z, line 253: Please put the comma at the end of the previous line, as is Python convention.

d3zd3z (approved):
I think this is a good idea. Do most flash tools, at least those that support IntelHex, do the right thing with skipped data like this?

scripts/imgtool/image.py
wcappelle: done

wcappelle:
> I think this is a good idea. Do most flash tools, at least those that support IntelHex, do the right thing with skipped data like this?

I think most of them do, although if not then this is a tool error, or a mcu/flash limitation. 
It might occur that the flash only supports so many separate page writes to the same page before erase (like 4), but that is more something for nand flash I think. That kind of flash isn't supported anyhow by mcuboot, so I think this is safe to do.

d3zd3z:
> It might occur that the flash only supports so many separate page writes to the same page before erase (like 4), but that is more something for nand flash I think.

Mostly what I've seen is that a very small number of multiple writes is allowed to the same write-aligned value (say 32-bits), often just 2. NOR doesn't usually have constraints on how many writes can be done within an erasable sector before erasing, probably limited by the two writes per 32-bit value.

PR #1290, opened 2022-01-25
Mbed port improvements

ATmobica:
These changes are important in the context of developing an application with the latest versions of mbed-os and mbed-tools.

Change list:
Add CMake support - compatible with a new command-line tool for MbedOS; `mbed-tools` (Mbed CLI 2)
Add logging enable parameter in mbed_lib.json - enable logging from .json configuration file
Add header size parameter in mbed_lib.json
Fix mbed files paths
Fix calling to BlockDevice::size before initialization - we need to initialize BlockDevice before getting its size. 

Signed-off-by: George Beckstein george.beckstein@gmail.com

### DIFF
--- a/boot/mbed/CMakeLists.txt
+++ b/boot/mbed/CMakeLists.txt
@@ -1,0 +1,61 @@
+# Copyright (c) 2021 ARM Limited. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Mbed-MCUboot Port
+
+cmake_minimum_required(VERSION 3.19.0 FATAL_ERROR)
+
+get_filename_component(BOOT_UTIL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../bootutil REALPATH)
+
+set(LIB_TARGET mbed-mcuboot)
+set(LIB_BOOTUTIL bootutil)
+
+add_library(${LIB_TARGET} STATIC)
+
+target_include_directories(${LIB_TARGET}
+    PUBLIC
+        include
+        ${BOOT_UTIL_DIR}/include
+        ${BOOT_UTIL_DIR}/src
+)
+
+target_sources(${LIB_TARGET}
+    PRIVATE
+        mcuboot_main.cpp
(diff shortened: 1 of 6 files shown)


utzig:
Why not ping the authors in the original PRs? At a minimum if a new PR is needed it should provide the original commits with the authorship retained...

ATmobica:
> Why not ping the authors in the original PRs? At a minimum if a new PR is needed it should provide the original commits with the authorship retained...

I assume that these PRs are stale if nothing has happened there for six months.  It's not a combination of original commits but solving the same issue for the newest version of mbed-os.

pan-:
@AGlass0fMilk I think this is derivative of your work.

boot/mbed/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+        mbed-mbedtls
+        mbed-storage-flashiap
+        mbed-storage-blockdevice
+)
+
+if("_RTE_" IN_LIST MBED_CONFIG_DEFINITIONS)
+    target_link_libraries(${LIB_TARGET}
+    PUBLIC
+        mbed-os
+    )
```
d3zd3z, line 42: Spurious space at the end of this line.

d3zd3z (commented):
Look for trailing whitespace, there a few of these.

There should be a blank line in the commit message after the first line before the summary. Generally, I would suggest these should probably be separate commits, but they aren't all that large, and all are targeted toward a single target.

However, I do think "Fix calling to BlockDevice::size before initialization" should be its own commit. This could probably use a bit more explanation in the commit text.

Have as a first line "mbed: Fix calling to BlockDevice::size before initialization", a blank line, and then a sentence or two explaining _why_ this is being done.

Each commit will need a Signed-off-by line, at the end of the commit. Have a look at `git log` to see what this looks like.

Thanks for the contribution.

ATmobica:
@d3zd3z Thanks for your comments.
I have to transfer these changes to a different PR due to a change of source branch. It will continue in #1301

PR #1293, opened 2022-01-26
zephyr: Use k_busy_wait if multithreading is not activated

kvaellning:
If CONFIG_MULTITHREADING is not set, k_sleep is not defined by Zephyr.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -67,7 +67,7 @@ const struct boot_uart_funcs boot_funcs = {
 #else
 #include <logging/log_ctrl.h>
 
-#define BOOT_LOG_PROCESSING_INTERVAL K_MSEC(30) /* [ms] */
+#define BOOT_LOG_PROCESSING_INTERVAL_MS     30 /* [ms] */
 
 /* log are processing in custom routine */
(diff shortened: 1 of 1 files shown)


d3zd3z (approved):

nvlsianpu (changes requested):
-1 boot_log_thread_func() is intended to be run in its own thread. See `zephyr_boot_log_start()`.

I see no point to make it compatible with the single thread mode

kvaellning:
@nvlsianpu Then maybe just activate multi-thread mode in the prj.conf per default and generating a better error message during compile time?

nvlsianpu:
> @nvlsianpu Then maybe just activate multi-thread mode in the prj.conf per default

CONFIG_MULTITHREADING=y is by default for all with excuse to some of configuration of the nRF devices.

> ... and generating a better error message during compile time?

Compilation will miss for instance k_thread_create() as well. Do you think that something like error on CONFIG_MULTITHREADING=n should be added?

kvaellning:
> Compilation will miss for instance k_thread_create() as well. Do you think that something like error on CONFIG_MULTITHREADING=n should be added?

That might be the best solution. I'll update the pull request in the next time.

PR #1295, opened 2022-01-26
readme.md: remove "IDF" from Espressif reference

almir-okato:
This change is to avoid misunderstanding as IDF is not supported
on top of MCUboot.
The Espressif port acts as a bare metal bootloader that can be
used on Espressif chips to boot supported OSes (like Zephyr or NuttX).

Signed-off-by: Almir Okato <almir.okato@espressif.com>


utzig (approved):
LGTM

PR #1296, opened 2022-01-27
boot_serial: allow configuring BOOT_SERIAL_INPUT_MAX from Zephyr

vaussard:
`BOOT_SERIAL_INPUT_MAX` is currently set to 512 bytes. This means that around 1kB of RAM is dedicated to `in_buf[]` and `dec_buf[]`, which is a significant amount on memory-constraint targets.

Zephyr can configure the maximum length of commands using the Kconfig BOOT_MAX_LINE_INPUT_LEN, whose default value is 512. But if one reduces this value to save RAM, the input and decoding buffers can also be reduced accordingly.

Make `BOOT_SERIAL_INPUT_MAX` to dependant on `CONFIG_BOOT_MAX_LINE_INPUT_LEN` when available. Use the default value otherwise.

Note: tested on STM32G0


boot/boot_serial/src/boot_serial.c
```diff
@@ -67,7 +67,11 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
+#ifdef CONFIG_BOOT_MAX_LINE_INPUT_LEN
```
nvlsianpu, line 70: `CONFIG_xxx` is a Kconfig generated properties. I would use mediate `MCUBOOT_xxx ` properties in order to make this OS agnostic

de-nordic: Is there Zephyr PR that enables the thing in Kconfig? (https://docs.zephyrproject.org/latest/guides/modules.html#process-for-submitting-changes-to-existing-modules)

nvlsianpu: > @nvlsianpu Aside from renaming, shouldn't we also move this to mcuboot_config.h?

yes

de-nordic: Ah ok, so the CONFIG_BOOT_MAX_LINE_INPUT_LEN is already in Kconfig.

vaussard: Thank you @nvlsianpu and @de-nordic for your suggestions. If I understand correctly:
- Rename `BOOT_SERIAL_INPUT_MAX` to `MCUBOOT_BOOT_SERIAL_INPUT_MAX` (and what about doing the same with `BOOT_SERIAL_OUT_MAX`?)
- Move `#define BOOT_SERIAL_INPUT_MAX   CONFIG_BOOT_MAX_LINE_INPUT_LEN` to boot/zephyr/include/mcuboot_config/mcuboot_config.h
- But where should be `#define BOOT_SERIAL_INPUT_MAX   512` moved? In all the other mcuboot_config.h? Or shall we keep it here and enclose it in `#ifndef BOOT_SERIAL_INPUT_MAX` to set the default value?

nvlsianpu: Keep define naming consistent what is here: defines in this file begins from BOOT_, defines in mcuboot_config.h begins from MCUBOOT_.  All ports but zephyr might define MCUBOOT_SERIAL_INPUT_MAX as 512 (or keep not-defined option her). Zephyr port should define MCUBOOT_SERIAL_INPUT_MAX as CONFIG_BOOT_MAX_LINE_INPUT_LEN.

vaussard: Thank you @nvlsianpu I will rework accordingly.

de-nordic: @vaussard Any changes to the PR?

de-nordic: @vaussard ping?

PR #1301, opened 2022-02-08
Mbed port improvements

ATmobica:
These changes are important in the context of developing an application with the latest versions of mbed-os and mbed-tools.

Change list:

Add CMake support - compatible with a new command-line tool for MbedOS; mbed-tools (Mbed CLI 2)
Fix mbed files paths - it's required for the new version of mbed-os (v6.15)

Add enable MCUboot logging parameter and fix logging configuration - add an option for a user to enable MCUboot logging from the project configuration level. Now, logging from MCUboot's sources will be printed in the same way as mbed-os logs.

Add the size of the MCU boot header area at the start of FLASH parameter - this parameter is not used in MCUboot source code directly but for example, is the argument for imgtool.py sign command which sign the binary image. It will be useful to have it in Mbed port settings to use during project building. For example post-build command in cmake which signs image.

Fix calling to BlockDevice::size before initialization - these changes fix a bug that can cause an uninitialized BlockDevice to be queried for its size. In some cases, this can return unexpected results (eg: 0). Move setting block device size after its initialization.

Signed-off-by: Artur Tynecki <artur.tynecki@mobica.com>


ATmobica:
@AGlass0fMilk @d3zd3z @utzig
I would like to ask for a review. Thank you

d3zd3z:
Just a few minor things, namely with the commit messages.

1. The first line should be of the form "tag: brief description". In this case, a good tag might be mbed. Try to keep this description fairly short (like not going past column 50-60) because it tends to be included in reports and such where it is just one column.
2. This should be followed by a blank line
3. The body of the commit text should describe the change, specifically why it is present. Try to wrap these lines at about 70 characters. Keep in mind they will sometimes be shown in other contexts, indented.
4. A blank line
5. The signed off by line. There should be a signed off by line from everyone that touches the patch.

If the code was derived from @AGlass0fMilk's earlier work, even if changed significantly, it is a good idea to give him credit in the commit text (in the body is fine). If it is enough to be considered derived, then you will need to include George's signed-off-by line in your commits as well.

ATmobica:
@d3zd3z 
I hope I understand your tips correctly and now it's ok.

ATmobica:
Is any further work on my side required, can we go forward?

d3zd3z (approved):
This looks fine from my perspective. It'd be nice to get someone else from mbed to take a look, though.

AGlass0fMilk:
I'll try to take a look.

I have changed roles and don't really work with Mbed anymore unfortunately...

I don't get a great feeling from ARM regarding future support for Mbed OS and have started using Zephyr.

PR #1302, opened 2022-02-08
single_loader: Fix hard coded buffer size in decryption function

mateo-ligne-netatmo:
Hello,
In https://github.com/mcu-tools/mcuboot/pull/1255, Buffer use to decrypt a flash section is hardcoded to a size of 1024. As far as I understood, as on some chip, erase can only be done per page (for instance on nordic nRF52840 which has flash pages of 4kB), buffer size should correspond to a page size. 
Signed-off-by: Mateo Ligne <mateo.ligne@netatmo.com>

### DIFF
--- a/boot/zephyr/single_loader.c
+++ b/boot/zephyr/single_loader.c
@@ -216,7 +216,7 @@ decrypt_region_inplace(struct boot_loader_state *state,
     uint32_t blk_sz;
     uint8_t image_index;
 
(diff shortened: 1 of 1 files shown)


d3zd3z:
We aren't likely to have stack space for this buffer, especially with larger erase sizes (they can be up to 128k).

This whole mode really needs to be rethought, since it isn't robust against power loss.

PR #1304, opened 2022-02-10
boot: serial: Adapt to Zephyr's new CRC APIs

carlescufi:
As part of https://github.com/zephyrproject-rtos/zephyr/pull/42457,
Zephyr recently changed its CRC APIs. The commit below
adapted the use of mcumgr in-tree, but MCUboot was missing the change:
https://github.com/zephyrproject-rtos/zephyr/commit/07c78e515c827c8346a758a5ce04cd56a46ae124

Note that although on other platforms the function called is
crc_ccitt(), the CRC16 actually used by MCUboot/mcumgr is:
https://reveng.sourceforge.io/crc-catalogue/16.htm#crc.cat.crc-16-xmodem


boot/boot_serial/src/boot_serial.c
```diff
@@ -589,9 +589,8 @@ boot_serial_output(void)
     bs_hdr->nh_group = htons(bs_hdr->nh_group);
 
 #ifdef __ZEPHYR__
-    crc =  crc16((uint8_t *)bs_hdr, sizeof(*bs_hdr), CRC_CITT_POLYMINAL,
-                 CRC16_INITIAL_CRC, false);
-    crc =  crc16(data, len, CRC_CITT_POLYMINAL, crc, true);
+    crc =  crc16_itu_t(CRC16_INITIAL_CRC, (uint8_t *)bs_hdr, sizeof(*bs_hdr));
+    crc =  crc16_itu_t(crc, data, len);
```
de-nordic, line 593: It looks that both the `crc16_itu_t` and `crc16_ccitt` calls are exactly the same now. Maybe we could get rid of the `#ifdef __ZEPHYR__` from the common code by providing definitions that would redirect, for Zephyr, calls from `crc16_ccitt` to `crc16_itu_t`, or some inline functions.

carlescufi: >It looks that both the crc16_itu_t and crc16_ccitt calls are exactly the same now.

In Zephyr they are not, but in other operating systems (NuttX, FreeRTOS, Mynewt) I have no idea.

nvlsianpu (approved):

de-nordic (approved):

de-nordic:
Tested on nrf52840dk_nrf5240, works ok.

utzig (approved):
LGTM

PR #1311, opened 2022-02-21
zephyr/single_loader: make decription working flash

### LINKED ISSUE #1310: zephyr: image decryption doesn't work with single-loader
mcuboot was build with `CONFIG_BOOT_SERIAL_ENCRYPT_EC256=y`, default encryption key was used  `CONFIG_BOOT_ENCRYPTION_KEY_FILE="enc-ec256-priv.pem"`, cryptolibrary was tinycrypt, tried with and without signature

Image was signed as follow:  `west sign -d build/hellow_2/ -t imgtool --  -E bootloader/mcuboot/enc-ec256-pub.pem`

target was `nrf52840dk_nrf52840`

Image is not running after was uploaded to the target.

nvlsianpu:
Decryption buffer was fixed to 1024 which imposes failure
while attempting to decrypt bigger image chunks, which is
expected on memories of bigger erase-block-size.

This patch attempt to get teh proper buffer size basing on
SOC's nv flash node.

fixes #1310

### DIFF
--- a/boot/zephyr/single_loader.c
+++ b/boot/zephyr/single_loader.c
@@ -221,9 +221,14 @@ done:
 }
 
 
+/* Get the SOC's flash erase block size from the DTS, fallback to 1024. */
+#define SOC_FLASH_ERASE_BLK_SZ \
+         DT_PROP_OR(DT_CHOSEN(zephyr_flash), erase_block_size,1024)
+
 /**
- * reads, decrypts in memory & write back the decrypted image in the same region
- * This function is NOT power failsafe since the image is decrypted in ram (stack)
+ * reads, decrypts in RAM & write back the decrypted image in the same region
+ * This function is NOT power failsafe since the image is decrypted in the RAM
+ * buffer.
  *
  * @param flash_area            The ID of the source flash area.
  * @param off_src               The offset within the flash area to
@@ -247,7 +252,7 @@ decrypt_region_inplace(struct boot_loader_state *state,
     uint32_t blk_sz;
     uint8_t image_index;
 
-    static uint8_t buf[1024] __attribute__((aligned));
+    static uint8_t buf[SOC_FLASH_ERASE_BLK_SZ] __attribute__((aligned));
     assert(sz <= sizeof buf);
 
     bytes_copied = 0;


nvlsianpu:
cc @wcappelle

wcappelle:
> cc @wcappelle

Yes, was indeed better to look it up for different erase block sizes. Most of them are 4K. 
Probably this is not the best use of the ram since only used once for a rather big buffer, but for most usecases/boards this should work afaik. I think this should also give a linker error if not enough ram, so I think this should be safe for boards where erase block sizes are too high to fit in ram.

Maybe good to check with @d3zd3z if stack is not preffered for this buffer, since then the space can be reused for other processes?

boot/zephyr/single_loader.c
```diff
@@ -221,6 +221,12 @@ read_image_size(const struct flash_area *fa_p,
...
 
+
+/* Get the SOC's flash erase block size from the DTS, fallback to 1024. */
+#define DT_DRV_COMPAT soc_nv_flash
+#define SOC_FLASH_ERASE_BLK_SZ \
+         DT_PROP_OR(DT_INST(0, soc_nv_flash), erase_block_size, 1024)
+
 /**
  * reads, decrypts in memory & write back the decrypted image in the same region
  * This function is NOT power failsafe since the image is decrypted in ram (stack)
```
wcappelle, line 232: can you also change the wrong comment of me? remove (stack)

boot/zephyr/single_loader.c
```diff
@@ -221,6 +221,12 @@ read_image_size(const struct flash_area *fa_p,
 }
 
 
+
+/* Get the SOC's flash erase block size from the DTS, fallback to 1024. */
+#define DT_DRV_COMPAT soc_nv_flash
+#define SOC_FLASH_ERASE_BLK_SZ \
+         DT_PROP_OR(DT_INST(0, soc_nv_flash), erase_block_size, 1024)
```
wcappelle, line 228: better to use the chosen "zephyr_flash" to get the flash properties?

`DT_PROP_OR(DT_CHOSEN(zephyr_flash), erase_block_size,1024)`

boot/zephyr/single_loader.c
nvlsianpu: sure

nvlsianpu:
@wcappelle BTW https://github.com/mcu-tools/mcuboot/blob/d8eff810ad2ce08c66764560a82f415ca710e3f9/boot/zephyr/single_loader.c#L354

This `#if 0` is something which looks wired to me - will register issue.

wcappelle:
> This `#if 0` is something which looks wired to me - will register issue.
This step is skipped since it was introducing some latency. This means that the image is not first validated before decryption is started. So decryption is done, and after decryption is finished, I rely on the normal boot process to validate the decrypted image. I don't think this is a bad thing since an attacker still needs to have the private key to sign the image if he wants to run this.

boot/zephyr/single_loader.c
```diff
@@ -221,9 +221,15 @@ read_image_size(const struct flash_area *fa_p,
 }
 
 
+/* Get the SOC's flash erase block size from the DTS, fallback to 1024. */
+#define DT_DRV_COMPAT soc_nv_flash
```
wcappelle, line 225: can be removed

wcappelle (approved):

PR #1313, opened 2022-02-23
workflow: Add workflow verifing integration with the upstream Zephyr

PerMac:
During the workflow a Zephyr repository is checked out (by default
main, but any version can be given when workflow is triggered
manually) and west update is run according to Zephyr's manifest.
However, the MCUboot version is then replaced by checking out
MCUboot main (triggered by push or cron), PR's head (triggered by
a PR to MCUboot) or custom SHA (manual trigger). Next, twister is
called for tests/builds in locations given in the workflow
(test_paths).

Signed-off-by: Maciej Perkowski <Maciej.Perkowski@nordicsemi.no>

### DIFF
--- a/.github/workflows/zephyr_build.yaml
+++ b/.github/workflows/zephyr_build.yaml
@@ -1,0 +1,103 @@
+# Copyright (c) 2022 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build Zephyr samples with Twister
+
+# Workflow triggers on PRs, pushes to main, once per day at midnight and can be started manually.
+on:
+  # By default, pull_request includes: opened, synchronize, or reopened
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: 0 0 * * *
+  # When triggered manually, ask for Zephyr and MCUBoot versions to check out
+  workflow_dispatch:
+    inputs:
+      version_zephyr:
+        description: 'Which Zephyr version to checkout?'
+        required: true
+        default: 'main'
+      version_mcuboot:
+        description: 'Which MCUBoot version to checkout?'
+        required: true
+        default: 'main'
+
+env:
+  ZEPHYR_VERSION: 'main'
+  MCUBOOT_VERSION: 'main'
+
+# Only cancel ongoing runs for PRs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build_zephyr_with_twister:
+    runs-on: ubuntu-latest
+    # Docker image from the zephyr upstream. Includes SDK and other required tools
+    container:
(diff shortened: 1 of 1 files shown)


PerMac:
In this version, twister is called with `--integration` flag which sets the testing scope at the test level. A test/sample will be built/executed for platforms in the corresponding testcase/sample.yaml in `integration_platforms` list. This flag also forbids skip statues and change such to errors. This is also the reason why we can see failures in this check:
- some scenarios have `platform_allow`  (which filters all platforms not included there) not including the platforms from `integration_platforms`.
This workflow correctly catches such errors. They will require a fix. `integration_platforms` entry can be added to individual scenarios (e.g here https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/mgmt/mcumgr/smp_svr/sample.yaml#L18), overwriting the one from `common` or `platform_allow` has to be expanded.

utzig:
@PerMac I suppose the plan was to squash those two commits, no?

PerMac:
Yes, I will squash them. I just realized after starting PR that it showed a problem with action "upload unit test results" which adds a comment with pass/fail/skip statistics. Apparently it requires extra steps to allow running on PRs from forked repos: https://github.com/mcu-tools/mcuboot/runs/5302376065?check_suite_focus=true#step:11:22
I haven't seen this when playing testing with this workflow on my fork, therefore I have to play here getting a proper fix. All commits will be squashed in the end

utzig:
@PerMac Btw, I would suggest you turn off all other workflows apart from the one you're adding, so no resources are spent while debugging the build issue.

nvlsianpu:
@PerMac Fix for smp_svr builds: https://github.com/zephyrproject-rtos/zephyr/pull/43155

PerMac:
All checks are green. I cleaned up commits. You can also see that the workflow this PR is adding was also executed and passed (Build samples with Twister / build_with_twister). @utzig @d3zd3z can you have a look?

.github/workflows/zephyr_build.yaml
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) 2022 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build samples with Twister
```
d3zd3z, line 4: I would suggest putting Zephyr in this line, maybe "Build Zephyr samples with Twister". It took me a little searching to find this build, as I was expecting to see Zephyr.

d3zd3z (commented):
This seems reasonable to me.

.github/workflows/zephyr_build.yaml
nvlsianpu: @PerMac I think that comment is reasonable.

nvlsianpu (approved):

.github/workflows/zephyr_build.yaml
PerMac: done

nvlsianpu:
@d3zd3z @utzig Can this go in?

d3zd3z (approved):

PR #1320, opened 2022-03-03
boot_serial: Fix buffer overflow in boot_serial_out

de-nordic:
The buf buffer set to collect total console payload consisting
of total size (two bytes), SMP header (eigth bytes), data payload
(bs_obuf, BOOT_SERIAL_OUT_MAX) and CRC (two bytes), pior to base64
encoding has been set to size of BOOT_SERIAL_OUT_MAX.
This means that if output data len, in bs_obuf, would be longer than
BOOT_SERIAL_OUT_MAX - 8 - 2 - 2, then composing of the output buffer
would overflow.


d3zd3z (approved):

de-nordic:
@nvlsianpu Can you take a look?

nvlsianpu (approved):

PR #1324, opened 2022-03-10
zephyr: add dropped sys_clock_disable definition

kvaellning:
Missing extern definition will result in a warning during build.
Dropped in e9c6b4d7af7598fbd915868cf89383ff854d2241.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -159,6 +159,8 @@ struct arm_vector_table {
     uint32_t reset;
 };
 
+extern void sys_clock_disable(void);
+
 static void do_boot(struct boot_rsp *rsp)
 {
     struct arm_vector_table *vt;


boot/zephyr/main.c
```diff
@@ -159,6 +159,8 @@ struct arm_vector_table {
     uint32_t reset;
 };
 
+extern void sys_clock_disable(void);
```
d3zd3z, line 162: I think we should `#include <drivers/timer/system_timer.h>` rather than having this extern ourselves. This way, the function will correctly be stubbed out on configurations that do not have this function.

d3zd3z (changes requested):

boot/zephyr/main.c
kvaellning: Unfortunately, `sys_clock_disable` is not part of the definition (at least not in Zephyr 2.7,0 and 2.7.1).
Is this version intended to run on Zephyr >= 3.0.0? If so, I'll drop the pull request.

nvlsianpu: @kvaellning Mcuboot is expected to run with zephyr more less of the same time.
Exact version is part of the zephyr manifest.

PR #1329, opened 2022-03-15
imgtool: Added support for providing the signature by 3rd party

nvlsianpu:
The sign command was extended so it now allow to provide the signature
as base64 formatted RAW file using --fix-sig along with relevant public key
--fix-sig-pubkey.

This patch is added for support the case where the party which produces
the image dose not have access to the signing image key but must request
third party for the signature.

Possible steps for get an image sign are:

1. Generate the public key form the private key using openssl:
``` Console
openssl rsa -in ./bootloader/mcuboot/root-rsa-2048.pem -pubout -out root-rsa-2048-priv.pem -outform PEM -RSAPublicKey_out
```

2. Get the payload against to which the  signature should be generated:
``` console
./bootloader/mcuboot/scripts/imgtool.py sign -v 1.0.0 -H 0x200 -S 0x67000 --align 8  --fix-sig-pubkey ./root-rsa-2048-priv.pem --vector-to-sign payload  ./build/hellow/zephyr/zephyr.hex data_to_sign.bin
```
Worth to notice that all but ed25519 signature are generated using the payload. For ed25519 signature are generated basing only on the image digest (so need to use `--vector-to-sign digest`)

3. Generate the signature and save it as base64 string
``` console
openssl dgst -sha256  -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:32 -sigopt rsa_mgf1_md:sha256 -sign ./bootloader/mcuboot/root-rsa-2048.pem -out openssl_sig.sig data_to_sig.bin

openssl base64 -in openssl_sig.sig -out openssl_sig_base64.sig
```

4. Add the the signature generated remotely to the image.
``` console
./bootloader/mcuboot/scripts/imgtool.py sign -v 1.0.0  -H 0x200 -S 0x67000 --align 8 --fix-sig openssl_sig_base64.sig --fix-sig-pubkey ./root-rsa-2048-priv.pem --public-key-format full  ./build/hellow/zephyr/zephyr.hex signed_app_image.hex
```

------
------

Additionally this patch introduces a way for exporting the signature while signing the image in the standard manner.

It can b exported as fallow
``` console
 ./bootloader/mcuboot/scripts/imgtool.py sign -k ./bootloader/mcuboot/root-rsa-2048.pem -v 1.0.0 -H 0x200 -S 0x67000 --align 8 --public-key-format full --sig-out out_sig.sig ./build/hellow/zephyr/zephyr.hex normal_signed.hex```

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -305,12 +305,17 @@ class Image():
         return cipherkey, ciphermac, pubk
 
     def create(self, key, public_key_format, enckey, dependencies=None,
-               sw_type=None, custom_tlvs=None, encrypt_keylen=128, clear=False):
+               sw_type=None, custom_tlvs=None, encrypt_keylen=128, clear=False, fixed_sig=None, pub_key=None):
         self.enckey = enckey
 
         # Calculate the hash of the public key
         if key is not None:
             pub = key.get_public_bytes()
+            sha = hashlib.sha256()
+            sha.update(pub)
+            pubbytes = sha.digest()
+        elif pub_key is not None:
+            pub = pub_key.get_public_bytes()
             sha = hashlib.sha256()
             sha.update(pub)
             pubbytes = sha.digest()
@@ -428,7 +433,7 @@ class Image():
 
         tlv.add('SHA256', digest)
 
-        if key is not None:
+        if key is not None and fixed_sig is None:
             if public_key_format == 'hash':
                 tlv.add('KEYHASH', pubbytes)
             else:
@@ -442,6 +447,14 @@ class Image():
             else:
                 sig = key.sign_digest(digest)
             tlv.add(key.sig_tlv(), sig)
+        elif fixed_sig is not None and key is None:
+            if public_key_format == 'hash':
+                tlv.add('KEYHASH', pubbytes)
+            else:
+                tlv.add('PUBKEY', pub)
+            tlv.add(pub_key.sig_tlv(), fixed_sig['value'])
+        else:
+            raise click.UsageError("Can not sign using key and provide fixed-signature at the same time")
 
         # At this point the image was hashed + signed, we can remove the
         # protected TLVs from the payload (will be re-added later)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/imgtool/image.py [hunk 1]
```diff
@@ -442,6 +442,10 @@ def create(self, key, public_key_format, enckey, dependencies=None,
             else:
                 sig = key.sign_digest(digest)
             tlv.add(key.sig_tlv(), sig)
+        elif fixed_sig is not None and key is None:
```
nvlsianpu, line 445: according to bootutil_img_validate() a TLV with public-key/public-key-has is aslo required.

hellesvik-nordic:
I created a [keygen_detached.sh](https://github.com/hellesvik-nordic/samples_for_nrf_connect_sdk/blob/main/bootloader_samples/mcuboot_detached_signature/keygen_detached.sh) to test this PR.
However, it did not work. Did I miss anything?
For reference, I tested signing an image directly, the old way in [keygen_normal.sh](https://github.com/hellesvik-nordic/samples_for_nrf_connect_sdk/blob/main/bootloader_samples/mcuboot_detached_signature/keygen_normal.sh), and it works as expected.
Both scripts are part of a nRF Connect SDK sample which should work standalone (as long as path to this PR in the scripts are correct)

hellesvik-nordic:
Making each image use the default key format fixed it. (Removed `--public-key-format full` from line 15 of [keygen_detached.sh](https://github.com/hellesvik-nordic/samples_for_nrf_connect_sdk/blob/main/bootloader_samples/mcuboot_detached_signature/keygen_detached.sh).)

scripts/imgtool/main.py
```diff
@@ -356,10 +379,32 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
         else:
             custom_tlvs[tag] = value.encode('utf-8')
 
+    #allow signature calculated externally
```
d3zd3z, line 382: Minor nit, there should be a space after the '#', the Allow capitalized, and a period at the end. See [PEP-8](https://peps.python.org/pep-0008/#comments).

d3zd3z (changes requested):

scripts/imgtool/main.py
nvlsianpu: fixed

d3zd3z (approved):

PR #1332, opened 2022-03-23
boot: zephyr: Add MCUboot status callback support

lairdjm:
Adds an optional callback when the MCUboot status changes which can
allow components to react.


lairdjm:
Might need more handling for the DFU checking/timing out in the other files where this is handled

boot/bootutil/src/loader.c
```diff
@@ -1998,6 +2002,10 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
     }
 #endif
 
+#if defined(CONFIG_MCUBOOT_ACTION_HOOKS)
```
d3zd3z, line 2005: I think it would be a lot cleaner if, instead of an ifdef around every call of the status change, we made the function always defined, but when the feature is not defined, make it an inline function with an empty body (and the incantation to ignore the argument).

boot/zephyr/include/mcuboot_status.h
```diff
@@ -0,0 +1,30 @@
...
+	MCUBOOT_STATUS_BOOTABLE_IMAGE_FOUND,
+	MCUBOOT_STATUS_NO_BOOTABLE_IMAGE_FOUND,
+	MCUBOOT_STATUS_BOOT_FAILED,
+	MCUBOOT_STATUS_USB_DFU_WAITING,
+	MCUBOOT_STATUS_USB_DFU_ENTERED,
+	MCUBOOT_STATUS_USB_DFU_TIMED_OUT,
+	MCUBOOT_STATUS_SERIAL_DFU_ENTERED,
+} mcuboot_status_type_t;
+
+extern void mcuboot_status_change(mcuboot_status_type_t status);
```
d3zd3z, line 26: So, here, have an #else, and something like:
```c
#else
static inline void mcuboot_status_change(mcuboot_status_type_t status)
{
   (void)status;
}
```

and move the `#ifdef` down so that the enum is always defined.

boot/zephyr/main.c
```diff
@@ -535,11 +561,16 @@ void main(void)
         /* at least one check if time was expired */
         timeout_in_ms = 1;
     }
-   boot_serial_check_start(&boot_funcs,timeout_in_ms);
+    boot_serial_check_start(&boot_funcs,timeout_in_ms);
```
d3zd3z, line 564: This is a legitimate fix, but probably should be in a separate commit.

d3zd3z (changes requested):

boot/zephyr/main.c
lairdjm: Moved to separate commit

lairdjm:
Is there anywhere else this would need to be applied to that is outside of the main.c file?

boot/bootutil/src/loader.c
```diff
@@ -55,6 +55,10 @@
 #include <os/os_malloc.h>
 #endif
 
+#if defined(CONFIG_MCUBOOT_ACTION_HOOKS)
```
d3zd3z, line 58: This use of `mcuboot_status.h` also should no longer need ifdefs. Just include it, and call `mcuboot_status_change` below, like done in main.

lairdjm: Fixed

boot/bootutil/include/bootutil/mcuboot_status.h
```diff
@@ -0,0 +1,30 @@
...
+	MCUBOOT_STATUS_USB_DFU_WAITING,
+	MCUBOOT_STATUS_USB_DFU_ENTERED,
+	MCUBOOT_STATUS_USB_DFU_TIMED_OUT,
+	MCUBOOT_STATUS_SERIAL_DFU_ENTERED,
+} mcuboot_status_type_t;
+
+#if defined(CONFIG_MCUBOOT_ACTION_HOOKS)
+extern void mcuboot_status_change(mcuboot_status_type_t status);
+#else
+#define mcuboot_status_change(status)
```
d3zd3z, line 27: It would be better to define this as:
```c
#define mcuboot_status_change(_status) do {} while 0
```
instead of just as an empty define. That way it will always be usable where a function would be, and people won't be surprised by spurious errors from the stray ';' that otherwise would be there.

d3zd3z (changes requested):
Just a minor little point. Otherwise looks good.

boot/bootutil/include/bootutil/mcuboot_status.h
lairdjm: Fixed

d3zd3z (approved):

PR #1334, opened 2022-03-28
boot: zephyr: Disable CONFIG_USB_DEVICE_REMOTE_WAKEUP

MarekPieta:
Change disables CONFIG_USB_DEVICE_REMOTE_WAKEUP, because it is not supported by the MCUboot bootloader source code.


MarekPieta:
Hi @jfischer-no
I think we need a follow-up change here too.

nvlsianpu (approved):

MarekPieta:
@nvlsianpu, can we merge it?

PR #1335, opened 2022-03-31
boot_serial: Upgrade from cddl-gen 0.1.0 to zcbor 0.4.0

oyvindronningstad:
cddl-gen has been renamed to zcbor.
Update regenerate_serial_recovery_cbor.sh and regenerate/recopy all
files.

Remove the submodule in ext/ since it is no longer necessary when
the zcbor package is installed (only needed for regeneration, not
for building).

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>


de-nordic:
Tested with echo and image upload, on nrf52840dk, works fine

d3zd3z (approved):

PR #1326, opened 2022-03-10
espressif:esp32: Add multi image support

almir-okato:
Changes on configuration and flash area organization for supporting multi image and implementation for booting on different processors on ESP32.

PR #1339, opened 2022-04-01
[RFC] MCUboot's multi image and boot sequence with multi processors (multi boot)

almir-okato:
**MCUboot's multi image and boot sequence with multi processors (multi boot)**

MCUboot bootloader is designed to boot a single image after `context_boot_go` verification is performed. This call returns the boot information exclusively for the referred single image that is required by `do_boot`, that loads it (if needed) and jumps to the entry point of the image. This happens even using multi images, where MCUboot can manage, however does not boot a second image, leaving this responsibility to the first application image.

MCUBoot Espressif's port can now handle multi image booting (PR https://github.com/mcu-tools/mcuboot/pull/1326), which allows both ESP32 CPUs (APP and PRO) to run different applications.

This RFC is intended to start a discussion to add multi boot feature (name open to suggestions) in order to handle both multi images and boot sequence if supported by the SoC.

The goal is:
- Transfer from OS to the bootloader the responsibility of starting the second image/processor.
- Have better support for AMP architecture, where images can run in stand-alone format.
- If chip design allows, separate/protect critical resources between the OSes before they start.

Possible drawbacks:
- Increase of boot time and size.
- More image and resources manual managing.

This PR presents an initial proposal of implementation:

a) Add a configuration for multi boot to `mcuboot_config.h`
```
#define MCUBOOT_MULTI_BOOT 1
```
b) Add information about the image index and the CPU index to `boot_rsp`.
```
struct boot_rsp {
    /** A pointer to the header of the image to be executed. */
    const struct image_header *br_hdr;

    /**
     * The flash offset of the image to execute.  Indicates the position of
     * the image header within its flash device.
     */
    uint8_t br_flash_dev_id;
    uint32_t br_image_off;

#ifdef MCUBOOT_MULTI_BOOT
    /**
     * Indicates in which CPU the image must be executed.
     */
    uint8_t img_index;
    uint8_t boot_cpu_index;
#endif
};
```
c) Add TLV to the image for indicating in which CPU it shall run.
```
#define IMAGE_TLV_BOOT_CPU_INDEX    0x70   /* CPU in which the image will be executed */
```
d) Add implementation to retrieve the boot information from other images. It would be used by the bootloader port if it supports the multi boot, so it can boot the referred image on the correct CPU.
```
fih_int boot_rsp_by_image_id(struct boot_rsp *rsp, uint32_t image_id);
```
As an alternative, `boot_go` could return the info instead of creating another function, however it would need changes on the APIs and its implementations (like in `context_boot_go` and `fill_boot_rsp`), which may not be desirable.

e) Each port would handle how to start the image.

What are your thoughts about that?

Thanks!

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil.h
+++ b/boot/bootutil/include/bootutil/bootutil.h
@@ -59,6 +59,14 @@ struct boot_rsp {
      */
     uint8_t br_flash_dev_id;
     uint32_t br_image_off;
+
+#ifdef MCUBOOT_MULTI_BOOT
+    /**
+     * Indicates in which CPU the image must be executed.
+     */
+    uint8_t img_index;
+    uint8_t boot_cpu_index;
+#endif
 };
 
 /* This is not actually used by mcuboot's code but can be used by apps
@@ -80,6 +88,8 @@ struct image_trailer {
 /* you must have pre-allocated all the entries within this structure */
 fih_int boot_go(struct boot_rsp *rsp);
 fih_int boot_go_for_image_id(struct boot_rsp *rsp, uint32_t image_id);
+
+fih_int boot_rsp_by_image_id(struct boot_rsp *rsp, uint32_t image_id);
 
 struct boot_loader_state;
 void boot_state_clear(struct boot_loader_state *state);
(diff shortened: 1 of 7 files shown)


almir-okato:
@d3zd3z @utzig 
Could you please take a look on this proposal?

d3zd3z:
MCUboot has been used for multiple images with separate CPUs (see the Cypress port).

In general, the assumption we make is that we want to keep as much complexity as possible out of the bootloader. Generally, the bootloader is something that cannot be upgraded, and we'd like as few possible opportunities for bugs to show up in it. As such, there will be a primary image that will be considered trusted enough to perform the rest of the initialization that is needed, including memory protection, etc. Presumably, with a multi-CPU setup, the first CPU will still be responsible for setting things up, and causing the second CPU to start. When you think about it, this isn't all that different than something like TF-M where the secure image could be thought of as the first CPU, and the non-secure as the second CPU.

But, I'm open to arguments as to why this complexity should be in the bootloader instead.

PR #1344, opened 2022-04-09
zephyr/main: configure DFU detection GPIO-pin using DTS

### LINKED ISSUE #965: boot/zephyr/main: swich gpio pin configuration to DT
GPIO PIN for serial recovery is configured using Kconfig.
While ago practical reasons determined that better to configure gpio PIN using DTS.

- Configure PIN using new `bootloader-button0` PIN alias.
  - If `bootloader-button0` not exist configure PIN using the `sw0` PIN alias (common in zephyr).

nvlsianpu:
Added support for configure GPIO-pin used for triggering the
recovery DFU using DTS. That pin is assigned thanks to
`mcuboot-button0` pin-allias.

DTS configuration takes precedence over the Kconfig pin configurations.

Kconfig hw pin properties get deprecated.

This patch reduces technical debt against zephyr-rtos we have.

fixes #965

~~[PR for introduce bootloader-button0 alias to the zephyr-rtos](https://github.com/zephyrproject-rtos/zephyr/pull/44741)~~ [Merged]

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -373,13 +373,27 @@ void zephyr_boot_log_stop(void)
 #if defined(CONFIG_MCUBOOT_SERIAL) || defined(CONFIG_BOOT_USB_DFU_GPIO)
 
 #ifdef CONFIG_MCUBOOT_SERIAL
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_SERIAL_DETECT_DELAY
+#else
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_USB_DFU_DETECT_DELAY
+#endif
+
+
+#define BUTTON_0_NODE DT_ALIAS(mcuboot_button0)
+
+#if DT_NODE_EXISTS(BUTTON_0_NODE) && DT_NODE_HAS_PROP(BUTTON_0_NODE, gpios)
+
+static const struct gpio_dt_spec button0 = GPIO_DT_SPEC_GET(BUTTON_0_NODE, gpios);
+
+#else /* fallback to legacy configuration */
+
+#if defined(CONFIG_MCUBOOT_SERIAL)
 
 #define BUTTON_0_GPIO_LABEL CONFIG_BOOT_SERIAL_DETECT_PORT
 #define BUTTON_0_GPIO_PIN CONFIG_BOOT_SERIAL_DETECT_PIN
 #define BUTTON_0_GPIO_FLAGS ((CONFIG_BOOT_SERIAL_DETECT_PIN_VAL) ?\
                                 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN) :\
                                 (GPIO_ACTIVE_LOW | GPIO_PULL_UP))
-#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_SERIAL_DETECT_DELAY
 
 #elif defined(CONFIG_BOOT_USB_DFU_GPIO)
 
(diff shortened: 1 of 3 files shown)


boot/zephyr/main.c
```diff
@@ -371,30 +372,62 @@ void zephyr_boot_log_stop(void)
...
-static bool detect_pin(const char* port, int pin, uint32_t expected, int delay)
+
+#ifdef CONFIG_MCUBOOT_SERIAL
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_SERIAL_DETECT_DELAY
+#else
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_USB_DFU_DETECT_DELAY
+#endif
+
+
+#define BUTTON_0_NODE DT_ALIAS(bootloader_button0)
```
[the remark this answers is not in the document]
mbolivar-nordic: > This is kind of a generic name. I am surprised the alias doesn't have "mcuboot" in the name.

This relates to a discussion we're having about configuring software specific features in devicetree. (See zephyr-devel; https://lists.zephyrproject.org/g/devel/message/8456). 

I suppose mcuboot could do the same thing, maybe with an `/mcuboot` node with compatible `mcuboot-configuration`. Then you could start putting whatever you want in there, not just GPIOs. Another option would be perhaps to use the `/options` node that @sjg20 mentioned in this response, since that's used by u-boot, with a quite similar use case: https://lists.zephyrproject.org/g/devel/message/8460

d3zd3z:
Looks like Marti's comments are worth addressing.

boot/zephyr/main.c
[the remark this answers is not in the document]
nvlsianpu: Appling your request will mean that It will be not a DT alias anymore.

mbolivar-nordic:
Discussed offline with @nvlsianpu. Agreed to stick with aliases that are prefixed with `mcuboot-` for now.

If a large number of these aliases start to accumulate, it may be worthwhile to create a dedicated node for configuring MCUboot itself. However, @nvlsianpu thinks this is unlikely, and I agree it makes sense to defer that until there is a real need.

nvlsianpu:
>> Looks like Marti's comments are worth addressing.

> Discussed offline with @nvlsianpu. Agreed to stick with aliases that are prefixed with mcuboot- for now.

> If a large number of these aliases start to accumulate, it may be worthwhile to create a dedicated node for configuring MCUboot itself. However, @nvlsianpu thinks this is unlikely, and I agree it makes sense to defer that until there is a real need.

@d3zd3z Can you revisit this?

mbolivar-nordic (approved):
Nice!

d3zd3z (approved):

nvlsianpu:
@anangl Can you recheck?

boot/zephyr/main.c
```diff
@@ -371,30 +371,70 @@ void zephyr_boot_log_stop(void)
...
-    int detect_value;
-    struct device const *detect_port;
+    int pin_active;
 
-    detect_port = device_get_binding(port);
-    __ASSERT(detect_port, "Error: Bad port for boot detection.\n");
+#ifdef BUTTON_0_LEGACY
+    button0.port = device_get_binding(BUTTON_0_GPIO_LABEL);
+    __ASSERT(button0.port, "Error: Bad port for boot detection.\n");
+#endif
```
anangl, line 426: For non-legacy setup, `device_is_ready(button0.port)` should be called before `gpio_pin_configure_dt(&button0, ...)`.

boot/zephyr/main.c
```diff
@@ -371,30 +371,70 @@ void zephyr_boot_log_stop(void)
...
+#ifdef CONFIG_MCUBOOT_SERIAL
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_SERIAL_DETECT_DELAY
+#else
+#define BUTTON_0_DETECT_DELAY CONFIG_BOOT_USB_DFU_DETECT_DELAY
+#endif
+
+
+#define BUTTON_0_NODE DT_ALIAS(mcuboot_button0)
+
+#if DT_NODE_HAS_STATUS(BUTTON_0_NODE, okay) && DT_NODE_HAS_PROP(BUTTON_0_NODE, gpios)
```
anangl, line 384: Why not just use `DT_NODE_EXISTS`? Status is not usually checked (nor set) for button and LED nodes. Sure, the above will work fine, but it looks inconsistent with other pieces of code that use a button.

```suggestion
#if DT_NODE_EXISTS(BUTTON_0_NODE) && DT_NODE_HAS_PROP(BUTTON_0_NODE, gpios)
```

boot/zephyr/main.c
nvlsianpu: fixed

boot/zephyr/main.c
nvlsianpu: fixed

anangl (approved):

PR #1346, opened 2022-04-12
Add downgrade prevention for swaps

kasjer:
Currently, downgrade prevention was limited to overwrite only
builds (version check) or devices with hardware storage for
security counter.

This extends downgrade prevention to be used when swap update
is selected.
Unlike MCUBOOT_HW_ROLLBACK_PROT option it does not require user
code to provide external way to store security counter.
Security counter from slot 1 image is used for comparison.
With security counter usage it is possible to have limited
software rollback if security counter was not incremented.

It is possible to use image version where strict rule for
image version comparison prevents any downgrades.

Downgrade prevention is also added to mynewt configuration.

If image in slot 1 is marked as pending and downgrade prevention
is in place, image will be deleted to avoid check on next boot.


boot/bootutil/src/loader.c
```diff
@@ -1902,6 +1902,43 @@ boot_update_hw_rollback_protection(struct boot_loader_state *state)
...
+static int
+check_downgrade_prevention(struct boot_loader_state *state)
+{
+#if defined(MCUBOOT_DOWNGRADE_PREVENTION) && (defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_SCRATCH))
+    uint32_t security_counter[2];
+    int rc;
+
+    if (MCUBOOT_DOWNGRADE_PREVENTION_SECURITY_COUNTER) {
+        /* If there was security no counter in slot 0, allow swap */
+        rc = bootutil_get_img_security_cnt(&(BOOT_IMG(state, 0).hdr), BOOT_IMG(state, 0).area, &security_counter[0]);
```
d3zd3z, line 1914: Please try to keep lines to 80-90 characters. These lines, including the #if`` above should be split across two or more lines.

boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h
```diff
@@ -85,6 +85,16 @@
 #if MYNEWT_VAL(BOOTUTIL_BOOTSTRAP)
 #define MCUBOOT_BOOTSTRAP 1
 #endif
+#if MYNEWT_VAL_CHOICE(BOOTUTIL_DOWNGRADE_PREVENTION, version)
+#define MCUBOOT_DOWNGRADE_PREVENTION                     1
```
d3zd3z, line 89: Maybe a comment here. It is not clear that `MCUBOOT_DOWNGRADE_PREVENTION` is used in a `#if defined` construct, whereas `MCUBOOT_DOWNGRADE_PREVENTION_SECURITY_COUNTER` is used as a boolean.

d3zd3z (commented):
It would be nice to also have support in the Zephyr kconfig files.

boot/bootutil/src/loader.c
kasjer: lines length limited as requested

boot/mynewt/mcuboot_config/include/mcuboot_config/mcuboot_config.h
kasjer: comment added

kasjer:
> It would be nice to also have support in the Zephyr kconfig files.

zephyr Kconfig updated as requested

d3zd3z (approved):

sjanc:
gentle ping :)

PR #1349, opened 2022-04-13
boot: Check that hash TLV is sufficiently large

d3zd3z:
Add a check for the size of the TLV entry for the boot record.
Although, this is in the protected area, and should be protected by a
signature, check that we don't induce an integer underflow if that entry
is somehow too short.


utzig (approved):
LGTM

nvlsianpu (approved):

PR #1351, opened 2022-04-18
X509 prototype

d3zd3z:
This is a snapshot of my x509 work from last year. It doesn't merge, doesn't very that the signature matches the one in the cert, and possibly has numerous other problems.


.github/workflows/sim.yaml
```diff
@@ -12,19 +12,24 @@ jobs:
     strategy:
       matrix:
         features:
-        - "sig-ecdsa,sig-ed25519,enc-kw,bootstrap"
+        - "sig-ecdsa,sig-ecdsa-mbedtls,sig-ed25519,enc-kw,bootstrap"
```
prayassamriya, line 15: @d3zd3z we are having difficulty understanding some pieces and this is blocking us in making real progress on top of your work. Is this possible to have a meeting and get our doubts cleared. we are very much interested in adding this support but need some guidance to proceed further.

d3zd3z: Sorry I missed seeing this. If you're still interested, let me know and we can find a time to talk. Are you on Slack?

prayassamriya: no problem. actually made a good progress later and able to achieve x509 cert based code signing. give us few days and will submit the pull request.

prayassamriya: @d3zd3z we want to raise the pull request for additional work we just finished. can you suggest how do we submit the same? should we create a fresh pull request from our repo having the latest mcuboot code, or do you want us to commit on your branch itself?

PR #1352, opened 2022-04-26
zephyr: Add a bool Kconfig option for swap move

d3zd3z:
The Kconfig system used by Zephyr does not allow the defaults for choice
options to be overridden.  To compensate for this, create a new boolean
config option that will determine what the default is for the boot mode.

This allows the kconfig override file for various Zephyr boards to
change the default to swap move.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -194,10 +194,21 @@ config BOOT_VALIDATE_SLOT0_ONCE
 	  low end devices with as a compromise lowering the security level.
 	  If unsure, leave at the default value.
 
+config BOOT_PREFER_SWAP_MOVE
+	bool "Prefer the newer swap move algorithm"
+	default y if SOC_FAMILY_NRF
+	default n
+	help
+	  If y, the BOOT_IMAGE_UPGRADE_MODE will default to using
+	  "move" instead of "scratch".  This is a separate bool config
+	  option, because Kconfig doesn't allow defaults to be
+	  overridden in choice options.  Most devices should be using
+	  swap move.
+
 if !SINGLE_APPLICATION_SLOT
 choice BOOT_IMAGE_UPGRADE_MODE
 	prompt "Image upgrade modes"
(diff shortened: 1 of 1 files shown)


d3zd3z:
Making this a draft for additional integration testing with Zephyr.

d3zd3z:
I think this should be benign until other changes are made in Zephyr, so it'd be nice to get this in, so we can work toward getting Zephyr transitioned to using it.

d3zd3z:
> For named choice (like that one you modified) you can modify it using additional Kconfig.

I'd be open to suggestions on how to make this work, but what you've suggested definitely doesn't work when used in a Kconfig.defconfig file. It is stated in the Kconfig documentation that this is a limitation.

nvlsianpu (approved):

boot/zephyr/Kconfig
```diff
@@ -194,10 +194,21 @@ config BOOT_VALIDATE_SLOT0_ONCE
 	  low end devices with as a compromise lowering the security level.
 	  If unsure, leave at the default value.
 
+config BOOT_PREFER_SWAP_MOVE
```
nvlsianpu, line 197: I would name this option with prefix MCUBOOT instead of BOOT - reason is that this will be used from boards defconfig files in zephyr, this would be right visible that it is for MCUBOOT.

nvlsianpu: @d3zd3z  One nit ^^

d3zd3z: Hmm, it makes it different than the rest of the options, but I can certainly do it that way if that makes it clearer.

d3zd3z: Thinking more about this. I don't think it makes sense to add an MCUBOOT prefix to this, unless we change all of the rest of the options as well. This is the Zephyr config file, and any of these will be showing up in board defconfigs.

nvlsianpu: okay, I won't insist

nvlsianpu (approved):

PR #1359, opened 2022-05-05
 espressif: Add warning for unsupported chip revision

almir-okato:
This MR adds checking and warning for ESP32, ESP32-S2, ESP32-C3, ESP32-S3 unsupported chip revisions on their initialization. 
Also it creates bootloader.conf for each chip and made respectively changes for build system and documentation.
Besides that, some common functions among the chips have been grouped.


almir-okato:
@gustavonihei @utzig PTAL

gustavonihei (approved):

PR #1362, opened 2022-05-09
espressif:esp32: Move app entry point call back to iram_loader_seg region

almir-okato:
Entry point call was moved back from main to esp_loader, so it is called from iram_loader_seg memory region.


almir-okato:
@gustavonihei @utzig PTAL

gustavonihei (approved):
LGTM

PR #1363, opened 2022-05-09
imgtool: Fix imgtool sign command without key

almir-okato:
Using imgtool sign command without a key was resulting the following error:
`Error: Can not sign using key and provide fixed-signature at the same time`
Neither a key nor fixed signature were being used.

The fix adds a condition that checks if either `key` or `fixed_sig` are `not None` before payload signing and TLV addition.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -435,38 +435,39 @@ class Image():
 
         tlv.add('SHA256', digest)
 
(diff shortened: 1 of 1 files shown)


almir-okato:
@gustavonihei @d3zd3z @nvlsianpu PTAL

nvlsianpu (approved):

PR #1372, opened 2022-05-21
boot: zephyr: add Kconfig for arm cortex-m that implements a cache

XenuIsWatching:
The Cache is an optional configuration of both the ARM Cortex-M7 and
Cortex-M55. Previously, it was just checking that it was just an M7
rather than knowing that the CPU actually was built with the cache.

This just follows the pull request into zephyr https://github.com/zephyrproject-rtos/zephyr/pull/44815

Signed-off-by: Ryan McClelland ryanmcclelland@fb.com


utzig (approved):
LGTM

PR #1382, opened 2022-06-09
boot_serial: zephyr: Use snprintf to format version string

de-nordic:
Move formatting of version string to use snprintf, which Zephyr
provides, instead defining utility function for that purpose.


boot/boot_serial/src/boot_serial.c
```diff
@@ -124,6 +124,10 @@ extern int bs_peruser_system_specific(const struct nmgr_hdr *hdr,
                                       const char *buffer,
                                       int len, zcbor_state_t *cs);
 
+#define zcbor_tstr_put_lit_cast(state, string) \
+	zcbor_tstr_encode_ptr(state, (uint8_t *)string, sizeof(string) - 1)
+
+#ifndef __ZEPHYR__
```
nvlsianpu, line 130: I would make this optional depends on additional mediate definition check for existence of `snprintf()` (or stdio.h). Thanks to that others os-platform might use this code as well.

```suggestion
#ifdef __ZEPHYR__
#define __SNPRINTF_EXISTS__ 1
#endif
#ifndef __SNPRINTF_EXISTS__
```

nvlsianpu (changes requested):

boot/boot_serial/src/boot_serial.c
de-nordic: Decided to define `MCUBOOT_USE_SNPRINTF` configuration identifier for Zephyr, to follow how other parts of code are configured.

PR #1385, opened 2022-06-13
boot: mynewt: Add boot data sharing config

kasjer:
There is code for sharing data between booloader and application.
So far only zephyr had config flags to enable this.

This adds configuration that allows to turn on data sharing code
in mynewt version.


utzig (approved):
LGTM

PR #1386, opened 2022-06-13
mynewt: Allow custom boot to be provided by pkg

kasjer:
BOOT_CUSTOM_START is defined by APP package (here).
This changes default value from 0 to empty
to allow for constructing packege that provides function
boot_custom_start() and automatically sets BOOT_CUSTOM_START to 1
for easy setup.


utzig (approved):
LGTM

PR #1387, opened 2022-06-13
boot_serial: Upload confirmation stalling

de-nordic:
The PR consists of two commits:
 1) Adding support for returning MGMT_ERR_EBUSY SMP code and supporting code; also adding busy flag checking around SMP reset command to prevent reset to be performed when SMP processing is in busy state;
 2) Adding stalling mechanism into SMP image processing that will re-request few of last image bytes for specific timeout to allow background task, for example performed by other CPU, to complete.


d3zd3z (changes requested):

boot/boot_serial/src/boot_serial.c
```diff
@@ -468,25 +505,27 @@ bs_upload(char *buf, int len)
...
+        flash_area_close(fap);
+    }
     BOOT_LOG_INF("RX: 0x%x", rc);
     zcbor_map_start_encode(cbor_state, 10);
     zcbor_tstr_put_lit_cast(cbor_state, "rc");
     zcbor_uint32_put(cbor_state, rc);
     if (rc == 0) {
         zcbor_tstr_put_lit_cast(cbor_state, "off");
-        zcbor_uint32_put(cbor_state, curr_off);
+	zcbor_uint32_put(cbor_state, next_offset(curr_off, img_size));
```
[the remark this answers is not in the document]
de-nordic: Sorry for that. Wrong vimrc used.

nvlsianpu (commented):
This PR will need some refactoring after https://github.com/mcu-tools/mcuboot/pull/1395 will gets in.

boot/boot_serial/src/boot_serial.c
```diff
@@ -267,6 +271,31 @@ bs_list(char *buf, int len)
     boot_serial_output();
 }
 
+static uint32_t next_offset(uint32_t curr_off, uint32_t img_size)
```
nvlsianpu, line 274: I would like to have function for checking the stalling timeout explicitly called in bs_uploda().
I think calling that while constructing the response is a forbidden black magic.

de-nordic:
> This PR will need some refactoring after #1395 will gets in.

Yes, unfortunately.

boot/boot_serial/src/boot_serial.c
```diff
@@ -468,25 +507,27 @@ bs_upload(char *buf, int len)
...
+	zcbor_uint32_put(cbor_state, next_offset(curr_off, img_size));
     }
     zcbor_map_end_encode(cbor_state, 10);
 
-    boot_serial_output();
-    flash_area_close(fap);
-
 #ifdef MCUBOOT_ENC_IMAGES
-    if (curr_off == img_size) {
+    if (!img_upload_stalling && curr_off == img_size) {
```
nvlsianpu, line 524: why to stall decryption at all?

nvlsianpu (changes requested):
This feature conception looks fine to me. PR needs some more work.

boot/boot_serial/src/boot_serial.c
```diff
@@ -267,6 +271,31 @@ bs_list(char *buf, int len)
...
+            img_upload_stalling = true;
+            img_upload_stalling_timeout =
+                k_uptime_get_32() + MCUBOOT_BOOT_SERIAL_IMG_UPLOAD_STALLING_TIMEOUT
+            curr_off -= 4;
+        }
+    } else {
+        if (img_upload_stalling_timeout <= k_uptime_get_32()) {
+            img_upload_stalling = false;
+        } else {
+            curr_off -= 4;
```
nvlsianpu, line 292:
```suggestion
            img_size -= 4;
```

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -231,6 +231,12 @@
 #define MCUBOOT_MAX_IMG_SECTORS       128
 #endif
 
+#ifdef CONFIG_BOOT_SERIAL_IMG_UPLOAD_STALLING
+#define MCUBOOT_BOOT_SERIAL_IMG_UPLOAD_STALLING 1
+#define MCUBOOT_BOOT_SERIAL_IMG_UPLOAD_STALLING_TIMEOUT	\
+	(CONFIG_BOOT_SERIAL_IMG_UPLOAD_STALLING_TIMEOUT);
```
d3zd3z, line 237: Why does this macro have a semicolon after it? It makes the use in the code strange, because there is no semicolon (which messes up editor indenting and stuff).

boot/boot_serial/src/boot_serial.c
```diff
@@ -313,6 +317,31 @@ static off_t erase_range(const struct flash_area *fap, off_t start, off_t end)
...
+#ifndef MCUBOOT_BOOT_SERIAL_IMG_UPLOAD_STALLING
+    (void)img_size;
+#else
+    static uint32_t img_upload_stalling_timeout;
+
+    if (!img_upload_stalling) {
+        if (curr_off == img_size) {
+            img_upload_stalling = true;
+            img_upload_stalling_timeout =
+                k_uptime_get_32() + MCUBOOT_BOOT_SERIAL_IMG_UPLOAD_STALLING_TIMEOUT
```
d3zd3z, line 331: This is strange without the semicolon. Why not just have the number in the macro and use it like a regular expression?

d3zd3z (changes requested):

boot/boot_serial/src/boot_serial.c
de-nordic: OK.... so this is programmers error and the only reason this even compiles is the other programmers error you have pointed out here: https://github.com/mcu-tools/mcuboot/pull/1387/files#r905364351.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
de-nordic: Fixed

boot/boot_serial/src/boot_serial.c
de-nordic: Fixed

boot/boot_serial/src/boot_serial.c
de-nordic: This skips decryption until stalling ends. If stalling is still happening then the image may not be something valid to decrypt yet.
There is also additional work needed in the future to actually add the decryption as "installation" step, so that the transfer reporting could also be stalled for the period it takes to decrypt an image.
Currently immediately when you get last offset of image the mcumgr-cli will get notified that upload has been successful, but what really happens is that decryption starts. This is actually wrong because you can no issue reset while in the middle of decrypting an image.

de-nordic:
I have decided to convert it to draft, as I consider to replace the feature with image post processing feature, that will have similar stalling mechanism.

PR #1396, opened 2022-06-22
zephyr: dropped __BOOTSIM__ ifdef-ry from the port

nvlsianpu:
Since #1383 the configuration translation header and
the logging translation header are not used by the simiulator.

Thanks to that __BOOTSIM__ dependence might be removed - what is done
in this patch.


de-nordic (approved):
Looks OK. Thanks!

utzig (approved):
LGTM

PR #1398, opened 2022-06-24
ptest: Update dependencies

d3zd3z:
Update some of the dependencies to newer versions, and run `cargo
update`.  This will resolve some potential security warnings about
issues found in older versions of dependencies.  I don't believe any of
these affect this code.


utzig (approved):
LGTM

PR #1404, opened 2022-07-05
zephyr: flash: support 32byte block write size

heinwessels:
This PR adds support STM32 devices with 32byte flash write alignment. Possible after [#1217](https://github.com/mcu-tools/mcuboot/pull/1217)

**Does it work**
Why yes! I tested with a `nucleo_h743zi` with the following steps:
 - Built and flashed the adapted bootloader. Built two
 - Built two demo applciations with bootloader support. Signed them by omitting `--align` so it automatically detects to use `32`.
 - Flash two demo applications. 
 - MCUBoot detects the correct magic in the second slot, and does a succesful swap.

**Old Behaviour:**
The STM32H7 device only supports writing to flash at a 32byte alignment. However, currently `BOOT_MAX_ALIGN` will always be set to `8` for a Zephyr build. This means the bootloader will not find the correctly signed image's magic number, because it's looking at the wrong location.

**New Behaviour:**
A Zephyr build will always look in the device tree if there's a chosen `zephyr,flash`, and if that node has a `write-block-size` property. If it exists, then it will set ultimately set the `BOOT_MAX_ALIGN` value to that `write-block-size`. It also checked that it's bigger that `8`.

**Typical STM32H7 DTS snippets**:
```
            chosen {
		zephyr,flash = &flash0;
```
```
			flash0: flash@8000000 {
				compatible = "st,stm32-nv-flash", "soc-nv-flash";
				label = "FLASH_STM32";
				write-block-size = < 0x20 >;
				erase-block-size = < 0x20000 >;
				max-erase-time = < 0xfa0 >;
				reg = < 0x8000000 0x200000 >;
				partitions {
```

**What about other Zephyr devices?**
- If the properties doesn't exist the default behaviour will be used.
- If the value is too small, like for example the `STM32F1`, then default behaviour will be used.

### DIFF
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -9,6 +9,8 @@
 #ifndef __MCUBOOT_CONFIG_H__
 #define __MCUBOOT_CONFIG_H__
 
+#include <zephyr/devicetree.h>
+
 #ifdef CONFIG_BOOT_SIGNATURE_TYPE_RSA
 #define MCUBOOT_SIGN_RSA
 #  if (CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN != 2048 && \
@@ -231,8 +233,15 @@
 #define MCUBOOT_MAX_IMG_SECTORS       128
 #endif
 
+/* Support 32-byte aligned flash sizes */
+#if DT_HAS_CHOSEN(zephyr_flash)
+    #if DT_PROP_OR(DT_CHOSEN(zephyr_flash), write_block_size, 0) > 8
+        #define MCUBOOT_BOOT_MAX_ALIGN \
+            DT_PROP(DT_CHOSEN(zephyr_flash), write_block_size)
+    #endif
+#endif
+
 #if CONFIG_BOOT_WATCHDOG_FEED
-#include <zephyr/devicetree.h>
 #if CONFIG_NRFX_WDT
 #include <nrfx_wdt.h>
 


nicokorn (approved):
Just tested it and it works.

nicokorn (approved):

utzig:
@heinwessels I tested this on the Nucleo-H753ZI and confirmed it upgrades successfully. Could you do a bit of fixup on the commit, because it seems that now `<zephyr/devicetree.h>` is also included by the NRFX watchdog.

heinwessels:
> @heinwessels I tested this on the Nucleo-H753ZI and confirmed it upgrades successfully. Could you do a bit of fixup on the commit, because it seems that now `<zephyr/devicetree.h>` is also included by the NRFX watchdog.

I'm not sure how to make it better, since the `DT_HAS_CHOSEN` is not always defined inside `mcuboot_config.h`, which fails compilation if I remove the include. I have several options of course, but I don't think there's a better one.
- Add a `#ifdef DT_HAS_CHOSEN` around the logic, and then be sure to include the devicetree everywhere that the config is used, and the alignment is required. However this is not secure, because it's too easy to forget to include the device tree and then not have expected behaviour.
```C
#include <zephyr/devicetree.h>   // Included so that it's available in the next include
#include <mcuboot_config/mcuboot_config.h>
```
- Place the logic inside `bootutil_public.h`. This would be out of place, reading device trees in generic code. And most devicetree and configurations are all handled inside `mcuboot_config`. 
- Keep code as is. Since the device tree is used elsewhere in that file, and could also be used by NRFX specifc code.

I vote to keep the code as is, but I'm willing to change it.

utzig:
> ...

All I was trying to say is that you can remove the current `#include <zephyr/devicetree.h>` which exists here: https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/include/mcuboot_config/mcuboot_config.h#L235. Sorry, I have not communicated it well. Please rebase on latest `main` and it will show up on your PR's diff.

heinwessels:
> > ...
> 
> All I was trying to say is that you can remove the current `#include <zephyr/devicetree.h>` which exists here: https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/include/mcuboot_config/mcuboot_config.h#L235. Sorry, I have not communicated it well. Please rebase on latest `main` and it will show up on your PR's diff.

Ah, now I understand. Thanks, and fixed :)

utzig (approved):
LGTM

PR #1406, opened 2022-07-06
Adding new macro for region copy buffer size

kail:
This addresses the issue opened here: https://github.com/mcu-tools/mcuboot/issues/1407

## Background:
[This PR](https://github.com/mcu-tools/mcuboot/pull/1217) introduced a change which appeared to be intended to [allow the buffer size to be configurable](https://github.com/mcu-tools/mcuboot/pull/1217/files#diff-a004fdf5907914a360ce1e7da94221c0165d1c4c8ba7b8c4eaf67bd63af395c1R83).

That change, however, is frivolous because `MCUBOOT_BOOT_MAX_ALIGN` is guaranteed to be [between 8 and 32](https://github.com/mcu-tools/mcuboot/commit/73c38c6fde489746d5307a691dc7006234372279#diff-531d0cc2251e606642938c20aeec3920ad187962f69c8e231f93aee73d3277c2R86)

## This PR:
This PR adds a new configuration macro to allow the region copy buffer size to be a value other than 1024, which should allow for a more efficient use of block devices when the block size is a different value.

It is backwards compatible with all version, as the default when that macro is unset is still 1024

Signed-off-by: Mikhail Skobov <mskobov@lyft.com>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -81,8 +81,8 @@ static struct boot_loader_state boot_data;
 #define TARGET_STATIC
 #endif
 
-#if BOOT_MAX_ALIGN > 1024
-#define BUF_SZ BOOT_MAX_ALIGN
+#ifdef BOOT_COPY_REGION_SIZE
+#define BUF_SZ BOOT_COPY_REGION_SIZE
 #else
 #define BUF_SZ 1024
 #endif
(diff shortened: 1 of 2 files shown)


boot/bootutil/src/loader.c
```diff
@@ -81,8 +81,8 @@ static struct boot_loader_state boot_data;
 #define TARGET_STATIC
 #endif
 
-#if BOOT_MAX_ALIGN > 1024
```
nvlsianpu, line 84: Previous values for undefined MCUBOOT_BOOT_COPY_REGION_SIZE should be kept

kail: I'm not sure that I follow. How are they being discarded?

In all prior versions `BOOT_MAX_ALIGN > 1024` is _never true_, so this condition can be removed entirely. And `BOOT_COPY_REGION_SIZE` is only defined iff `MCUBOOT_BOOT_COPY_REGION_SIZE` is defined, so by default,  if a user does not define `MCUBOOT_BOOT_COPY_REGION_SIZE` then 1024 will be used

nvlsianpu: OK, Can you investigate CI failure?

kail: Sure thing, I will take a look later today

PR #1415, opened 2022-07-12
RFC: boot_serial: Post update, installation progress hook and usage

de-nordic:
The RFC PR consists of two commits:
 1) Adding specification for `boot_img_install_progress_hook` that allows to report post processing status of uploaded image (for example decryption or processing on another core);
 2) Invocation points in image list and image upload processing, implementing RFC change to SMP protocol: https://github.com/zephyrproject-rtos/zephyr/pull/47253


sfeutrier:
I see this pull request has not been merged. Is there another solution to be able to track the percentage/progress of the update then?

PR #1428, opened 2022-07-25
zephyr/boards: increase bootloader size for nrf52840 big partition

nvlsianpu:
sample.bootloader.mcuboot.usb_cdc_acm_recovery_log test build was
failing.

This particular configuration needs some more
flash that before.

This patch gives it 4 kB more, so build can pass


nvlsianpu:
superseded by #1429

PR #1429, opened 2022-07-26
zephyr: disable qspi-nor on nrf52840

nvlsianpu:
Sine https://github.com/zephyrproject-rtos/zephyr/pull/47629 was
merged the qspi nod driver is enabled by default. This cause unwanted
flash footprint rise for configurations which are not using the qspi.

This patch disables it back.


gustavonihei (approved):
LGTM.

PR #1431, opened 2022-07-27
zephyr: remove flash_device_get_binding

galak:
remove flash_device_get_binding as we can utilizes DEVICE_DT_GET to
get the flash_dev pointer set at build time.  This removes usage of
device_get_binding and handles the lack of 'label' properties in
the devicetree better.

Signed-off-by: Kumar Gala <galak@kernel.org>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -506,21 +506,6 @@ void main(void)
 
     mcuboot_status_change(MCUBOOT_STATUS_STARTUP);
 
-#if (!defined(CONFIG_XTENSA) && DT_HAS_CHOSEN(zephyr_flash_controller))
-    if (!flash_device_get_binding(DT_LABEL(DT_CHOSEN(zephyr_flash_controller)))) {
-        BOOT_LOG_ERR("Flash device %s not found",
-		     DT_LABEL(DT_CHOSEN(zephyr_flash_controller)));
-        while (1)
-            ;
-    }
-#elif (defined(CONFIG_XTENSA) && defined(JEDEC_SPI_NOR_0_LABEL))
-    if (!flash_device_get_binding(JEDEC_SPI_NOR_0_LABEL)) {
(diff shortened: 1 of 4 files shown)


boot/zephyr/main.c
```diff
@@ -506,20 +506,11 @@ void main(void)
...
 
-#if (!defined(CONFIG_XTENSA) && DT_HAS_CHOSEN(zephyr_flash_controller))
-    if (!flash_device_get_binding(DT_LABEL(DT_CHOSEN(zephyr_flash_controller)))) {
-        BOOT_LOG_ERR("Flash device %s not found",
-		     DT_LABEL(DT_CHOSEN(zephyr_flash_controller)));
+    if (!flash_device_get_binding()) {
+        BOOT_LOG_ERR("Flash device not ready");
         while (1)
             ;
     }
```
de-nordic, line 513: Do we even need this part?
How does the rest of booting process depend on this?
Is it only for checking whether there is device to boot from? If yes, can we do that at compile time?

The boot process will have flash_area_open invocation somewhere below that point, that will do the same check anyway, to get access to partition to boot.

galak: I'm no expert here, I'm happy to remove the code if those that work on mcuboot say its not needed.

de-nordic: > I'm no expert here, I'm happy to remove the code if those that work on mcuboot say its not needed.

@nvlsianpu @utzig  Can you provide advise here?

nvlsianpu: I think flash_device_get_binding() and its usage can be dropped. MCUboot will fail later on first attempt to the flash anyway. Since device is provide directly for DT-2-C API checking of that dosne't make bi sense anymoe.

galak: will update to remove `flash_device_get_binding`

de-nordic (approved):
I have tested the change on nrf52840dk_nrf52840 and it works correctly.

Looks OK, thanks!

nvlsianpu (approved):
looks fine to me.

PR #1433, opened 2022-07-29
boot_serial: fix serial recovery mode with timeout

pepe2k:
If `BOOT_SERIAL_WAIT_FOR_DFU` is selected, the CPU shouldn't enter idle state, waiting for interrupt from the console because we expect booting if no mcumgr command is received within a configured timeout (with the `CONFIG_BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT`). Without this fix, when using `BOOT_SERIAL_WAIT_FOR_DFU` the boot process hangs forever, waiting for input from console.

This fixes original commit where optional timeout support in serial recovery was included: e3822f8180ed.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -848,7 +848,9 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
 
     off = 0;
     while (timeout_in_ms > 0 || bs_entry) {
+#ifndef MCUBOOT_SERIAL_WAIT_FOR_DFU
         MCUBOOT_CPU_IDLE();
+#endif
         MCUBOOT_WATCHDOG_FEED();
 #ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
         uint32_t start = k_uptime_get_32();


nvlsianpu:
@pepe2k This fix is relevant to configuration CONFIG_MULTITHREADING=n. Issue doesn't harming when MCUboot uses multithreading. Am I right?

pepe2k:
@nvlsianpu thanks for the comment. So far I've been testing this only with some nRF52 (`MULTITHREADING=n`) and indeed, with multi-threading enabled everything works as expected (without the fix).
Now I'm not sure whether my fix is proper (or, whether the commit description is full and correct). Check if multi-threading is enabled is "hide" within `MCUBOOT_CPU_IDLE` definition (https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/include/mcuboot_config/mcuboot_config.h#L286).

nvlsianpu:
@pepe2k Your fix is proper. Whe Multitihreding is enable issue doesn't harm boot process as rtos managing idle time on its own (and the macro is empty).

Can you just add commnet to the ifdefry you added for explain the reason behind this conditional code?

nvlsianpu (approved):

de-nordic (approved):

pepe2k:
@nvlsianpu 
> Can you just add commnet to the ifdefry you added for explain the reason behind this conditional code?

I was AFK for a few days and forgot about this one.
Do you still want me to add some extra explanation to the code in a separate PR/commit?

Thanks!

nvlsianpu:
@pepe2k Would be great If you can add a commen.

PR #1434, opened 2022-07-29
zephyr: Explicitly include cmsis.h

yonsch:
main.c uses CMSIS functions such as __set_MSP, which require
cmsis.h to be included. Up until now, that file was included
indirectly through other ARM headers. This patch explicitly
includes cmsis.h, for platforms on which those indirect includes
do not work.

Signed-off-by: Yonatan Schachter <yonatan.schachter@gmail.com>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -27,6 +27,12 @@
 #include <soc.h>
 #include <zephyr/linker/linker-defs.h>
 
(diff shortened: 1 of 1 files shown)


d3zd3z:
Rebased to bring in some fixes affecting CI.

d3zd3z (approved):

PR #1442, opened 2022-08-16
boot:zephyr: should clear mpu config

lenghonglin:
Hello , everyone. 

I use mcuboot those day, i met a very strange things that if not call `z_arm_clear_arm_mpu_config`, it will hard fault.
image

I config SRAM MPU in zephyr
image

So , we should `z_arm_clear_arm_mpu_config` if `CONFIG_CPU_HAS_ARM_MPU` = y . 

Signed-off-by: honglin leng <a909204013@gmail.com>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -183,6 +183,11 @@ static void do_boot(struct boot_rsp *rsp)
     /* Disable the USB to prevent it from firing interrupts */
     usb_disable();
 #endif
+
+#if CONFIG_CPU_HAS_ARM_MPU || CONFIG_CPU_HAS_NXP_MPU
+    z_arm_clear_arm_mpu_config();
+#endif
+
 #if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
     cleanup_arm_nvic(); /* cleanup NVIC registers */
 
@@ -192,9 +197,6 @@ static void do_boot(struct boot_rsp *rsp)
     SCB_DisableICache();
 #endif
 
-#if CONFIG_CPU_HAS_ARM_MPU || CONFIG_CPU_HAS_NXP_MPU
-    z_arm_clear_arm_mpu_config();
-#endif
 
 #if defined(CONFIG_BUILTIN_STACK_GUARD) && \
     defined(CONFIG_CPU_CORTEX_M_HAS_SPLIM)


lenghonglin:
@nvlsianpu  could you please review this prm thanks

lenghonglin:
image
twister bug with nrf52840 ?

nvlsianpu (changes requested):
Hi,
I think you should rather enable CONFIG_MCUBOOT_CLEANUP_ARM_CORE option for your build instead of this patch.
Or explain more on reason for which you want always clear the MPU (zephyr-rtos application does this on its own).

Build issue you encountered was related to zephyr -rtos main branch API change

lenghonglin:
> Hi, I think you should rather enable CONFIG_MCUBOOT_CLEANUP_ARM_CORE option for your build instead of this patch. Or explain more on reason for which you want always clear the MPU (zephyr-rtos application does this on its own).
> 
> Build issue you encountered was related to zephyr -rtos main branch API change

i test frdm_k64 and my board. the same thing is both disable `CONFIG_MCUBOOT_CLEANUP_ARM_CORE `,  but frdm_k64 is work fine , my board cannot boot. enable `CONFIG_MCUBOOT_CLEANUP_ARM_CORE ` is not a good way. 

So i think the difference is about `sram`. my board config mpu with sram.

It is good to clear mpu config before reboot, this will solve some mpu config that cannot boot success. 

In other words, this mr maybe sovle some issue, but does not introduce new problems, right?

nvlsianpu:
@lenghonglin I don't understand the reason for which you are not using  `CONFIG_MCUBOOT_CLEANUP_ARM_CORE=y`

lenghonglin:
> @lenghonglin I don't understand the reason for which you are not using `CONFIG_MCUBOOT_CLEANUP_ARM_CORE=y`

This is a good question,  enable `CONFIG_MCUBOOT_CLEANUP_ARM_CORE` although can solve this issue, but this is not a really solution.

let me think about this issue.

both `frdm_k64` and `my_board ` is disable  `CONFIG_MCUBOOT_CLEANUP_ARM_CORE`, but only `frdm_k64` can reboot success. this is can explain that there is other reason make `my_board` cannot reboot.  

I found the difference is about `mpu config`, so the best solution is should clear mpu config if config `CONFIG_CPU_HAS_ARM_MPU `

nvlsianpu:
@lenghonglin As you mention before you are using the custom with peculiar SRM MPU configuration - you should enable CONFIG_MCUBOOT_CLEANUP_ARM_CORE for your board - This option allows to cleanup mpu and more.

Let me know if you can't do that (with explanation).

Issue #1444, opened 2022-08-17
imgtool: Sign writes image_ok to wrong offset for hex output format


abxy:
To reproduce, sign and `--confirm` an image with output format hex. Then convert the output to .bin and inspect using xxd.
```
imgtool sign --key <path to key> --header-size 0x200 --align 8 --version 1.0 \
  --slot-size 0x67000 --confirm zephyr.bin signed-fw.hex -x 0xc000
objcopy -I ihex -O binary signed-fw.hex signed-fw.hex.tobin
xxd signed-fw.hex.tobin
```

See example output below (ommissions marked with `...`)
```
...
000663a0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000663b0: 0000 0000 0000 0000 01ff ffff ffff ffff  ................
000663c0: ffff ffff ffff ffff 0000 0000 0000 0000  ................
000663d0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
...
00066fd0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00066fe0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00066ff0: 77c2 95f3 60d2 ef7f 3552 500f 2cb6 7980  w...`...5RP.,.y.
```

You'll find that the magic is written correctly, but the `image_ok`flag which is expected immediately preceding the magic has been written to a completely different offset, `0x000663b8` in the example.

Running the same commands but outputting to a .bin file directly results in correct output.
Note that the difference in padding bytes being `ff` instead of `00` is expected
```
...
00066fe0: ffff ffff ffff ffff 01ff ffff ffff ffff  ................
00066ff0: 77c2 95f3 60d2 ef7f 3552 500f 2cb6 7980  w...`...5RP.,.y.
```

I believe https://github.com/mcu-tools/mcuboot/issues/1390 is related and I have a patch ready to submit as a PR.

utzig:
> Note that the difference in padding bytes being `ff` instead of `00` is expected

Thanks for the issue + fix. Btw, you could also do `objcopy -I ihex -O binary --gap-fill=0xff signed-fw.hex signed-fw.hex.tobin` which should generate the same binary.

PR #1445, opened 2022-08-17
imgtool: Fix output of confirmed image in HEX format

abxy:
This PR fixes https://github.com/mcu-tools/mcuboot/issues/1444

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -243,12 +243,11 @@ class Image():
                                                   self.enctlv_len)
                 trailer_addr = (self.base_addr + self.slot_size) - trailer_size
                 if self.confirm and not self.overwrite_only:
-                    magic_size = 16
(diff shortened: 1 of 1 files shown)


scripts/imgtool/image.py
```diff
@@ -246,9 +246,7 @@ def save(self, path, hex_addr=None):
                     magic_size = 16
                     magic_align_size = align_up(magic_size, self.max_align)
```
utzig, line 247: If you don't mind the unrelated changed, please remove the `magic_size` above and use `len(self.boot_magic)` in the call to `align_up` above. Otherwise the change looks good!

abxy: Sure thing! It's in the updated PR (force pushed)

utzig (approved):
LGTM

nvlsianpu (approved):

PR #1458, opened 2022-08-30
boot: zephyr: return NULL if stm32 watchdog is not enabled

bbilas:
Return the device pointer if the node identifier refers to a watchdog node with status “okay”, otherwise return NULL.

Signed-off-by: Bartosz Bilas <bartosz.bilas@hotmail.com>


bbilas:
@nvlsianpu PTAL

bbilas:
Depends on https://github.com/mcu-tools/mcuboot/pull/1459.

mniestroj (approved):

utzig (approved):
LGTM

PR #1466, opened 2022-09-04
imgtool: add option to export public PEM

utzig:
Update `getpub` with new `lang` option, "pem", which allows exporting a public key as a PEM file. This can later be distributed to be used for encrypting an image, and gets away with having to use openssl for this step.


nvlsianpu (commented):
LGTM

d3zd3z (approved):
Ahh, the good old pem programming language.

utzig:
> Ahh, the good old pem programming language.

Yes it's the best, doesn't even need a compiler.

PR #1471, opened 2022-09-10
zephyr: ci: enable testing more watchdog paths

utzig:
Enable tests to be run on frdm_k64f and disco_l475_iot1. The l475 uses the STM32 IWDG by default, and the k64f can be used for the generic watchdog path. Both boards also received a config to enable the watchdog.

Fix watchdog device typos.


utzig:
This should fail 2 twister tests until the typos are fixed (in a followup commit).

utzig:
> This should fail 2 twister tests until the typos are fixed (in a followup commit).

As expected fails happened on this CI run: https://github.com/mcu-tools/mcuboot/runs/8288161320

fabiobaltieri (approved):

bbilas (approved):

boot/zephyr/sample.yaml
```diff
@@ -5,9 +5,10 @@ sample:
 tests:
   sample.bootloader.mcuboot:
     tags: bootloader_mcuboot
-    platform_allow: nrf52840dk_nrf52840 frdm_k64f
```
nvlsianpu, line 8: I have to consult this line. I'll response soon.

nvlsianpu (changes requested):

boot/zephyr/sample.yaml
nvlsianpu: Twister will run this testcase for all, listed bellow integration platform (if run with `--integration`). This change causes that it will be possible to runt this testcase for any platform (specified in commndaline).

nvlsianpu (approved):

PR #1472, opened 2022-09-13
boot: zephyr: Remove legacy serial recovery GPIO configuration

### LINKED ISSUE #1473: zephyr/serial_recovery: build failure if CONFIG_BOOT_SERIAL_DETECT_PIN has undefined value
zephyr-rtos build system fails because kconfig property has no default value on some boards.
This become the issue as user might use DTS gpio aliases so legacy configuration is not provided.

MarekPieta:
Change removes the legacy configuration. The configuration can no longer be used, because GPIO DTS nodes no longer support labels. The GPIO must be configured by board's DTS.

PR also adds default values of boot detect pins to prevent build failures.

fixes #1473


boot/zephyr/boards/conexio_stratus.conf
```diff
@@ -11,9 +11,6 @@ CONFIG_BOOT_MAX_IMG_SECTORS=256
 
 # MCUboot serial recovery
 CONFIG_MCUBOOT_SERIAL=y
-CONFIG_BOOT_SERIAL_DETECT_PORT="GPIO_0"
```
nvlsianpu, line 14: drop this board patch
this is a unknown board.

nvlsianpu (changes requested):

nvlsianpu:
2nd commit message:

Change removes the legacy configuration. The legacy configuration become problematic as can not
longer use GPIO DTS nodes labels, because GPIO DTS nodes no longer support labels. Therefore need to use GPIO DTS node name with the legacy approach.

The GPIO is preferred to be configured by board's DTS, which is loot more easier,

boot/zephyr/boards/conexio_stratus.conf
MarekPieta: Dropped. Btw. should we keep configuration of non-zephyr board in here?

nvlsianpu (approved):

PR #1489, opened 2022-10-12
Feature/1410 min refactor zephyr port

gregshue:
This PR contains a new sample demonstrating a standard Zephyr use case not currently supported.  The successive commits demonstrate the successive build FAILURES encountered until a minimal(?) solution was found to support the new sample.

Fundamentally these reveal that in order to support an externally configurable, composable solution the MCUBoot source build file cannot contain the `find_package(Zephyr ...)` command.  In other words, the reusable sources build file (CMakeLists.txt) must be in a different directory than the project configuration build file (CMakeLists.txt). Similarly, the Kconfig files within the module sources tree cannot also be a top-level project Kconfig definition file (otherwise causing recursive sourcing of Kconfig.zephyr).

This may not be the actual minimal solution for the current implementation of the Zephyr build system, but it seems to be the smallest implementation that is consistent with the Zephyr build system features, paradigms, and recommended/expected behaviors.


gregshue:
I don't like having to copy the local *.conf and board/ files into the new sample, but a general pattern must work across modules.  Once [Zephyr PR #47638](https://github.com/zephyrproject-rtos/zephyr/pull/47638) is fully approved and merged, then these files could be referenced rather than copied.

gregshue:
This PR implementation has been cleaned up and existing functional builds verified and documented at every commit.  (NOTE: Make-ing mcuboot/sample/zephyr did not succeed at the base of this PR, so it was not part of the verification process.)

gregshue:
The push above takes advantage of the just-merged of the fix for Zephyr's issue #41830, enabling use of ZEPHYR_<modulename>_MODULE_DIR in CONF_FILE, OVERLAY_CONFIG, and DTC_OVERLAY_FILE before find_package().  

It also takes advantage of the APPLICATION_CONFIG_DIR.  

Together these reduced that solution to 17 files (touched and created).

samples/zephyr_external_config/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+# find_package(Zephyr) in order to load application boilerplate:
+# http://docs.zephyrproject.org/application/application.html
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(mcuboot_exernal_config)
+
+target_sources(app PRIVATE
+  keys.c
+  )
+
+if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
```
d3zd3z, line 49: Can we have a commit before this that refactors cmake in mcuboot a bit, so this code can be shared with the other cmake files instead of duplicated here. Perhaps `zephyr/cmake`.

d3zd3z (changes requested):
Regarding the copy of the board files, before committing this, can we make a Zephyr issue, so we can reference that in this commit. Then, once this merges, we can modify the Zephyr issue to point back to this commit, to demonstrate why the fix is needed.

samples/zephyr_external_config/CMakeLists.txt
gregshue: Done.

gregshue:
The following changes address comments above:

- The Zephyr build issue has been captured in https://github.com/zephyrproject-rtos/zephyr/issues/51621.
- `mcuboot_svc` has been renamed `mcuboot_app`
- The `mcuboot/boot/zephyr/CMakeLists.txt` logic has been extracted out into files in `mcuboot/zephyr/cmake/`.
- The new sample has been relocated to `mcuboot/zephyr/samples/mcuboot_external_config/`

I think that addresses all feedback so far.

gregshue:
@nvlsianpu Since you missed today's TSC meeting where we reviewed the minimum implementation in detail, do you have any feedback?

nordicjm (changes requested):
This introduces the follow when configuring `boot/zephyr`:
```
warning: MBEDTLS_CFG_FILE (defined at modules/mbedtls/Kconfig:50,
/tmp/aa/bootloader/mcuboot/zephyr/Kconfig:170, modules/mbedtls/Kconfig:50) was assigned the value
'mcuboot-mbedtls-cfg.h' but got the value 'config-tls-generic.h'. See
http://docs.zephyrproject.org/latest/kconfig.html#CONFIG_MBEDTLS_CFG_FILE and/or look up
MBEDTLS_CFG_FILE in the menuconfig/guiconfig interface. The Application Development Primer, Setting
Configuration Values, and Kconfig - Tips and Best Practices sections of the manual might be helpful
too.
```
This configuration should not be failing to be set

gregshue:
@nordicjm What is the command line/test case you were building that shows the warning?

nordicjm:
> @nordicjm What is the command line/test case you were building that shows the warning?

`cmake -GNinja -DBOARD=nrf52dk_nrf52832 ..`

gregshue:
> cmake -GNinja -DBOARD=nrf52dk_nrf52832 ..

Which was built from where?

nordicjm:
> Which was built from where?

`boot/zephyr`

gregshue:
🤔 I'm sorry, but I'm not following you very well.

In my environment:
- `mcuboot` is the manifest repository, so for me `mcuboot` is directly below the workspace root rather than in the `bootloader/` subdirectory
- I only build using `west`, so I don't know what other parameters are missing from a `cmake` command.
- When I run the following command from my workspace root the build completes without warnings: `west build -p always -b nrf52dk_nrf52832 mcuboot/boot/zephyr/`
- When I run the following command from my workspace root the build does not find a top-level CMakeLists.txt: `cd mcuboot/boot/zephyr && cmake -GNinja -DBOARD=nrf52dk_nrf52832 ..`

The commit comments list all the build commands I was aware I needed to verify still worked (west, twister).  What additional commands do I need to verify?

nordicjm:
```
<get zephyr and get these changes to mcuboot>
cd bootloader/mcuboot/boot/zephyr
mkdir _build
cd _build
cmake -GNinja -DBOARD=nrf52dk_nrf52832 ..
ninja
```

gregshue:
The symptom (warning behavior) was different depending on whether mcuboot or zephyr was the manifest repository - even with the same workspace tree and SHAs.  (The config system parses Kconfig files in a different order for these two cases.)

The root cause was the original prj.conf was missing a setting.  This has been fixed.

nordicjm:
I'm not sure what's happened here, but I'm trying to compare a .config generated using the current version in main of zephyr vs this, in the zephyr one, the configuration file makes sense, here's an excerpt:
```
#
# MCUboot-specific configuration options
#
CONFIG_MCUBOOT=y
CONFIG_BOOT_USE_MBEDTLS=y

#
# MCUBoot settings
#
# CONFIG_SINGLE_APPLICATION_SLOT is not set
# CONFIG_BOOT_SIGNATURE_TYPE_NONE is not set
CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN=2048
# CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256 is not set
# CONFIG_BOOT_SIGNATURE_TYPE_ED25519 is not set
CONFIG_BOOT_SIGNATURE_KEY_FILE="root-rsa-2048.pem"
# CONFIG_MCUBOOT_CLEANUP_ARM_CORE is not set
CONFIG_MBEDTLS_CFG_FILE="mcuboot-mbedtls-cfg.h"
...
CONFIG_BOOT_WATCHDOG_FEED=y
# CONFIG_BOOT_IMAGE_ACCESS_HOOKS is not set
# CONFIG_MCUBOOT_ACTION_HOOKS is not set
# end of MCUBoot settings

CONFIG_MCUBOOT_DEVICE_SETTINGS=y

#
# Zephyr configuration options
#
# CONFIG_MULTITHREADING is not set
CONFIG_USB_DEVICE_PRODUCT="MCUBOOT"
CONFIG_GPIO=y
# CONFIG_KSCAN is not set
# CONFIG_WIFI is not set
# CONFIG_SPI is not set
CONFIG_GPIO_INIT_PRIORITY=40
...
```

For some reason, with these changes, the output is now scrambled:
```
CONFIG_GPIO=y  <-- Yes, the header here is now gone
# CONFIG_KSCAN is not set
# CONFIG_WIFI is not set
# CONFIG_SPI is not set
CONFIG_GPIO_INIT_PRIORITY=40
# CONFIG_MODEM is not set
# CONFIG_UART_INTERRUPT_DRIVEN is not set
...
#
# mcuboot (/tmp/aa/bootloader/mcuboot)
#
CONFIG_MCUBOOT=y
CONFIG_MCUBOOT_APP=y
CONFIG_BOOT_USE_MBEDTLS=y

#
# MCUBoot settings
#
# CONFIG_SINGLE_APPLICATION_SLOT is not set
# CONFIG_BOOT_SIGNATURE_TYPE_NONE is not set
...
# CONFIG_MCUBOOT_ACTION_HOOKS is not set
# end of MCUBoot settings

CONFIG_MCUBOOT_DEVICE_SETTINGS=y

#
# Zephyr configuration options  <-- what's going on here?
#
# CONFIG_MULTITHREADING is not set
CONFIG_USB_DEVICE_PRODUCT="MCUBOOT"
CONFIG_MCUBOOT_BOOTUTIL_LIB=y
CONFIG_ZEPHYR_MCUBOOT_MODULE=y
# end of mcuboot (/tmp/aa/bootloader/mcuboot) <-- what's going on here?
```

gregshue:
> For some reason, with these changes, the output is now scrambled:

Yes.  The Zephyr config system lists configs from the top-level Kconfig before it lists those from the modules.  You will need to relocate the block of settings before comparison. ;-)

When we moved the Kconfig declarations from `$mcuboot/boot/zephyr/Kconfig` into `$mcuboot/zephyr/Kconfig` we put the declarations in a place they can be reused by other project configurations (e.g., under $mcuboot/zephyr/tests/, $mcuboot/zephyr/samples/, $my_application/apps/`). This is a necessary fix to work as a composable system. IIRC, this move revealed that the exiting `$mcuboot/boot/zephyr/` Kconfig settings are incomplete - lacking `CONFIG_CUSTOM_MBED_CFG_FILE=y` and does not pick up some of the logging declarations.

Note: I do not know whether the module kconfigs are ordered by `$module/zephyr/module.yml:name` or by `basename $module`.  If the latter, then we cannot control the declaration order anyway. Either way, we cannot control which is the first `default` declaration to be seen for a Kconfig - but that is a separate issue that should lead Zephyr to extend or reinterpret the Kconfig spec such that later conditional defaults override unconditional defaults.

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	  boot/bootutil/include/bootutil/mcuboot_status.h
+
+endmenu
+
+config MCUBOOT_DEVICE_SETTINGS
+	# Hidden selector for device-specific settings
+	bool "Non-hidden selector for device-specific settings"
+	default n
+        # CPU options
+	select MCUBOOT_DEVICE_CPU_CORTEX_M0 if CPU_CORTEX_M0
```
nvlsianpu, line 625: this should be kept as automatic hidden selection  as it used to be.

zephyr/module.yml
```diff
@@ -1,5 +1,10 @@
+# Copyright (c) 2022 Legrand North America, LLC.
```
nvlsianpu, line 1: If you are gong to add copyright list here, please add Nordic semiconductor to it as well.
`# Copyright (c) 2019-2021 Nordic Semiconductor ASA`

zephyr/Kconfig
gregshue: @nvlsianpu 
When relocating the Kconfig definitions from `$mcuboot/boot/zephyr/Kconfig` (project-specific) to `$mcuboot/zephyr/Kconfig` (module-specific), the definition of `MCUBOOT_DEVICE_SETTINGS` is now visible to the `$zephyr/samples/tfm_integration/psa-crypto/` build. This build with existing default 'y' now fails with:

> warning: FLASH_MAP (defined at subsys/storage/flash_map/Kconfig:10) has direct dependencies FLASH_HAS_DRIVER_ENABLED with value n, but is currently being y-selected by the following symbols:
> - MCUBOOT_DEVICE_SETTINGS (defined at /home/zephyr/mcubbootwksp/mcuboot/zephyr/Kconfig:620), with value y, direct dependencies y (value: y)

In other words, refactoring these Kconfigs for reuse also exposes them to more existing projects. Defaulting anything to 'y' impacts those projects too.  

AFAICT to not effect existing projects the default needs to be 'n' and the control point (e.g., `MCUBOOT_DEVICE_SETTINGS`) needs to be selectable by the projects that need it.  Do you see a better solution?

zephyr/module.yml
gregshue: Added.

gregshue:
> > Add prompt to MCUBOOT_DEVICE_SETTINGS Kconfig and default 'n'.
> > These device settings rules need to not be applied for the psa_crypto build.
> 
> What is conflicting here?

Unknown.  This is supposed to be a refactor, so the goal is to not change the existing settings behavior.

gregshue:
> What is conflicting here?

See the warning about FLASH_MAP in an earlier response.

nordicjm (changes requested):
I've built mcuboot for 2 boards prior to this patch, and with this patch, then MD5's the output bin files and checked the configs for differing values. The output MD5's do not match, so that is an instant NACK from me. The configuration files show these differences:
```
Left in A:
Array
(
    [204] => CONFIG_CUSTOM_MBEDTLS_CFG_FILE=y
    [222] => # mcuboot (/tmp/aa/bootloader/mcuboot)
    [225] => CONFIG_MCUBOOT_APP=y
    [288] => CONFIG_ZEPHYR_MCUBOOT_MODULE=y
    [289] => # end of mcuboot (/tmp/aa/bootloader/mcuboot)
)
Left in B:
Array
(
    [2] => # MCUboot-specific configuration options
    [271] => # CONFIG_CUSTOM_MBEDTLS_CFG_FILE is not set
)
```
I've rebased this to main to ensure it wasn't missing any configuration. So why is this generating a different output than prior to it being applied? The output should be the same.

gregshue:
@nordicjm 

> So why is this generating a different output than prior to it being applied? The output should be the same.

During this refactoring I discovered that the prior configuration was incomplete.  Addressing this will change the MD5s of the output bin files. 

What evidence do we have that the prior configuration was actually correct and complete?

gregshue:
BTW, MD5 calculations will also change when the object files are linked in a different order (giving a different binary stream feeding the MD5 calculation)

gregshue:
Here are the reasons for the differences:

- CONFIG_CUSTOM_MBEDTLS_CFG_FILE is defined by mbedtls and is supposed to be set when a custom mbedtls config file is used.  This was lacking from the original prj.conf because the project had a local Kconfig that set the default MBEDTLS_CFG_FILE to a value before the module declared it.  Basically, the original project did not follow the design of the MBEDTLS Kconfigs.

- CONFIG_MCUBOOT_APP is a new Kconfig to control building in the MCUboot application.  Without this control it is not possible to configure in an external project when the MCUboot app is built in to the image.

- CONFIG_ZEPHYR_MCUBOOT_MODULE is generated by the build system when it recognizes mcuboot as a module.  This read-only configuration has no effect on the build.

zephyr/cmake/mcuboot_zephyr_runner_mass_erase.cmake
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2022 Legrand North America, LLC.
```
nordicjm, line 1: This isn't your copyright, this is a cut and paste of the original file, the copyright is `# Copyright (c) 2017 Open Source Foundries Limited`

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+# Copyright (c) 2022 Legrand North America, LLC.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+
+# Option to build the project with the MCUBoot application
+config MCUBOOT
+	default n
+	bool  "MCUBoot interfaces"
```
nordicjm, line 12: Why is this selectable?

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	select MCUBOOT_APP
+
+config MCUBOOT_APP
+	default n
+	bool  "MCUBoot application"
```
nordicjm, line 17: Why is this selectable?

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	select USE_DT_CODE_PARTITION if HAS_FLASH_LOAD_OFFSET
+	select MCUBOOT_BOOTUTIL_LIB
+
+config MCUBOOT_BOOTUTIL_LIB_OWN_LOG
+	default n if MCUBOOT_APP
+
+config BOOT_USE_MBEDTLS
+	bool
+	# Hidden option
+	default n
```
nordicjm, line 28: `default n` is not needed

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	help
+	  Use mbedTLS for crypto primitives.
+
+config BOOT_USE_TINYCRYPT
+	bool
+	# Hidden option
+	default n
```
nordicjm, line 35: As above

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	# When building for ECDSA, we use our own copy of mbedTLS, so the
+	# Zephyr one must not be enabled or the MBEDTLS_CONFIG_FILE macros
+	# will collide.
+	help
+	  Use TinyCrypt for crypto primitives.
+
+config BOOT_USE_CC310
+	bool
+	# Hidden option
+	default n
```
nordicjm, line 45: As above

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+
+config NRF_CC310_BL
+	bool
+	default n
+
+menu "MCUBoot settings"
+
+config SINGLE_APPLICATION_SLOT
+	bool "Single slot bootloader"
+	default n
```
nordicjm, line 68: etc.

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
+# Copyright (c) 2017-2020 Linaro Limited
+# Copyright (c) 2020 Arm Limited
+# Copyright (c) 2022 Legrand North America, LLC.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+
+# Option to build the project with the MCUBoot application
```
nordicjm, line 9: This now appears at `Modules → mcuboot` in menuconfig instead of at the top level... mcuboot isn't a module, it's an application. I don't find this to be an enhancement, and this is going to confuse people that have been used to mcuboot, the application, being at the top level, where other applications and zephyr best practice states to put it at.

After having a play with it, I don't understand this change. If I configure hello_world without sysbuild, I see these options - why? I can't enable mcuboot from here, and if I do enable these options, all I get is a broken build with compile errors about `The flash driver lacks page layout support; falling back on hacks.` which doesn't make sense. And if I configure the application using sysbuild with mcuboot, these settings aren't even set, again, if I enable them, the build suddenly breaks with the same error.

zephyr/Kconfig
```diff
@@ -0,0 +1,680 @@
...
+	  during application firmware boot.
+
+	  Zephyr applications on Cortex-M will perform this register clean-up
+	  by default, if they are chain-loadable by MCUboot, so MCUboot does
+	  not need to perform such a cleanup itself.
+
+config MBEDTLS_CFG_FILE
+	default "mcuboot-mbedtls-cfg.h" if MCUBOOT
+
+config BOOT_HW_KEY
```
nordicjm, line 173: note to self: stopped review at this point

nordicjm (changes requested):

zephyr/cmake/mcuboot_zephyr_runner_mass_erase.cmake
gregshue: Thank you for pointing this out.  I missed the copyright in the file this was extracted from.  Now fixed.

zephyr/Kconfig
gregshue: Good question.  Looking back from an evolved PR, this probably shouldn't be.  I'll rework.

zephyr/Kconfig
gregshue: Project configurations outside the mcuboot repository need a Kconfig to control whether or not the MCUBoot application is built into the executable.  MCUBOOT_APP provides that control.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2022 Legrand North America, LLC.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+
+
+# Detect the platform reliably
+if(ZEPHYR_BASE)
```
de-nordic, line 10: Why not `if(NOT ZEPHYR_BASE)`

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+  message(FATAL_ERROR "ZEPHYR_BASE needs to be defined for Zephyr builds")
+endif()
+
+# Verify the module name is set correctly
+set(expected_module_name "mcuboot")
+if(NOT "${module_name}" STREQUAL ${expected_module_name})
+  message(FATAL_ERROR "A module configuration error has occurred.
+This module is named '${module_name}' when '${expected_module_name}' \
+is expected. 
+Verify `module.yml` contains the line 'name: ${expected_module_name}'.")
```
de-nordic, line 21: Can this be better formatted, looks jagged.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+endif()
+unset(expected_module_name)
+
+#Do not allow in source builds
```
de-nordic, line 25: Missing space after #.

zephyr/Kconfig
gregshue: It pre-existed in the moved file.  Since this was a "refactor" I wanted to minimize the number of unnecessary changes.  Cleaning out 'default n' is not a necessary change for this "refactor".

zephyr/Kconfig
gregshue: As above.

zephyr/Kconfig
gregshue: etc.

zephyr/subsys/mcuboot_app/CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
...
+zephyr_library_sources(
+  ${BOOT_DIR}/bootutil/src/fault_injection_hardening_delay_rng_mbedtls.c
+  )
+endif()
+
+if(CONFIG_SINGLE_APPLICATION_SLOT)
+zephyr_library_sources(
+  ${BOOT_DIR}/zephyr/single_loader.c
+  )
+zephyr_library_include_directories(${BOOT_DIR}/bootutil/src)
```
de-nordic, line 94: Missing indentation.

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
```diff
@@ -0,0 +1,30 @@
...
+.. code-block:: console
+
+  west build -b nrf52840dk_nrf52840 \
+    bootloader/mcuboot/samples/modules/mcuboot/mcuboot_external_config
+
+
+Sample Output
+=============
+
+None.
```
de-nordic, line 30: Waste of section, if there is nothing to present here.

zephyr/cmake/mcuboot_signature_key_file_generation.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+  if(IS_ABSOLUTE ${CONFIG_BOOT_SIGNATURE_KEY_FILE})
+    set(KEY_FILE ${CONFIG_BOOT_SIGNATURE_KEY_FILE})
+  elseif((DEFINED mcuboot_signature_file_dir) AND
+	 (EXISTS ${mcuboot_signature_file_dir}/${CONFIG_BOOT_SIGNATURE_KEY_FILE}))
+    set(KEY_FILE ${mcuboot_signature_file_dir}/${CONFIG_BOOT_SIGNATURE_KEY_FILE})
+  else()
+    set(KEY_FILE ${ZEPHYR_MCUBOOT_MODULE_DIR}/${CONFIG_BOOT_SIGNATURE_KEY_FILE})
+  endif()
+  message("MCUBoot bootloader key file: ${KEY_FILE}")
```
de-nordic, line 32: You can probably check here if the `${KEY_FILE}` exists and report that it could not be found.

zephyr/Kconfig
gregshue: From [Modules (External projects)](https://docs.zephyrproject.org/latest/develop/modules.html):

> Zephyr relies on the source code of several externally maintained projects in order to avoid reinventing the wheel and to reuse as much well-established, mature code as possible when it makes sense. In the context of Zephyr’s build system those are called modules. 

And the only specific requirement on all modules is:

> The module repository codebase shall include a module.yml file in a zephyr/ folder at the root of the repository.

@nordicjm wrote:
> mcuboot isn't a module, ...

Based on this definition and criteria, `mcuboot` was a downstream module before I ever started. It just wasn't represented that way in the build system.

> ... it's an application.

Consider:
From [Zephyr's Glossary entry for "application"](https://docs.zephyrproject.org/latest/glossary.html#term-application):

> The set of user-supplied files that the Zephyr build system uses to build an application image for a specified board configuration. It can contain application-specific code, kernel configuration settings, and at least one CMakeLists.txt file.

In order for mcuboot to live up to the reuse intent, the sources are not user-supplied any more than the drivers are user-supplied. 

The use case that prompted this change is my need to control and build boot code for my project without having to fork or change repositories in `zephyrproject-rtos`.  My "application" is simply one CMakeLists.txt, configuration settings files, and one 'keys.c' source file.  (FYI, all my other projects either provide an empty `main(void){}` or link in the weakly defined one provided by the kernel.)

From this perspective, the mcuboot app is a service that can be built into an application, just like a UI, or BLE beacon (or TCP/IP "ping"/"echo"/"data source").

> this is going to confuse people that have been used to mcuboot, the application, being at the top level, where other applications and zephyr best practice states to put it at.

People will be confused adjusting to the Zephyr definition of "application", which does not require any logic.  In fact, Zephyr kernel provides a weakly defined, default `void main(void)` definition, and the Zephyr build system works with the application providing only a zero-byte .c file (no symbols).

The top-level CMakeLists.txt still provides a title for the "application" configuration.

> If I configure hello_world without sysbuild, I see these options - why?

The better question is "For a certifiable image should `main()` contain any application-level functionality?"  For certifiable images every line of code needs to be traced back to requirements and verified that it meets the requirements.  The most efficient way to verify this is with automated unit testing. The Zephyr testing framework provides the definition of `void main(void)`, so the code under test must not. So, though `main()` _can_ contain logic, it probably shouldn't.

Separately, notice that nothing prevents building an image that combines outputing "hello world!" with booting an image - other than that both currently define the symbol "main()".  As we have noticed, neither should define that symbol.

Q: Would you prefer me to change the mcuboot source code to no longer define main() in this PR or in one that introduces a testsuite for main?

>  I can't enable mcuboot from [hello_world sample", and if I do enable these options, all I get is a broken build with compile errors about The flash driver lacks page layout support; falling back on hacks. which doesn't make sense

Actually, the broken build makes perfect sense.  You are trying to build a feature set for a target+drivers that do not support it. Why do you expect it would always work?

> And if I configure the application using sysbuild with mcuboot, these settings aren't even set, again, if I enable them, the build suddenly breaks with the same error.

Sysbuild is a convenience tool that supports building and flashing two images at once.  One _very_ reasonable scenario that should be supported by Sysbuild is to build and reflash an immutable boot image with a mutable one with a different feature set (e.g., one that drives a graphics display).  If Sysbuild can't support this scenario then it is lacking functionality.

zephyr/Kconfig
gregshue: This pre-existed in the moved file. In order to reduce conflict and minimize the size of the PR, I chose to only make necessary changes.  Cleaning out unnecessary `default n` would simply add noise to this PR.

zephyr/Kconfig
gregshue: As above

zephyr/Kconfig
gregshue: As above

zephyr/Kconfig
gregshue: etc.

zephyr/Kconfig
gregshue: In retrospective, this probably shouldn't be.  I'll fix it shortly.

zephyr/Kconfig
gregshue: Projects in other repositories need to be able to configure whether or not the MCUboot application is built into their image.  This is the primary use case that motivated this refactoring.

zephyr/CMakeLists.txt
gregshue: No particular reason.  What is your reasoning for preferring one over the other?

zephyr/CMakeLists.txt
gregshue: Of course.  Please point me to an example of what you would like.

zephyr/subsys/mcuboot_app/CMakeLists.txt
gregshue: This repository is not the Zephyr repository nor under the control of the Zephyr Project. Has the MCUboot project defined a source file formatting standard or policy?

zephyr/cmake/mcuboot_signature_key_file_generation.cmake
gregshue: This is pre-existing logic in the moved file.  This was not modified for the same rationale as not removing unnecessary `default n`.

zephyr/CMakeLists.txt
gregshue: Since this is a brand new file I am fine with fixing formatting. Has the MCUboot project defined a source file formatting standard or policy?

zephyr/cmake/mcuboot_signature_key_file_generation.cmake
gregshue: IIRC, the only changes I made in these files was to use `${ZEPHYR_MCUBOOT_MODULE_DIR}` instead of relative paths.

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
gregshue: Q: This really has a different purpose than a sample (conceptual, non-error handling) and a test suite (verifying functionality and error handling).  It is a real reference application (configuration, etc).  The closes guidance we have from Zephyr Project is to locate this in `$mcuboot/zephyr/app/` (from Zephyr's example-application), but that directory (tree) naming doesn't scale to support multiple applications.  Where should this reference application configuration go?

zephyr/Kconfig
gregshue: I **must** be able to build and flash the boot code for release independent of Sysbuild.  Running boot code without a valid image is also a necessary test case. Sysbuild is an efficiency enhancement, not a base case of usage.

zephyr/Kconfig
nordicjm: > From this perspective, the mcuboot app is a service that can be built into an application, just like a UI, or BLE beacon (or TCP/IP "ping"/"echo"/"data source").

No it is not, mcuboot is a bootloader, it provides services/libraries that an application can use and zephyr does, but mcuboot itself is not just something you add onto an application, it is a separate application altogether. Modules in zephyr in Kconfig control features for a build, for the mcuboot repository this would include the libraries that mcuboot includes as these can and are used from an application.

> The better question is "For a certifiable image should main() contain any application-level functionality?"

I really do not care about that, that sort of thinking is for the safety auditable branches of zephyr and other repositories, the main branches used by zephyr should not bend to the need of a safety critical/certifiable need if it is detrimental to the users that do not need that.

> Q: Would you prefer me to change the mcuboot source code to no longer define main() in this PR or in one that introduces a testsuite for main?

I think you are grossly misunderstanding how mcuboot works, mcuboot does and absolutely SHOULD have a main, that's how the bootloader works, it must operate without multi-tasking/threading support. This comment makes no sense about mcuboot.

> Actually, the broken build makes perfect sense. You are trying to build a feature set for a target+drivers that do not support it. Why do you expect it would always work?

So mcuboot doesn't work on nordic devices? Why does it build and work fine when I build mcuboot for these myself, or build an application and mcuboot for the same boards using sysbuild?

> Sysbuild is a convenience tool that supports building and flashing two images at once. One very reasonable scenario that should be supported by Sysbuild is to build and reflash an immutable boot image with a mutable one with a different feature set (e.g., one that drives a graphics display). If Sysbuild can't support this scenario then it is lacking functionality.

Sysbuild can and does support that (as in, sysbuild right now as it is in-tree). We currently have sysbuild configuring multiple images per core for a dual-core MCU (nRF53) and that is working fine.

nordicjm: And just to show how simple sysbuild is, I can do this right now in zephyr:
```
cd zephyr/samples/hello_world
mkdir _build
cd _build
cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -DSB_CONFIG_BOOTLOADER_MCUBOOT=y -DAPP_DIR=.. ../../../share/sysbuild
ninja flash
```

Which will build hello_world and mcuboot for my board, and will then flash both images to my board. I can even change mcuboot's configuration by altering the cmake like with e.g. `-Dmcuboot_CONFIG_LOG=n`

zephyr/Kconfig
nordicjm: > Projects in other repositories need to be able to configure whether or not the MCUboot application is built into their image. This is the primary use case that motivated this refactoring.

This is just wrong, mcuboot is a immutable bootloader, it is separate from the application and should not be combined, the point in mcuboot is it's small and does as little as possible to expose the least possible attack surface to ensure root of trust. If you're building it into your application, you might as well just disable security entirely at that point.

zephyr/CMakeLists.txt
de-nordic: https://github.com/mcu-tools/mcuboot/pull/1489/files#diff-00695564eb9dc6eac91ebf6608e7b236de613e6e0a329565c92df88056db7645R30, within this review.

zephyr/Kconfig
gregshue: > mcuboot is a bootloader, it provides services/libraries that an application can use and zephyr does, but mcuboot itself is not just something you add onto an application, ...

Please restate this following the definition of "application" from the Zephyr glossary. Either mcuboot is or is not an application.  If it can provide services/libraries to other applications, then it is more than an application (just like example-application is a manifest-module rather than just a module.

> the main branches used by zephyr should not bend to the need of a safety critical/certifiable need if it is detrimental to the users that do not need that.

Please review the [Zephyr Project statement,](https://www.zephyrproject.org/wp-content/uploads/sites/38/2020/09/CLEAN-LF-Zephyr-Charter-20200624-effective-20200901.pdf).  Item 1.b states:
> maintain an auditable code base, while taking advantage of community
participation; this auditable code base is open source;

Item 4e [f] states (reformatted, emphasis mine):
> The Security [Safety] Architect shall
> - serve on the TSC as a “security maintainer,” ["safety maintainer"] and 
> - is responsible for verifying that the **Project** is following the processes for selected
standards that the Security [Safety] Committee recommends and
> - that the Project is appropriately documenting features to move from the long term support (“LTS”)
code base to the auditable code base. 
> It is intended that Maintainers will seek the sign-off from the Security [Safety] Architect prior to implementing significant changes to the LTS code base **where such changes may result in significant changes in
functionality between the LTS and auditable code bases**.

The above items state that developers **should** bend to the need of a safety critical/certifiable need unless working on functionality or design that has been granted an exception by both the Safety Architect and the Security Architect.       

> I think you are grossly misunderstanding how mcuboot works

Actually I understand it better than most, having heavily designed and refactored single- and multi-stage boot code for an MCU-based product line for high volume, long-lived devices.

> it must operate without multi-tasking/threading support

That is one of the use cases.  Another requires that that it work when in a multi-tasking/threading environment (e.g., when MCUboot zephyr port is using the USB interface for serial recovery, or any situation a where an end user integrates a multi-threaded solution for any of the support APIs like UI for CGD).

> mcuboot does and absolutely SHOULD have a main,

You are confusing design with implementation.  MCUboot does provide top-level logic that must automatically run once the lower layers are initialized.  The implementation does not have to provide the symbol `main()`.  It could be achieved by providing the symbol `mcuboot_main()` which is called by either Zephyr's `main()`, ztest suite, or the last executed `SYS_INIT()`.

> So mcuboot doesn't work on nordic devices?

Not what I said.  The build error shows that the subset of configuration settings used were insufficient and/or incorrectly set for the target. I never claimed that integrating MCUBOOT_APP into the hello_world sample only required setting the one Kconfig.  

BTW, if you look at the commit comments you will see that I explicitly verified my changes built correctly for nrf52840dk_nrf52840 as well as for all the platforms pulled in by twister for the referenced test suites.

> Sysbuild can and does support that (as in, sysbuild right now as it is in-tree).

As an end user, I need to build and release my boot image independent of building other images. MCUboot needs to support building this straight from CMake/ninja without using west or sysbuild.

Separately, working on MCUboot issue #1499 I found the Zephyr testing framework/structure does not work with the current mcuboot zephyr port build files, but it does work with these changes. This is completely independent of sysbuild.

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
de-nordic: > This is a standard section in Zephyr documentation. Explicitly stating there is no output is useful information.

Fair point. 

Someday I am going to write sample returning literally `None.`, to make this section confusing.

zephyr/Kconfig
nordicjm: > The above items state that developers should bend to the need of a safety critical/certifiable need unless working on functionality or design that has been granted an exception by both the Safety Architect and the Security Architect.

No it doesn't, it talks about introducing changes into LTS branches, this is nothing to do with LTS.

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
de-nordic: > Q: This really has a different purpose than a sample (conceptual, non-error handling) and a test suite (verifying functionality and error handling). It is a real reference application (configuration, etc). The closest guidance we have from Zephyr Project is to locate this in `$mcuboot/zephyr/app/` (from Zephyr's example-application), but that directory (tree) naming doesn't scale to support multiple applications. Where should this reference application configuration go?

Good question. Maybe we should have some place for "reference board configurations", or templates. We probably lack some configuration management here.

zephyr/cmake/mcuboot_signature_key_file_generation.cmake
de-nordic: OK

zephyr/subsys/mcuboot_app/CMakeLists.txt
de-nordic: No, but it does not mean that indentations have to be chaotic.
This is moved code anyway, so it is how it is. No change required.

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
de-nordic: The problem with these reference configurations is that they do not explain much, regarding selected Kconfig options or DTS definitions.

zephyr/Kconfig
gregshue: > mcuboot is a immutable bootloader

That is one use case. Another necessary one is for mcuboot to be a secondary (and mutable) bootloader - needed when the immutable block of the device is too small for the full bootloader functionality.

> it is separate from the application and should not be combined,

You cannot logically claim that there is only one application, that it is separate from the bootloader, AND that the bootloader is an application.

> the point in mcuboot is it's small and does as little as possible to expose the least possible attack surface to ensure root of trust.

This description is not unique to mcuboot. It is also true for many certifiably secure firmware images.  The unique point of bootloaders is to load and execute another image. Doing "as little as possible" varies based on user needs.  Some need serial recovery. Some need it to run from ram. Some need it to support multiple partitions.  Some need it to prevent version downgrades. Each feature combination could be supported by unique source files, but it is more efficient to support them with reusable, configurable code. Using Zephyr's definition of "application", each configuration combination to build mcuboot is a distinct application in its own right.

BTW, it is actually very reasonable to implement outputing/logging "hello_world" as a trivial subsystem launched by `SYS_INIT()`, and integrate this into a bootloader. It makes even more sense if this subsystem existed in within a proprietary module and logged the executable version rather than the string "hello world". ;-)

zephyr/samples/modules/mcuboot/mcuboot_external_config/nrf52840dk_nrf52840_cc310_ecdsa.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=y
```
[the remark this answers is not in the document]
gregshue: This is located in the same place as in `$mcuboot/boot/zephyr/`.  Changing this was not necessary for this minimal refactor (just like cleaning out unnecessary `default n` in Kconfigs).

zephyr/samples/modules/mcuboot/mcuboot_external_config/README.rst
gregshue: According to Zephyr's glossary, reference **configurations** are a type of "application", whether or not it also contains local logic or data. The configurations could simply be a configuration of overlays located in other modules.  

The explanations of the Kconfig settings and the overlays needs to be outside the applications.  There needs to be a module-level location for reusable overlays.  We can already use `\${ZEPHYR_<moduleNameUpper>_MODULE_DIR}` in our CONF_FILE assignments _before_ calling `find_project()`.

I would like to see Zephyr Project identify a tree parallel to `samples/` and `tests/` for multiple non-sample, non-test applications.  This could be named `apps/` or `applications/`.  It should not be named with a singular noun (e.g `app/`).

zephyr/subsys/mcuboot_app/CMakeLists.txt
gregshue: Agreed.  It should be cleaned up in a different PR.

zephyr/CMakeLists.txt
gregshue: Fair point, given this is a new file.  I'll clean it up shortly.

zephyr/Kconfig
gregshue: In order to release a certified bootloader the processes and tools to build that must exist in auditable branch, and exceptions to this in LTS should be signed off by the Safety & Security Architects.  Until that is signed off then the feature should exist in LTS too. This includes building artifacts verifying functionality in the mcuboot top-level logic.

gregshue: > this is nothing to do with LTS

I agree that this has nothing to do with introducing changes to the LTS branch, but it does have to do with whether the existing mcuboot build files work with the Zephyr LTS branch build features that meet my user needs.  Sysbuild does not meet my user needs and does not address the need to build correctly with the Zephyr testing structure.

nvlsianpu (changes requested):
[DNM] This need be considered by zephyr-rtos/zephyr maintainers

tejlmand (changes requested):
This is a very intrusive change in the build system in my opinion.

It moves a lot of code around, apparently because of this motivation:
> Fundamentally these reveal that in order to support an externally configurable, composable solution the MCUBoot source build file cannot contain the find_package(Zephyr ...) command. In other words, the reusable sources build file (CMakeLists.txt) must be in a different directory than the project configuration build file (CMakeLists.txt). Similarly, the Kconfig files within the module sources tree cannot also be a top-level project Kconfig definition file (otherwise causing recursive sourcing of Kconfig.zephyr).

Zephyr already today supports the use of `APPLICATION_CONFIG_DIR` which means that any user of MCUboot can create a custom configuration dir outside MCUboot for local tweaking of MCUboot.

For example by creating `app_conf_test` folder and pass that information to the build system, like this:
```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -Bbuild boot/zephyr/ -DAPPLICATION_CONFIG_DIR=(pwd)/app_conf_test
...
-- Found BOARD.dts: /projects/github/ncs/zephyr/boards/arm/nrf52840dk_nrf52840/nrf52840dk_nrf52840.dts
-- Found devicetree overlay: /projects/github/ncs/bootloader/mcuboot/app_conf_test/app.overlay
-- Generated zephyr.dts: /projects/github/ncs/bootloader/mcuboot/build/zephyr/zephyr.dts
-- Generated devicetree_generated.h: /projects/github/ncs/bootloader/mcuboot/build/zephyr/include/generated/devicetree_generated.h
-- Including generated dts.cmake file: /projects/github/ncs/bootloader/mcuboot/build/zephyr/dts.cmake
Parsing /projects/github/ncs/bootloader/mcuboot/boot/zephyr/Kconfig
Loaded configuration '/projects/github/ncs/zephyr/boards/arm/nrf52840dk_nrf52840/nrf52840dk_nrf52840_defconfig'
Merged configuration '/projects/github/ncs/bootloader/mcuboot/app_conf_test/prj.conf'
Merged configuration '/projects/github/ncs/bootloader/mcuboot/app_conf_test/boards/nrf52840dk_nrf52840.conf'
...
```

Notice how devicetree overlay and config files are being picked up from the `app_conf_test/` and `app_conf_test/boards` folders in above log messages.

One can even create a custom CMakeLists.txt file and do customization in there before sourcing `boot/zephyr/CMakeLists.txt` file if such is needed.

For time being, I see this PR as a `I don't understand / like the current structure therefore I would like this structure instead`. type of PR.

Maybe I have missed something fundamental regarding this PR, so if that is the case I urge @gregshue to instead create much smaller PRs with a clear objective of what is the exact goal.
For example a PR that take a subpart of the CMake code into a dedicate library for re-usability with a clear purpose.

Note, I have only made a few code comments, but I have discovered a lot of places that I cannot approve as they are proposed currently, but imo it make little sense to point out all those places atm, as the overall state of PR is far from going into such details.

Note: regarding the use of `APPLICATION_CONFIG_DIR` there is a flaw in MCUboot which I have addressed here: #1522

nvlsianpu:
@gregshue I see your points on that mcuboot is a downstream project for zephyr. Having a fork for mcuboot in zephyr-rtos GH doesn't determine how the maintenance is preformed. I want to have 1:1 relation, so fare this approach have been working fine. On every synchronization of zephyr-rtos/mcuboot with the origin the code diff is expected to be empty.

I design this PR with DNM as this change have critical severity for the zephyr-project and every projects which leverage the mcuboot zephyr port.

As you can see we have already review from build system expert.

gregshue:
Nothing in this change breaks the zephyr builds. As Ana's has said, Zephyr needs to make decisions independent of downstream concerns. In fact, this change came about because the current implementation does not fully support the range of user needs supported by Zephyr. 

Specifically, What changes here are critically severe?

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    
+    This module is named '${module_name}' when '${expected_module_name}' is expected. 
+    Verify `module.yml` contains the line 'name: ${expected_module_name}'.
+    ------------------------------------------------------------------------")
+endif()
+unset(expected_module_name)
+
+# Do not allow in source builds
+set(CMAKE_DISABLE_SOURCE_CHANGES ON)
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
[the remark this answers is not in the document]
gregshue: Everybody begins knowing nothing. I'll remove them.

tejlmand:
> Nothing in this change breaks the zephyr builds.

In which case any PR should just blindly be approved, or what is your point ?

This PR is not simply trying to improve current code, but moving code around irrespective of the current design.
Although part of existing code can surely be improved, the move proposed in this PR just makes things even more messy.

> As Ana's has said, Zephyr needs to make decisions independent of downstream concerns.

and that is a carte-blanche to request changes directly in the mcuboot repository, or what is your point ?
Try to compare the change @nashif is requesting, which is a single config line in a `prj.conf` to the major changes requested in this PR.

> In fact, this change came about because the current implementation does not fully support the range of user needs supported by Zephyr.
> 

Exactly which use-case is not properly handled by the description provided here:
https://github.com/mcu-tools/mcuboot/pull/1489#pullrequestreview-1186232501

and why can such use-case not be done in a less-intrusive way ?

gregshue:
> Maybe I have missed something fundamental regarding this PR, ...

You have.

> I urge @gregshue to instead create much smaller PRs with a clear objective of what is the exact goal.

Good suggestion. Look at PR #1524 

MCUboot is the manifest repository of my workspace. Building this this PR from the root of my workspace with either current mcuboot `main` branch or after manually applying the changes of #1522 generates the same build failures.

'''
west build -p always -b nrf52840dk_nrf52840 mcuboot/zephyr/samples/modules/mcuboot/mcuboot_external_config/
'''

gives two sets of warnings (even though building mcuboot/boot/zephyr builds cleanly):
- (1x) warning: MBEDTLS_CFG_FILE (defined at modules/mbedtls/Kconfig:50, modules/mbedtls/Kconfig:50) was assigned the value 'mcuboot-mbedtls-cfg.h' but to the value ''. Check these unsatisfied dependencies: ((MBEDTLS_BUILTIN && MBEDTLS) || (MBEDTLS_BUILTIN && MBEDTLS && 0)) (=n).
- (12x) warning: attempt to assign the value 'n' to the undefined symbol ...

I cannot even reach the point of verifying my local board settings cascade with those in APPLICATION_CONFIG_DIR.

> MCUboot is a Zephyr module containing the MCUboot application.

The MCUboot zephyr module also contains the mcuboot bootutil library.

> The current design builds a bootloader independent of the application.
With your design changes, then you start building and linking the application together with the bootloader in the process, and thus they start to be tightly coupled with all the cons of address spacing and separating the two parts in the linker script.

This is where you misunderstood what this does.  This still builds and flashes the bootloader independently of the application. According to the zephyr.map files, this refactoring only does the following w.r.t. linking:
- `app/libapp.a` is reduced to have `keys.c.obj`.  All the other object files go into `zephyr/libzephyr.a`.
- `mcuboot/libmcuboot_util.a` becomes `mcuboot/subsys/mcuboot_util/libmcuboot_util.a`.

> This is an application specific CMakeLists.txt file suddenly being moved into Zephyr module CMake processed code.

AFAICT the only part of the original CMakeLists.txt that makes this an application CMakeLists.txt is 'find_project(Zephyr ...)'.  The rest could be refactored into `include()`ed files, or into subsystems.  From my many years of experience in this area, this will be much easier to maintain if the "application" logic is refactored as a subsystem.

> just moving the code really doesn't improve anything.

... except for fixing the use cases that the current configuration doesn't support.

> In which case any PR should just blindly be approved, or what is your point ?

The answer didn't identify what changes "have critical severity for the zephyr-project".  It doesn't seem to be a build issue.

> but moving code around irrespective of the current design.

The current design doesn't meet the breadth of user needs (e.g., mine, as shown by the samples).

> the move proposed in this PR just makes things even more messy.

I actually find it far more orderly, consistent, scalable, and applicable to other downstream modules than the current design.

> and why can such use-case not be done in a less-intrusive way ?

It may be able to be done in a less intrusive way, but this is a way that follows existing patterns in Zephyr and example-application, introduces more alignment with the Zephyr build/configuration system, and can be applied in general to downstream repositories.

Show me another way to meet this use case and then let's evaluate the approaches based on overall merit.

nordicjm:
> MCUboot is the manifest repository of my workspace

Then your project is completely wrong from the start.

tejlmand:
> > I urge @gregshue to instead create much smaller PRs with a clear objective of what is the exact goal.
> 
> Good suggestion. Look at PR #1524

Thanks.

This PR actual reveals that this PR (#1489) originates from a lack of understanding on how MCUboot is integrated into Zephyr and how it works.

gregshue:
> did you read the comments I wrote and tried it out ?

Apparently not well enough.  Now that I have here's what I found:

- I successfully built from my mcuboot-manifested workspace with the following command:
'''
west build -p always -b nrf52840dk_nrf52840 mcuboot/boot/zephyr \
-DAPPLICATION_CONFIG_DIR=`(pwd)`/mcuboot/samples/zephyr/mcuboot_external_configuration
'''
- I adding a local CMakeLists.txt (include(${CMAKE_CURRENT_LIST_DIR}/../../../boot/zephyr/CMakeLists.txt)` and running the following command:
'''
west build -p always -b nrf52840dk_nrf52840 mcuboot/samples/zephyr/mcuboot_external_configuration
'''
The build FAILED, reporting warnings on:
  1. Kconfigs being assigned non-'n' values but having value 'n'
  2. attempt to assign the value 'n' to the undefined symbol ...;
 
- I copied `$mcuboot/boot/zephyr/Kconfig` locally, fixed up the `rsource "Kconfig.serial_recovery"` to include the relative path, and attempted the command immediately above. The build failed with "No such file or directory: TINYCRYPT_DIR: '/home/zephyr/mcuboot_review/mcuboot/samples/ext/tinycrypt/lib'

**Clearly, `$mcuboot/boot/zephyr/CMakeLists.txt` is not designed to be included`**

> So use ExternalProject_Add() or add_subdirectory().

Notice that both require a path. **This is a problem for my use case**, where the top-level CMakeLists.txt and configurations exist in a different repository than mcuboot or zephyr, and no command line options other than project path is allowed, and I am not allowed to change the mcuboot repository. (In my case even the board needs to be set through a file under SCM control.)

AFAICT, your suggestions _still_ do not support my use case (building from a config entirely in a different repo).

> You can create your own custom CMakeLists.txt and fully control MCUboot any way you like

I can, but I shouldn't have to. I am following a widely-presumed use case.

> But https://github.com/mcu-tools/mcuboot/pull/1489 is a PR where you propose yet one more variant in customization.

Unlike APPLICATION_CONFIG_DIR, the pattern of customization in this PR is not new to the Zephyr Project. Rather, this PR **refactors** the implementation to properly support **Zephyr's standard pattern of customization**. It also aligns with [Zephyr's definition of "application"](https://docs.zephyrproject.org/latest/glossary.html), which does not require any application-unique logic.

How about this: Please show me how build and configure a bootloader build from a CMakeLists.txt and prj.conf under `example-application` (e.g., `west build -b nrf52840dk_nrf52840 $example-application/mcuboot_external_configuration`), because I don't think you can do it without modifying `zephyr/` or `mcuboot/`, or putting restrictions on the workspace directory tree.

tejlmand:
> How about this: Please show me how build and configure a bootloader build from a CMakeLists.txt and prj.conf under example-application (e.g., west build -b nrf52840dk_nrf52840 $example-application/mcuboot_external_configuration), because 

This is a PR review, so basically such a request should be posted at the proper channel, but as it just requires 4 CMake commands `cmake_minimum_required()`, `find_package()`, `include(),` and `ExternalProject_Add()` then I have done you the courtesy of creating a repo with a freestanding application.
If it can work as a freestanding application, then it can also work as a workspace application.
You need to add this commit on top of `mcuboot` #1522 .
If this fails for you, please read the getting started on how to properly install Zephyr, or take any further questions to the proper channel (Discord).

See here:
https://github.com/tejlmand/oot_mcuboot_showcasing

```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -Bbuild .
Loading Zephyr module(s) (Freestanding): west zephyr_module
...
-- Build files have been written to: /tmp/oot_mcuboot_external_configuration/build
$ ninja -C build
ninja: Entering directory `build'
[5/8] Performing configure step for 'mcuboot'
...
-- Found devicetree overlay: /tmp/oot_mcuboot_external_configuration/app.overlay
...
Loaded configuration '/projects/github/ncs/zephyr/boards/arm/nrf52840dk_nrf52840/nrf52840dk_nrf52840_defconfig'
Merged configuration '/tmp/oot_mcuboot_external_configuration/prj.conf'
Merged configuration '/tmp/oot_mcuboot_external_configuration/boards/nrf52840dk_nrf52840.conf'
...
271/271] Linking C executable zephyr/zephyr.elf
           FLASH:       40092 B        48 KB     81.57%
[8/8] Completed 'mcuboot'
```

> I don't think you can do it without modifying zephyr/ or mcuboot/, or putting restrictions on the workspace directory tree.

As I have said before, I don't think you have fully understood the power of the current design and infrastructure of the build system.

Can we close this topic now ?

gregshue:
Thank you for the example. I did not want to assume that `${ZEPHYR_<moduleNameUpper}_MODULE_DIR} was always going to be available for use in `ExternalProject_Add()`.

I was able to successfully build this stand-alone, when added into `example-application` as the manifest repositoryl

BUT ...

The build failed when I also passed in `-DCONF_FILE=...`. Here are the results I got:
- "-DCONF_FILE=prj.conf": picks up local `prj.conf`; does NOT pick up either local or remote `boards/nrf52840dk_nrf52840.conf`; did not generate the proper size image.
- "-DCONF_FILE=${ZEPHYR_MCUBOOT_MODULE_DIR}/boot/zephyr/prj.conf": picks up `$mcuboot/boot/zephyr/prj.conf`; does NOT pick up either local or remote `boards/nrf52840dk_nrf52840.conf`; did not generate the proper size image
- "-DCONF_FILE=${ZEPHYR_MCUBOOT_MODULE_DIR/boot/zephyr/prj.conf;prj.conf": picks up `$mcuboot/boot/zephyr/prj.conf`; does NOT pick up local `prj.conf` does not pick up remote and local `boards/nrf52840dk_nrf52840.conf`; did not generate the proper size image.

So ... both approaches suffer from common failures in the build system forcing the duplication of config files/settings.

Now we're down to comparing the actual approaches to support reuse of logic currently private to an application.  We're making headway!

First, notice that Zephyr's glossary definition of [application](https://docs.zephyrproject.org/latest/glossary.html) does not require any logic or data unique to it. As you know, the build system works fine if an application provides a zero-byte C source file. We need to discuss the reuse of _logic_ (e.g., library/subsystem) separate from the reuse of an _application_ (configuration and build files).

Second, in general no one is able to predict when logic in one application will need to be reused in another. Having experienced the evolution and growth of a software product line over many generations. I am a witness that functionality we _never_ expected to be used in a different application needed to be refactored out for sharing. The strategic direction is to expect any/all logic will end up being shared across multiple applications.

Third, the mcuboot module already exposes bootutil as a subsystem through `module.yml`. It is already configurable through Kconfigs. The bootutil subsystem/library already has a zephyr unique CMakeLists.txt. None of these concepts are introduced with this PR. In fact, this PR is refactoring the `boot/zephyr/` build files to follow these existing concepts and patterns to expose the logic to be reused in the existing/expected way.

Fourth, it sounds like `APPLICATION_CONFIG_DIR` was introduced as a hotfix for a specific use case. The pattern implied by this functionality is to attempt to reuse logic and build files that may not have been designed for reuse. (See above issues.) I expect this tactic was straight forward but strategically leads to a more fragile ecosystem than one where all code to reuse is created and maintained as intentionally reusable code.

With all that I have learned about your recommendations and experienced myself over the years, it seems to me having MCUboot refactor out reusable code into subsystems (as done in this PR) is the wiser solution for both the short and long term.

Evaluating the complexity of this PR in terms of touched file count is really inappropriate. Whether one file is moved or an entire directory is should not matter. This PR is simply:
1. splitting files apart based on fragments of reuse
2. locating the reusable portions in the expected location for reusable content
3. providing a standard sample/test case for automatic (re)verification
4. putting all this in the organizational structure that is naturally understood/expected/debuggable by all working in the Zephyr ecosystem.

Honestly, your steadfast advocacy of `APPLICATION_CONFIG_DIR` sounds like `I don't understand / like the current zephyr platform architecture therefore I'll promote this structure instead.` This will likely cause you problems when the build system has to be "qualified" for the auditable branch. You, my friend, will have to tackle the whole of software engineering to do that. You might as well take the free course [LFD-116](https://training.linuxfoundation.org/training/software-engineering-basics-for-embedded-systems-lfd116/) sooner rather than later.

Now, are you ready to approve this PR so we can move on to filling out mcuboot test cases?

tejlmand:
> "-DCONF_FILE=prj.conf": picks up local prj.conf; does NOT pick up either local or remote boards/nrf52840dk_nrf52840.conf; did not generate the proper size image.

Did you fail to read this:
> If this fails for you, please read the getting started on how to properly install Zephyr, or take any further questions to the proper channel (Discord).

that's how `-DCONF_FILE` is documented to work.
Using CONF_FILE is opting-out of the automatic process.
go read the docs:
https://docs.zephyrproject.org/latest/build/kconfig/setting.html#the-initial-configuration
or read the comment I gave you here:
https://github.com/mcu-tools/mcuboot/pull/1524#issuecomment-1323478044

tejlmand:
>  it sounds like APPLICATION_CONFIG_DIR was introduced as a hotfix for a specific use case.

Quite the opposite, it's generic and scalable as it works for any app, from hello world, to hci_rpmsg, and even to MCUboot.
And can be used for any board.

Your proposal is MCUboot specific and tailored exactly to your needs.
It makes application specific Kconfigs available in the general Kconfig tree, not only MCUboot app but all applications now has those settings available.

Image each and every sample in Zephyr want to have their settings available in the global Kconfig space.
What a mess and nightmare that will become.

gregshue:
> that's how -DCONF_FILE is documented to work.

From the perspective of multiple stakeholders in Zephyr, your design is not valid no matter how much it has been verified. See below.

(The design also doesn't fully meet your specification, as it didn't properly parse a list of multiple prj.conf.)

> Quite the opposite, it's generic and scalable as it works for any app, from hello world, to hci_rpmsg, and even to MCUboot.
And can be used for any board.

The design of CONF_FILE rules do not support the [SOLID design principles](https://www.digitalocean.com/community/conceptual-articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design), specifically the Liskov Substitution Principle. Thus it is not scalable to multiple levels of project derivation. In case you hadn't noticed it the Zephyr ecosystem architecture is a client-server organized architecture of objects. Notice how:
- In https://docs.zephyrproject.org/latest/develop/api/design_guidelines.html
  - the first bullet is related to providing an instance reference to the object related to the callback,
  - the third bullet a reference to is originator data.
- the SYS_INIT() interface hooks in the object constructor

> Your proposal is MCUboot specific and tailored exactly to your needs.

How does aligning with the general pattern of Zephyr modularity make it **specific and tailored** to my needs? Try again.

>It makes application specific Kconfigs available in the general Kconfig tree, not only MCUboot app but all applications now has those settings available.

Your approach does the same thing in a less obvious way.  The oot configuration and CMakeLists.txt constitutes a separate application per Zephyr's definition. ANY/all applications already have them available. Once the code is reused the settings are no longer application specific. My approach exposes the Kconfigs controlling the reusable logic in exactly the same way as happens in other downstream modules - following the solution supported by Zephyr.

Anyone need to be able to combine the MCUboot app logic with a ztest suite. By the Zephyr definition this is a different application. I also need to be able to combine MCUboot app logic with other proprietary logic - without touching the MCUboot repository. Fundamentally, all certifiable logic will be shared with test applications. The typical case will be sharing that logic with a ztest application.

> Imagine each and every sample in Zephyr want to have their settings available in the global Kconfig space.
> What a mess and nightmare that will become.

It doesn't have to be, and Zephyr header include guards have already illustrated the solution.  The Kconfigs simply need to be prefixed with the moduleName+pathToDefinition. This pattern of solution has been practiced for decades.

tejlmand:
> From the perspective of multiple stakeholders in Zephyr, your design is not valid no matter how much it has been verified. 

Take it to Discord.
That functionality pre-dates my time with Zephyr, it goes all the way back to Kbuild days (way before CMake).

If you read the docs then there might be a feature doing what you ask, but either rtfm or ask politely on Discord.

PR #1501, opened 2022-10-28
boot: zephyr: prefer swap move if scratch_partition not enabled

str4t0m:
With the exception of nrf targets `BOOT_SWAP_USING_SCRATCH` mode was still the preferred algorithm.

Changing the preferred mode in cases where no `scratch_partition` is defined will allow successfully building mcuboot for such boards without the need for any board specific overlays.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -197,7 +197,7 @@ config BOOT_VALIDATE_SLOT0_ONCE
 config BOOT_PREFER_SWAP_MOVE
 	bool "Prefer the newer swap move algorithm"
 	default y if SOC_FAMILY_NRF
-	default n
+	default y if !$(dt_nodelabel_enabled,scratch_partition)
 	help
 	  If y, the BOOT_IMAGE_UPGRADE_MODE will default to using
 	  "move" instead of "scratch".  This is a separate bool config


str4t0m:
Will need approval for the workflows to run in this case.

boot/zephyr/Kconfig
```diff
@@ -197,6 +197,7 @@ config BOOT_VALIDATE_SLOT0_ONCE
 config BOOT_PREFER_SWAP_MOVE
 	bool "Prefer the newer swap move algorithm"
 	default y if SOC_FAMILY_NRF
+	default !$(dt_nodelabel_enabled,scratch_partition)
 	default n
```
utzig, line 201: Does this line have to stay after the new one?

str4t0m: I slightly updated the line and removed the `default n`.
In case a scratch_partition is available it will implicitly be set to `default n` and could theoretically be overwritten later on.

https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults

str4t0m:
The build failure is unrelated.
Only sample `sample.mcumgr.smp_svr.bt` fails on `nrf52dk_nrf52832` due limited resources (flash overflow).

nvlsianpu (approved):

nvlsianpu (approved):

PR #1502, opened 2022-10-31
boot_serial: use image number only from first SMP upload request packet

pepe2k:
According to the [SMP protocol documentation](https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_1.html), `image number` value in `image upload request` is optional and can be included only in packet with `off` (data offset) set to `0` (first packet in upload request).

Current implementation of the `direct image upload` in MCUboot first, on every received `image upload request` packet, sets the image number to `0` and then, if available, reads value from received CBOR data. In extreme case, when the client side of SMP follows the documentation and provides target `image number` only in the initial packet, this leads to destruction of both images (if SMP client requests upload of image number == 2, only data from first packet will be written to selected slot and everything else will end in the first one).

This change fixes the problem, making MCUboot implementation aligned with Zephyr documentation of the SMP protocol by considering image number only from first (off == 0) `image upload request` SMP packet.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -349,7 +349,9 @@ bs_upload(char *buf, int len)
     size_t img_chunk_off = SIZE_MAX;    /* Offset of image chunk within image  */
     uint8_t rem_bytes;                  /* Reminder bytes after aligning chunk write to
                                          * to flash alignment */
-    int img_num;
+    int img_num_tmp = 0;                /* Temp variable for image number */
+    static int img_num = 0;
+    bool img_num_requested;
     size_t img_size_tmp = SIZE_MAX;     /* Temp variable for image size */
     const struct flash_area *fap = NULL;
     int rc;
@@ -363,8 +365,6 @@ bs_upload(char *buf, int len)
                                          */
     static struct flash_sector status_sector;
 #endif
-
-    img_num = 0;
 
     /*
      * Expected data format.
@@ -384,11 +384,13 @@ bs_upload(char *buf, int len)
         goto out_invalid_data;
     }
 
+    img_num_requested = false;
     for (int i = 0; i < upload._Upload_members_count; i++) {
         struct Member_ *member = &upload._Upload_members[i]._Upload_members;
         switch(member->_Member_choice) {
             case _Member_image:
-                img_num = member->_Member_image;
+                img_num_tmp = member->_Member_image;
+                img_num_requested = true;
                 break;
             case _Member_data:
                 img_chunk = member->_Member_data.value;
@@ -412,6 +414,15 @@ bs_upload(char *buf, int len)
          * Offset must be set in every block.
          */
         goto out_invalid_data;
+    }
+
+    /* Use image number only from packet with offset == 0. */
+    if (img_chunk_off == 0) {
+        if (img_num_requested) {
+            img_num = img_num_tmp;
+        } else {
+            img_num = 0;
+        }
     }
 
 #if !defined(MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD)


pepe2k:
Rebased on latest `main`.

boot/boot_serial/src/boot_serial.c
```diff
@@ -414,6 +413,14 @@ bs_upload(char *buf, int len)
         goto out_invalid_data;
     }
 
+    /* According to the specification, only requests with 'off' set to '0' can
+     * contain image number. Use image number value provided only in first SMP
+     * upload request packet.
+     */
```
de-nordic, line 419: You do not have to mention doc here, just state that image number is only taken from packets with offset == 0.
If we would mention documentation, in comments, as a reason to every change, the code would get really wired. Unless documentation requires something that is counter-intuitive, you do not have to put justification citing documentation in comments.

pepe2k: @de-nordic makes sense. I fixed that and rebased on latest 'main'.

Thanks for your comment!

boot/boot_serial/src/boot_serial.c
```diff
@@ -414,6 +413,11 @@ bs_upload(char *buf, int len)
         goto out_invalid_data;
     }
 
+    /* Use image number only from packet with offset == 0. */
+    if (img_chunk_off == 0) {
+        img_num = img_num_tmp;
```
nordicjm, line 418: Doesn't this then cause an issue - if someone uploads an image with n set to 2, it uploads to that slot, if they then upload another image without specifying n, it will upload again to slot 2, but should fall back to the default of img_num being 0?

pepe2k: @nordicjm

Yes, you're right... that's another side of the generic issue I didn't consider. I think adding `img_num` reset to zero at the end of a successful upload would cover this case but it's not perfect. Another extreme case is what should happen if upload for slot 2 is canceled on the client side (for whatever reason) and then a new upload without provided slot number starts. It would end up in slot 2 from previous upload.

```diff
@@ -573,20 +573,23 @@ bs_upload(char *buf, int len)
                 rc = MGMT_ERR_EUNKNOWN;
                 goto out;
             }
 #endif
             rc = BOOT_HOOK_CALL(boot_serial_uploaded_hook, 0, img_num, fap,
                                 img_size);
             if (rc) {
                 BOOT_LOG_ERR("Error %d post upload hook", rc);
                 goto out;
             }
+
+            /* Upload done, reset image number for consecutive uploads */
+            img_num = 0;
         }
     } else {
     out_invalid_data:
         rc = MGMT_ERR_EINVAL;
     }
 
 out:
     BOOT_LOG_INF("RX: 0x%x", rc);
     zcbor_map_start_encode(cbor_state, 10);
     zcbor_tstr_put_lit_cast(cbor_state, "rc");
```

A better solution I can think about is a bool flag stating whether image number field was provided in upload request. If yes and upload offset == 0, use that value for image number, otherwise set image number to 0.

```diff
@@ -351,6 +351,7 @@ bs_upload(char *buf, int len)
                                          * to flash alignment */
     int img_num_tmp = 0;                /* Temp variable for image number */
     static int img_num = 0;
+    bool img_num_requested;
     size_t img_size_tmp = SIZE_MAX;     /* Temp variable for image size */
     const struct flash_area *fap = NULL;
     int rc;
@@ -383,11 +384,13 @@ bs_upload(char *buf, int len)
         goto out_invalid_data;
     }
 
+    img_num_requested = false;
     for (int i = 0; i < upload._Upload_members_count; i++) {
         struct Member_ *member = &upload._Upload_members[i]._Upload_members;
         switch(member->_Member_choice) {
             case _Member_image:
                 img_num_tmp = member->_Member_image;
+                img_num_requested = true;
                 break;
             case _Member_data:
                 img_chunk = member->_Member_data.value;
@@ -415,7 +418,11 @@ bs_upload(char *buf, int len)
 
     /* Use image number only from packet with offset == 0. */
     if (img_chunk_off == 0) {
-        img_num = img_num_tmp;
+        if (img_num_requested) {
+            img_num = img_num_tmp;
+        } else {
+            img_num = 0;
+        }
     }
 
 #if !defined(MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD)
```

pepe2k:
Rebased on latest `main`, included fix for extreme case mentioned by @nordicjm and proposed in https://github.com/mcu-tools/mcuboot/pull/1502#discussion_r1114058288.

pepe2k:
Closing this one in favor of updated one with a regression fix: https://github.com/mcu-tools/mcuboot/pull/1735

Cheers,
Piotr

PR #1507, opened 2022-11-08
boot_serial: support fragmentation for outgoing SMP packets

pepe2k:
The mcumgr packet sent over serial should be fragmented into frames of up to 127 bytes, including 2-bytes frame start header and 1-byte for terminating newline [1], resulting in up to 124 bytes for the base64 encoded payload.

Current implementation doesn't fulfill above requirement and transmits single frames, without considering their length. This change introduces support for fragmenting as defined in 'SMP over console' specification.

[1] https://github.com/apache/mynewt-mcumgr/blob/master/transport/smp-console.md


pepe2k:
Rebased on latest `main`.

nordicjm (approved):

pepe2k:
Rebased on latest `main`.

nordicjm (approved):

de-nordic (approved):

PR #1508, opened 2022-11-08
Feature/1499 mcuboot util unit testing

gregshue:
This PR creates an initial unit test based on the Zephyr test framework and build system for a portion of bootutil common code. The purpose of this PR is to establish a scalable pattern for thoroughly testing common MCUboot functionality within the Zephyr environment.  Future PRs will be needed to extend the common code testing coverage.


gregshue:
What is the controversy?

tejlmand (commented):
I wonder, doesn't MCUboot has its own tests that ensure the functionality of MCUboot code ?

As MCUboot is user by many projects, then I don't see why MCUboot should use the Zephyr test framework, unless of course there are some re-use benefits.
But so far the PR message / commit messages doesn't really go into such details.

If those tests are considered integration tests with Zephyr, as to ensure that MCUboot correctly works inside Zephyr, then I believe such tests belongs in Zephyr itself, so that Zephyr can ensure that each update of MCUboot revision does indeed match the expectations / requirements of Zephyr.

tejlmand (changes requested):
Rejecting because this PR depends on #1489 which is not approved so this should not be considered until an agreement is reached on #1489.

gregshue:
> I wonder, doesn't MCUboot has its own tests that ensure the functionality of MCUboot code ?

It seems not at the level needed for certification.

> As MCUboot is user by many projects, then I don't see why MCUboot should use the Zephyr test framework, unless of course there are some re-use benefits.

Ideally the reused MCUboot code would be tested with a common framework.  The port-specific code should be tested somewhere.  In either case
, for certification the tests need to be run against the target hardware with the target build system and toolchain - or justification provided for each test not run there.  At this point in time it is actually easier and more expedient to flesh out mcuboot tests for the Zephyr environment.

> If those tests are considered integration tests with Zephyr, as to ensure that MCUboot correctly works inside Zephyr, then I believe such tests belongs in Zephyr itself, ...

Most of the tests are internal unit/integration tests.  These belong in mcuboot.

Since mcuboot is **downstream** of zephyr, even integration tests belong in MCUboot to avoid circular dependencies. This type of **must** be supported for users like Brix and myself who need to not touch the Zephyr release we are basing our products on top of.

> ... so that Zephyr can ensure that each update of MCUboot revision does indeed match the expectations / requirements of Zephyr.

Following ISO 12207 Systems and Software Engineering - Software life cycle processes, you have these requirement dependencies backwards.  The requirements come from the users to the system to the system elements to the product line platform.  MCUboot executable is a system element, whereas Zephyr is the product line platform. MCUboot will eventually identify requirements it has of any RTOS, build system, etc.  Zephyr will need to identify how it is meeting those requirements.

PR #1513, opened 2022-11-12
Feature/1512 zephyr doc build

### LINKED ISSUE #1512: zephyr: documentation build w/ mcuboot repo content
User Need:
As a task-based contracted firmware engineer,
I need to generate comprehensive documentation to go along with the solution I deliver to customers,
so that I can minimize the support I need to give when my customer eventually tries to use, port, and/or modify the delivered code.

Proposal:
For the Zephyr port, add documentation configuration files to support building Zephyr-specific MCUboot documentation following the Zephyr document generation processes and tools.

gregshue:
Fixes #1512 .


nvlsianpu (changes requested):
[DNM] for now change is controversial

gregshue:
What is the controversy?

tejlmand (changes requested):
Why should the MCUboot docs change ?
There is already a Readme for MCUboot with Zephyr here: https://github.com/mcu-tools/mcuboot/blob/main/docs/readme-zephyr.md
which follows the same pattern as any other project:
https://github.com/mcu-tools/mcuboot/blob/main/docs/readme-espressif.md
https://github.com/mcu-tools/mcuboot/blob/main/docs/readme-mbed.md
https://github.com/mcu-tools/mcuboot/blob/main/docs/readme-mynewt.md
etc.

As it's seen then MCUboot is used by many projects, and starting to have such integration documentation for each project in MCUboot is going to be a never ending task.

MCUboot provides a nice overview here:
https://github.com/mcu-tools/mcuboot/#mcuboot-how-tos

If a project wants to adjust the look of the MCUboot doc pages, then I think it should be the responsibility of that project.
See for example how nRF Connect SDK compiles several docsets into one, including Zephyr and MCUboot:
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/mcuboot/index-ncs.html
image

--------

One controversial commit is 8e667d398c79db55684c30764a85c8b77986886d .

That commit is a pure copy of files from Zephyr ,why ?

At some point in future, Zephyr will progress and then those files are no longer in sync with their Zephyr counterparts.
Who will ensure those files are kept in sync ?

Duplicating files between repos should generally be avoided / minimized, and in above commit we are talking about copying 14 files / >5200 LOC.

I fail to see what extra knowledge readers are getting with this commit:
https://github.com/mcu-tools/mcuboot/pull/1513/commits/84a451691d00f85fc669546ca734bad6f7d27f6f
compared to the existing MCUboot docs and project specific docs linked above.

I propose to reject this PR due to the duplication of massive loc, and the fact that the extra doc provides no extra information to the reader.

The request for #1512:
> I need to generate comprehensive documentation to go along with the solution I deliver to customers,
so that I can minimize the support I need to give when my customer eventually tries to use, port, and/or modify the delivered code.

is already possible in today's solution without this amount of duplicated files, as it has been shown in above screenshot and link.

gregshue:
> is already possible in today's solution without this amount of duplicated files, as it has been shown in above screenshot and link.

That screenshot is for NCS documentation system, not Zephyr's documentation system.  IIRC from Carles' comments a while ago, Nordic documentation system had to introduce substantial changes from Zephyr to support what you reference. AFAICT, this is the simplest way to extend __Zephyr's documentation system for **workspace-level** documentation__.

> the duplication of massive loc

The duplications are there because some content needed to be tailored and the current system required content to come over in relatively large blocks.  I reused whatever I didn't expect to need tweaking. I copied only what I must, given the current Zephyr documentation system design.

> I fail to see what extra knowledge readers are getting with this commit <> compared to the existing MCUboot docs and project specific docs linked above

The documentation process is also involved in requirements tracing, so it is not only about "knowledge".

> At some point in future, Zephyr will progress and then those files are no longer in sync with their Zephyr counterparts.
Who will ensure those files are kept in sync ?

All the copied data is now custom to the receiving repository.  It is up to the maintainers to decide if they care if it is in sync with the original source.  If they care, it is up to them to do the update.  This is **downstream** of Zephyr, so Zephyr Project isn't concerned with this.

Issue #1516, opened 2022-11-14
imgtool: Improvement to P256 key generation feature


adeaarm:
Hello,

I think that there is a bug currently in the private key that gets generated by the imgtool when the `--minimal` option is passed to the tool: the ASN.1 encoding is not valid as not all the correct bytes are removed nor all the length fields are updated to reflect the reduced size of the key once you remove the OPTIONAL public part.

At the same time, we wanted to add support in the tool for `TraditionalOpenSSL` key formats, i.e. the format described in RFC 5915 [1] / SEC1 [2] documentation, which normally ends up producing smaller keys than the PKCS#8 format that is currently used by the tool.

[1] https://www.rfc-editor.org/rfc/rfc5915
[2] https://www.secg.org/sec1-v2.pdf

PR #1517, opened 2022-11-14
imgtool: Improve ECDSA key generation tool

adeaarm:
This patch improves imgtool in the area of key generation of ECDSA keys for P256, more precisely:

* fixes a bug in the ASN.1 encoding of the "minimal" type of key for PKCS8 format
* Add support to the tool for both PKCS8 format and RFC5915/SEC1, i.e. the traditionalOpenSSL format. A P256 minimal key in this format takes 51 bytes. This is done by adding a command line option called -f or --format which can be either pkcs8 or openssl. The default behaviour remains unchanged so if no option is passed a pkcs8 key will be produced.

This is the proposed fix for issue #1516 

Signed-off-by: Antonio de Angelis <antonio.deangelis@arm.com>


scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
     def _get_public(self):
         return self.key.public_key()
 
-    def _build_minimal_ecdsa_privkey(self, der):
+    def _build_minimal_ecdsa_privkey(self, der, format):
+        if (format == serialization.PrivateFormat.OpenSSH):
```
utzig, line 106: You don't need to use parentheses here.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
+            print(os.path.basename(__file__) + \
+                ": Warning: --minimal is supported only for PKCS8 or TraditionalOpenSSL formats")
+            return bytearray(der)
         '''
         Builds a new DER that only includes the EC private key, removing the
         public key that is added as an "optional" BITSTRING.
         '''
```
utzig, line 113: This was a docstring for the function, if you want to leave it here changed it to comments.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
-            raise ECDSAUsageError(EXCEPTION_TEXT)
-        b[offset_SEQ + 1] -= len_PUB
-        offset_OCT_STR = 27
-        if b[offset_OCT_STR] != 0x04:
-            raise ECDSAUsageError(EXCEPTION_TEXT)
-        b[offset_OCT_STR + 1] -= len_PUB
-        if b[0] != 0x30 or b[1] != 0x81:
-            raise ECDSAUsageError(EXCEPTION_TEXT)
-        b[2] -= len_PUB
+        if (format == serialization.PrivateFormat.PKCS8):
```
utzig, line 115: No need for parentheses.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
+                raise ECDSAUsageError(EXCEPTION_TEXT)
+            b[offset_OCT_STR + 1] -= len_PUB
+            if b[0] != 0x30 or b[1] != 0x81:
+                raise ECDSAUsageError(EXCEPTION_TEXT)
+            # as b[1] has bit7 set, the length is on b[2]
+            b[2] -= len_PUB
+            if b[2] < 0x80:
+                b = b[0:1] + b[2:]
+
+        elif (format == serialization.PrivateFormat.TraditionalOpenSSL):
```
utzig, line 136: No need for parantheses.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
+            offset_OCT_STR = 27
+            if b[offset_OCT_STR] != 0x04:
+                raise ECDSAUsageError(EXCEPTION_TEXT)
+            b[offset_OCT_STR + 1] -= len_PUB
+            if b[0] != 0x30 or b[1] != 0x81:
+                raise ECDSAUsageError(EXCEPTION_TEXT)
+            # as b[1] has bit7 set, the length is on b[2]
+            b[2] -= len_PUB
+            if b[2] < 0x80:
+                b = b[0:1] + b[2:]
```
utzig, line 134: I don't think this is needed unless you are using another ASN.1 decoder. The first byte being `0x81` is just telling there is one byte with the length. Although for strict DER this might be more *correct*, but then the Mbed TLS ASN.1 decoder we use doesn't seem to have issues with it the way it is. I will do some testing of this tomorrow or the next day, if it works I have no problem with merging it. PS: you could also have done `del b[1]`!

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
-        b[offset_SEQ + 1] -= len_PUB
-        offset_OCT_STR = 27
-        if b[offset_OCT_STR] != 0x04:
-            raise ECDSAUsageError(EXCEPTION_TEXT)
-        b[offset_OCT_STR + 1] -= len_PUB
-        if b[0] != 0x30 or b[1] != 0x81:
-            raise ECDSAUsageError(EXCEPTION_TEXT)
-        b[2] -= len_PUB
+        if (format == serialization.PrivateFormat.PKCS8):
+            offset_PUB = 68 # where the context specific TLV starts (tag: 0xA1)
```
utzig, line 116: Please add another space before the `#`. see: https://peps.python.org/pep-0008/#inline-comments

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
+            if der[offset_PUB] != 0xa1:
+                raise ECDSAUsageError(EXCEPTION_TEXT)
+            len_PUB = der[offset_PUB + 1] + 2 # take into account also the 0xA1 0x44 Type-Length bytes
```
utzig, line 119: Add another space before `#`.

scripts/imgtool/keys/ecdsa.py
```diff
@@ -102,37 +102,57 @@ def generate():
...
+            b = bytearray(der[:offset_PUB]) # remove the context specific TLV with the PUB key
```
utzig, line 120: Add another space before `#`.

scripts/imgtool/keys/ecdsa.py
utzig, line 119: Good one (`+ 2`)!

utzig (commented):
I suggest breaking the changes where you add the filename in the messages into its own commit, and squashing the one adding the default into the main one. What will RFC5915/SEC1 encoding be used for?

scripts/imgtool/keys/ecdsa.py
adeaarm: You're right, this is not strictly necessary, the resulting ASN.1 encoding is still compliant, just 1 byte bigger than necessary. But given that I was making changes in this area, I thought of making it strictly adhere to the spec. It shouldn't break any compatibility as both encodings are valid. Will change to del b[1], looks better.

adeaarm:
> I suggest breaking the changes where you add the filename in the messages into its own commit, and squashing the one adding the default into the main one. What will RFC5915/SEC1 encoding be used for?

Will rework breaking into different commits as suggested, and assess the comments below as part of this rework. We are working on extending support for ECDSA to the P384 curve and support for a new cryptography backend based on PSA Crypto APIs [1], for some of the crypto-related operations done in the bootloader. We plan to contribute these patches back once ready, in a separate PR.
PKCS#8 key format, unless there are some specific requirements (i.e. PrivateKey encryption) is more or less just a wrapping of the SEC1 representation, which normally ends up being smaller, hence we thought having the option in the tool to export the key in that (smaller) format could be useful, especially when you use higher order curves where the key sizes are bigger.

[1] https://armmbed.github.io/mbed-crypto/html/

scripts/imgtool/keys/ecdsa.py
adeaarm: Done. I run PEP8 so I fixed all these issues related to parentheses. I also submitted a separate commit where I fixed the remaining PEP8 warnings in the files that I touched in this PR.

scripts/imgtool/keys/ecdsa.py
adeaarm: It was done by error, moved it back to the correct place.

scripts/imgtool/keys/ecdsa.py
adeaarm: Done as part of PEP8 validation.

scripts/imgtool/keys/ecdsa.py
adeaarm: Done as part of PEP8 validation.

scripts/imgtool/keys/ecdsa.py
adeaarm: Done as part of PEP8 validation.

scripts/imgtool/keys/ecdsa.py
adeaarm: Done as part of PEP8 validation.

scripts/imgtool/keys/ecdsa.py
adeaarm: Done as part of PEP8 validation.

utzig:
@adeaarm Tested this with Zephyr, works well, thanks for the fixes. Btw, I didn't know but `format` is a Python reserved word: https://docs.python.org/3.8/library/functions.html#format. Not a problem here though. I will merge once the CI passes.

utzig:
> @adeaarm Tested this with Zephyr, works well, thanks for the fixes. Btw, I didn't know but `format` is a Python reserved word: https://docs.python.org/3.8/library/functions.html#format. Not a problem here though. I will merge once the CI passes.

Oops, signed email checks failed because of a case mismatch. You have to either fix your `user.email` or the sign-off-by for you commits.

adeaarm:
> > @adeaarm Tested this with Zephyr, works well, thanks for the fixes. Btw, I didn't know but `format` is a Python reserved word: https://docs.python.org/3.8/library/functions.html#format. Not a problem here though. I will merge once the CI passes.
> 
> Oops, signed email checks failed because of a case mismatch. You have to either fix your `user.email` or the sign-off-by for you commits.

Thanks, did not realize I had camel case in my user.email. I have changed them now, it should be ok.

adeaarm:
Fixes #1516

PR #1521, opened 2022-11-20
boot: zephyr: allow timeout based recovery with CDC ACM

pepe2k:
This makes it possible to enable timeout (`BOOT_SERIAL_WAIT_FOR_DFU`) mode for the serial recovery when using `CDC ACM` based serial device. This was runtime tested on `nRF52840-Dongle`.

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -148,7 +148,7 @@ config BOOT_SERIAL_ENCRYPT_EC256
 
 config BOOT_SERIAL_WAIT_FOR_DFU
 	bool "Wait for a prescribed duration to see if DFU is invoked by receiving a mcumgr comand"
-	depends on BOOT_SERIAL_UART
+	depends on BOOT_SERIAL_UART || BOOT_SERIAL_CDC_ACM
 	help
 	  If y, MCUboot waits for a prescribed duration of time to allow
 	  for DFU to be invoked. The serial recovery can be entered by receiving any


matteoscordino:
I also runtime tested this on a custom board, and I can confirm it works. It would be nice to have this merged, so it can trickle downstream to the Nordic SDK

pepe2k:
@nvlsianpu, @de-nordic, I'm kindly asking for a review, it's very simple but useful change.

Cheers,
Piotr

nordicjm (commented):
Not sure of the implications of this, does mcuboot clean up USB if it's been enabled if using this mode before booting the application? Otherwise it would be a guess if a supported board will work with this configuration (some boards might not do a USB reset when the application boots) @d3zd3z

nvlsianpu (approved):

de-nordic:
> Not sure of the implications of this, does mcuboot clean up USB if it's been enabled if using this mode before booting the application? Otherwise it would be a guess if a supported board will work with this configuration (some boards might not do a USB reset when the application boots) @d3zd3z

@desowin any thoughts?

desowin:
> > Not sure of the implications of this, does mcuboot clean up USB if it's been enabled if using this mode before booting the application? Otherwise it would be a guess if a supported board will work with this configuration (some boards might not do a USB reset when the application boots) @d3zd3z
> 
> @desowin any thoughts?

I am not too sure about what the original concern is. There is really no "moving over" the USB state from bootloader to application, like there could be with e.g. UART or other interface. Before application boots, USB is disabled, which from host perspective is equivalent to user disconnecting the device https://github.com/mcu-tools/mcuboot/blob/8f289ba5f9610214a70af06615ff756ca5a74025/boot/zephyr/main.c#L195-L198

de-nordic (approved):

nordicjm (approved):

PR #1524, opened 2022-11-21
New sample: mcuboot_external_config

gregshue:
Created a sample of how to configure MCUBoot build from an external repository.  This sample provides no `main()` definition and reuses all logic from MCUBoot and other repositories.  The functionality matches the existing functionality of mcuboot/boot/zephyr.

The sample is located within `$mcuboot/zephyr/samples/modules/mcuboot/` so that the directory path matches that used in Zephyr for samples of external modules.  The directory name `mcuboot_external_config` intentionally begins with the module name to help avoid futures conflict with samples that may be defined in other repositories. Together, this pattern supports easy integration of this sample documentation with sample documentation in Zephyr.

NOTE: This commit required copying board and .conf files from mcuboot/boot/zephyr because the _current_ zephyr build system (about v3.2.0) does not fully support using APPLICATION_CONFIG_DIR with a list of files/directories in CONF_FILE and with boards/ beneath those settings.  When APPLICATION_CONF_DIR was set to reference mcuboot/boot/zephyr/ and CONF_FILE referenced APPLICATION_CONFIG_DIR then here, the boot config did not pick up a board-specific configuration setting from APPLICATION_CONFIG_DIR, resulting in FLASH going from ~40000 to ~45000 bytes.

  This is captured in Zephyr Issue #51621.

Attempted Verification:

1. (Pass) west build -p always -b nrf52840dk_nrf52840 \ mcuboot/zephyr/samples/modules/mcuboot/mcuboot_external_config/
2. (Pass) ./zephyr/scripts/twister --testsuite-root mcuboot/zephyr/samples/

Signed-off-by: Gregory Shue <gregory.shue@legrand.com>


zephyr/samples/modules/mcuboot/mcuboot_external_config/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Add a common dts overlay necessary to ensure mcuboot is linked into,
+# and fits inside, the boot partition. (If the user specified a
+# DTC_OVERLAY_FILE on the CMake command line, we need to append onto
+# the list).
+set(DTC_OVERLAY_FILE
+  "${DTC_OVERLAY_FILE};${APPLICATION_CONFIG_DIR}/dts.overlay"
+  CACHE STRING "" FORCE
+  )
```
tejlmand, line 22: setting CONF_FILE and DTC_OVERLAY_FILE may be a easy, convenient and acceptable way for a custom downstream application, but it's not a principle that should be used used in upstream samples.

A number of reasons, such as it's not scalable.
Users cannot opt-out of the forced `dts.overlay`.
Other flaws are seen here: #1522

zephyr/samples/modules/mcuboot/mcuboot_external_config/keys.c
```diff
@@ -0,0 +1,209 @@
+/*
```
tejlmand, line 1: This is not a file you should add.

It already included by the application and can be considered a skeleton file.
When building MCUboot with Zephyr the key is specified using `CONFIG_BOOT_SIGNATURE_KEY_FILE` and extracted into `autogen-pubkey.c` which is included in the build.
Ref:
https://github.com/zephyrproject-rtos/mcuboot/blob/cfec947e0f8be686d02c73104a3b1ad0b5dcf1e6/boot/zephyr/CMakeLists.txt#L295-L307

tejlmand (changes requested):
I start to see your problem.

It's a fundamental lack of understanding of how the system is organized.

Instead of requesting big changes to the existing system, could you please try to read the docs and understand how the existing infrastructure works.
Also discord is a good place to ask for help or clarification on topics when docs are unclear, or raise issues if you believe you have found a bug, such as #1522 .

Much better way than to open PRs like #1489 and spawn big discussions.

The `APPLICATION_CONFIG_DIR` is intended to allow users to create out-of-tree configuration directories for in-tree samples.
And when using this setting, you are in complete control of the sample.

The `OVERLAY_CONFIG` are suitable for smaller adjustments.

I pushed a branch here:
https://github.com/tejlmand/mcuboot/tree/mcuboot/review_1524
with commit:
c3faa70dd666e1e4ffe98d588e4ecf2492cb4006

That show-cases how configuration can be kept independent of the sample, and can be built as:
```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -Bbuild boot/zephyr/ -DAPPLICATION_CONFIG_DIR=`pwd`/samples/zephyr/mcuboot_external_configuration
...
-- Found devicetree overlay: /projects/github/ncs/bootloader/mcuboot/samples/zephyr/mcuboot_external_configuration/app.overlay
...
Merged configuration '/projects/github/ncs/bootloader/mcuboot/samples/zephyr/mcuboot_external_configuration/prj.conf'
Merged configuration '/projects/github/ncs/bootloader/mcuboot/samples/zephyr/mcuboot_external_configuration/boards/nrf52840dk_nrf52840.conf'
Configuration saved to '/projects/github/ncs/bootloader/mcuboot/build/zephyr/.config'
...
MCUBoot bootloader key file: /projects/github/ncs/bootloader/mcuboot/root-rsa-2048.pem
-- Configuring done
-- Generating done
-- Build files have been written to: /projects/github/ncs/bootloader/mcuboot/build
```

Notice how all conf and overlay files are picked up correctly according to `APPLICATION_CONFIG_DIR`.

Configuration and bu8ild passes as expected, and users can add custom board configs if they so wish.

The key is extracted from `CONFIG_BOOT_SIGNATURE_KEY_FILE` and generated into `autogen-pubkey.c`.

Please read the docs:
https://docs.mcuboot.com/readme-zephyr (Especially section: `Extracting the public key`)
https://docs.zephyrproject.org/latest/develop/application/index.html#application-configuration

This is **not** a PR to be merged.
It is an attempt to present the intended use-case.

tejlmand:
also note, config files are being picked up according to the description here:
https://docs.zephyrproject.org/latest/build/kconfig/setting.html#the-initial-configuration

> 1. If CONF_FILE is set, the configuration file(s) specified in it are merged and used as the application configuration. CONF_FILE can be set in various ways:
>      ....
> 2. Otherwise if CONF_FILE is set, and a single configuration file of the form `prj_<build>.conf` is used, then if file `boards/<BOARD>_<build>.conf` exists in same folder as file `prj_<build>.conf`, the result of merging `prj_<build>.conf` and `boards/<BOARD>_<build>.conf` is used.
> 3. Otherwise, `prj_<BOARD>.conf` is used if it exists in the application configuration directory.
> 4. Otherwise, if `boards/<BOARD>.conf` exists in the application configuration directory, the result of merging it with prj.conf is used.


Notice the `otherwise`.
This means as soon as you manually specify a list of `CONF_FILES` then the selection of config files returns at step 1., and never reach step 4.
The intention behind that is that a user specifying `-DCONF_FILES` manually are in full control and has opt-out of any automatic selection of board specific config files.
So when using `CONF_FILES` as setting, you are responsible for passing **all** files you want to be using in the build.

gregshue:
> I start to see your problem.
>
> It's a fundamental lack of understanding of how the system is organized.

I understand why you think that, but that's usually not it.

The [Zephyr Project Mission Statement, et., al.](https://zephyrproject.org/wp-content/uploads/sites/38/2020/09/CLEAN-LF-Zephyr-Charter-20200624-effective-20200901.pdf) discusses "secure", "safe", "auditable codebase", and "certification artifacts". These imply a safe/secure software engineering process will be followed. These processes (e.g., foundationally based on ISO 12207) prescribe a requirements-driven, top-down design, beginning with Stakeholder requirements & end user needs. Fulfilling Zephyr's mission and vision statements is dependent on engineering the software (e.g., having to qualify our build system).

Most of the participants in Zephyr are thinking about what _can_ be done (e.g., APPLICATION_CONFIG_DIR), while I am thinking about what _should_ be done to fulfill the larger context/foreseeable future, and why (e.g., refactor so that mcuboot is configurable and sustainable exactly like every other project and module). Most come from the perspective of "developers/programmers" while I come from the perspective of "systems & software engineer/architect".

My fundamental problem is when stakeholder requirements & user needs are not driving the solution towards consistency, reusability, and extensibility.  I want to submit user needs & requirements, but Zephyr Project is not ready for that yet.  Instead, I submit PRs for addressing my user needs and to help align the solution for a scalable, composable, reusable, extensible platform.

In this particular case the Zephyr Project could serve itself and the community better by adding a `project` configuration/CMakeLists.txt into `example-application` for building a bootloader without TFM. This could serve as a test case and be a very useful reference for those of us needing to _reuse_ an upstream repository at a specific SHA.

gregshue:
@tejlmand If you want to understand the difference between "developer" and "engineer" then let's talk after you take the following free course:

https://training.linuxfoundation.org/training/software-engineering-basics-for-embedded-systems-lfd116/.

PR #1525, opened 2022-11-22
Revert commits 900da1386523405f6b77071665638667bf782dd9 and e82b370ca57a539db59c463c1f4c21d8113640b7

nordicjm:
Reverts 2 commits:

    Revert "Explicitly set module name in `zephyr/module.yml`"
    
    This reverts commit e82b370ca57a539db59c463c1f4c21d8113640b7.

    Revert "Add west.yml to make manifest-module."
    
    This reverts commit 900da1386523405f6b77071665638667bf782dd9.

The reason for these commits is that forcing a module name is not desired when that module can and is used in multiple projects and sources - for example with zephyr-integrated projects such as TF-M where the utility code is used as part of the build - if the name is forced then this will cause conflicts with module that themselves pull mcuboot in for parts they need, if they do not use the in-tree mcuboot sources.
The other issue is that this wrongly treats mcuboot as a library depending upon zephyr whilst zephyr treats mcuboot as an application - these are at odds with one another. mcuboot itself is an application, a bootloader, it has libraries that applications can use, but it itself should not be pulling zephyr in, zephyr does and should pull mcuboot in. If an application wants to use parts of mcuboot with zephyr, it can do similar to what TF-M does or it can have it's own downstream (not upstream) manifest repository created which pulls in zephyr (importing mcuboot) and the custom application which will include the mcuboot libraries and/or code.


tejlmand (approved):

d3zd3z (changes requested):
> mcuboot itself is an application, a bootloader, it has libraries that applications can use, but it itself should not be pulling zephyr in, zephyr does and should pull mcuboot in.

I'm not sure I can say I agree with this. MCUboot is an application, and when used as an application it does "pull in" zephyr, as it is a zephyr app.

But as I understand from Greg's needs (and mine as well) is that it is not possible currently to have my own directory (that I can build with `west build ./mypath`) that builds the mcuboot application. I'd like to see being able to do this, while still satisfying what Nordic's needs are here.

The comment also suggests you are using multiple versions of mcuboot at the same time. Is that the issue with setting the module name? Could one of those copies have a change in it that changes the name in the module.

d3zd3z:
> if the name is forced then this will cause conflicts with module that themselves pull mcuboot in for parts they need, if they do not use the in-tree mcuboot sources.

Is this something that is actually happening and needed? TFM, for example, uses the same source directory of mcuboot that is otherwise used by zephyr and stuff. It seems unusual to me to want to have multiple versions of mcuboot in the same working tree.

nordicjm:
> I'm not sure I can say I agree with this. MCUboot is an application, and when used as an application it does "pull in" zephyr, as it is a zephyr app.

That is an application. Applications use code from zephyr.

> But as I understand from Greg's needs (and mine as well) is that it is not possible currently to have my own directory (that I can build with west build ./mypath) that builds the mcuboot application. I'd like to see being able to do this, while still satisfying what Nordic's needs are here.

@tejlmand has posted many examples to greg including a whole sample repo that shows it can be done as-is today

> The comment also suggests you are using multiple versions of mcuboot at the same time. Is that the issue with setting the module name? Could one of those copies have a change in it that changes the name in the module.

Why should the name of the module be forced? TF-M pulls in mcuboot files for the libraries, they shouldn't need to fork mcuboot and then have to maintain their own fork just to be able to do that

> Is this something that is actually happening and needed? TFM, for example, uses the same source directory of mcuboot that is otherwise used by zephyr and stuff. It seems unusual to me to want to have multiple versions of mcuboot in the same working tree.

It is, see lines 108 and 124 of https://github.com/nrfconnect/sdk-nrf/blob/main/west.yml#L108

tejlmand:
> But as I understand from Greg's needs (and mine as well) is that it is not possible currently to have my own directory (that I can build with `west build ./mypath`) that builds the mcuboot application. I'd like to see being able to do this, while still satisfying what Nordic's needs are here.

@d3zd3z please see this oot example which is even a freestanding application (would need to be moved to a `west` workspace in order to be built using `west build`, but that would just require to add that repository to a manifest:
https://github.com/tejlmand/oot_mcuboot_showcasing
https://github.com/tejlmand/oot_mcuboot_showcasing/blob/oot_mcuboot/CMakeLists.txt

```
$ cmake -GNinja -DBOARD=nrf52840dk_nrf52840 -Bbuild .
Loading Zephyr module(s) (Freestanding): west zephyr_module
...
-- Build files have been written to: /tmp/oot_mcuboot_external_configuration/build
$ ninja -C build
ninja: Entering directory `build'
[5/8] Performing configure step for 'mcuboot'
...
-- Found devicetree overlay: /tmp/oot_mcuboot_external_configuration/app.overlay
...
Loaded configuration '/projects/github/ncs/zephyr/boards/arm/nrf52840dk_nrf52840/nrf52840dk_nrf52840_defconfig'
Merged configuration '/tmp/oot_mcuboot_external_configuration/prj.conf'
Merged configuration '/tmp/oot_mcuboot_external_configuration/boards/nrf52840dk_nrf52840.conf'
...
271/271] Linking C executable zephyr/zephyr.elf
           FLASH:       40092 B        48 KB     81.57%
[8/8] Completed 'mcuboot'
```

Please note that this show-case requires #1522.

tejlmand:
> > if the name is forced then this will cause conflicts with module that themselves pull mcuboot in for parts they need, if they do not use the in-tree mcuboot sources.
> 
> Is this something that is actually happening and needed? TFM, for example, uses the same source directory of mcuboot that is otherwise used by zephyr and stuff. It seems unusual to me to want to have multiple versions of mcuboot in the same working tree.

Partly true, but that is only because Zephyr asks it to do so here and only if asked to do so (default is `y`):
https://github.com/zephyrproject-rtos/zephyr/blob/78cf49e0fd4520f043c47cec8035ba80fe2472af/modules/trusted-firmware-m/CMakeLists.txt#L255-L257

and Zephyr can only do so because the same revision of MCUboot works with both TF-M and Zephyr.
TF-M has it's own build of TF-M and will actually download it as part of the build if the path is not already given, see this:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/887798f6e67203e6a77059a86edf41bb136a3c0b/lib/ext/mcuboot/CMakeLists.txt#L24
and the build of MCUboot inside TF-M is done here:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/master/secure_fw/partitions/firmware_update/bootloader/mcuboot/CMakeLists.txt

Although the current state is that we have a MCUboot revision that is working with both Zephyr and TF-M in the revision used by Zephyr, then we cannot rely on the fact that this will always be true.
As an example, when TF-M downloads a specific mbedTLS version they apply a number of patches on top:
https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/lib/ext/mbedcrypto

Although Nordic is active in both Zephyr and TF-M it doesn't mean we can ensure that a similar situation will never happen in TF-M, that is TF-M requires revision A, which is not compatible with revision B required by a regular MCUboot build in Zephyr.
In such a case there are only two options:
- Allow TF-M to download it's required revision at build time (generally not accepted in Zephyr, as repos should be in the manifest)
- Add the TF-M required revision of MCUboot to the manifest file (which then would result in two MCUboot revisions).

As a reference, this sample:
https://docs.zephyrproject.org/latest/samples/tfm_integration/psa_firmware/README.html
will actually build MCUboot using the TF-M build system, not the Zephyr port.

In NCS we actively points to two MCUboot revisions in our manifest:
Regular Zephyr used revision:
https://github.com/nrfconnect/sdk-nrf/blob/8c63b618bf8d0f5fe3669dcacdd1221bc0cb0ae4/west.yml#L108-L111
TF-M revision:
https://github.com/nrfconnect/sdk-nrf/blob/8c63b618bf8d0f5fe3669dcacdd1221bc0cb0ae4/west.yml#L124-L127

I'm not into the details on the exact reasons so I would need to consult the security experts, but iirc part of it has to do with security and what has been examined and guaranteed to be secure wrt. TF-M.

But the main point is that there may be reasons for having two (or more) revisions of MCUboot in a manifest file.
When no module name is given in `zephyr/module.yml`, then the module will be named based on the folder, and you can refer to them using that name (as has always been the case for mcuboot).

For the Zephyr revision, then MCUboot should be placed in a folder named `mcuboot`, as it is today, so that Zephyr can refer to this folder using `ZEPHYR_MCUBOOT_MODULE_DIR`.
TF-M itself doesn't use the Zephyr build infrastructure, but if the TF-M MCUboot is placed in a folder named `tfm_mcuboot`, then the Zephyr build system can pass that information to the TF-M build system using `-DMCUBOOT_PATH=${ZEPHYR_TFM_MCUBOOT_MODULE_DIR}`.

But that is no longer possible because of: #1492.
Hence the reason I believe that commit should be reverted.

tejlmand:
@d3zd3z although you haven't requested specifically a description of the problem with https://github.com/mcu-tools/mcuboot/commit/900da1386523405f6b77071665638667bf782dd9 then let me explain where I see issues.

First, the simple user facing part.
What uses what / who verifies and tests that things are working together.
Should I start at MCUboot or should I start out at Zephyr repository ?

If only having a manifest in Zephyr, then things are much clearer to the end-user.
Having it in MCUboot, then I would expect some documentation in MCUboot as to how I should clone MCUboot and use it as a manifest repo. No such documentation exists today.

As for the manifest file itself, then the biggest issue I see is that it is pointing to revision `main`.
That means if I use MCUboot as my manifest repository the if I do a `west update` today, then I get one Zephyr revision, but if I do it tomorrow I get another Zephyr revision checked-out.
That means I have no way of predicting what I get.
It breaks bisectability because Zephyr revision `main` is too dynamic.

In Zephyr, shas or tags are used as revisions in Zephyr's manifest file.
And during release testing it is verified that the MCUboot revision pointed to by that SHA or tag is working correctly.

But I can not use the manifest in MCUboot at a given revision and be sure that the corresponding Zephyr revision is working, cause it points at `main`:
https://github.com/mcu-tools/mcuboot/blob/08a716dc4ed71a96105053880c959145a9da547b/west.yml#L25

Let's assume we change this into a SHA instead of `main` ?
Then who will maintain a proper revision in MCUboot manifest file ?
Who will test and validate that the Zephyr revision is actually working as expected ?

I believe the safest is to keep current approach of having Zephyr pointing at a specific MCUboot revision and ensuring this is working.
This gives a clear responsibility.
But of course, if anyone has the bandwidth to maintain and verify the opposite direction (MCUboot manifest -> Zephyr revision), that can surely be handled, but using `main` as revision is not a good approach.

So at current state I propose to revert 900da1386523405f6b77071665638667bf782dd9.

nordicjm:
@d3zd3z can this be merged?

nordicjm:
@d3zd3z this can be merged

d3zd3z (approved):

PR #1528, opened 2022-11-24
fih: Hardening of the fault injection countermeasures

mimok:
FIH countermeasures should be hardened, this PR brings the following changes:
- Success and error return values are hardcoded and returned through `fih_ret` type
- When a function needs to return an interger, `fih_int` type is used
- fih_eq() and fih_not_eq() are replaced by macro
- a loop in `loader.c` is hardenned to prevent bypass
- `fih_rc` variables are now securely set (affectation followed by a check)
- `fih_rc` are now explicitly set to `FIH_FAILURE` after a `FIH_NOT_EQ` test

Some functionnal tests have been done but additional tests may be needed.


boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -93,9 +93,13 @@ extern "C" {
 #ifndef MCUBOOT_FIH_PROFILE_OFF
 #define FIH_POSITIVE_VALUE 0x1AAAAAAA
 #define FIH_NEGATIVE_VALUE 0x15555555
+#define FIH_CONST1 0x1FCDEA88
+#define FIH_CONST2 0x19C1F6E1
```
utzig, line 97: Is this random or is there some meaning to these constants?

utzig:
@mimok You need to sign-off all your commits.

boot/bootutil/include/bootutil/fault_injection_hardening.h
mimok: Yes, there are random values

mimok:
> @mimok You need to sign-off all your commits.

sign-off added (had to amend my commits to fix author email)

mimok:
FIH hardening tests have failed because tf-m need to be patched (i can provide the patch)...

mimok:
fix mcuboot panic when BOOT_IMAGE_NUMBER==1

RcColes:
Removing the duplicated variables is interesting - it seems to go against the prevailing wisdom, but I can see why. I think it might be worth experimenting with:
 * Removing the volatile keyword from fih_int
 * Reducing the size of the members of the struct to uint16_t
 
 As far as I can tell, the 32-bit Arm PCS (https://github.com/ARM-software/abi-aa/releases/download/2022Q3/aapcs32.pdf) allows the return of a 32 byte composite value in r0, so this _should_ eliminate the stacking of the return value and allow us to keep the validation.
 
 Will do a more thorough review, but the rest looks very nice :) I particularly like FIH_SET

mimok:
I'm pretty sur I first intended to encode `fih_int` with two  `uint16_t` but it was not really more secure than the current implementation (I don't remember why though ).

Removing volatile keyword should cause double if to be optimized which is not what we want.

I will try to publish my fault inhection tool soon to allow others devs to perform their own tests.

RcColes:
I remember chatting to someone (I want to say Arnaud de Gradmaison) about him having a better replacement to `volatile`. I don't remember exactly, but it was based on the inline assembly preventing optimisation, and was better since it didn't generate the extra load/stores related to a variable being volatile.

I suspect it was something like `asm volatile("noop" : "" : "" : "r" (var))`, though I remember this not working when I originally did the FIH library and I haven't got round to testing it yet

mimok:
To get rid of 'volatile' seems appealing but what about the portability of such a construction? I believe it would be limited, isn't it? When I developed my PR, I was thinking in using only C code without any line of assembly (contrary to what wolfssl do in wolfboot).

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -214,19 +214,9 @@ fih_int fih_int_encode(int x)
...
-//     return (x == y) && fih_delay() && !(x != y);
-// }
-
-// __attribute__((always_inline)) inline
-// int fih_not_eq(fih_ret x, fih_ret y)
-// {
-//     return (x != y) && fih_delay() && !(x == y);
-// }
+#define FIH_EQ(x, y) ((x == y) && fih_delay() && !(y != x))
+#define FIH_NOT_EQ(x, y) ((x != y) || !fih_delay() || !(y == x))
```
tamasban, line 218: Does it improve security? If do so then how?

boot/bootutil/src/image_validate.c
```diff
@@ -497,7 +499,8 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
              */
             fih_rc = fih_ret_encode_zero_equality(img_security_cnt <
                                    fih_int_decode(security_cnt));
-            if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
+            if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+                FIH_SET(fih_rc, FIH_FAILURE);
```
tamasban, line 503: should not be line 529 the same? FIH_SET(fih_rc, FIH_FAILURE);

What about the declaration of the fih_ret variables? Should not all `fih_rc` be initialised by FIH_SET?

Might a macro would handy:
`FIH_DECLEAR(x) (fih_ret x; FIH_SET(x, FIH_FAILURE));`

boot/bootutil/src/image_validate.c
```diff
@@ -488,7 +489,8 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
 
             FIH_CALL(boot_nv_security_counter_get, fih_rc, image_index,
                                                            &security_cnt);
-            if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
+            if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+                FIH_SET(fih_rc, FIH_FAILURE);
```
tamasban, line 493: Is it intended to prevent flipping bits in `fih_rc` and might turn it to `FIH_SUCCESS`?

boot/bootutil/src/loader.c
```diff
@@ -2137,10 +2137,17 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
      * have finished. By the end of the loop each image in the primary slot will
      * have been re-validated.
      */
+    volatile int fih_cnt;
+    FIH_SET(fih_cnt, 0);
     IMAGES_ITER(BOOT_CURR_IMG(state)) {
 #if BOOT_IMAGE_NUMBER > 1
+        //Hardenned to prevent from skipping check of a given image, img_mask is declared volatile
```
tamasban, line 2144: nit: please use /* */ insted of //

boot/bootutil/src/loader.c
```diff
@@ -2137,10 +2137,17 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
      * have finished. By the end of the loop each image in the primary slot will
      * have been re-validated.
      */
+    volatile int fih_cnt;
```
tamasban, line 2140: Variables are usually declared at the beginning of a function. Please move it there to be aligned with the existing pattern.

boot/bootutil/src/loader.c
```diff
@@ -2201,7 +2221,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 out:
     close_all_flash_areas(state);
 
-    if (rc) {
+    if (rc != 0) {
```
tamasban, line 2225: What about using FIH_SET always when a value is assigned to a `fih_ret`?

boot/bootutil/src/loader.c
```diff
@@ -2149,6 +2156,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
              */
             rc = boot_read_image_headers(state, false, &bs);
             if (rc != 0) {
+                FIH_SET(fih_rc, FIH_FAILURE);
```
tamasban, line 2160: seems redundant with lines 2224-2226. Is it on purpose?  The same applies to all changes below where `FIH_SET` is used

boot/bootutil/src/loader.c
```diff
@@ -3185,7 +3205,8 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
     while (true) {
 #endif
         FIH_CALL(boot_load_and_validate_images, fih_rc, state);
-        if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
+        if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+            fih_rc = FIH_FAILURE;
```
tamasban, line 3210: `FIH_SET`?

boot/bootutil/src/loader.c
```diff
@@ -3211,11 +3232,13 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 #endif
         rc = boot_update_hw_rollback_protection(state);
         if (rc != 0) {
+            fih_rc = FIH_FAILURE;
```
tamasban, line 3236: same as above about redundancy

boot/bootutil/src/loader.c
```diff
@@ -3230,7 +3253,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 out:
     close_all_flash_areas(state);
 
-    if (fih_eq(fih_rc, FIH_SUCCESS)) {
+    if (rc != 0) {
```
tamasban, line 3257: FIH_SET in the next line? then might the individual calls in front of the goto statements could be removed

boot/bootutil/src/loader.c
```diff
@@ -2146,6 +2146,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
             if(!state->img_mask[BOOT_CURR_IMG(state)]) {
                 FIH_PANIC;
             } else {
+                ++fih_cnt;
```
tamasban, line 2149: this change could be squashed with the previous commit to get a cleaner history

tamasban (changes requested):
Thanks for the fixes! I have started reviewing. Here is the first round of comments and a few questions to help me better understand the changes and the security improvements.

mimok:
@tamasban Maybe the simplest way to take into account some of your remarks would be to shash all the commits. What is your opinion on this point?

Also even if it is not really required using systematically FIH_SET might be both cleaner, simpler and safer. Adding a FIH_DECLARE macro is indeed a good idea.

tamasban:
I'm fine to squash all the commits into a single one and incorporate all the review fixes there. But please also merge the commit messages. Describe all the changes in that commit message.

I think mandating the usage of FIH_SET/FIH_DECLEAR as a pattern would result in a less error-prone code. However, it might add a significant delay in the boot if FIH_PROFILE HIGH is chosen. What do you think?

tamasban:
Just let you know that working on a patch to fix the FIH CI job with the updated FIH library.

Related patch in TF-M:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/18309

Will post a patch for MCUboot here which should be included in this PR to fix FIH CI with a recent TF-M version.

mimok:
For sure, I will concatenate the comments of all the commits. I'm planning to squash the commits before the end of the next week.

Not really sure what would be the overhead in term of CPU clock cycles if we add FIH_SET/DECLARE everywhere. I'm going to take into account your remarks and I will double check wether it is useful to add additional countermeasures or not.

tamasban:
I have updated the TF-M version in the FIH CI job:
https://github.com/mcu-tools/mcuboot/pull/1545

You can rebase your commits to `main` and add this single-line change to yours to enable the FIH CI with the updated library:

```
diff --git a/ci/fih-tests_run.sh b/ci/fih-tests_run.sh
index 1a1f26d6..f45a0284 100755
--- a/ci/fih-tests_run.sh
+++ b/ci/fih-tests_run.sh
@@ -19,7 +19,7 @@ set -e
 pushd .. &&\
    git clone https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git &&\
    pushd trusted-firmware-m &&\
-   git checkout TF-Mv1.7.0 &&\
+   git fetch https://review.trustedfirmware.org/TF-M/trusted-firmware-m refs/changes/09/18309/1 && git checkout FETCH_HEAD &&\
    popd
```

Proposed way forward due to cross references between TF-M and MCUboot:
- Get to shape this PR. You can use the WIP change in TF-M to run the FIH CI job. I can think we can merge first this PR to MCUboot.
- I will update the WIP change in TF-M with the reference to the MCUboot version, which includes the updated library. I will merge this as second.
- Update MCUboot FIH CI job with the merged commit hash in TF-M.

mimok:
Thank you Tamas, I will try to do that this weekend.

boot/bootutil/src/image_validate.c
```diff
@@ -445,7 +445,7 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
             if (rc) {
                 goto out;
             }
-            key_id = bootutil_find_key(image_index, key_buf, len);
+            key_id = bootutil_find_key(image_index, key_buf, len); //TODO: should be called with FIH_CALL
```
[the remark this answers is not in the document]
mimok: Not really, I was thinking this could be problematic to not use FIH_CALL here but FI tool does not reveal any problem here.

boot/bootutil/include/bootutil/fault_injection_hardening.h
mimok: I have seen that using inlined functions lead to more complex (and vulnerable) code at assembly level than using macros.  .

boot/bootutil/src/loader.c
mimok: Previously the construction where `rc` was set to -1 was weak. Setting `fih_rc` to FIH_FAILURE solved the problem. But you are right there is some redundacy here. We could replace FIH_SET with a simple affectation or remove the last test to save some clock cycles.

boot/bootutil/src/loader.c
mimok: seems to be not really needed here (according to fault injection simulations)

boot/bootutil/src/loader.c
mimok: To obtain the best security, you have to switch as soon as possible from weak `rc` return values to more robust `fih_rc`. affectations before got were added to improve the code security and therefore cannot be removed without introducing vulnerabilities.

boot/bootutil/include/bootutil/boot_hooks.h
```diff
@@ -80,9 +80,9 @@ int boot_read_image_header_hook(int img_index, int slot,
  * 
  * @retval FIH_SUCCESS: image is valid, skip direct validation
  *         FIH_FAILURE: image is invalid, skip direct validation
- *         fih encoded BOOT_HOOK_REGULAR: follow the normal execution path.
+ *         FIH_BOOT_HOOK_REGULAR: follow the normal execution path.
```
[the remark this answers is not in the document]
mimok: all commits are now squashed

mimok:
@tamasban, i believe its ok now. According to your answaers, we may still have to slightly change the code.

mimok:
Declaring `img_mask` volatile causes a compilation error during CI script execution. Maybe we could use a temporary volatile variable to store the `img_mask` value which has to be tested?

tamasban:
Thanks for the fixes! Could you upload a version with a temporary variable? 

Please also check the FIH CI results here:
https://github.com/mcu-tools/mcuboot/actions/runs/3725460271/jobs/6325335913
You can check the result in the `FIH hardening test run` menu. If you scroll to the very end then you can see that this section of code is reported as vulnerable more than once:
```
Boot success: address: 0x100023a0 skipped: 2
100023a0 <FIH_LABEL_FIH_CALL_START_236+0x18>:
    bootutil_sha256_context shactx;
    uint8_t em[MBEDTLS_MPI_MAX_SIZE];
    uint8_t db_mask[PSS_MASK_LEN];
    uint8_t h2[PSS_HLEN];
    int i;
    fih_ret fih_rc = FIH_FAILURE;
100023a0:	9205      	str	r2, [sp, #20]

    if (ctx->MBEDTLS_CONTEXT_MEMBER(len) != PSS_EMLEN ||
100023a2:	d008      	beq.n	100023b6 <FIH_LABEL_FIH_CALL_END_299+0x2>
    /* Step 14.  If H = H', output "consistent".  Otherwise, output
     * "inconsistent". */
    FIH_CALL(boot_fih_memequal, fih_rc, h2, &em[PSS_HASH_OFFSET], PSS_HLEN);
```

`skipped: 2` means that PC was increased by two so a single instruction is skipped. Could you take a look at this section of the code? Previously it was not reported. 
However, it is clear that the updated library performs better. It reduces the number of successful boots with a tampered image. Most boots only happen with skipping multiple instructions (PC is increased by 8 or 10).
I will try to recheck the code before the Christmas break.

mimok:
I have seen that changing some part of the code has an impact on other, sometimes unrelated, parts of the code. Compiler may slightly change the way it compiles instructions leading to new vulnerabilities. Here, i don't know what is actually hapening but its clear that one of the tests (before call to `boot_fih_memequal()`) will likely fail if signature is modified. This ends up by executing goto out while `fih_rc` value has not been properly set.

A solution would be to use FIH_SET for each sensitive affectation which should solve this problem and all the others "hidden" problems but performances would be affected.

I'm pushing a fix for the previous problem. Lets see if your FI tool find new problems before fixing `bootutil_cmp_rsasig`.

mimok:
I see that I  should have declared the volatile variable inside the loop... I will check if there is an easy way to execute CI script locally before pushing. @tamasban , do you have an idea how to do that?

tamasban:
> A solution would be to use FIH_SET for each sensitive affectation which should solve this problem and all the others "hidden" problems but performances would be affected.

What about tweaking `FIH_SET` in a way to make it cheaper and decrease the runtime penalty?
Currently, FIH_SET is defined like this:
``` 
#define FIH_SET(x, y) x = y; if(fih_delay() && (x != y)) FIH_PANIC
```
This adds a random delay between the variable init and the value check operation. This could take too many CPU cycles. Furthermore, the delay is added only if the FIH profile HIGH is set. 

I wondering whether creating a simpler delay loop could be beneficial. It would add const delay.  The amount of delay would depend on some values which are determined at build time or vary from place to place where the `FIH_SET` macro is used.

``` 
#define DELAY_MASK 0x000000FF

/* Delay would depend on the memory address of the function and the location of the expression in the source code */
#define FIH_CONST_DELAY \
    do { \
        const void *func_addr = &(__func__); \
        volatile int delay = ((int)func_addr +  __LINE__) & DELAY_MASK; \
        volatile int i; \
        for (i = 0; i < delay;) { \
            ++i; \
        } \
    } while(0)

#define FIH_SET(x, y) x = y; FIH_CONST_DELAY; if(x != y) FIH_PANIC
```

https://godbolt.org/z/nf6qjqExa

Determining the actual delay is not possible without dumping the actual bootloader binary. I think this is expected in the case of a product which is hardened against fault injection attacks.  Even though the delays are the same in between consecutive executions this might add enough hardening to the FIH_SET macro to justify the usage in all cases where a `fih_ret` variable is used. However, the runtime penalty is not huge and this construct could be used in all FIH profiles.

What do you think?

tamasban:
Found that this simple change eliminates the boot issue in `bootutil_cmp_rsasig`, which is shown by the CI tests.
```
diff --git a/boot/bootutil/src/image_rsa.c b/boot/bootutil/src/image_rsa.c
index 0c357338..7d29d434 100644
--- a/boot/bootutil/src/image_rsa.c
+++ b/boot/bootutil/src/image_rsa.c
@@ -170,7 +170,8 @@ bootutil_cmp_rsasig(mbedtls_rsa_context *ctx, uint8_t *hash, uint32_t hlen,
     uint8_t db_mask[PSS_MASK_LEN];
     uint8_t h2[PSS_HLEN];
     int i;
-    fih_ret fih_rc = FIH_FAILURE;
+    fih_ret fih_rc;
+    FIH_SET(fih_rc, FIH_FAILURE);
```

mimok:
From my experience as hardware security evaluator, I would say that a delay which does not change across reset is not really a problem for the attacker. Indeed, most of the time, attackers blindly scan a given time interval by injecting fault by steps of xx nanoseconds and as long as there is no random delays in the execution flow this approach gives good results.

If we don't care about double fault injections, we could simply remove the call to FIH_DELAY. However, if we are considering a fault model where several instructions are skipped a constant delay which would be the same for every FIH_SET calls might be sufficient (eg. wolfboot from wolfssl use a fixed number of cmp instructions).

Here, if performance is a concern, I would simply remove FIH_DELAY from FIH_SET. Having some feedbacks from someone using mcuboot with HIGH profile would help us to understand if FIH_DELAY overhead is really a problem.

boot/bootutil/include/bootutil/fault_injection_hardening.h
```diff
@@ -45,7 +45,7 @@
  *
  * fih_int fih_rc = FIH_FAILURE;
```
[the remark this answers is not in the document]
mimok: done

mimok:
I pushed several commits to take into account your comments.

ci/fih-tests_run.sh
```diff
@@ -19,7 +19,7 @@ set -e
 pushd .. &&\
    git clone https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git &&\
    pushd trusted-firmware-m &&\
-   git checkout TF-Mv1.7.0 &&\
+   git fetch https://review.trustedfirmware.org/TF-M/trusted-firmware-m refs/changes/09/18309/1 && git checkout FETCH_HEAD &&\
```
tamasban, line 22: I have updated the patch in TF-M. You can check out the latest version with this command:
`git fetch https://review.trustedfirmware.org/TF-M/trusted-firmware-m refs/changes/09/18309/2 && git checkout FETCH_HEAD`

boot/bootutil/src/loader.c
```diff
@@ -2173,21 +2183,32 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 
         rc = boot_add_shared_data(state, BOOT_PRIMARY_SLOT);
         if (rc != 0) {
+            FIH_SET(fih_rc, FIH_FAILURE);
             goto out;
         }
+        ++fih_cnt;
+    }
+    /*
+    * fih_cnt should be equal to BOOT_IMAGE_NUMBER now.
```
tamasban, line 2205: nit: add one more space to make the `*` vertically aligned.

tamasban:
Multiple fault injection is not really in the protection scope, due to its higher bar and the complexity of properly testing it. However, the possibility to introduce random delays is targeting this fault model. I have never heard back from anyone who is using the HIGH profile. So, unfortunately, there is no feedback on its effectiveness or on cost.  Therefore, I would leave the code as is, we can change it later anytime if there is a real need.

Regarding skipping multiple instructions, we always thought it is a valid scenario and better to prepare for it and test for it.  Our assumption is based on whitepapers on this topic.  Therefore, the fault injection test script in the CI are prepared to increase the PC by 2, 4, 6, 8 or 10 bytes. Which simulates skipping max 10 (theoretical maximum) instructions. Usually, it is less because instructions are encoded either on 2 or 4 bytes and they are mixed. Currently, only the instruction skip fault model is tested.

Please apply this small fix which targets a code snippet where protection can be defeated by multiple instruction skip:
```
diff --git a/boot/bootutil/src/loader.c b/boot/bootutil/src/loader.c
index dd3480c6..a75ded5d 100644
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2169,7 +2169,8 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 
 #ifdef MCUBOOT_VALIDATE_PRIMARY_SLOT
         FIH_CALL(boot_validate_slot, fih_rc, state, BOOT_PRIMARY_SLOT, NULL);
-        if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+        if (FIH_EQ(fih_rc, FIH_FAILURE) ||
+            FIH_EQ(fih_rc, FIH_NO_BOOTABLE_IMAGE)) {
             FIH_SET(fih_rc, FIH_FAILURE);
             goto out;
```

Overall the changes are LGTM. Please squash them and rebase to the master branch. 
This CI job is failing, but do not know why:
https://github.com/mcu-tools/mcuboot/actions/runs/3838505965/jobs/6535247115
Let's see whether the rebase resolve it. If not then we can ask for help from the Zephyr folks.

mimok:
@tamasban What happen if someone manages to introduce a fault in `fih_rc` (e.g. random value)? `fih_rc` value would be seen as equivalent to `FIH_SUCCESS`, isn't it ?
```
-        if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+        if (FIH_EQ(fih_rc, FIH_FAILURE) ||
+            FIH_EQ(fih_rc, FIH_NO_BOOTABLE_IMAGE)) {
```

tamasban:
True. I have added a check for all possible values. It seems a bit overkill in normal operation, but avoid the boot issue in case of multiple instruction skip.

```
-        if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+        /* Check for all possible values is redundant in normal operation it
+         is meant to prevent FI attack.
+         */
+        if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS) ||
+            FIH_EQ(fih_rc, FIH_FAILURE) ||
+            FIH_EQ(fih_rc, FIH_NO_BOOTABLE_IMAGE)) {
```

mimok:
I have rebased, squashed and force pushed. Now, everything should be up to date.

tamasban (approved):
Changes LGTM! Thanks for the contribution!
@RcColes: Could you do another round of review?

boot/bootutil/src/image_rsa.c
```diff
@@ -170,22 +170,19 @@ bootutil_cmp_rsasig(mbedtls_rsa_context *ctx, uint8_t *hash, uint32_t hlen,
     uint8_t db_mask[PSS_MASK_LEN];
     uint8_t h2[PSS_HLEN];
     int i;
-    int rc = 0;
-    fih_int fih_rc = FIH_FAILURE;
+    fih_ret fih_rc;
+    FIH_SET(fih_rc, FIH_FAILURE);
```
d3zd3z, line 174: Any reason this is set separately with an FIH_SET instead of just assigned like the other uses seem to be?

boot/bootutil/src/loader.c
```diff
@@ -2173,21 +2188,32 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 
         rc = boot_add_shared_data(state, BOOT_PRIMARY_SLOT);
         if (rc != 0) {
+            FIH_SET(fih_rc, FIH_FAILURE);
             goto out;
         }
+        ++fih_cnt;
+    }
+    /*
+     * fih_cnt should be equal to BOOT_IMAGE_NUMBER now.
```
d3zd3z, line 2210: It doesn't show here, but this line and the next have a trailing space. Please fix that.

d3zd3z (changes requested):

ci/fih-tests_run.sh
```diff
@@ -19,7 +19,7 @@ set -e
 pushd .. &&\
    git clone https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git &&\
    pushd trusted-firmware-m &&\
-   git checkout TF-Mv1.7.0 &&\
+   git fetch https://review.trustedfirmware.org/TF-M/trusted-firmware-m refs/changes/09/18309/2 && git checkout FETCH_HEAD &&\
```
RcColes, line 22: Will this link remain valid?

RcColes (commented):
Looks good, couple of minor things and one main one

boot/bootutil/src/loader.c
```diff
@@ -2270,7 +2291,7 @@ split_go(int loader_slot, int split_slot, void **entry)
     free(sectors);
 
     if (rc) {
-        fih_rc = fih_int_encode(rc);
+        fih_rc = FIH_FAILURE;
```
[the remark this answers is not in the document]
mimok: Actually, I'm totaly agree with you. If I had to implement a fault resistant product based on MCUboot, i would use `FIH_SET()` everywhere. Indeed, a small change in the way the compiler operates might lead to new vulnerabilities coming from vulnerable affectations.

If everybody is okay (@tamasban and @d3zd3z), i will enhance the code in this way.

boot/bootutil/src/image_validate.c
```diff
@@ -343,7 +343,7 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
     uint16_t type;
     int sha256_valid = 0;
 #ifdef EXPECTED_SIG_TLV
-    fih_int valid_signature = FIH_FAILURE;
```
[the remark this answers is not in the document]
mimok: I'm agree, see next comment

ci/fih-tests_run.sh
mimok: I don't think so. Its only a temporary workaround because of the dependencies between MCUboot and TF-M, ask @tamasban  for details.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
tamasban: No objection, I agree.  I have the same thoughts [earlier](https://github.com/mcu-tools/mcuboot/pull/1528#issuecomment-1339398591).

If FIH is disabled then there is no overhead. If your fault model includes FI then you need the best available protection. An increase in code size is an acceptable trade-off.

ci/fih-tests_run.sh
tamasban: Not sure what happens with the link if the change got merged to TF-M.

I think this flow could work:
- First, merge this PR with the link to the change on review in TF-M repo.
- Then I can create another change in TF-M with exact same content as the current change + update the MCUboot version (which includes the FIH enhancement) in TF-M repo. That can be merged and in this way, the current link remains valid.
- Finally, a new PR in MCUboot to update the TF-M version in the FIH CI job to be aligned.
- Current WIPchnage in TF-M can be abandoned.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
mimok: @RcColes , @tamasban I have modified all affectations in two commits. The first on targets potentially security critical affectations while the second one targets less critical affectations or affectations that could be easily bypassed. This two commits are only here to ease the source code modification reading. They can be squashed before merging.

boot/boot_serial/src/boot_serial.c
```diff
@@ -235,7 +235,8 @@ bs_list(char *buf, int len)
                 flash_area_read(fap, 0, &hdr, sizeof(hdr));
             }
 
-            fih_ret fih_rc = FIH_FAILURE;
+            fih_ret fih_rc;
```
tamasban, line 238: The same pattern occurs multiple times. What about FIH_DECLARE?
https://github.com/mcu-tools/mcuboot/pull/1528#discussion_r1039593644

boot/bootutil/src/loader.c
```diff
@@ -3207,7 +3216,7 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 #endif
         FIH_CALL(boot_load_and_validate_images, fih_rc, state);
         if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
-            fih_rc = FIH_FAILURE;
+            FIH_SET(fih_rc,FIH_FAILURE);
```
tamasban, line 3219: nit: missing space  `fih_rc, FIH_FAILURE`

tamasban (changes requested):

boot/boot_serial/src/boot_serial.c
mimok: You are right, let's implement the `FIH_DECLARE()` macro. I will propose a new commit with this change.

utzig:
1) I don't see any new CI workflow, how is this supposed to be tested?
2) If the previous version was passing the qemu based tests, how can we check this PR fixes anything (not only "theoretically"!)?
3) Should the qemu based tests stay here in the first place?
4) Assuming this PR completely fixes the fault issues, what guarantees that new code in the future doesn't break it again?

tamasban:
> 1. I don't see any new CI workflow, how is this supposed to be tested?

Same CI as before. 

> 2. If the previous version was passing the qemu based tests, how can we check this PR fixes anything (not only "theoretically"!)?

It is not entirely "passing" currently on the CI. It does not provide a passed/failed response. It tells you how many times could the system successfully boots up with a corrupted image (the signature does not match)  with and without enabled fault injection countermeasures.

The latest result from `main` (4427e4c9775938a9617536ca6cb4d368368f36db):
FIH disabled: https://github.com/mcu-tools/mcuboot/actions/runs/3929054765/jobs/6717394710
FIH enabled (MEDIUM): https://github.com/mcu-tools/mcuboot/actions/runs/3929054765/jobs/6717395135

Scroll to the bottom and search for: `System booted (undesired behaviour)`

> 3. Should the qemu based tests stay here in the first place?
I think so. 

> 4. Assuming this PR completely fixes the fault issues, what guarantees that new code in the future doesn't break it again?

I do not think that "completely fixes" is the right term. The QEMU-based test does not provide full test coverage. Only the critical parts, which are marked with the FIH_START/FIH_STOP asm labels, are tested. It could be that outside of these ranges of the code there are vulnerabilities.  Plus only the instruction skip fault model is implemented, the register corruption and other fault models are not tested.

Plus this is just a simulation.  How a real chip reacts to fault injection is a different story. That is nondeterministic. SW countermeasures are just a level of defence, but without HW countermeasures, they are not good enough. 

The FIH library helps the developer to identify the critical parts and harden those.  But being sure that a particular chip is resistant to fault injection requires dedicated testing (SW simulation and real HW testing) and binary-level investigations of critical code sections.

With the recent changes the number of successful boots with a corrupted image is decreased to 0. So, we can transform the CI result to a binary pass/failed result. Then it can give an early indication if some new code snippet introduces a new vulnerability, which can be exploited with instruction skip. 

@trustngotech Could you elaborate on the topic? I think you have a very good insight.

utzig:
> The latest result from `main` ([4427e4c](https://github.com/mcu-tools/mcuboot/commit/4427e4c9775938a9617536ca6cb4d368368f36db)): FIH disabled: https://github.com/mcu-tools/mcuboot/actions/runs/3929054765/jobs/6717394710 FIH enabled (MEDIUM): https://github.com/mcu-tools/mcuboot/actions/runs/3929054765/jobs/6717395135
> 
> Scroll to the bottom and search for: `System booted (undesired behaviour)`

Thanks, makes sense, but shouldn't we see an error? If there is an "undesired behavior" that we are not informed of what's the point of having a CI running? Shouldn't it be fixed so it properly fails?

mimok:
I think @tamasban did a pretty good summary of the current status. Automated detection of fault injection vulnerabilities is still an on-going research topic. We could improve the coverage of the current CI system by performing tests on critical function bodies but this would take a really long time (probably hours).

In my opinion, MCUboot maintainers should determine whether FI resistance is mandatory or not. In the first case, CI system should be improved to extend the code covergage and any discovered fault according to the elected fault models shall lead to a CI failure. In the second case, the current CI system do the job pretty well, it could still be improved but this is not so urgent.

Anyway as @tamasban says, proper validation of fault injection countermeasures requires tests on real board using for exemple something like the chipwhisperer test tool.

tamasban:
>Thanks, makes sense, but shouldn't we see an error? If there is an "undesired behavior" that we are not informed of what's the point of having a CI running? Shouldn't it be fixed so it properly fails?

I think it is not an easy question (at least for me): 
- In the first place, FIH CI test the build configs with various level FIH protection profiles. 
- Secondly, it gives feedback about the FI resistance of the code.  When the original PR was merged then CI showed that the PR increased the resistance a lot, but it was not perfect. It reduced the number of places significantly where only a single instruction skip can defeat MCUboot. But still, there were places where skipping multiple instructions worked. How many instruction skips happen in a real chip?  AFAIK usually only a single instruction skip is considered in the security evaluation, but it is not always true in the reality. Depends on the microarchitecture of the chip. 
- Sometimes unrelated code changes could lead to rearrangement in the binary that makes it error-prone against FI.  How to signal this to the developer? The new code is functionally correct but FI CI fails. Then should we merge the PR or not?  Might a later change will remove the vulnerability. 
- We only testing with GCC. Other compilers probably show a bit different result.
- For these reasons, we decided to not have a passed/failed result of the CI. Instead, just report the success/failure ratio.

I think in any case if someone wants the MCUboot code used in an FI-certified product he needs a downstream/snapshot version which is tested on the dedicated HW and the revealed vulnerabilities are fixed.  Might be on the assembly level, not sure just thinking. However, those vulnerabilities are device-specific and the fix can't be applied in the generic code.

Just an idea of how to fine-tune the FIH CI: 
- We can make the FIH CI fail if any is successful boot happens below a certain threshold. I mean if the PC is advanced less than 4. This can 1-2 instruction skip, depending on the actual instruction.  But the result of FIH CI won't block the merging. If a successful boot happens values above the threshold then CI will pass, because we think that in the majority of the cases skipping that much instruction is unrealistic. OR we can just decrease how many instructions skip to test.  

I'm happy to take any advice.

mimok:
In my opinion, we should just try to ease the work of the persons wanting to implement FI resistance in MCUboot. The actual assessment of the resistance should be their work because we don't known which MCU they will use and don't have the ressources needed to make these tests ourselves.

tamasban:
@trustngotech  Thanks, for the fixes!

The changes LGTM. Unfortunately for some reason, a CI job failed:
https://github.com/mcu-tools/mcuboot/actions/runs/3986927070/jobs/6836163469

Could you squash and rebase the changes to the `main` branch?

mimok:
Squash and force push done

mimok:
> The changes LGTM. Unfortunately for some reason, a CI job failed: https://github.com/mcu-tools/mcuboot/actions/runs/3986927070/jobs/6836163469

@tamasban , I believe the problem come from this commit:
https://github.com/zephyrproject-rtos/zephyr/commit/22a8af5e63b7790e3b9277ca561c851002b5c3a9

An empty static function was previously declared which is no more the case now.

tamasban (approved):

tamasban:
@d3zd3z @RcColes @utzig Could you take another look? Do you have another comment/change request?

utzig (commented):
Looks OK, a few style issues to fix...

utzig:
> * We can make the FIH CI fail if any successful boot happens below a certain threshold. I mean if the PC is advanced less than 4. This can 1-2 instruction skip, depending on the actual instruction.  But the result of FIH CI won't block the merging. If a successful boot happens with values above the threshold then CI will pass, because we think that in the majority of the cases skipping that much instruction is unrealistic. Or we can just decrease how many instructions skip to test and ignore the high numbers.

Maybe there's a way to set a workflow to not *be required*, so at merges are still OK even when it fails. At least in that case we would know...

mimok:
Defining when and how to merge is your policy but I'm agree with @tamasban proposition.

> We can make the FIH CI fail if any successful boot happens below a certain threshold. I mean if the PC is advanced less than 4. This can 1-2 instruction skip, depending on the actual instruction. But the result of FIH CI won't block the merging. If a successful boot happens with values above the threshold then CI will pass, because we think that in the majority of the cases skipping that much instruction is unrealistic.

CI *shall fail* if a sucessful boot is obtained by bypassing one (or two instructions) as it would *completly defeat* the purpose of the FIH mechanisms

Above this threshold, as there is problably no way to mark a workflow as "almost failed", i would let it pass.

boot/zephyr/single_loader.c
```diff
@@ -403,7 +403,7 @@ int
 boot_handle_enc_fw()
 {
     int rc = -1;
-    fih_int fih_rc = FIH_FAILURE;
+   FIH_DECLARE(fih_rc, FIH_FAILURE);
```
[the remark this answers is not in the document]
mimok: done

boot/zephyr/single_loader.c
```diff
@@ -439,11 +439,11 @@ boot_handle_enc_fw()
  *
  * @return		FIH_SUCCESS on success; nonzero on failure.
  */
-fih_int
+fih_ret
 boot_go(struct boot_rsp *rsp)
 {
     int rc = -1;
-    fih_int fih_rc = FIH_FAILURE;
+   FIH_DECLARE(fih_rc, FIH_FAILURE);
```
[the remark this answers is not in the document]
mimok: done

tamasban:
@utzig @mimok: Let's handle the CI changes in a separate PR. I will take care of it. Follow up it in #1584

tamasban (approved):

utzig:
@mimok Please squash the commits.

mimok:
@utzig PR squashed and rebased

utzig (approved):

RcColes (approved):
LGTM, agree with fixing CI in another patch

PR #1531, opened 2022-11-29
imgtool: fix tlv encoding

mzagrabski:
As mcuboot supports 2 byte tlv type, where range 0x00a0 - 0xfffe is vendor reserved, allow setting tlv with non-zero upper byte.

Also fix potential big-endian issues with type encoding.


mniestroj (approved):

utzig (approved):
LGTM

PR #1533, opened 2022-11-30
serial recovery: Add unaligned stack buffer writing

### LINKED ISSUE #1532: Serial recovery does not respect alignment requirements of flash driver
It seems that the serial recovery system does not respect the requirements of the underlying flash driver, in this specific example, for the QSPI driver on nordic devices. It needs a word aligned length and address, however the serial recovery mechanism is using a buffer with LSB address 0x9 which is not word-aligned, as a result when trying to upload firmware, the go mcumgr application sits in an endless loop at trying to upload to 0x0 because of this issue.

nordicjm:
Fixes a bug when writing to devices which have memory alignment requirements with data being using directly from a zcbor-response whereby the alignment of the buffer data does not meet the requirements of the flash driver.

Fixes #1532

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -511,7 +511,38 @@ bs_upload(char *buf, int len)
 
     BOOT_LOG_INF("Writing at 0x%x until 0x%x", curr_off, curr_off + img_chunk_len);
     /* Write flash aligned chunk, note that img_chunk_len now holds aligned length */
+#if defined(MCUBOOT_SERIAL_UNALIGNED_BUFFER_SIZE) && MCUBOOT_SERIAL_UNALIGNED_BUFFER_SIZE > 0
+    if (flash_area_align(fap) > 1 &&
+        (((size_t)img_chunk) & (flash_area_align(fap) - 1)) != 0) {
+        /* Buffer address incompatible with write address, use buffer to write */
+        uint8_t write_size = MCUBOOT_SERIAL_UNALIGNED_BUFFER_SIZE;
+        uint8_t wbs_aligned[MCUBOOT_SERIAL_UNALIGNED_BUFFER_SIZE];
+
+        while (img_chunk_len >= flash_area_align(fap)) {
+            if (write_size > img_chunk_len) {
+                write_size = img_chunk_len;
+            }
+
+            memset(wbs_aligned, flash_area_erased_val(fap), sizeof(wbs_aligned));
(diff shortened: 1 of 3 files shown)


nordicjm:
@de-nordic @nvlsianpu

de-nordic (changes requested):

boot/boot_serial/src/boot_serial.c
```diff
@@ -508,7 +508,38 @@ bs_upload(char *buf, int len)
 
     BOOT_LOG_INF("Writing at 0x%x until 0x%x", curr_off, curr_off + img_chunk_len);
     /* Write flash aligned chunk, note that img_chunk_len now holds aligned length */
+#if defined(__ZEPHYR__) && CONFIG_BOOT_SERIAL_UNALIGNED_BUFFER_SIZE > 0
```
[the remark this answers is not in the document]
nordicjm: Fixed

de-nordic (approved):

boot/boot_serial/src/boot_serial.c
```diff
@@ -508,7 +508,38 @@ bs_upload(char *buf, int len)
...
+                write_size = img_chunk_len;
+            }
+
+            memset(wbs_aligned, flash_area_erased_val(fap), sizeof(wbs_aligned));
+            memcpy(wbs_aligned, img_chunk, write_size);
+
+            rc = flash_area_write(fap, curr_off, wbs_aligned, write_size);
+
+            if (rc != 0) {
+                break;
```
nvlsianpu, line 529:
```suggestion
              goto out;
```

nvlsianpu (changes requested):

boot/boot_serial/src/boot_serial.c
nvlsianpu: maybe it is ok

nordicjm: Fixed

nvlsianpu (changes requested):
I'm afraid that the patch doesn't take into account the  `rem_bytes` and its procedure.

nordicjm:
> I'm afraid that the patch doesn't take into account the rem_bytes and its procedure.

That part of the code is unchanged, rem_bytes is set before this code runs, so if it is set then it will run like usual after the main write loop

nordicjm:
@nvlsianpu any comments?

nvlsianpu (approved):

PR #1536, opened 2022-12-01
boot_serial: Allow using a buffer larger than 512 bytes

### LINKED ISSUE #1535: Zephyr: Serial recovery has Kconfig for setting size but then has a static size of 512 bytes
This code makes no sense, it seems to buffer the input from the serial device, which is of configurable size via Kconfig, then copies it into a fixed 512 byte buffer, then decodes it into another 512 byte buffer, that's 3 buffers for the same data, and the final two are fixed at 512 bytes for reasons unknown

nordicjm:
There are 3 levels of buffers and only the first one seems to be configurable, this fixes that issue.

Fixes #1535

I don't know if 512 is a minimum so have only allowed changing the size if a larger value is set

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -78,7 +78,10 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
-#define BOOT_SERIAL_INPUT_MAX   512
+#ifndef MCUBOOT_SERIAL_MAX_RECEIVE_SIZE
+#define MCUBOOT_SERIAL_MAX_RECEIVE_SIZE 512
+#endif
+
 #define BOOT_SERIAL_OUT_MAX     (128 * BOOT_IMAGE_NUMBER)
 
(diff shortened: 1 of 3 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -78,7 +78,12 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
+#if defined(CONFIG_BOOT_MAX_LINE_INPUT_LEN) && CONFIG_BOOT_MAX_LINE_INPUT_LEN > 512
```
de-nordic, line 81: I would avoid adding `CONFIG_` things here.
I would rather have `MCUBOOT_` identifier defined in mcuboot_config.h and used here instead of the `CONFIG_`.
So the same logic here but `MCUBOOT_` used.

de-nordic (changes requested):

de-nordic (approved):

nvlsianpu (approved):

PR #1541, opened 2022-12-07
bootutil: Fixing memset not beeing called

de-nordic:
Memset could have been out optimized by compiler and also
not called in error path.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -919,6 +919,21 @@ boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
 }
 
 #if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
+
+#if defined(MCUBOOT_ENC_IMAGES) || defined(MCUBOOT_SWAP_SAVE_ENCTLV)
+/* Replacement for memset(p, 0, sizeof(*p) that does not get
+ * optimized out.
+ */
+static void like_mbedtls_zeroize(void *p, size_t n)
+{
+    volatile unsigned char *v = (unsigned char *)p;
+
+    for (size_t i = 0; i < n; i++) {
+        v[i] = 0;
+    }
+}
+#endif
+
 /**
  * Copies the contents of one flash region to another.  You must erase the
  * destination region prior to calling this function.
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/loader.c
```diff
@@ -919,6 +919,20 @@ boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
 }
 
 #if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
+
+#if defined(MCUBOOT_ENC_IMAGES) || defined(MCUBOOT_SWAP_SAVE_ENCTLV)
+/* Replacement for memset(p, 0, sizeof(*p) that should not get
```
SebastianBoe, line 924:
```suggestion
/* Replacement for memset(p, 0, sizeof(*p) that does not get
```

Let's make a stronger statement unless you have reason to believe it can be optimized out.

boot/bootutil/src/loader.c
```diff
@@ -919,6 +919,20 @@ boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
...
+ * optimized out.
+ */
+static void like_mbedtls_zeroize(void *v, size_t n)
+{
+    for(volatile unsigned char *p = (unsigned char *)v; n > 0; --n)
+    {
+        *p++ = 0;
+    }
```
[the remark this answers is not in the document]
de-nordic: Eh.. ok. I liked the bad looking one.

SebastianBoe (approved):

sigvartmh (approved):

nordicjm (approved):

de-nordic:
@d3zd3z This should also go in 1.10

boot/bootutil/src/loader.c
```diff
@@ -2214,17 +2229,22 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
+
+    fih_rc = FIH_SUCCESS;
+out:
     /*
      * Since the boot_status struct stores plaintext encryption keys, reset
      * them here to avoid the possibility of jumping into an image that could
      * easily recover them.
      */
+#if defined(MCUBOOT_ENC_IMAGES) || defined(MCUBOOT_SWAP_SAVE_ENCTLV)
+    like_mbedtls_zeroize(&bs, sizeof(bs));
```
utzig, line 2243: I think you should also test if Mbed-TLS is being used, and in that case call `mbedtls_platform_zeroize`. Otherwise just use `memset` like below. If I remember correctly Tinycrypt uses `memset` without any "volatile" tricks or anything, so I assume it will also leave behind its state, which includes the encryption keys. I have to double check this, if my setup also gets rid of the `memset` calls! Apart from this PR is good, but it's just a suggestion anyway...

utzig: Which means `like_mbedtls_zeroize` can be removed, or it should be something like `bootutil_zeroize` but then we have to use it in all places where it's required, including the Tinycrypt code.

de-nordic:
@d3zd3z Any objections against this? You are the security guy here, as far as I can tell.

de-nordic:
@de-nordic Is putting his thumb on the scale and merging. If there are any requests to make this thing better, do a PR for that.

PR #1545, opened 2022-12-09
Update TF-M version in CI FIH test

tamasban:
Update from `TF-Mv1.4.0` to `TF-Mv1.7.0`.

### DIFF
--- a/ci/fih_test_docker/execute_test.sh
+++ b/ci/fih_test_docker/execute_test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-# Copyright (c) 2020-2021 Arm Limited
+# Copyright (c) 2020-2022 Arm Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


ci/fih_test_docker/fi_tester_gdb.sh
```diff
@@ -40,6 +40,13 @@ function skip_instruction {
 
     cat >commands.gdb <<EOF
 target remote localhost: 1234
+file $IMAGE_DIR/bl2.axf
+b boot_go_for_image_id
```
RcColes, line 44: Nit: I think this should be able to be a `b boot_go_for_image_id if image_id != 0`, which makes the logic slightly nicer. Could well be wrong though, I hardly ever use conditional breakpoints these days.

ci/fih_test_docker/execute_test.sh
```diff
@@ -34,6 +34,9 @@ else
 fi
 
 # build TF-M with MCUBoot
+
+export GIT_SSL_NO_VERIFY=1
```
RcColes, line 38: What's the purpose of this? I'd be concerned about disabling SSL

tamasban: When I tried to run the scripts in my local environment I faced the issue below during TF-M build. This is a workaround for that:

```
Cloning into 'tfm_test_repo-src'...
fatal: unable to access 'https://git.trustedfirmware.org/TF-M/tf-m-tests.git/': server certificate verification failed. CAfile: none CRLfile: none

Interestingly I can pull the code from the same machine, but the scripts run in a docker container. Might this cause the issue?

tamasban: Removed since it not an issue in GitHub CI.

RcColes (approved):

nvlsianpu (approved):

Issue #1165, opened 2021-10-05
Swap using scratch checks for swap size being less than primary/secondary slot sizes but this is incorrect


nordicjm:
At https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/swap_scratch.c#L247 there is a check for if the primary slot or secondary slot sizes are greater than the scratch size and if so, it fails. The whole point in having the scratch area as far as I understand it is so that data from one slot can be put there whilst the other slot's data goes to the main slot and that the scratch area can be filled up so that one swap can contain many sectors of either area, speeding up the process of swapping as it doesn't do it on a sector by sector basis (after spending time shifting all sectors in one slot up by one)

We have the primary slot on main flash of an nRF52840 and secondary slot in QSPI - both have 4KB sector sizes, the scratch for testing is placed on QSPI also, current head of tree zephyr is being used at commit ef0d955758e549f39d97b07089dedc4bdd05bc0e - with this check in, upgrade using scratch is not possible (testing with 64KB slot sizes and 16KB scratch size), with that check commented out, the upgrade can proceed and works, image upgrading works, image reverting works, so it seems like a bug in that check

lairdjm:
Seems to be an issue with this however in that mcuboot will always swap the images at startup, even if you confirm the image from the application, on reboot mcuboot will swap them over.

de-nordic:
> 
> 
> At https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/swap_scratch.c#L247 there is a check for if the primary slot or secondary slot sizes are greater than the scratch size and if so, it fails. The whole point in having the scratch area as far as I understand it is so that data from one slot can be put there whilst the other slot's data goes to the main slot and that the scratch area can be filled up so that one swap can contain many sectors of either area, speeding up the process of swapping as it doesn't do it on a sector by sector basis (after spending time shifting all sectors in one slot up by one)

The code you are pointing to is check whether scratch area is big enough to cover largest sector of primary/secondary slot; the check is done because flash, for example in stm32 devices, does not have to have sectors of the same size.
In case of uniform size sectors the `sz?` will always have size of sector here, and will be reset to 0 before leaving this if.

Something is wrong with your DTS configuration of slots. I have tested the feature with 4k (1 page/sector) size scratch on QSPI and secondary slot on QSPI and everything seems to work.
Are you sure that last sectors of your secondary slot are not overlapping with the scratch?

[...]

PR #1550, opened 2022-12-19
docs: zephyr: Add note on swap-using-scratch

nordicjm:
Adds details on how to use the swap-using-scratch algorithm and associated risks.

### DIFF
--- a/docs/readme-zephyr.md
+++ b/docs/readme-zephyr.md
@@ -15,6 +15,12 @@ partitions defined in its device tree. These partitions are:
 - `boot_partition`: for MCUboot itself
 - `slot0_partition`: the primary slot of Image 0
 - `slot1_partition`: the secondary slot of Image 0
+
+It is not recommended to use the swap-using-scratch algorithm of MCUboot, but
+if this operating mode is desired then the following flash partition is also
+needed (see end of this help file for details on creating a scratch partition
+and how to use the swap-using-scratch algorithm):
+
 - `scratch_partition`: the scratch slot
 
 Currently, the two image slots must be contiguous. If you are running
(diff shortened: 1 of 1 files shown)


nordicjm:
@d3zd3z can this be reviewed please?

docs/readme-zephyr.md
```diff
@@ -156,3 +162,50 @@ although MCUboot's key management infrastructure supports multiple keypairs.
...
+with an 8KB scratch area, the scratch area would be erased and programmed 8
+times per image upgrade/revert). To configure MCUboot to work in
+swap-using-scratch mode, the Kconfig value must be set when building it:
+`CONFIG_BOOT_SWAP_USING_SCRATCH=y`.
+
+Note that it is possible for an application to get into a stuck state when
+swap-using-scratch is used whereby an application has loaded a firmware update
+and marked it as test/confirmed but MCUboot will not swap the images and
+erasing the secondary slot from the zephyr application returns an error
+because the slot is marked for upgrade.
```
utzig, line 211: Is this related to #1165? Was anyone else able to reproduce it or you were the only person that has ever seen the error? Does it also happen when using only internal flash?

nordicjm: Everyone at the company I worked at on multiple boards (pinnacle 100, bt610), people on discord with varying boards. Both boards I've seen it on had primary slots in internal flash, with secondary and scratch partitions in external QSPI memory (all devices with 4KB sector sizes)

utzig: Pinnacle 100 and bt610 are also nrf52840, correct? Do they also use QSPI flash? MCUboot doesn't really know if the flash is internal or external, that's up to the OS backend, but I am wondering if there is some pattern here. How many upgrades in average I would have to run to see the issue? Do you know if uploading new firmware with MCUmgr in between tests is necessary to trigger the issue, or just doing test/revert of the same images is enough?

nordicjm: Not sure on that, it was completely random as to when it happened, as bad as it sounds there was no easily reproducible method to get a device into this state (it was all during development, might have been reported post production too but I do not remember) - images would have been loaded via MCUmgr (bluetooth and/or shell transports) then marked as test from that, the image wasn't left and flipped between test and none.

d3zd3z: Do we need to create an issue to track this? Ideally, we would be able to figure out how to reproduce this in the simulator. Does it happen only with the write sizes differing between internal and external? But, I don't think creating the issue should block merging this issue. Let me know in a comment that there is an issue created, and this can proceed.

nordicjm: I gave it a quick try to reproduce it and was not able to in 20 minutes or so. The write and erase sizes for both memories were the same (mx25... QSPI flash - 4KB sector/erase size, 4 byte minimum write size, nrf52840 flash - 4KB sector/erase size, 4 byte minimum write size), and it was completely random as to how it got into that mode, but once it was in that mode, you were unable to get it out of that mode or load a new update, mcuboot would not do any swapping and the app would not allow uploading a new image because the image was marked for test

utzig: Yes, please open an issue with the same description. I have used swap with scratch and newtmgr together for many years with Mynewt, and to some extent with Zephyr as well, and never bumped into this bug. I am somewhat suspicious it might be related to the QSPI driver, like an error write that does not raise an error or something similar, but it's just a guess. Btw, I've never tried scratch on the QSPI flash, but for MCUboot flash is just flash, wherever it is located...

d3zd3z:
I'll approve this once the tracking issue is created for the problem described about unreliability. Please reference the issue in a comment here.

d3zd3z (approved):

PR #1554, opened 2022-12-21
Remove MCUBOOT_SWAP_USING_SCRATCH being set when MCUBOOT_SINGLE_APPLICATION_SLOT has been selected

de-nordic:
Scratch is not used when MCUBOOT_SINGLE_APPLICATION_SLOT  is used but definition of FLASH_AREA_IMAGE_SCRATCH has been required anyway because it would be default algorithm if nothing else has been selected.


utzig (approved):
LGTM

d3zd3z (approved):

PR #1556, opened 2022-12-21
Missing logic fo MCUBOOT_SWAP_USING_SCRATCH

de-nordic:
There was missing logic for checking whether MCUBOOT_SWAP_USING_SCRATCH is selected, as it was assumed default.
Zephyr Kconfig CONFIG_BOOT_SWAP_USING_SCRATCH has not been propagated to MCUBOOT_SWAP_USING_SCRATCH.


boot/bootutil/src/bootutil_priv.h
```diff
@@ -56,9 +56,10 @@ struct flash_area;
 
 #if (defined(MCUBOOT_OVERWRITE_ONLY) + \
      defined(MCUBOOT_SWAP_USING_MOVE) + \
+     defined(MCUBOOT_SWAP_USING_SCRATCH) + \
      defined(MCUBOOT_DIRECT_XIP) + \
      defined(MCUBOOT_RAM_LOAD)) > 1
-#error "Please enable only one of MCUBOOT_OVERWRITE_ONLY, MCUBOOT_SWAP_USING_MOVE, MCUBOOT_DIRECT_XIP or MCUBOOT_RAM_LOAD"
+#error "Please enable only one of MCUBOOT_OVERWRITE_ONLY, MCUBOOT_SWAP_USING_MOVE, MCUBOOT_SWAP_USING_SCRATCH, MCUBOOT_DIRECT_XIP or MCUBOOT_RAM_LOAD"
```
utzig, line 62: `MCUBOOT_SWAP_USING_SCRATCH` is enabled by default when no other mode is selected, just a few lines below. If you want to also add it to the `#error` message put this `#if` block below the next one.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -73,6 +73,10 @@
 #define MCUBOOT_SWAP_USING_MOVE 1
 #endif
 
+#ifdef CONFIG_BOOT_SWAP_USING_SCRATCH
+#define MCUBOOT_SWAP_USING_SCRATCH 1
```
utzig, line 77: There's no need to have this, this mode is already the default.

utzig (changes requested):
This PR is not required.

utzig:
I just tried swap using scratch with Zephyr on the nrf52840dk and had no issues. What are you trying to achieve?

boot/bootutil/src/bootutil_priv.h
de-nordic: Then the code that defines `MCUBOOT_SWAP_USING_SCRATCH` should be moved out of this file.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
de-nordic: For example I want to have something like this:
```
#ifdef MCUBOOT_SWAP_USING_SCRATCH
...
#endif
```
Now I have to bring in the boot/bootutil/src/bootutil_priv.h to have this defined, or bring the logic that "defines" it, to an unit where I try to do that code.
So maybe the logic from boot/bootutil/src/bootutil_priv.h should be here to auto-magically define the `MCUBOOT_SWAP_USING_SCRATCH`.

utzig: Sure, sounds good. You'll probably want to make it explicit for all ports then, also for the simulator.

PR #1558, opened 2022-12-21
ci: Fix imgtool CI workflow on `main` branch

gustavonihei:
## Summary

**imgtool** relies on `packaging.version.LegacyVersion` to parse image version numbers, and this has been removed in packaging 22.0.
This results in the following error in the **Publish** step:
```python
+ python ../ci/compare_versions.py --old 1.9.0 --new 1.9.0
Traceback (most recent call last):
  File "/home/runner/work/mcuboot/mcuboot/scripts/../ci/compare_versions.py", line 15, in <module>
    from packaging.version import parse, LegacyVersion
ImportError: cannot import name 'LegacyVersion' from 'packaging.version' (/usr/local/lib/python3.10/dist-packages/packaging/version.py)
```
https://github.com/mcu-tools/mcuboot/actions/runs/3741361421/jobs/6350869462

Proposed fix based on https://github.com/pypa/pip-audit/pull/427.

## Impact
Should have no impact. Patch should bring CI back to normal.

## Testing
CI build pass.

PR #1564, opened 2023-01-03
ci: Fix compatibility with packaging==22

utzig:
packaging >= 22 dropped support for LegacyVersion, which was the usual result of an invalid version number being parsed. Now it is PEP-440 strict and throws an exception on fails. This fixes the script to work with both older and newer releases.


utzig:
This fixes the same issue as #1558, but instead of forcing the use of older packaging (<22), this patch adds support for both old and newer releases.

gustavonihei (approved):

PR #1566, opened 2023-01-05
Replace flash_area_sector_from_off with flash_area_get_sector

de-nordic:
The flash_area_sector_from_off  uses hardcoded flash_area object pointer, the flash_area_get_sector takes flash_area pointer.

Commits for:
1.  Zephyr implementation;
2. sim implementation;
3. espressif implementation;
4. general code switch;
5. zephyr/single_loader switch.

### DIFF
--- a/boot/zephyr/flash_map_extended.c
+++ b/boot/zephyr/flash_map_extended.c
@@ -140,3 +140,24 @@ __weak uint8_t flash_area_erased_val(const struct flash_area *fap)
     (void)fap;
     return ERASED_VAL;
 }
+
+int flash_area_get_sector(const struct flash_area *fap, off_t off,
+                          struct flash_sector *fsp)
+{
+    struct flash_pages_info fpi;
+    int rc;
+
+    if (off >= fap->fa_size) {
+        return -ERANGE;
+    }
+
+    rc = flash_get_page_info_by_offs(fap->fa_dev, fap->fa_off + off,
+            &fpi);
+
+    if (rc == 0) {
+        fsp->fs_off = fpi.start_offset - fap->fa_off;
+        fsp->fs_size = fpi.size;
+    }
+
+    return rc;
+}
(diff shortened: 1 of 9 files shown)


boot/zephyr/flash_map_extended.c
```diff
@@ -140,3 +140,24 @@ __weak uint8_t flash_area_erased_val(const struct flash_area *fap)
...
+
+    if (off >= fap->fa_size) {
+        return -ERANGE;
+    }
+
+    rc = flash_get_page_info_by_offs(fap->fa_dev, fap->fa_off + off,
+            &fpi);
+
+    if (rc == 0) {
+        fsp->fs_off = fpi.start_offset - fap->fa_off;
```
nordicjm, line 158: Trying to work back and check this is correct, should this be take away? If start offset is e.g. 0x0, and desired offset is 0x200, the offset would be 0x0 - 0x200 which would underflow to 0xFFFFFE00

de-nordic: Let me think.
The `flash_get_page_info_by_offs` returns page info in context of device, not flash area. So the `fpi.start_offset` will be greater or equal `fap->fa_off`, unless area is not aligned to flash page.
So even if `off` is set to 0, the `fsp->fs_off` should be 0 or next page start offset, unless area is not aligned to flash page.
I should return error here in case when `fpi.start_offset` is less than `fap->fa_off` and replace type of `off` as size_t to prevent negative offsets.
@nordicjm  That is what you want here?

nordicjm (approved):

boot/zephyr/flash_map_extended.c
de-nordic: @nordicjm Do you want me to do the error check for the case with bad flash page alignment, where page starts before area offset?

nordicjm: Don't think there's a need for now

d3zd3z (approved):
This looks good to me. As long as @nordicjm is ok with the underflow concern.

d3zd3z:
Please open an issue if this is discovered to break espressif.

PR #1569, opened 2023-01-05
WIP: RFC: Preliminary work on flash_area open once.

de-nordic:
Work in progress, although Zephyr seems to already work and flash usage has been reduced.


boot/zephyr/include/flash_map_backend/flash_map_backend.h
```diff
@@ -62,6 +62,21 @@ int flash_area_id_to_multi_image_slot(int image_index, int area_id);
...
+/* Get sectors for given flash_area object. This function is similar to
+ * flash_area_get_sectors but takes flash_area object pointer instead
+ * of flash area identifier.
+ *
+ * @param fa    pointer to flash_area object.
+ * @param count in: size of array for returned sectors, out: number
+ *              of sectors filled in.
+ * @param ret   array of sectors.
+ *
+ * Returns 0 on success, -ERANGE when there is not enough space in
```
nordicjm, line 76: ERANGE looks wrong

boot/zephyr/flash_map_extended.c
```diff
@@ -135,3 +135,33 @@ __weak uint8_t flash_area_erased_val(const struct flash_area *fap)
     (void)fap;
     return ERASED_VAL;
 }
+
+int flash_area_get_sectors_fa(const struct flash_area *fa, uint32_t *count,
+    struct flash_sector *ret)
+{
+    off_t fa_off = fa->fa_off;
+    off_t offset = 0;
+    size_t fa_size = fa->fa_size;
```
nordicjm, line 155: fa_off and fa_size only seem to be used for reading, can't the fa structure be used directly instead?

boot/bootutil/src/loader.c
```diff
@@ -240,12 +240,7 @@ boot_read_image_size(struct boot_loader_state *state, int slot, uint32_t *size)
...
 
-    area_id = flash_area_id_from_multi_image_slot(BOOT_CURR_IMG(state), slot);
-    rc = flash_area_open(area_id, &fap);
-    if (rc != 0) {
-        rc = BOOT_EFLASH;
-        goto done;
-    }
+    fap = flash_area_from_multi_image_slot(BOOT_CURR_IMG(state), slot);
 
     off = BOOT_TLV_OFF(boot_img_hdr(state, slot));
```
nordicjm, line 246: Would need to check if there is any way that this function (from any call) could be passed a different value, e.g. by fault injection, because if so then the bootloader operation becomes undefined

boot/zephyr/flash_map_extended.c
de-nordic: Yes it can.

boot/zephyr/include/flash_map_backend/flash_map_backend.h
d3zd3z: ERANGE seems appropriate here.

PR #1574, opened 2023-01-11
zephyr: Disable and deprecate flash erase

### LINKED ISSUE #1573: zephyr: Disable flash erash when flashing
This setting should be disabled by default, or entirely removed. In zephyr, west allows erasing targets when they are flashed, an image should not be deciding that for the user. This causes headaches for sysbuild because depending upon flash order, mcuboot can nuke what's already been programmed to flash, therefore it makes sense to control this as it does with every other zephyr image by having this option available in west, not forcing it at an application level.

nordicjm:
This deprecates the flash erase Kconfig for zephyr, if this action is required then the board should be flashed using west with the `--erase` argument supplied instead.

Fixes #1573

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -481,8 +481,8 @@ config BOOT_USB_DFU_DETECT_DELAY
 endif # BOOT_USB_DFU_GPIO
 
 config ZEPHYR_TRY_MASS_ERASE
-	bool "Try to mass erase flash when flashing MCUboot image"
-	default y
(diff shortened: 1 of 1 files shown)


nordicjm:
@d3zd3z can this be reviewed please?

de-nordic (approved):
Looks OK.

de-nordic:
When synced to Zephyr, this will require release notes entry.

mbolivar-nordic (commented):
I believe it was me who originally added this option and I support its deprecation and eventual removal now that we have better mechanisms like `west flash --erase`.

PR #1576, opened 2023-01-13
provide for non-console recovery uart

brian-level:
In some situations such as having multiple CPUs that can upgrade another it is nice to allow another uart to use mcuboot serial recovery while (for example) keeping a console on another.

This change allows someone to specify a recover uart to use other than chosen zephyr-console  by adding to a device tree as in the examples below

aliases {
    recovery-uart=&uart1;
       or
    mcuboot-recovery-uart=&uart0;
 };

OR adding mcuboot-recovery-uart to the chosen node 

If not specified, the chosen zephyr-console uart will be used

### DIFF
--- a/boot/zephyr/serial_adapter.c
+++ b/boot/zephyr/serial_adapter.c
@@ -22,9 +22,24 @@
 #include "bootutil/bootutil_log.h"
 #include <zephyr/usb/usb_device.h>
 
-#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE)
-#error Zephyr UART console must been disabled if serial_adapter module is used.
+#if defined(CONFIG_BOOT_SERIAL_UART)
+	#if (DT_NODE_HAS_STATUS(DT_CHOSEN(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_CHOSEN(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(recovery_uart)
+	#else
+		/* note that if a recovery uart is specified, and it happens to be
+		 * the same node as zephyr-console, we don't detect that at build time
+		 */
+		#define RECOVERY_UART DT_CHOSEN(zephyr_console)
+		#if defined(CONFIG_UART_CONSOLE)
+			#error Zephyr UART console must be disabled if serial_adapter module is used.
+		#endif
+	#endif
 #endif
+
 
 BOOT_LOG_MODULE_REGISTER(serial_adapter);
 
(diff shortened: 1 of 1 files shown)


boot/zephyr/serial_adapter.c
```diff
@@ -22,10 +22,25 @@
...
-#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE)
-#error Zephyr UART console must been disabled if serial_adapter module is used.
+#if defined(CONFIG_BOOT_SERIAL_UART)
+	#if (DT_NODE_HAS_STATUS(DT_CHOSEN(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_CHOSEN(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(recovery_uart)
+	#else
```
nordicjm, line 32: There should not be an alias and a chosen for this, that becomes confusing, can stick with chosen since it is a singular node device for a whole board.

boot/zephyr/serial_adapter.c
```diff
@@ -22,10 +22,25 @@
...
 
-#if defined(CONFIG_BOOT_SERIAL_UART) && defined(CONFIG_UART_CONSOLE)
-#error Zephyr UART console must been disabled if serial_adapter module is used.
+#if defined(CONFIG_BOOT_SERIAL_UART)
+	#if (DT_NODE_HAS_STATUS(DT_CHOSEN(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_CHOSEN(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(mcuboot_recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(mcuboot_recovery_uart)
+	#elif (DT_NODE_HAS_STATUS(DT_ALIAS(recovery_uart), okay))
+		#define RECOVERY_UART DT_ALIAS(recovery_uart)
```
nordicjm, line 31: Why is there another one?

nordicjm (changes requested):

boot/zephyr/serial_adapter.c
brian-level: This was to be backwards compatible with device trees we already had.  The general solution could just be chosen.

d3zd3z (changes requested):
The commits will need a sign-off-by line, and I recommend configuring git so that your email has your full name, rather than a shortened handle. You'll need to do an interactive rebase, and use `git commit --amend --reset-author` to fix the author, once you have changed the git configuration.

brian-level:
My email doesn't have my full name, it is just what it is.  Can you please
provide more detailed instructions for these comments (assuming they apply
to me)

kmeinhar:
This is very similar to this PR #1155 right?

d3zd3z:
> My email doesn't have my full name, it is just what it is.

It is set within git, and the username can be set to whatever you want. As far as adding the signed off by, that can be done by either inserting the text into the commit message manually, or passing `-s` to the git commit command (possibly with --amend to update the top commit).

d3zd3z:
> This is very similar to this PR #1155 right?

@de-nordic can you take a look at this?

de-nordic:
> > This is very similar to this PR #1155 right?
> 
> @de-nordic can you take a look at this?

I would rather go with #1155, because I do not want to introduce MCUboot chosen, while this is Zephyr specific code, and we have the chosen for MCUmgr anyway. If we want to have the MCUboot chosen, it should be used as override for zephyr chosen.

de-nordic:
https://github.com/mcu-tools/mcuboot/pull/1155 went in, is this PR still relevant?

PR #1578, opened 2023-01-19
loader.c: fill_rsp - Fix possible array index overflow - L1 CERT-C

DOAR-Infineon:


boot/bootutil/src/loader.c
```diff
@@ -182,6 +182,11 @@ fill_rsp(struct boot_loader_state *state, struct boot_rsp *rsp)
             break;
         }
     }
+    /* At least one image must be active, otherwise skip the execution */
+    if(BOOT_CURR_IMG(state) >= BOOT_IMAGE_NUMBER)
```
d3zd3z, line 186: Should be a space between the 'if' and the '('.

boot/bootutil/src/loader.c
```diff
@@ -182,6 +182,11 @@ fill_rsp(struct boot_loader_state *state, struct boot_rsp *rsp)
...
+    {
```
d3zd3z, line 187: Brace should be at the end of the previous line, preceded by a space.

d3zd3z (changes requested):

de-nordic (approved):

d3zd3z:
The commits will need Signed off by lines. I recommend the author using their real name in the Author field. Let us know if you need help with this, as it is a fairly simple change that one of the maintainers could do, but we do need your permission to fix the author email address and add the signed-off-by.

d3zd3z (changes requested):
These should probably just be squashed as well.

boot/bootutil/src/loader.c
DOAR-Infineon: Fixed

boot/bootutil/src/loader.c
DOAR-Infineon: Fixed

d3zd3z (approved):

PR #1580, opened 2023-01-23
espressif:ESP32-S3: Fix multiboot APP CPU start

almir-okato:
Add missing function for starting the APP CPU when booting the second image (multi image).


almir-okato:
@gustavonihei PTAL

gustavonihei (approved):
LGTM

PR #1582, opened 2023-01-24
MCUboot on ESP32 build as Zephyr application

marekmatej:
This PR provides building MCUboot for ESP32 targets as any other Zephyr application.


boot/zephyr/CMakeLists.txt
```diff
@@ -252,6 +252,26 @@ if(CONFIG_MCUBOOT_SERIAL)
     )
 endif()
 
+if(CONFIG_SOC_ESP32 OR CONFIG_SOC_ESP32S2 OR CONFIG_SOC_ESP32C3 OR CONFIG_SOC_ESP32S3)
```
sylvioalves, line 255: We plan to have some `CONFIG_FAMILY_SOC_ESP32` soon. For now, this works.

boot/zephyr/CMakeLists.txt
```diff
@@ -252,6 +252,26 @@ if(CONFIG_MCUBOOT_SERIAL)
...
+
+  if(CONFIG_SOC_ESP32)
+    zephyr_compile_definitions(CONFIG_ESP32_XTAL_FREQ=40)
+  endif()
+  zephyr_compile_definitions(BOOTLOADER_BUILD=1)
+  zephyr_compile_definitions(CONFIG_MCUBOOT=1)
```
sylvioalves, line 261: `CONFIG_MCUBOOT` this should be 1 by the buildsystem. Is it required in here?

boot/zephyr/CMakeLists.txt
```diff
@@ -252,6 +252,26 @@ if(CONFIG_MCUBOOT_SERIAL)
...
+  zephyr_compile_definitions(NDEBUG=1)
```
sylvioalves, line 262: If `BOOTLOADER_BUILD` and `NDEBUG` are fixed, could it be placed somewhere else? sdkconfig.h?

boot/zephyr/main.c
```diff
@@ -42,6 +42,35 @@
 #include "bootutil/mcuboot_status.h"
 #include "flash_map_backend/flash_map_backend.h"
 
+/* Check if Espressif target is supported */
+#if defined(CONFIG_SOC_ESP32) || \
```
sylvioalves, line 46: What if you add a new kconfig at https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/Kconfig defining 
```c
config ESPRESSIF_SOC
    depends on CONFIG_SOC_ESP32 || (...)
```

boot/zephyr/CMakeLists.txt
```diff
@@ -252,6 +252,26 @@ if(CONFIG_MCUBOOT_SERIAL)
...
+
+  zephyr_include_directories(${BOOT_DIR}/espressif/include)
+  zephyr_include_directories(${BOOT_DIR}/espressif/hal/include)
+
+  zephyr_sources (
+    ${BOOT_DIR}/espressif/port/esp_loader.c
+    ${BOOT_DIR}/espressif/hal/src/bootloader_init_common.c
+    ${BOOT_DIR}/espressif/hal/src/bootloader_wdt.c
+    ${BOOT_DIR}/espressif/hal/src/${CONFIG_SOC}/bootloader_init.c
+  )
```
gustavonihei, line 272: As convenient as it may be, it seems conceptually wrong for one port (`boot/zephyr`) to reference files from any other port (`boot/espressif`).

marekmatej: First I used `hal_espressif` versions of `boot/espressif/hal/src/bootloader_.*.c` files, but any change/fix in the `bootloader_init` function asked to expose port private/static members which ends up colliding/creating multiple definitions due to IDF sources organization. Then I noticed very same sources are available in the `boot/espressif/` which lacks all the disadvantages and solves all previous issues. The `esp_loader.c` reference is different and much easier to remove from the Zephyr port. 

I feel your concerns, but think about this: It is wrong to reference the code from different SOC port just to use some function I am lazy to write. But here we share the same SOC early init functions (loaded by the the same ROM loader), then continuing entirely Zephyr way. So I think it is close to DRY principle.

But, maybe there are other reasons for not doing it and I'm open to discuss :)

gustavonihei: > I feel your concerns, but think about this: It is wrong to reference the code from different SOC port just to use some function I am lazy to write. But here we share the same SOC early init functions (loaded by the the same ROM loader), then continuing entirely Zephyr way. So I think it is close to DRY principle.

Applying DRY principle here comes at the cost of introducing coupling between two distinct ports, which in my opinion makes it prohibitive. So for now I believe duplicating the code is a better alternative.

boot/zephyr/CMakeLists.txt
marekmatej: I added `CONFIG_FAMILY_SOC_ESP32` into related Zephyr PR#54835

almir-okato:
LGTM

marekmatej:
depends on https://github.com/zephyrproject-rtos/zephyr/pull/56756

marekmatej:
The workflow needs approval, @gustavonihei can you take a look please?

gustavonihei (approved):

marekmatej:
@utzig, @d3zd3z  could you please take a look? The CI is failing on the Sim and my attempt to fix it was not successful.

utzig:
> @utzig, @d3zd3z could you please take a look? The CI is failing on the Sim and my attempt to fix it (PR#1689) was not successful.

#1693

utzig:
@marekmatej Rebase.

gustavonihei (approved):

PR #1587, opened 2023-01-30
Add config for rsa signature support

Vfgb:


almir-okato:
@Vfgb you may squash your commits and avoid left "Merge commits" when opening a pull request.

boot/nuttx/include/mcuboot_config/mcuboot_config.h
```diff
@@ -39,6 +39,18 @@
...
+       CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN != 3072)
+#    error "Invalid RSA key size (must be 2048 or 3072)"
+#  else
+#    define MCUBOOT_SIGN_RSA_LEN CONFIG_BOOT_SIGNATURE_TYPE_RSA_LEN
+#  endif
+#endif
+
 /* Uncomment for RSA signature support */
 
 /* #define MCUBOOT_SIGN_RSA */
```
almir-okato, line 56: As you added the above config, you may remove these lines.

boot/nuttx/keys.c
```diff
@@ -0,0 +1,55 @@
...
+    0xc9, 0x02, 0x03, 0x01, 0x00, 0x01,
+};
+const unsigned int rsa_pub_key_len = 270;
+#endif
+
+#if defined(HAVE_KEYS)
+const struct bootutil_key bootutil_keys[] = {
+    {
+        .key = rsa_pub_key,
+        .len = &rsa_pub_key_len,
```
almir-okato, line 50: Don't forget to wrap this with `#if defined(MCUBOOT_SIGN_RSA)`.

utzig:
@Vfgb Please update your PR here, there's no need to open new PRs, and we can't track what reviews where already done. Also don't use `main` as your PR branch, make your changes on a new branch, eg: `git checkout -b nuttx-rsa-config`. Then always force push that branch to fix changes suggested in the review. You can change the origin branch of the PR with the `edit` button at the right of the title.

boot/nuttx/keys.c
Vfgb: yeah you are right!

Vfgb:
@utzig Thank you for your guidance, i am sorry for my wrong operation. And i can't to change the origin branch with the edit button. I can only do that with a new pr. I wonder if I can do in this way？

utzig:
> @utzig Thank you for your guidance, i am sorry for my wrong operation. And i can't to change the origin branch with the edit button. I can only do that with a new pr. I wonder if I can do in this way？

Oh my bad, sorry! I remember doing it, but it was the target branch which can be changed. Well you can decide if you are able to work using the same branch name and then re-open this PR, or otherwise reopen the one you had opened from your new branch today.

PR #1589, opened 2023-01-30
espressif: ci: Add new building jobs configs for Espressif chips 

almir-okato:
Add Github jobs for testing build ESP32-XX within more features: Serial Recovery, Multi Image and Multi Boot

This PR also adds a special case on the `check-signed-off-by.sh` script for skipping the parent merge commit for Gitlab fork usage.


ci/check-signed-off-by.sh
```diff
@@ -12,17 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# this retrieves the merge commit created by GH
-parents=(`git log -n 1 --format=%p HEAD`)
+giturl=(`git config --get remote.origin.url`)
+repo='gitlab'
+if [[ "$giturl" == *"$repo"* ]]; then
+  parent=(`git log -n 1 --format=%p HEAD`)
```
utzig, line 18: Does this result in a single element array on Gitlab?

ci/check-signed-off-by.sh
```diff
@@ -12,17 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# this retrieves the merge commit created by GH
-parents=(`git log -n 1 --format=%p HEAD`)
+giturl=(`git config --get remote.origin.url`)
```
utzig, line 15: I think you don't need to use parenthesis here, either use the back quote to enclose the command, or `$(...)`

ci/check-signed-off-by.sh
```diff
@@ -12,17 +12,24 @@
...
+repo='gitlab'
+if [[ "$giturl" == *"$repo"* ]]; then
```
utzig, line 17: I'd rather just use `*"gitlab"*` here, since `repo` is too generic a name.

utzig (commented):
Commit b66ee68 is fine. I'm not sure why Gitlab support needs to be added here, do you use a fork there? How about maintaining the fix only there?

ci/check-signed-off-by.sh
almir-okato: Yes, I was wondering what I could do about that. I've enabled Gitlab Runners on our fork and wanted to use mostly of the existent scripts like this one instead of duplicating them. I'm thinking about the alternatives.

ci/check-signed-off-by.sh
almir-okato: I changed that and used "github" as first case, so the `else` statement would be for "others".

utzig: If you don't mind a suggestion, I would rather not hard-code the name of the CI, but instead check how many parents there are and make decisions based on that, something like:

```sh
parents=(`git log -n 1 --format=%p HEAD`)

if [[ "${#parents[@]}" -eq 1 ]]; then
  # CI doesn't use a merge commit
  commits=$(git show -s --format=%h ${parents})
elif [[ "${#parents[@]}" -eq 2 ]]; then
  # CI uses a merge commit, eg GH / Travis
  from="${parents[0]}"
  into="${parents[1]}"
  commits=$(git show -s --format=%h ${from}..${into})
else
  echo "CI has 3 parents? Don't know what to do!"
  exit 1
fi
```

What do you think? Btw, if this works for your CI and you are OK with the change, please add a temporary "non signed" commit, so we see the fail!

almir-okato: Thanks for the suggestion!
It could work, however, it would remove this verification and maybe affect the script purpose:
```
  if [[ "${#parents[@]}" -ne 2 ]]; then
    echo "This PR's merge commit is missing a parent!"
    exit 1
  fi
```
Are you more familiar with that? Would it be an issue if we change this verification?

utzig: It's not a problem, after this PR we'll deal with values 1 and 2, and fail for other values.

almir-okato: Ok, I've changed that as you suggested.
Thanks!

utzig (approved):
LGTM

PR #1591, opened 2023-02-01
Add config fot rsa support in nuttx

Vfgb:
Added configuration to enable nuttx to use rsa sign.


utzig:
Seems to be a "duplicate" of #1587

PR #1593, opened 2023-02-03
Add config for rsa sign in nuttx

Vfgb:


utzig:
The commit needs to be signed to be accepted. Also why would you prefer RSA over ECSA?

Vfgb:
> The commit needs to be signed to be accepted. Also why would you prefer RSA over ECSA?
because in my project , use ECSA takes up more flash spac. And this is the first time I've used signature technology, I don't know much about the difference between them. So I used RSA.

utzig:
> > because in my project , use ECSA takes up more flash spac. And this is the first time I've used signature technology, I don't know much about the difference between them. So I used RSA.

ECDSA should be smaller at least if using Tinycrypt. Keys are also smaller.

Vfgb:
> > > because in my project , use ECSA takes up more flash spac. And this is the first time I've used signature technology, I don't know much about the difference between them. So I used RSA.
> 
> ECDSA should be smaller at least if using Tinycrypt. Keys are also smaller.

Thanks for your advice, I have been using MBEDTLS. I tried Tinycrypt, ECDSA is indeed smaller.

Vfgb:
@utzig  is this can be reviewed please?

utzig:
> @utzig is this can be reviewed please?

It will. I wanted to test it locally before approving/merging, but haven't had much spare time.

Vfgb:
> > @utzig is this can be reviewed please?
> 
> It will. I wanted to test it locally before approving/merging, but haven't had much spare time.

ok, thank you!

Vfgb:
@d3zd3z can you help me to review it please?

d3zd3z (changes requested):
The changes themselves look ok, but the commits themselves need a bit more descriptions. The commit text should explain that this is putting in the public keys from the development keys, and that this isn't useful for production.

Also, the last commit seems to just change the key from the previous commit. Can these commits be squashed.

Also, please add release notes, as per `docs/contributing.md`.

Vfgb:
> The changes themselves look ok, but the commits themselves need a bit more descriptions. The commit text should explain that this is putting in the public keys from the development keys, and that this isn't useful for production.
> 
> Also, the last commit seems to just change the key from the previous commit. Can these commits be squashed.
> 
> Also, please add release notes, as per `docs/contributing.md`.

I modify the commit texts and squashe these commits, but i can't find the file contributing.md. 
Could you look over the changes for me first, and tell me where the contributing.md is?
Thank you very much for your help!

PR #1601, opened 2023-02-10
bootutil: Little rework of boot_find_status

de-nordic:
In rework:
  - size has been removed from array, as it has been making array at least of two elements, even if scratch partition not used;
  - some variables have been localized and some removed;
  - no longer returning rc codes from flash_area_ functions on failure, just single point of failure-exit - it made no sense anyway as we can not recover from flash failure;
  - flash area pointer is now NULLified on failure, or when status not found.

When building for SWAP_MOVE on nrf52840dk_nrf52840 this reduces flash usage by 16bytes, which is not so great but ate least 16bytes less.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -242,16 +242,13 @@ boot_enc_key_off(const struct flash_area *fap, uint8_t slot)
 static int
 boot_find_status(int image_index, const struct flash_area **fap)
 {
-    uint8_t magic[BOOT_MAGIC_SZ];
-    uint32_t off;
-    uint8_t areas[2] = {
+    uint8_t areas[] = {
 #if MCUBOOT_SWAP_USING_SCRATCH
         FLASH_AREA_IMAGE_SCRATCH,
 #endif
         FLASH_AREA_IMAGE_PRIMARY(image_index),
     };
     unsigned int i;
-    int rc;
 
     /*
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/bootutil_misc.c
```diff
@@ -262,26 +259,26 @@ boot_find_status(int image_index, const struct flash_area **fap)
...
+        if (flash_area_read(*fap, boot_magic_off(*fap), magic, BOOT_MAGIC_SZ)) {
+            flash_area_close(*fap);
+            break;
         }
 
         if (BOOT_MAGIC_GOOD == boot_magic_decode(magic)) {
             return 0;
         }
 
+        flash_area_close(*fap);
```
utzig, line 277: Shouldn't this `flash_area_close` happen before the test for good magic above?

de-nordic: No, the area needs to remain open. I have  have introduced bug with this commit https://github.com/mcu-tools/mcuboot/pull/1183/files, as `boot_find_status` should leave the area open.
Should I revert the change first and then continue with the rework?

utzig: I don't remember #1183, but the change makes sense if `fap` is the function's "return". I suppose it didn't break anything because "close" doesn't really do much! :-P

utzig (approved):
LGTM

d3zd3z (approved):

PR #1604, opened 2023-02-14
warningfix: bootutil_priv.h

xuzihan351:
- fix warning while MCUBOOT_SWAP_USING_MOVE is not defined

Signed-off-by: Zihan XU <zihan.xu@hpmicro.com>

### DIFF
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -181,8 +181,7 @@ _Static_assert(sizeof(boot_img_magic) == BOOT_MAGIC_SZ, "Invalid size for image 
                  (hdr)->ih_ver.iv_minor,                                  \
                  (hdr)->ih_ver.iv_revision,                               \
                  (hdr)->ih_ver.iv_build_num)
-
-#if MCUBOOT_SWAP_USING_MOVE
+#if defined(MCUBOOT_SWAP_USING_MOVE) && MCUBOOT_SWAP_USING_MOVE
 #define BOOT_STATUS_MOVE_STATE_COUNT    1
 #define BOOT_STATUS_SWAP_STATE_COUNT    2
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/bootutil_priv.h
```diff
@@ -180,15 +180,17 @@ _Static_assert(sizeof(boot_img_magic) == BOOT_MAGIC_SZ, "Invalid size for image
                  (hdr)->ih_ver.iv_minor,                                  \
                  (hdr)->ih_ver.iv_revision,                               \
                  (hdr)->ih_ver.iv_build_num)
-
+#if defined(MCUBOOT_SWAP_USING_MOVE)
 #if MCUBOOT_SWAP_USING_MOVE
```
d3zd3z, line 184: Can we not just change line 184 to the if defined, instead of wrapping with an additional #if?

xuzihan351: Yes, it is better to do that. I have changed line 184 to [#if defined(MCUBOOT_SWAP_USING_MOVE) && MCUBOOT_SWAP_USING_MOVE].
Because someone may set MCUBOOT_SWAP_USING_MOVE to 0.

utzig:
This is fine, but it shouldn't be required, what compiler are you using?

utzig (approved):

utzig:
> This is fine, but it shouldn't be required, what compiler are you using?

Looking at the submodule added with #1605, looks like gcc for riscv32!

xuzihan351:
Yes, it is gcc for riscv32.

de-nordic (approved):

nvlsianpu (approved):

utzig:
The committer email is correct now, but author email is still "@hpmciro.com". Please fix.

PR #1610, opened 2023-02-15
ci: Skip sign-off checks for dependabot

de-nordic:
Dependabot uses different e-mails for signoff and commit.

### DIFF
--- a/ci/check-signed-off-by.sh
+++ b/ci/check-signed-off-by.sh
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+DEPENDABOT_COMMITER='GitHub <noreply@github.com>'
+DEPENDABOT_AUTHOR='dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>'
+
+# this retrieves the merge commit created by GH
 parents=(`git log -n 1 --format=%p HEAD`)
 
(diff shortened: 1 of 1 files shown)


ci/check-signed-off-by.sh
```diff
@@ -32,6 +48,10 @@ for sha in $commits; do
   author="Signed-off-by: $(git show -s --format="%an <%ae>" ${sha})"
   committer="Signed-off-by: $(git show -s --format="%cn <%ce>" ${sha})"
 
+  if commit_by_dependabot "${sha}"; then
+    continue
+  fi
+
```
utzig, line 54: Instead of creating a function and re-doing the git work why not just check for the expected text appended in the already existing variables?

I would suggest something on the lines of:

```sh
# XXX: somewhere at the beginning
BOT_AUTHOR="dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>"
BOT_COMMITER="GitHub <noreply@github.com>

# XXX: The skip dependabot test
if [[ $author == *$BOT_AUTHOR && $committer == *$BOT_COMMITTER ]]; then
  continue
fi
```

```

de-nordic: Removed the function.

utzig (approved):

d3zd3z (approved):

PR #1615, opened 2023-02-17
ci: Fix missing range for signoff check with single parent

de-nordic:
Only parent commit has been checked instead of pr commits, due to missing range in git show.

### DIFF
--- a/ci/check-signed-off-by.sh
+++ b/ci/check-signed-off-by.sh
@@ -16,7 +16,7 @@ parents=(`git log -n 1 --format=%p HEAD`)
 
 if [[ "${#parents[@]}" -eq 1 ]]; then
   # CI doesn't use a merge commit
-  commits=$(git show -s --format=%h ${parents})
+  commits=$(git show -s --format=%h ${parents}...HEAD)
 elif [[ "${#parents[@]}" -eq 2 ]]; then
   # CI uses a merge commit, eg GH / Travis
   from="${parents[0]}"


de-nordic:
@utzig I am not sure about what is the context the CI check is called in but it seems that single parent path has been missing range.
When I have been invoking the unmodified script locally only the parent commit, not the one I have been adding, would have been checked by the signoff check logic.

utzig:
> @utzig I am not sure about what is the context the CI check is called in but it seems that single parent path has been missing range. When I have been invoking the unmodified script locally only the parent commit, not the one I have been adding, would have been checked by the signoff check logic.

Makes sense, also the way it's written seems to check only a single commit, no matter how many commits are in the PR. Could you take a look @almir-okato?

utzig:
> When I have been invoking the unmodified script locally only the parent commit, not the one I have been adding, would have been checked by the signoff check logic.

If you want to properly simulate GH actions you can always create a merge commit between the origin branch and the branch you are using for testing, so it will execute the same steps.

de-nordic:
> 
> If you want to properly simulate GH actions you can always create a merge commit between the origin branch and the branch you are using for testing, so it will execute the same steps.

Thanks!

almir-okato:
> > @utzig I am not sure about what is the context the CI check is called in but it seems that single parent path has been missing range. When I have been invoking the unmodified script locally only the parent commit, not the one I have been adding, would have been checked by the signoff check logic.
> 
> Makes sense, also the way it's written seems to check only a single commit, no matter how many commits are in the PR. Could you take a look @almir-okato?

Sorry, I didn't notice that issue before when making changes to this script.
Yes, I agree.

de-nordic:
@d3zd3z @utzig Are we going forward with this?

utzig:
> @d3zd3z @utzig Are we going forward with this?

I don't think this is a good solution. The way I thought of doing it, was to pass an optional parameter which is the starting sha of the list of commits to check, so without a parameter go for a merge commit (must have two parents), if a sha is passed use `sha...HEAD` as parents. When I get back from class I'll try to do it, otherwise if I don't send a PR and you agree with the solution feel free to update your PR tomorrow.

de-nordic:
Closed as #1640 has been merged to fix the same issue.

Issue #1584, opened 2023-01-27
CI: Improve FIH job result assesment


tamasban:
Continuation of the thread in the #1528 PR.

> Defining when and how to merge is your policy but I'm agree with @tamasban proposition.
> 
> We can make the FIH CI fail if any successful boot happens below a certain threshold. I mean if the PC is advanced less than 4. This can 1-2 instruction skip, depending on the actual instruction. But the result of FIH CI won't block the merging. If a successful boot happens with values above the threshold then CI will pass, because we think that in the majority of the cases skipping that much instruction is unrealistic.
> 
> CI shall fail if a sucessful boot is obtained by bypassing one (or two instructions) as it would completly defeat the purpose of the FIH mechanisms
> 
> Above this threshold, as there is problably no way to mark a workflow as "almost failed", i would let it pass.

PR #1618, opened 2023-02-20
CI: Improve FIH job result assesment

Roolli:
This pull request modifies the FIH CI job to fail in case a successful boot happens.

The CI should fail if a successful boot is achieved by bypassing one or two instructions as it would defeat the purpose of the FIH mechanisms. The relevant workflow has also been modified to disable the auto-cancel feature if one of the configs fail.

relevant issue: #1584

### DIFF
--- a/.github/workflows/fih_tests.yaml
+++ b/.github/workflows/fih_tests.yaml
@@ -13,13 +13,17 @@ concurrency:
 jobs:
   config:
     strategy:
+      fail-fast: false
       matrix:
         fih_env:
         # FIH environment must use the following space separated format:
         #   BUILD_TYPE SKIP_SIZE DAMAGE_TYPE FIH_LEVEL(optional)
-        - "RELEASE 2,4,6,8,10 SIGNATURE"
-        - "RELEASE 2,4,6,8,10 SIGNATURE LOW"
-        - "RELEASE 2,4,6,8,10 SIGNATURE MEDIUM"
+        - "RELEASE 2,4,6 SIGNATURE"
+        - "RELEASE 2,4,6 SIGNATURE LOW"
+        - "RELEASE 2,4,6 SIGNATURE MEDIUM"
+        - "RELEASE 8,10 SIGNATURE"
+        - "RELEASE 8,10 SIGNATURE LOW"
+        - "RELEASE 8,10 SIGNATURE MEDIUM"
         - "MINSIZEREL 2,4,6 SIGNATURE"
         - "MINSIZEREL 2,4,6 SIGNATURE LOW"
         - "MINSIZEREL 2,4,6 SIGNATURE MEDIUM"
--- a/ci/fih_test_docker/generate_test_report.py
+++ b/ci/fih_test_docker/generate_test_report.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 Arm Limited
+# Copyright (c) 2020-2023 Arm Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


davidvincze (approved):

tamasban (approved):

ci/fih_test_docker/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        # The address was successfully skipped by the debugger
+        'SKIPPED': 'Address was skipped',
+        'NO_BOOT': 'System not booted (desired behaviour)',
+        'BOOT': 'System booted (undesired behaviour)'
+}
+
+
+def parse_yaml_file(filepath):
+    with open(filepath) as output_yaml_file:
+        results = yaml.safe_load(output_yaml_file)
```
utzig, line 33: Not really an output file if you are loading it, right? In a limited scope like this just use `f`.

ci/fih_test_docker/validate_output.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+from utils import CATEGORIES, parse_yaml_file
+
+
+def validate_output(test_stats, skip_size, fih_level):
+    if(test_stats[CATEGORIES['BOOT']] > 0
```
utzig, line 20: Add a space between `if` and `(`.

ci/fih_test_docker/validate_output.py
```diff
@@ -0,0 +1,39 @@
...
+def main():
+    parser = argparse.ArgumentParser(description='''Process a FIH test output yaml file,
+     and validate no sucessfull boots have happened''')
+    parser.add_argument('filename', help='yaml file to process')
+    parser.add_argument('skip_size', help='instruction skip size')
+    parser.add_argument('fih_level', nargs="?",
+                        help='fault injection hardening level')
+
+    args = parser.parse_args()
+    (test_stats, _, _) = parse_yaml_file(args.filename)
```
utzig, line 34: No need for parenthesis on the left side, also you could just do:

```python
test_stats = parse_yaml_file(args.filename)[0]
```

ci/fih_test_docker/utils.py
```diff
@@ -0,0 +1,62 @@
...
+                continue
+
+            if test["boot"]:
+                test_stats.update([CATEGORIES['BOOT']])
+                continue
+
+            failed_boot_last_lines.update([test["last_line"]])
+        else:
+            exec_fail_reasons.update([test["test_exec_fail_reason"]])
+    return (test_stats, failed_boot_last_lines, exec_fail_reasons)
```
utzig, line 62: Add on empty line above the return, and the return does not need parenthesis here.

ci/fih_test_docker/generate_test_report.py
```diff
@@ -13,47 +13,12 @@
...
-                continue
-
-            if test["boot"]:
-                test_stats.update([CATEGORIES['BOOT']])
-                continue
-
-            failed_boot_last_lines.update([test["last_line"]])
-        else:
-            exec_fail_reasons.update([test["test_exec_fail_reason"]])
+    (test_stats, failed_boot_last_lines, exec_fail_reasons) = results
```
utzig, line 21: No need for parenthesis.

.github/workflows/fih_tests.yaml
```diff
@@ -13,13 +13,17 @@ concurrency:
 jobs:
   config:
     strategy:
+      fail-fast: false
```
utzig, line 16: Why?

ci/fih_test_docker/validate_output.py
Roolli: Fixed

ci/fih_test_docker/validate_output.py
Roolli: Fixed

ci/fih_test_docker/utils.py
Roolli: Fixed

ci/fih_test_docker/generate_test_report.py
Roolli: Fixed

.github/workflows/fih_tests.yaml
Roolli: This reduces the amount of times you need to run the whole CI, if there are different errors between different configs, but @tamasban can clarify it more.

utzig: I have never used this option, but I think that failing fast is better than failing slow! :rofl:

utzig: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Maybe what you want to achieve is to let other workflows run to the end even if the FIH tests fail? @tamasban

tamasban: AFAIU this config option has no effect on other workflows like sim, Zephyr, etc. Its scope is restricted only to the FIH jobs.

The goal is to let the other jobs run until the end within the FIH test suite even though one fails. With this approach, one can see all build or FI-related issues after a single CI execution and failures are not revealed only one by one in consecutive CI executions.

This is not aligned with the rest of the CI job configs. We can revert this change if you want to keep all the jobs config aligned.

utzig (approved):
LGTM

d3zd3z (approved):

PR #1625, opened 2023-02-23
imgtool: option to read key passphrase from environment

crazyscot:
In a CI chain it is sometimes useful to automatically sign an image (e.g. once qualification tests have passed). Naturally, it is important to protect the private key in such cases.

This patch provides that capability.

The private key should be stored in a passphrase-protected PEM file in the usual way. The CI provider should be configured, through its secrets mechanism, to inject the passphrase as an environment variable of your choice. A new imgtool option is then used to specify that variable name, e.g.:
   imgtool --password-env=DEV_KEY_PASSPHRASE sign infile.hex outfile.hex

This option also works for the other verbs that load keys (getpub, getpriv, imgfile).

Note that argument ordering is critical. --password-env must appear before the verb!

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -21,6 +21,7 @@ import re
 import click
 import getpass
 import imgtool.keys as keys
+import os
 import sys
 import base64
 from imgtool import image, imgtool_version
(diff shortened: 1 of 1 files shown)


utzig:
@crazyscot The commit needs to be signed.

crazyscot:
> @crazyscot The commit needs to be signed.

Done, thanks.

crazyscot:
My apologies, it appears that I messed up pushing the GPG signature on this. Commit updated, can we remove the stale label please?

de-nordic (approved):

d3zd3z:
> My apologies, it appears that I messed up pushing the GPG signature on this. Commit updated, can we remove the stale label please?

Oh, I think what @utzig meant to say was the the commit needs a sign-off-by footer in the commit text. There is not a requirement to sign commits in this project.

d3zd3z (approved):

PR #1626, opened 2023-02-23
imgtool: allow verify mode to use public ed25519 & x25519 keys

crazyscot:
Verify mode already works for the public halves of RSA and ECDSA keypairs. This patch corrects an apparent oversight, enabling that functionality in the Ed25519 and X25519 family.

### DIFF
--- a/scripts/imgtool/keys/ed25519.py
+++ b/scripts/imgtool/keys/ed25519.py
@@ -57,6 +57,13 @@ class Ed25519Public(KeyClass):
     def sig_len(self):
         return 64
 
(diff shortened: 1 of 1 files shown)


utzig:
@crazyscot The commit needs to be signed, also please only change `ed25519`, `x25519` is not a signing algorithm, otherwise looks OK.

crazyscot:
> @crazyscot The commit needs to be signed, also please only change `ed25519`, `x25519` is not a signing algorithm, otherwise looks OK.

Done, thanks.

utzig (approved):
LGTM. Thanks!

PR #1628, opened 2023-02-23
MCUBoot: Implement multi memory boot feature

DOAR-Infineon:
Implement multi memory boot feature

In the current implementation, the ability to launch several applications using different memory modes has been added.
Use cases: multi core and ram app boot alongside with flash based applications

- MCUBOOT_MULTI_MEMORY_LOAD 
--- Feature: partial multi-image boot for RAM based applications 
--- Feature: partial multi-image boot for Flash applications

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil.h
+++ b/boot/bootutil/include/bootutil/bootutil.h
@@ -83,7 +83,14 @@ fih_ret boot_go_for_image_id(struct boot_rsp *rsp, uint32_t image_id);
 
 struct boot_loader_state;
 void boot_state_clear(struct boot_loader_state *state);
-fih_ret context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp);
+
+fih_ret context_boot_go_flash(struct boot_loader_state *state, struct boot_rsp *rsp);
+
+#if defined(MCUBOOT_RAM_LOAD)
+fih_ret context_boot_go_ram(struct boot_loader_state *state, struct boot_rsp *rsp);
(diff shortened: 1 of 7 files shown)


boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -83,7 +83,13 @@ fih_ret boot_go_for_image_id(struct boot_rsp *rsp, uint32_t image_id);
 
 struct boot_loader_state;
 void boot_state_clear(struct boot_loader_state *state);
-fih_ret context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp);
+
+fih_int context_boot_go_flash(struct boot_loader_state *state, struct boot_rsp *rsp);
```
tamasban, line 87: `fih_int` was replaced by `fih_ret` in this change:
https://github.com/mcu-tools/mcuboot/commit/5047f032c93fd8957be6529f8f1ad7a3fe626057

tamasban (changes requested):
What is the purpose of this change, could you elaborate? I'm just guessing that the goal might is to enable a single MCUBoot image to be able to process images whether they are residing in RAM or flash.

DOAR-Infineon:
> What is the purpose of this change, could you elaborate? I'm just guessing that the goal might is to enable a single MCUBoot image to be able to process images whether they are residing in RAM or flash.

These changes were made to make MCUBoot able to work with flash and ram images independently.
The use case that we need to support, a multi image configuration:
* Application_1 boots from RAM
* Application_2 boots from FLASH
* Application_N boots from FLASH

DOAR-Infineon:
@d3zd3z @tamasban 
Please review

PR #1633, opened 2023-02-24
imgtool: sign: allow using 16-bit custom TLV types

pepe2k:
The TLV type field in trailer TLV struct is defined as unsigned 16-bit (in LE byte order). Currently, due to Python's struct format specified as `BBH`, the `imgtool` accepts only single byte for this field.

This results in error when trying to use 16-bit TLV type with option `--custom-tlv`:

> struct.error: ubyte format requires 0 <= number <= 255

This changes format to `HH` which allows using 16-bit TLV types.


scripts/imgtool/image.py
```diff
@@ -111,7 +111,7 @@ def add(self, kind, payload):
         """
         e = STRUCT_ENDIAN_DICT[self.endian]
         if isinstance(kind, int):
-            buf = struct.pack(e + 'BBH', kind, 0, len(payload))
+            buf = struct.pack(e + 'HH', kind, len(payload))
```
utzig, line 114: Makes sense. For correctness with the description in https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L100, I suggest you add code to test the value is between `0x00a0` and `0xfffe`. If it's not you can: `raise click.UsageError("Invalid TLV value")`

pepe2k: @utzig thanks for your comment and suggestion but as far as I can see, the reserved range for vendors isn't continuous, e.g. 0x00a0 - 0x00ff, 0x01a0 - 0x01ff ... Wouldn't a simple 0x00a0 ~ 0xfffe range test be misleading in that case? Or am I missing something here?

utzig: I think that's just convention; it should also work for something like 0x0100 as far as I can tell. What's more important is to not accept 0xffff and values already defined.

pepe2k: OK, I've added range validity check, updated commit message, title and rebased on latest main.

The overlapping with predefined values is checked earlier: https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/main.py#L394

utzig (approved):
LGTM

pepe2k:
Rebased on latest `main`.

PR #1635, opened 2023-02-27
boot: zephyr: Clean up before chainloading by default

### LINKED ISSUE #1634: Zephyr: Chain loaded application has possibility of not working and faulting
There is an issue with the default mcuboot configuration when chainloading other (zephyr generally but any application is impacted) applications which is that by default, mcuboot does not clear the MPU configuration. When configured for an nrf52840, this enables a MPU stack guard by default which it places part the way through ram at 0x20002540 for 64 bytes, this memory is then made read only by the MPU. When the application starts, if it touches this ram, the module will fault - it is not known if an application will touch memory at this address or not but regardless the MPU _must_ be cleared because leaving read only regions leads to undefined behaviour.

nordicjm:
This changes the default mcuboot configuration for zephyr applications to clean up before booting the user application. The reason for this change is that mcuboot may configure protection (e.g. MPU stack guard) which is then used by the user application during its startup code prior to cleaning the configuration up, this can lead to a unbootable application and potentially irrecoverable module, therefore cleaning up is now being enabled by default.

Fixes #1634

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -146,20 +146,18 @@ config BOOT_SIGNATURE_KEY_FILE
 config MCUBOOT_CLEANUP_ARM_CORE
 	bool "Perform core cleanup before chain-load the application"
 	depends on CPU_CORTEX_M
-	default y if !ARCH_SUPPORTS_ARCH_HW_INIT
+	default y
 	help
 	  This option instructs MCUboot to perform a clean-up of a set of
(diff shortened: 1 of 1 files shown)


nordicjm:
@de-nordic @d3zd3z @utzig This is an urgent fix, can you please review.

de-nordic (approved):

carlescufi (approved):

PR #1636, opened 2023-02-28
imgtool: fix boot_magic when -e big and max-align > 8

RaphaelDupont:
Currently if max-align > 8, magic_boot only works in little endian

For example :
With max-align = 16 and endian = big, boot_magic starts with 0x10 0x00, but it should be 0x00 0x10


utzig (approved):
LGTM

utzig:
@RaphaelDupont There is a mismatch in the commit sign, please fix.

RaphaelDupont:
> @RaphaelDupont There is a mismatch in the commit sign, please fix.

Done

utzig:
@RaphaelDupont Please rebase on latest `main`.

scripts/imgtool/image.py
```diff
@@ -166,11 +166,18 @@ def __init__(self, version=None, header_size=IMAGE_HEADER_SIZE,
...
+                    align_msb, align_lsb, 0x2d, 0xe1,
+                    0x5d, 0x29, 0x41, 0x0b,
+                    0x8d, 0x77, 0x67, 0x9c,
+                    0x11, 0x0f, 0x1f, 0x8a, ])
+            else :
+                self.boot_magic = bytes([
+                    align_lsb, align_msb, 0x2d, 0xe1,
+                    0x5d, 0x29, 0x41, 0x0b,
+                    0x8d, 0x77, 0x67, 0x9c,
+                    0x11, 0x0f, 0x1f, 0x8a, ])
```
utzig, line 180: It's OK as is, but to avoid the copy pasta in the magic you could also have done something like:

```python
lsb = self.max_align & 0x00ff
msb = (self.max_align & 0xff00) >> 8
align = bytes([msb, lsb]) if self.endian == "big" else bytes([lsb, msb])
self.boot_magic = align + bytes([0x2d, 0xe1,
                                 0x5d, 0x29, 0x41, 0x0b,
                                 0x8d, 0x77, 0x67, 0x9c,
                                 0x11, 0x0f, 0x1f, 0x8a, ])
```

RaphaelDupont: Thank you for the feedback, I changed the code according to your suggestion

PR #1638, opened 2023-02-28
boot: zephyr: Remove deprecated GPIO Kconfig entries

nordicjm:
Removes the old deprecated Kconfig items for configuring GPIO ports and pins for entering serial recovery/USB DFU mode. These were deprecated almost a year ago and should not be used as device tree has long replaced them.


nordicjm:
@de-nordic

de-nordic (approved):

PR #1641, opened 2023-03-07
P224 curve removal

Roolli:
There has been an ongoing discussion regarding p224 curve removal. There's no imgtool support for it currently, and the likelihood of it being used without it is very low. This pull request removes the p224 curve from MCUBoot and all the references that are tied to it.


boot/bootutil/include/bootutil/image.h
```diff
@@ -85,7 +85,6 @@ struct flash_area;
 #define IMAGE_TLV_PUBKEY            0x02   /* public key */
 #define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
 #define IMAGE_TLV_RSA2048_PSS       0x20   /* RSA2048 of hash output */
-#define IMAGE_TLV_ECDSA224          0x21   /* ECDSA of hash output */
```
utzig, line 88: I think it's better to leave the TLV definition here, so it's not reused, just change the comment to state it's not supported anymore.

docs/design.md
```diff
@@ -107,7 +107,6 @@ struct image_tlv {
 #define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
 #define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
 #define IMAGE_TLV_RSA2048_PSS       0x20   /* RSA2048 of hash output */
-#define IMAGE_TLV_ECDSA224          0x21   /* ECDSA of hash output */
```
utzig, line 110: Don't remove, just comment it's not supported anymore.

boot/bootutil/include/bootutil/image.h
davidvincze: Hi Fabio,
I have no problem with marking it as not supported, I just would like to understand the reasoning behind it.
I don't see how reusing the value 0x21 for a different TLV type would cause any problems.
Thanks.

utzig: Why would you break compatibility? `newt` for example is able to create images using p224. If someone is actually using that curve, why would I parse this TLV as some other type of data instead of just failing? How can you know who else might be using other tooling with MCUboot without publishing it to the world?

davidvincze: Thanks for the answer, it gave me a new perspective.
I understand, but there won't be any compatibility anyway - if we remove the P224 support.
However, I completely agree with this: it is much better to fail during TLV parsing with a missing TLV error than fail because of its unexpected content.

docs/design.md
Roolli: I've put it back, and marked it with not supported anymore.

boot/bootutil/include/bootutil/image.h
Roolli: I've put it back, and marked it with not supported anymore.

utzig (approved):

davidvincze:
> @davidvincze Don't wait forever!

It is a non-blocking PR so I wanted to wait a week for feedback (nothing so far, no DM replies), so I think we merge it. 👍

davidvincze:
@Roolli please add a release note snippet for this change.

PR #1642, opened 2023-03-10
zephyr: serial_recovery: Use Zephyr manifest zcbor files

nordicjm:
Switches from using the zcbor files in-tree to using the ones that are part of the zephyr manifest, this prevents using old and potentially buggy versions of the zcbor library.


utzig (approved):
LGTM

de-nordic (approved):

PR #1644, opened 2023-03-10
boot: serial_recovery: Add image hash support

### LINKED ISSUE #1581: zephyr: serial recovery: Missing required hash field in state of images response
On https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_1.html it states that for the "get image state" response, hash is a required field, however in mcuboot with serial recovery, this field is in fact not sent, therefore either the field *should* be sent or the documentation needs to be updated to specify that it is not a required field in the response.

nordicjm:
Adds support for outputting the image hash TLV in serial recovery mode, which is needed to comply with the img_mgmt MCUmgr group requirements.

Fixes #1581

Test file upload and listing from mcuboot:
```
Images:
 image=0 slot=0
    version: 0.0.0.0
    bootable: false
    flags: 
    hash: 73fc86f2aee87d13620ca718b851e524a01cd6d20aede1672ed3e31457b35d24
Split status: N/A (0)
```
And checking from actual booted image:
```
Images:
 image=0 slot=0
    version: 0.0.0
    bootable: true
    flags: active confirmed
    hash: 73fc86f2aee87d13620ca718b851e524a01cd6d20aede1672ed3e31457b35d24
Split status: N/A (0)
```

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -82,7 +82,7 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #define MCUBOOT_SERIAL_MAX_RECEIVE_SIZE 512
 #endif
 
-#define BOOT_SERIAL_OUT_MAX     (128 * BOOT_IMAGE_NUMBER)
+#define BOOT_SERIAL_OUT_MAX     (160 * BOOT_IMAGE_NUMBER)
 #define BOOT_SERIAL_FRAME_MTU   124 /* 127 - pkt start (2 bytes) and stop (1 byte) */
 
 #ifdef __ZEPHYR__
@@ -120,6 +120,11 @@ static bool bs_entry;
 static char bs_obuf[BOOT_SERIAL_OUT_MAX];
 
 static void boot_serial_output(void);
+
+#ifdef MCUBOOT_SERIAL_IMG_GRP_HASH
+static int boot_serial_get_hash(const struct image_header *hdr,
+                                const struct flash_area *fap, uint8_t *hash);
+#endif
 
 static zcbor_state_t cbor_state[2];
 
@@ -217,6 +222,9 @@ bs_list(char *buf, int len)
     uint32_t slot, area_id;
     const struct flash_area *fap;
     uint8_t image_index;
+#ifdef MCUBOOT_SERIAL_IMG_GRP_HASH
+    uint8_t hash[32];
+#endif
 
     zcbor_map_start_encode(cbor_state, 1);
     zcbor_tstr_put_lit_cast(cbor_state, "images");
@@ -261,6 +269,11 @@ bs_list(char *buf, int len)
                 }
             }
 
+#ifdef MCUBOOT_SERIAL_IMG_GRP_HASH
+            /* Retrieve SHA256 hash of image for identification */
+            rc = boot_serial_get_hash(&hdr, fap, hash);
+#endif
+
             flash_area_close(fap);
 
             if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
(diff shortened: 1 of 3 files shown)


nordicjm:
@de-nordic

boot/boot_serial/src/boot_serial.c
```diff
@@ -82,7 +82,7 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #define MCUBOOT_SERIAL_MAX_RECEIVE_SIZE 512
 #endif
 
-#define BOOT_SERIAL_OUT_MAX     (128 * BOOT_IMAGE_NUMBER)
+#define BOOT_SERIAL_OUT_MAX     (168 * BOOT_IMAGE_NUMBER)
```
pepe2k, line 85: > Mcumgr packets sent over serial are fragmented into frames of 127 bytes or fewer. This 127-byte maximum applies to the entire frame, including header, CRC, and terminating newline.

https://github.com/apache/mynewt-mcumgr/blob/master/transport/smp-console.md#overview

Fix available in https://github.com/mcu-tools/mcuboot/pull/1507

boot/boot_serial/src/boot_serial.c
```diff
@@ -260,6 +265,9 @@ bs_list(char *buf, int len)
                 }
             }
 
+            /* Retrieve SHA256 hash of image for identification */
+            rc = boot_serial_get_hash(&hdr, fap, hash);
```
utzig, line 269: It would be better to check `rc` and only append the CBOR data if it is 0.

de-nordic:
> Adds support for outputting the image hash TLV in serial recovery mode, which is needed to comply with the img_mgmt MCUmgr group requirements.
> 

We should start relaxing SMP requirements for MCUboot or have some alternative simplified version, because although this completes SMP protocol requirement for image status in reality it does not add that much of a value, as this is recovery protocol and should be rarely used, at the cost of flash.

nordicjm:
> We should start relaxing SMP requirements for MCUboot or have some alternative simplified version, because although this completes SMP protocol requirement for image status in reality it does not add that much of a value, as this is recovery protocol and should be rarely used, at the cost of flash.

I did check the flash before and after and the size increase was minimal (will recheck and get concrete numbers). This is needed where updatable images numbers is > 1 but seems like a useful enough feature even when it is 1 to see if the image was uploaded properly

nordicjm:
`nrf52840dongle_nrf52840` with:
```
           FLASH:       57188 B        60 KB     93.08%
             RAM:       36928 B       256 KB     14.09%
```
without:
```
           FLASH:       57072 B        60 KB     92.89%
             RAM:         36 KB       256 KB     14.06%
```
116 byte flash increase
64 byte RAM increase

utzig:
I don't think it's a problem adding the hash, I just think it's useless in this scenario, which is probably why it was not added originally. I think it's even debatable if `list` should be implemented at all, In my opinion, and I might be missing something here, only `upload` with a pre-erase would be required for recovery. That said I am fine with the change.

nordicjm:
It's probably not useful for serial recovery for the main core, but serial recovery can be used for more than that, e.g. storing images for other cores or other CPUs, in those instances it's useful as you can then query which images are presently loaded on a device, especially if there is a requirement for e.g. 2 cores to have a set of link images because of incompatibilities with different versions and there is a possibility that power can be lost whilst one of the updates is being transferred.

de-nordic:
@nordicjm Can we have some MCUBOOT_CONFIG option turning this off? 100 bytes in flash may not be much, but we struggle to fit other stuff and things keep to add up.
I do not think RAM in MCUboot is that important, as we have the whole device until app boots.

nordicjm:
> @nordicjm Can we have some MCUBOOT_CONFIG option turning this off? 100 bytes in flash may not be much, but we struggle to fit other stuff and things keep to add up.

Added

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -205,6 +205,10 @@
 #define MCUBOOT_SERIAL_WAIT_FOR_DFU
 #endif
 
+#ifdef CONFIG_BOOT_SERIAL_IMG_GRP_HASH
+#define BOOT_SERIAL_IMG_GRP_HASH
```
de-nordic, line 209: Use the same prefixes as the entire mcuboot_config.h

nordicjm: Fixed

de-nordic (approved):

utzig (approved):

PR #1645, opened 2023-03-10
bootutil: Fix erase of trailer when located in scratch area

jamike:
Trailer is temporary located in scratch when data of image goes up to last sector of primary slot. At the end of the swap procedure, the temporary trailer in scratch must be erased to prevent mcuboot from reading a stale status from the scratch area. The erase tentative was not correct, this is fixed.Issue is easily reproduced with an image size of one sector.

### DIFF
--- a/boot/bootutil/src/swap_scratch.c
+++ b/boot/bootutil/src/swap_scratch.c
@@ -673,7 +673,7 @@ boot_swap_sectors(int idx, uint32_t sz, struct boot_loader_state *state,
         BOOT_STATUS_ASSERT(rc == 0);
 
         if (erase_scratch) {
-            rc = boot_erase_region(fap_scratch, 0, sz);
+            rc = boot_erase_region(fap_scratch, 0, flash_area_get_size(fap_scratch));
             assert(rc == 0);
         }
     }


utzig:
What issue do you see without the fix?

Stephane-LeRoy:
Without fix, the trailer in scratch area remains present after swap completion, in case of small image  (smaller than scratch area).
Then at next boot, mcuboot detect the trailer in scratch area and then try to resume a swap operation, whereas it is already completed.

Refer to the comment already present in the code few lines before the modified code line => trailer has to be erased.
        /* If we wrote a trailer to the scratch area, erase it after we persist
         * a trailer to the primary slot.  We do this to prevent mcuboot from
         * reading a stale status from the scratch area in case of immediate
         * reset.
         */

utzig:
> Without fix, the trailer in scratch area remains present after swap completion, in case of small image (smaller than scratch area).

Interesting, I have tested this situation a long time ago on a Nucleo-F746ZG, which due to limitations in the flash size and layout, only supported slots and scratch area each with 256KB, and it worked, at least at the time.

utzig:
Just out of curiosity, what is the size of your slots and scratch?

utzig:
Also you need to properly sign off your commit.

jamike:
the scratch size is 56 kbytes : several images can be updated with different size (i.e some are  large 1Mbytes, and other are small  32kbytes only)

utzig (approved):
LGTM

PR #1650, opened 2023-03-16
boot_serial: Fix Mynewt tests build

sjanc:
__packed is Zephyr specific macro and it may not be available in all environments. Use more generic __attribute__((packed)) instead.

### DIFF
--- a/boot/boot_serial/src/boot_serial_priv.h
+++ b/boot/boot_serial/src/boot_serial_priv.h
@@ -54,6 +54,10 @@ extern "C" {
 #define NMGR_ID_CONS_ECHO_CTRL  1
 #define NMGR_ID_RESET           5
 
+#ifndef __packed
+#define __packed __attribute__((__packed__))
+#endif
+
 struct nmgr_hdr {
     uint8_t  nh_op;             /* NMGR_OP_XXX */
     uint8_t  nh_flags;


boot/boot_serial/src/boot_serial_priv.h
```diff
@@ -61,7 +61,7 @@ struct nmgr_hdr {
     uint16_t nh_group;          /* NMGR_GROUP_XXX */
     uint8_t  nh_seq;            /* sequence number */
     uint8_t  nh_id;             /* message ID within group */
-} __packed;
+} __attribute__((packed));
```
utzig, line 64: * Why do you think the Mynewt CI tests are passing?
* Is this compatible with every compiler Zephyr supports?
* Why not add a definition some lines above, like this:

```c
#ifndef __packed
#define __packed __attribute__((__packed__))
#endif
```

sjanc: I run them:)   but I've also updated CI to run mynewt tests (see https://github.com/mcu-tools/mcuboot/pull/1651)

utzig: #1651 Mynewt tests failed with: `multiple definition of '__packed'`.  Weird for a Zephyr specific macro, no?

utzig: Well, do you know why just building a target works, and building a test does not?

sjanc: since __packed is not defined it is treated as a variable name, so binary ends up with multiple __packed global symbols

sjanc: I think mynewt build works just because boot_serial_priv.h is included only in boot_serial.c, while in tests this is included also in boot_test.c

so this is a symptom of rather unexpected behavior if __packed is not a macro for __attribute__

utzig: That's reasonable. But I still find it better to just define it as I mentioned above.

sjanc: OK, I've updated PR with your suggestion. For now I just put this in boot_serial_priv.h but I think in longer term this should be in some common header..

nordicjm (approved):

utzig (approved):
LGTM

PR #1651, opened 2023-03-16
ci: mynewt: Sync with upstream Github Actions

sjanc:
Build Mynewt targets in similar way as Mynewt GH Actions. This also enables running Mynewt specific tests.

### DIFF
--- a/ci/mynewt_install.sh
+++ b/ci/mynewt_install.sh
@@ -57,9 +57,15 @@ arm_toolchain_install() {
     done
 }
 
+native_test_setup() {
+    sudo apt-get update
+    sudo apt-get install -y gcc-multilib
+}
+
 mkdir -p $HOME/bin
 export PATH=$HOME/bin:$PATH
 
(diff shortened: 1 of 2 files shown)


utzig:
Test reason those `xxx_install.sh` and `xxx_run.sh` scripts exist is that we used to run tests on Travis as well, so GH workflows and Travis could share the install/run steps. Who knows if in the future there won't be need to migrate to something else. I think you could just as easily have updated them.

sjanc:
well, I was looking at zephyr integration but I'm fine with just updating those scripts, will update PR tomorrow

utzig:
> well, I was looking at zephyr integration but I'm fine with just updating those scripts, will update PR tomorrow

Yep, I didn't remember those, but Zephyr has unlimited resources and can change everything tomorrow if needed.

utzig:
Rebase!

utzig (approved):
LGTM

PR #1660, opened 2023-03-24
Add support for a PSA Crypto APIs backend for RSA and SHA

adeaarm:
With this PR, we want to add optional support for a new backend to implement cryptographic APIs based on the [ PSA Crypto APIs](https://armmbed.github.io/mbed-crypto/html/). We start by adding support for it in the RSA encryption/signing APIs and in the SHA256 hashing APIs. A follow up PR enabling this new backend with the existing simulator tests will be opened later. This PR covers only:

* The cryptographic changes by adding a generic backend option called `MCUBOOT_USE_PSA_CRYPTO` for RSA and SHA256
* Refactors the RSA module which in the current implementation is strongly coupled with mbed TLS APIs, separating it similarly to other modules in a generic abstraction layer (rsa.h) that allows interfacing with arbitrary crypto libraries that follow the interface.

An additional small change to common prints in `bootutil` to show the `image_index` is added as well as, although unrelated from the rest of the patches, we found it was helping in our debugging efforts throughout development., so it probably makes no harm to have it here.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/common.h
+++ b/boot/bootutil/include/bootutil/crypto/common.h
@@ -7,9 +7,10 @@
 #ifndef __BOOTUTIL_CRYPTO_COMMON_H__
 #define __BOOTUTIL_CRYPTO_COMMON_H__
 
-/* TODO May need to update this in a future 3.x version of Mbed TLS.
- * Extract a member of the mbedtls context structure.
+/* The check below can be performed even for those cases
+ * where MCUBOOT_USE_MBED_TLS has not been defined
  */
+#include "mbedtls/version.h"
 #if MBEDTLS_VERSION_NUMBER >= 0x03000000
 #define MBEDTLS_CONTEXT_MEMBER(X) MBEDTLS_PRIVATE(X)
 #else
--- a/boot/bootutil/include/bootutil/crypto/rsa.h
+++ b/boot/bootutil/include/bootutil/crypto/rsa.h
@@ -1,0 +1,356 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Copyright (c) 2023 Arm Limited
+ */
+
+/*
+ * This module provides a thin abstraction over some of the crypto
+ * primitives to make it easier to swap out the used crypto library.
+ *
+ * At this point, the choices are: MCUBOOT_USE_MBED_TLS and
+ * MCUBOOT_USE_PSA_CRYPTO. Note that support for  MCUBOOT_USE_PSA_CRYPTO is
+ * still experimental and it might not support all the crypto abstractions
+ * that MCUBOOT_USE_MBED_TLS supports. For this reason, it's allowed to have
+ * both of them defined, and for crypto modules that support both abstractions,
+ * the MCUBOOT_USE_PSA_CRYPTO will take precedence.
+ */
+
+/*
+ * Note: The source file that includes this header should either define one of the
+ * two options BOOTUTIL_CRYPTO_RSA_CRYPT_ENABLED or BOOTUTIL_CRYPTO_RSA_SIGN_ENABLED
+ * This will make the signature functions or encryption functions visible without
+ * generating a "defined but not used" compiler warning
+ */
+
+#ifndef __BOOTUTIL_CRYPTO_RSA_H_
+#define __BOOTUTIL_CRYPTO_RSA_H_
+
+#include "mcuboot_config/mcuboot_config.h"
+
+#if defined(MCUBOOT_USE_PSA_CRYPTO) || defined(MCUBOOT_USE_MBED_TLS)
+#define MCUBOOT_USE_PSA_OR_MBED_TLS
+#endif /* MCUBOOT_USE_PSA_CRYPTO || MCUBOOT_USE_MBED_TLS */
+
+#if (defined(MCUBOOT_USE_PSA_OR_MBED_TLS)) != 1
+    #error "One crypto backend must be defined: either MBED_TLS/PSA_CRYPTO"
+#endif
+
+#if defined(MCUBOOT_USE_PSA_CRYPTO)
+
+#include <psa/crypto.h>
+#include "bootutil/enc_key_public.h"
+
+#elif defined(MCUBOOT_USE_MBED_TLS)
+
+#include "mbedtls/rsa.h"
+#include "mbedtls/version.h"
+#if defined(BOOTUTIL_CRYPTO_RSA_CRYPT_ENABLED)
+#if MBEDTLS_VERSION_NUMBER >= 0x03000000
+#include "rsa_alt_helpers.h"
+#else
+#include "mbedtls/rsa_internal.h"
+#endif
+#endif /* BOOTUTIL_CRYPTO_RSA_CRYPT_ENABLED */
+#include "mbedtls/asn1.h"
+#include "bootutil/crypto/common.h"
+
+#endif /* MCUBOOT_USE_MBED_TLS */
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if defined(MCUBOOT_USE_PSA_CRYPTO)
+
+typedef struct {
+    psa_key_id_t key_id;
+} bootutil_rsa_context;
+
+static inline void bootutil_rsa_init(bootutil_rsa_context *ctx)
+{
+    ctx->key_id = PSA_KEY_ID_NULL;
+}
+
+static inline void bootutil_rsa_drop(bootutil_rsa_context *ctx)
+{
+    if (ctx->key_id != PSA_KEY_ID_NULL) {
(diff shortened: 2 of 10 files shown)


davidvincze (changes requested):
Minor comments, otherwise it looks good to me.
Error handling for the bootutul_sha256_x functions are generally missing everywhere. In some cases the return value of the crypto backend is passed on (and then ingnored), in other cases a hardcoded return value (success) is used. 
This made me think if it would make sense to change the return type to `void` for each hash functions or to deploy proper error handling in the boot logic for these crypto operations. In most cases this error is non recoverable.. the device may try a reset by software/watchdog.

boot/bootutil/include/bootutil/crypto/rsa.h
```diff
@@ -0,0 +1,356 @@
...
+    bootutil_rsa_context *ctx,
+    size_t *olen,
+    const uint8_t *input,
+    uint8_t *output,
+    size_t output_max_len)
+{
+    psa_status_t status = PSA_ERROR_INVALID_ARGUMENT;
+
+    /* Perform an additional defensive check to compare the modulus of the RSA
+     * key to the expected input to the decryption function, i.e. TLV_ENC_RSA_SZ
```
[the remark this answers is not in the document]
adeaarm: Done

boot/bootutil/src/image_rsa.c
```diff
@@ -30,20 +30,24 @@
...
-#include "mbedtls/rsa.h"
-#include "mbedtls/asn1.h"
-#include "mbedtls/version.h"
+#define BOOTUTIL_CRYPTO_RSA_SIGN_ENABLED
+#include "bootutil/crypto/rsa.h"
 
-#include "bootutil_priv.h"
-#include "bootutil/fault_injection_hardening.h"
+/* PSA Crypto APIs provide an integrated API to perform the verification
+ * while for other crypto backend we need to implement each step at this
```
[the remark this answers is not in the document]
adeaarm: Done

boot/bootutil/src/loader.c
```diff
@@ -2813,15 +2813,15 @@ boot_load_image_to_sram(struct boot_loader_state *state)
 
         rc = boot_verify_ram_load_address(state);
         if (rc != 0) {
-            BOOT_LOG_INF("Image RAM load address 0x%x is invalid.", img_dst);
```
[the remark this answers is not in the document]
adeaarm: Done

adeaarm:
> Minor comments, otherwise it looks good to me.
> 
> Error handling for the bootutul_sha256_x functions are generally missing everywhere. In some cases the return value of the crypto backend is passed on (and then ingnored), in other cases a hardcoded return value (success) is used. This made me think if it would make sense to change the return type to `void` for each hash functions or to deploy proper error handling in the boot logic for these crypto operations. In most cases this error is non recoverable.. the device may try a reset by software/watchdog.

I have assessed the comments. The point about error handling for SHA256 is valid. I have modified the crypto abstraction functions to always return the state of the operation (mostly, it would affect PSA Crypto APIs as other APIs tend to just not return anything in init or drop. At the same time, I believe that building a proper error handling on the side of the caller is out of scope for this PR. But nevertheless, with these additions, it shold be possible to have a dedicated PR to rework the error handling without touching the crypto abstractions.

boot/bootutil/src/loader.c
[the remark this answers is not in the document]
davidvincze: Thank you.

davidvincze:
> > Minor comments, otherwise it looks good to me.
> > Error handling for the bootutul_sha256_x functions are generally missing everywhere. In some cases the return value of the crypto backend is passed on (and then ingnored), in other cases a hardcoded return value (success) is used. This made me think if it would make sense to change the return type to `void` for each hash functions or to deploy proper error handling in the boot logic for these crypto operations. In most cases this error is non recoverable.. the device may try a reset by software/watchdog.
> 
> I have assessed the comments. The point about error handling for SHA256 is valid. I have modified the crypto abstraction functions to always return the state of the operation (mostly, it would affect PSA Crypto APIs as other APIs tend to just not return anything in init or drop. At the same time, I believe that building a proper error handling on the side of the caller is out of scope for this PR. But nevertheless, with these additions, it shold be possible to have a dedicated PR to rework the error handling without touching the crypto abstractions.

Thank you Antonio for the additional modifications. Strictly speaking, it was also out of the scope of this PR, but I felt this was a good opportunity for that. I haven't found anything else, thank you for the quality work.

davidvincze (approved):

boot/bootutil/include/bootutil/crypto/rsa.h
```diff
@@ -0,0 +1,356 @@
+/*
```
d3zd3z, line 1: This entire file has DOS line endings. Please fix the commits that add it rather than a new commit.

boot/bootutil/include/bootutil/crypto/common.h
```diff
@@ -1,19 +1,20 @@
 /*
  * SPDX-License-Identifier: Apache-2.0
  *
- * Copyright (c) 2021 Arm Limited
```
d3zd3z, line 4: This file also has DOS line endings, which we should probably fix. Perhaps add a single commit at the front of this sequence that fixes the line endings in this file.

boot/bootutil/include/bootutil/crypto/rsa.h
```diff
@@ -0,0 +1,356 @@
...
+
+static inline void bootutil_rsa_drop(bootutil_rsa_context *ctx)
+{
+    if (ctx->key_id != PSA_KEY_ID_NULL) {
+        (void)psa_destroy_key(ctx->key_id);
+    }
+}
+
+#if defined(BOOTUTIL_CRYPTO_RSA_CRYPT_ENABLED)
+static inline int bootutil_rsa_oaep_decrypt(
```
d3zd3z, line 85: Why is this a static inline? This is a lot of code, and we really shouldn't be pulling it into multiple parts of the image. Realistically, most of these should probably just be implemented in a C file, with a header that points to an implementation. Or if there are multiple implementations, there could be a conditional static that just calls the right one.

Static inline can often result in multiple copies of the code. If the code is large, like this one is, it will sometimes even result in the function getting expanded multiple times.

For the most part, inline isn't going to be very helpful. In mcuboot, codesize is much more important than speed, especially for a function that takes hundreds of ms like this one does.

d3zd3z (changes requested):
Some initial comments. I'll dive a little more into the RSA interface once that is less staticy. Overall, the idea is good, though.

d3zd3z:
I've created a change #1686 to fix the dos line endings in `common.h`. We should probably let that merge, and rebase this (fixing the resulting conflict) so that we don't have to worry about more conflicts later.

boot/bootutil/include/bootutil/crypto/rsa.h
adeaarm: I agree, I think inline just slipped in the definition without much thought. I tried to mimic what already happens for other modules that are abstracted into the crypto/<headers.h>, but I agree in general that this should probably evolve into a set of C files (one per abstraction) to avoid static implementation of functions in header files (i.e. reduce possibility of having code duplication). I will go through the prototypes of the functions in rsa.h and remove inline when the code associated to it is _too_ big.

adeaarm:
> I've created a change #1686 to fix the dos line endings in `common.h`. We should probably let that merge, and rebase this (fixing the resulting conflict) so that we don't have to worry about more conflicts later.

I have updated this PR to fix the comments so far (remove superfluous inline, fix CRLF for rsa.h. I have also fixed "common.h" for the time being, so by the time #1686 gets merged this will rebase on top of it without conflicts

boot/bootutil/include/bootutil/crypto/rsa.h
davidvincze: A related open issue: [Crypto related source restructuring #1665](https://github.com/mcu-tools/mcuboot/issues/1665)

davidvincze:
> > I've created a change #1686 to fix the dos line endings in `common.h`. We should probably let that merge, and rebase this (fixing the resulting conflict) so that we don't have to worry about more conflicts later.
> 
> I have updated this PR to fix the comments so far (remove superfluous inline, fix CRLF for rsa.h. I have also fixed "common.h" for the time being, so by the time #1686 gets merged this will rebase on top of it without conflicts

#1686 has been merged.

adeaarm:
Hi @d3zd3z , any more comments on this?

davidvincze:
No new feedback, after raising the attention of the general MCUboot community on Slack. Is it okay if I merge this PR to move forward, and face with consequences later?

d3zd3z (approved):
I'm ok with the state of this change. Regarding error status, it is probably good to support that, as it is becoming common to include in cryptographic libraries. The SHA operations can't fail, but there are possibly operations that can.

PR #1663, opened 2023-03-28
espressif: Remove uncommon baudrate from flash command

mkettn:
esptool.py will select a sane default anyway.


almir-okato:
@mkettn please check your commit message style and also sign it (`git commit -s`).

Besides that, I would suggest you to add a parameter for the baud rate, so if desirable, one can define it before hand, similar as the `FLASH_PORT` param.

almir-okato:
Hi @mkettn, my last PR https://github.com/mcu-tools/mcuboot/pull/1756 also changed the flash parameters, so I think we can close this.

mkettn:
Great, that its fixed, Thank you very much!

PR #1666, opened 2023-04-03
Add a generic signature validation module for ECDSA

adeaarm:
This PR is a follow up on PR #1660 where we started to introduce crypto backend support for PSA Crypto APIs. In this PR we introduce an abstraction layer that support generic ECDSA signatures, based on PSA Crypto APIs in `crypto/ecdsa.h`. This abstraction layer is curve-agile in the sense that it can support multiple curve-sizes and operates accordingly by parsing the corresponding public key, unlike the current support for P224 and P256 curves which is tied specifically at build time for those particular curves. This abstraction will then be used in a following PR to support increased-size signature verification schemes based on the P384 curve, a curve size which is recommended by new specification such as the [Arm CCA](https://developer.arm.com/documentation/DEN0096/latest)


adeaarm:
> My concern is that this code would co-exist in parallel with the current ECDSA signature verification module. I think it's a preferable direction to create a generic (curve-agile) ECDSA abstraction layer, but I think it should be merged with the current ECDSA code in boot/bootutil/src/image_ec256.c.
> 
> And despite the fact that the PSA Crypto API provides a generic ECDSA interface, I believe there is no reason to introduce a general `MCUBOOT_SIGN_ECDSA` config option. MCUboot supports 1 type of signature at a time. There is a general `MCUBOOT_SIGN_RSA` option too, but it is supplemented by a `MCUBOOT_SIGN_RSA_LEN` option. The `MCUBOOT_SIGN_EC256` and a new `MCUBOOT_SIGN_EC384` config option (capability) would tell the build system and simulator which key file to select.

Just to confirm if my understanding is correct, here you would prefer to not have `image_ec256.c` and `image_ecdsa.c` both existing (and governed by `MCUBOOT_SIGN_EC256` and `MCUBOOT_SIGN_ECDSA`, but just have a single ECDSA verification module (let's call it `image_ecdsa.c`, which would be enabled when `MCUBOOT_SIGN_EC256 || MCUBOOT_SIGN_EC384`, and in that module, have this new ECDSA crypto abstraction called. The new abstraction must be called (i.e. drop the _p256_ from the function names in the existing ecdsa_p256.h abstraction), and hence have a single crypto abstraction for ECDSA, let's call it again ecdsa.h, but that allows for the PSA crypto based one and the existing ones, all as part of the same abstraction (i.e. under the same function names, essentially, governed by proper ifdefs). If this is what you mean, I think it makes sense and it's doable. Please let me know in case I misunderstood something, before I start the reworking

davidvincze:
>Just to confirm if my understanding is correct, here you would prefer to not have image_ec256.c and image_ecdsa.c both existing (and governed by MCUBOOT_SIGN_EC256 and MCUBOOT_SIGN_ECDSA, but just have a single ECDSA verification module (let's call it image_ecdsa.c, which would be enabled when MCUBOOT_SIGN_EC256 || MCUBOOT_SIGN_EC384, and in that module, have this new ECDSA crypto abstraction called. 

Yes, that's my suggestion.

>  The new abstraction must be called (i.e. drop the p256 from the function names in the existing ecdsa_p256.h abstraction), and hence have a single crypto abstraction for ECDSA, let's call it again ecdsa.h, but that allows for the PSA crypto based one and the existing ones, all as part of the same abstraction (i.e. under the same function names, essentially, governed by proper ifdefs). If this is what you mean, I think it makes sense and it's doable. Please let me know in case I misunderstood something, before I start the reworking

Exactly. :)

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -0,0 +1,585 @@
...
+    /* Indicate to the caller that the verify function needs the raw ASN.1
+     * signature, not a decoded one. */
+    #define MCUBOOT_ECDSA_NEED_ASN1_SIG
+#endif /* MCUBOOT_USE_MBED_TLS */
+
+/*TODO: remove this after cypress port mbedtls to abstract crypto api */
+#if defined(MCUBOOT_USE_CC310) || defined(MCUBOOT_USE_MBED_TLS)
+#define NUM_ECC_BYTES (256 / 8)
+#endif
+
```
davidvincze, line 66: Please include <stdint.h>

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -0,0 +1,585 @@
...
+static inline void bootutil_ecdsa_drop(bootutil_ecdsa_context *ctx)
+{
+    (void)ctx;
+}
+
+static inline int bootutil_ecdsa_verify(bootutil_ecdsa_context *ctx,
+                                             uint8_t *pk, size_t pk_len,
+                                             uint8_t *hash, size_t hash_len,
+                                             uint8_t *sig, size_t sig_len)
```
davidvincze, line 195: nit:
The parameter lists got misaligned when renaming the ecdsa functions.
Could you please fix it?

davidvincze (commented):
I seems okay to me, a few nits and I'll have a look at the documentation later and upstream a
patch to update it with the new SIGN_EC384 option/capability if necessary.
Thank you for the update, that's exactly how I imagined this contribution. 👍

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -0,0 +1,585 @@
...
+    psa_set_key_type(&key_attributes, PSA_KEY_TYPE_ECC_PUBLIC_KEY(PSA_ECC_FAMILY_SECP_R1));
+
+    return (int)psa_import_key(&key_attributes, *cp, key_size, &ctx->key_id);
+}
+
+/* Verify the signature against the provided hash. The signature gets parsed from
+ * the encoding first, then PSA Crypto has a dedicated API for ECDSA verification
+ */
+static inline int bootutil_ecdsa_verify(const bootutil_ecdsa_context *ctx,
+    uint8_t *hash, size_t hlen, uint8_t *sig, size_t slen)
```
davidvincze, line 438: The function signature is wrong. The `uint8_t *pk` and `size_t pk_len` parameters are missing.

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -0,0 +1,590 @@
...
+#if (defined(MCUBOOT_USE_TINYCRYPT) + \
+     defined(MCUBOOT_USE_CC310) + \
+     defined(MCUBOOT_USE_PSA_OR_MBED_TLS)) != 1
+    #error "One crypto backend must be defined: either CC310/TINYCRYPT/MBED_TLS/PSA_CRYPTO"
+#endif
+
+#if defined(MCUBOOT_USE_TINYCRYPT)
+    #include <tinycrypt/ecc_dsa.h>
+    #include <tinycrypt/constants.h>
+    #define BOOTUTIL_CRYPTO_ECDSA_P256_HASH_SIZE (4 * 8)
```
davidvincze, line 44: Could you move the BOOTUTIL_CRYPTO_ECDSA_P256_HASH_SIZE define
to a common macro section just like here?
https://github.com/mcu-tools/mcuboot/pull/1660/files#diff-e3bebec8c109d099a2dc6d5503ae58954a164c647d8831c5adc680fe9c045abdR37

boot/bootutil/include/bootutil/crypto/ecdsa.h
adeaarm: Done

boot/bootutil/include/bootutil/crypto/ecdsa.h
adeaarm: Done. I have asked @Roolli to test the PSA Crypto abstraction with the remaining of his patches as well (both for P256 and P384)

boot/bootutil/include/bootutil/crypto/ecdsa.h
adeaarm: Done

boot/bootutil/include/bootutil/crypto/ecdsa.h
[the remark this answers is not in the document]
adeaarm: Done.

davidvincze (approved):

PR #1668, opened 2023-04-04
Enable PSA Crypto API tests in the simulator

mdalzellarm:
In this PR, code is introduced to allow the simulator to interface with PSA Crypto functionality. When enabled with `--features psa-crypto-api`, the simulator will use all possible PSA enabled functions, and where PSA functionality is not available, will continue to use MBed TLS functions.

### DIFF
--- a/sim/mcuboot-sys/build.rs
+++ b/sim/mcuboot-sys/build.rs
@@ -10,6 +10,7 @@ use std::path::{Path, PathBuf};
 
 fn main() {
     // Feature flags.
+    let psa_crypto_api = env::var("CARGO_FEATURE_PSA_CRYPTO_API").is_ok();
     let sig_rsa = env::var("CARGO_FEATURE_SIG_RSA").is_ok();
     let sig_rsa3072 = env::var("CARGO_FEATURE_SIG_RSA3072").is_ok();
     let sig_ecdsa = env::var("CARGO_FEATURE_SIG_ECDSA").is_ok();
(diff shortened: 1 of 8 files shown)


sim/mcuboot-sys/src/c.rs
```diff
@@ -66,6 +69,8 @@ impl BootGoResult {
 pub fn boot_go(multiflash: &mut SimMultiFlash, areadesc: &AreaDesc,
                counter: Option<&mut i32>, image_index: Option<i32>,
                catch_asserts: bool) -> BootGoResult {
+    init_psa();
```
davidvincze, line 72: My first idea was that it would be better the introduce a crypto backend agnostic initialization function here and invoke the required functions based on the sim configuration (in this case psa-crypto-api). But then I had a look at the implementation of `invoke_boot_go()` and now I think that the crypto initialization step is better suited there.
Would it be possible or it needs to be here to ensure `psa_crypto_init()` and `mbedtls_test_enable_insecure_external_rng()` are called once?

sim/mcuboot-sys/build.rs
```diff
@@ -81,6 +82,64 @@ fn main() {
         panic!("mcuboot does not support more than one sig type at the same time");
     }
 
+    if psa_crypto_api {
+        if sig_ecdsa || enc_ec256 || enc_x25519 || enc_aes256_ec256 || sig_ecdsa_mbedtls || enc_aes256_x25519 || enc_kw  || enc_aes256_kw {
```
davidvincze, line 86: Please break this into multiple lines, most of the code does not exceed 80 characters.

davidvincze, line 86: I believe this condition shall be updated in the future with enabling new algorithms with the PSA API in the simulator environment.

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
```diff
@@ -0,0 +1,50 @@
...
+
+#if defined(MCUBOOT_ENCRYPT_RSA) || defined(MCUBOOT_SIGN_RSA)
+    #define MBEDTLS_PK_C
+    #define MBEDTLS_ASN1_PARSE_C
+    #define MBEDTLS_CTR_DRBG_C
+    #define MBEDTLS_CIPHER_C
+    #define MBEDTLS_ENTROPY_C
+    #define MBEDTLS_PK_PARSE_C
+    #define MBEDTLS_PK_WRITE_C
+#endif /* defined(MCUBOOT_ENCRYPT_RSA) */
```
davidvincze, line 27:
```suggestion
#endif /* MCUBOOT_ENCRYPT_RSA || MCUBOOT_SIGN_RSA */
```

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
```diff
@@ -0,0 +1,50 @@
...
+
+#if defined(MCUBOOT_ENCRYPT_EC256) || defined(MCUBOOT_ENCRYPT_X25519)
+    #define MBEDTLS_CIPHER_MODE_CTR
+    #define MBEDTLS_PLATFORM_FREE_MACRO free
+    #define MBEDTLS_PLATFORM_CALLOC_MACRO calloc
+    #define MBEDTLS_AES_C
+#endif /* defined(MCUBOOT_ENCRYPT_RSA) */
```
davidvincze, line 34:
```suggestion
#endif /* MCUBOOT_ENCRYPT_EC256 || MCUBOOT_ENCRYPT_X25519 */
```

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
```diff
@@ -0,0 +1,50 @@
...
+
+#if !defined(MCUBOOT_ENCRYPT_X25519)
+    #define MBEDTLS_PSA_BUILTIN_CIPHER 1
+#endif
```
davidvincze, line 38:
```suggestion
#endif /* MCUBOOT_ENCRYPT_X25519 */
```

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
```diff
@@ -0,0 +1,50 @@
...
+/* Enable PSA Crypto Core without support for the permanent storage
+ * Don't define MBEDTLS_PSA_CRYPTO_STORAGE_C to make sure that support
+ * for permanent keys is not enabled, as it is not usually required during boot
+ */
+#define MBEDTLS_PSA_CRYPTO_C
+#define MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
+
+#if defined(MCUBOOT_ENCRYPT_RSA) || defined(MCUBOOT_SIGN_RSA)
+    #define MBEDTLS_PK_C
+    #define MBEDTLS_ASN1_PARSE_C
```
davidvincze, line 21: This is already defined in config-rsa.h

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
```diff
@@ -0,0 +1,50 @@
...
+    #define MBEDTLS_ENTROPY_C
+    #define MBEDTLS_PK_PARSE_C
+    #define MBEDTLS_PK_WRITE_C
+#endif /* defined(MCUBOOT_ENCRYPT_RSA) */
+
+#if defined(MCUBOOT_ENCRYPT_EC256) || defined(MCUBOOT_ENCRYPT_X25519)
+    #define MBEDTLS_CIPHER_MODE_CTR
+    #define MBEDTLS_PLATFORM_FREE_MACRO free
+    #define MBEDTLS_PLATFORM_CALLOC_MACRO calloc
+    #define MBEDTLS_AES_C
```
davidvincze, line 33: `MBEDTLS_CIPHER_MODE_CTR` and `MBEDTLS_AES_C` are already defined in `config-ec.h` and `config-ed25519.h`. I believe these are duplications, but please correct me if I'm wrong.

sim/mcuboot-sys/csupport/config-rsa-kw.h
```diff
@@ -29,6 +29,11 @@
 #ifndef MCUBOOT_MBEDTLS_CONFIG_RSA_KW
 #define MCUBOOT_MBEDTLS_CONFIG_RSA_KW
 
+#if defined(MCUBOOT_USE_PSA_CRYPTO)
+#include "config-add-psa-crypto.h"
+#endif /* defined(MCUBOOT_USE_PSA_CRYPTO) */
+
```
davidvincze, line 35: Please delete the extra new line.

sim/mcuboot-sys/src/c.rs
mdalzellarm: I believe the two of them do need to be called at once. I seem to remember there being a strange behaviour which requires `psa_init()` to be called regardless of if it is used or not (which is why there was a stubbed version of the function); if it was removed from here then adding the option of psa-crypto-api was not possible.

sim/mcuboot-sys/build.rs
mdalzellarm: Is this anything which would require changing in this PR? Should I put a comment on it to signpost this, or will it just be as simple as changes being done in the future?

sim/mcuboot-sys/src/c.rs
oberon-sk: I would suggest to name this function `init_crypto` or `crypto_init` instead, because it has nothing to do with PSA here - only the implementation of the function calls `psa_crypto_init` if `feature = "psa-crypto-api"`

sim/mcuboot-sys/build.rs
davidvincze: No, I nothing needs to be done here, sorry.

sim/mcuboot-sys/csupport/config-add-psa-crypto.h
davidvincze: Thanks for updating it.

davidvincze:
@mdalzellarm may I ask you the split (if required) and squash the last commit which contains only fixes for the previous two?
After that can you rebase the patch stack? Force push is fine.

The CI tests were failing because the latest changes were pushed with an e-mail address other than matthew.dalzell@arm.com.

adeaarm:
Hi @mdalzellarm , could you sort out the rebase correctly? I can see 58 commits now and the CI is still failing

davidvincze (approved):

PR #1671, opened 2023-04-06
bootutil: Provide boot_set_next function

de-nordic:
Commit provides boot_set_next function that allows to set next application slot to boot by flash area object pointer, describing the slot.
The function also takes active which is supposed to indicate whether running application is being set for next boot and confirm parameter that allows to confirm the image.

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil_public.h
+++ b/boot/bootutil/include/bootutil/bootutil_public.h
@@ -39,6 +39,7 @@
 #define H_BOOTUTIL_PUBLIC
 
 #include <inttypes.h>
+#include <stdbool.h>
 #include <string.h>
 #include <flash_map_backend/flash_map_backend.h>
 #include <mcuboot_config/mcuboot_config.h>
@@ -265,6 +266,33 @@ boot_read_swap_state_by_id(int flash_area_id, struct boot_swap_state *state);
 int
 boot_read_swap_state(const struct flash_area *fa,
                      struct boot_swap_state *state);
+
+/**
+ * @brief Set next image application slot by flash area pointer
+ *
+ * @param fa pointer to flash_area representing image to set for next boot;
+ * @param active should be true if @fa points to currently running image
+ *        slot, false otherwise;
+ * @param confirm confirms image; when @p active is true, this is considered
+ *        true, regardless of passed value.
+ *
+ * It is users responsibility to identify whether @p fa provided as parameter
+ * is currently running/active image and provide proper value to @p active.
+ * Failing to do so may render device non-upgradeable.
+ *
+ * Note that in multi-image setup running/active application is the one
+ * that is currently being executed by any MCU core, from the pair of
+ * slots dedicated to that MCU core. As confirming application currently
+ * running on a given slot should be, preferably, done after functional
+ * tests prove application to function correctly, it may not be a good idea
+ * to cross-confirm running images.
+ * An application should only confirm slots designated to MCU core it is
+ * running on.
+ *
+ * @return 0 on success; non-zero error code on failure.
+ */
+int
+boot_set_next(const struct flash_area *fa, bool active, bool confirm);
 
 #ifdef __cplusplus
 }
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -4,7 +4,7 @@
  * Copyright (c) 2017-2019 Linaro LTD
  * Copyright (c) 2016-2019 JUUL Labs
  * Copyright (c) 2019-2021 Arm Limited
- * Copyright (c) 2020 Nordic Semiconductor ASA
+ * Copyright (c) 2020-2023 Nordic Semiconductor ASA
  *
  * Original license:
  *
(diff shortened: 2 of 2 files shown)


boot/bootutil/src/bootutil_public.c
```diff
@@ -458,6 +458,77 @@ boot_swap_type_multi(int image_index)
...
+                rc = boot_write_swap_info(fa, swap_type, 0);
+            }
+        }
+        break;
+
+    case BOOT_MAGIC_BAD:
+        if (active) {
+            rc = BOOT_EBADVECT;
+        } else {
+            /* The image slot is corrupt.  There is no way to recover, so erase the
```
nvlsianpu, line 515: corrupt -> coruptted

nordicjm: corrupt is correct

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,24 @@ int
 boot_read_swap_state(const struct flash_area *fa,
                      struct boot_swap_state *state);
 
+/**
+ * @brief Set next image application slot by flash area pointer
+ *
+ * @param fa pointer to flash_area representing image to set for next boot;
+ * @param active should be true if @fa points to active image, false if other
```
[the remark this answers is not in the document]
de-nordic: Added explanation.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,32 @@ int
...
+ *        true, regardless of passed value.
+ *
+ * It is users responsibility to identify whether @p fa provided as parameter
+ * is currently running/active image and provide proper value to @p active.
+ * Failing to do so may render device non-upgradeable.
+ *
+ * Note that in multi-image setup running/active application is the one
+ * that is currently being executed by any MCU core, from the pair of
+ * slots dedicated to that MCU core. As confirming running slot should
+ * be done, preferably,  after sanity checks done by an application running
```
nordicjm, line 286: double space after comma

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,32 @@ int
...
+ * on that core, * it may not be a good idea to cross-confirm running images.
```
nordicjm, line 287: looks like a second line appears part the way here

boot/bootutil/include/bootutil/bootutil_public.h
nordicjm, line 286: Also change from "sanity check" to something else, functional test, etc.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,32 @@ int
...
+ * An application should only confirm core it is running from or will
```
nordicjm, line 288: *running from

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,32 @@ int
...
+ * be run on it after reset.
```
nordicjm, line 289: Not sure what this final part means

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Fixed.

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Fixed.

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Reworded.

boot/bootutil/include/bootutil/bootutil_public.h
de-nordic: Clarified.

nordicjm (approved):

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -266,6 +267,33 @@ int
...
+ * is currently running/active image and provide proper value to @p active.
+ * Failing to do so may render device non-upgradeable.
+ *
+ * Note that in multi-image setup running/active application is the one
+ * that is currently being executed by any MCU core, from the pair of
+ * slots dedicated to that MCU core. As confirming application currently
+ * running on a given slot should be, preferably, done after functional
+ * tests prove application to function correctly, it may not be a good idea
+ * to cross-confirm running images.
+ * An application should only confirm slots designated to MCU core it is
```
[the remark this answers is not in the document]
de-nordic: The above not is advice rather than requirement: that why it is not written it is forbidden, rather that it may not be good idea.
Cross confirming images may not be good idea, unless images can somehow communicate their state, otherwise you may end up confirming something that does not work.
If you assume one image to be leading and other following, and leading would be the only one to confirm, this may work, but if both applications hold the same capability user will end up in complicated logic to avoid things falling apart.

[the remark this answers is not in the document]
nvlsianpu: I bet that in many cases only one executable has ability to confirm images and also there is the main application image which is able to verify whether dependent images are working correctly (centralized architecture).

nvlsianpu (approved):

boot/bootutil/include/bootutil/bootutil_public.h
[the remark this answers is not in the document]
de-nordic: Yes, there are many cases like that. In such cases running application is supposed to figure out what slot is running on other MCU it is trying to confirm and set `active` flag accordingly. Of course the primary app is the taking responsibility for confirming something that does not work.
Problem with confirming something on other MCU, is that you do not even know whether this thing still works. Because confirm happens directly to flash, the app may be actually rebooting from test, because it fails and resets, when it gets OK flag from primary application and then gets into reboot loop.

Issue #1659, opened 2023-03-24
zephyr: watchdog: setup is missing


jeppenodgaard:
If watchdog is enabled in Kconfig and watchdog alias exists in the device tree this macro is defined:
```
#define MCUBOOT_WATCHDOG_FEED()                               \
    do {                                                      \
        const struct device* wdt =                            \
            DEVICE_DT_GET(DT_ALIAS(watchdog0));               \
        if (device_is_ready(wdt)) {                           \
            wdt_feed(wdt, 0);                                 \
        }                                                     \
    } while (0)
```
Some watchdog drivers require setup to be called before the watchdog actually works. Some examples from the [zephyr](https://github.com/zephyrproject-rtos/zephyr) watchdog drivers:
- https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/watchdog/wdt_wwdg_stm32.c#L193
- https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/watchdog/wdt_ti_tps382x.c#L37
- https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/watchdog/wdt_mcux_wdog.c#L51

wdt_setup is also required by the [zephyr watchdog sample](https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/watchdog/src/main.c#L114).
It looks like wdt_setup() is missing.

**Suggestion**
Create a macro to setup the watchdog during init before feed is called:
```
#define MCUBOOT_WATCHDOG_SETUP()                              \
    do {                                                      \
        const struct device* wdt =                            \
            DEVICE_DT_GET(DT_ALIAS(watchdog0));               \
        if (device_is_ready(wdt)) {                           \
            wdt_setup(wdt, 0);                                \
        }                                                     \
    } while (0)
```
If the macro is not define in the one of the preprocessor if/else the define it emtpy:
```
#define MCUBOOT_WATCHDOG_SETUP()
```

PR #1673, opened 2023-04-11
boot: zephyr: setup watchdog

jeppenodgaard:
The Zephyr watchdog API defines a setup function. This function needs to be executed before the watchdog is functional in some cases. This commit adds MCUBOOT_WATCHDOG_SETUP when using the generic watchdog0 alias otherwise it is an empty define.

Fixes https://github.com/mcu-tools/mcuboot/issues/1659

### DIFF
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -287,20 +287,7 @@
 #error "No NRFX WDT instances enabled"
 #endif /* defined(CONFIG_NRFX_WDT0) && defined(CONFIG_NRFX_WDT1) */
 
-#elif CONFIG_IWDG_STM32 /* CONFIG_NRFX_WDT */
-#include <zephyr/device.h>
-#include <zephyr/drivers/watchdog.h>
-
-#define MCUBOOT_WATCHDOG_FEED() \
-    do {                        \
-        const struct device* wdt =                                  \
-            DEVICE_DT_GET_OR_NULL(DT_INST(0, st_stm32_watchdog));   \
-        if (device_is_ready(wdt)) {                                 \
-            wdt_feed(wdt, 0);                                       \
-        }                                                           \
-    } while (0)
-
-#elif DT_NODE_HAS_STATUS(DT_ALIAS(watchdog0), okay) /* CONFIG_IWDG_STM32 */
+#elif DT_NODE_HAS_STATUS(DT_ALIAS(watchdog0), okay) /* CONFIG_NRFX_WDT */
 #include <zephyr/device.h>
 #include <zephyr/drivers/watchdog.h>
 
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -304,6 +304,15 @@
 #include <zephyr/device.h>
 #include <zephyr/drivers/watchdog.h>
 
+#define MCUBOOT_WATCHDOG_SETUP()                              \
+    do {                                                      \
+        const struct device* wdt =                            \
+            DEVICE_DT_GET(DT_ALIAS(watchdog0));               \
+        if (device_is_ready(wdt)) {                           \
+            wdt_setup(wdt, 0);                                \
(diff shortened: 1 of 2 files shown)


jeppenodgaard:
@bbilas maybe you can have a look at this PR?

bbilas:
What about stm32 watchdog? As you have written, it also needs to execute `setup` function before.

jeppenodgaard:
> What about stm32 watchdog? As you have written, it also needs to execute `setup` function before.

I think that is the case but I do not have any hardware to test it on. I can only test if it builds.

jeppenodgaard:
> Maybe we should drop support for that completely and rely on watchdog alias and nrf custom solution only? If any stm32 user wanted to have support for wdg it would require adding a watchdog0 alias.

That makes sense to me.

Should I maybe create a build warning if the stm watchdog is enabled but the alias is not set to help people that rely on the current watchdog implementation?

bbilas:
My guess is that a few people (or perhaps nobody) use this feature since there are no issues reported for STM32 recently so I would just remove that in the follow-up commit with an explanation in the commit message why it was removed and how to use it properly.

jeppenodgaard:
@bbilas I added a commit based on our discussion.

bbilas:
Your commit message seems to be wrong

> To use the generic watchdog defines this (not tested) should be added to
the dts aliases: watchdog0 = &st_stm32_watchdog

Looking at the source code it should be

`watchdog0 = &iwdg;` but anyway - let's just remove that part. It's pretty clear how to use generic watchdog alias for all vendors

jeppenodgaard:
> Your commit message seems to be wrong
> 
> > To use the generic watchdog defines this (not tested) should be added to
> > the dts aliases: watchdog0 = &st_stm32_watchdog
> 
> Looking at the source code it should be
> 
> `watchdog0 = &iwdg;` but anyway - let's just remove that part. It's pretty clear how to use generic watchdog alias for all vendors

Ok. I removed that part from the commit message.

bbilas (approved):

de-nordic:
@jeppenodgaard  Since this modifies Zephyr port, can you make PR in Zephyr, that modifies west.yaml to pick and test the change?

jeppenodgaard:
@de-nordic I created a PR here:
https://github.com/zephyrproject-rtos/zephyr/pull/57095
There's a lot of failures but I cannot see anything wrong with the change. It should align with https://docs.zephyrproject.org/latest/develop/modules.html#process-for-submitting-changes-to-existing-modules

nordicjm:
You are using a pull link on the mirror repo of https://github.com/zephyrproject-rtos/mcuboot and not https://github.com/mcu-tools/mcuboot you will need to add a remote to the upstream for that to work - see https://github.com/zephyrproject-rtos/zephyr/pull/56355/files for an example

jeppenodgaard:
Thanks @nordicjm :+1:

jeppenodgaard:
@de-nordic it looks like [the PR](https://github.com/zephyrproject-rtos/zephyr/pull/57095) passed.

nordicjm (commented):
Enabled watchdog on nrf52840 and put breakpoint on `wdt_nrf_setup` and it never seems to be called, so I'm not sure this fixes the problem

bbilas:
> Enabled watchdog on nrf52840 and put breakpoint on `wdt_nrf_setup` and it never seems to be called, so I'm not sure this fixes the problem

The nRF platform uses a custom solution instead, please see:
https://github.com/mcu-tools/mcuboot/commit/ab03b91afd2e1fda4bf74c3690785dfbc77ad235

I have no idea why the wdog implementation is different than the other vendors. Unfortunately, the commit message does not explain it at all.

nordicjm:
Indeed I had a look and can't understand why it was merged when the implementation is wrong, have added a bug about it.

nordicjm (approved):

bbilas:
> Indeed I had a look and can't understand why it was merged when the implementation is wrong, have added a bug about it.

It can be easily fixed in the follow-up commit by removing that implementation entirely then the generic `watchdog0` alias will be used instead.

jeppenodgaard:
@de-nordic does this PR need anything before a merge?

PR #1676, opened 2023-04-13
Boot set next direct xip

de-nordic:
bootutil: Add DirectXIP version of boot_set_next

The commit adds DirectXIP version of bootutil boot_set_next
function.
The function is enabled by configuration option:
MCUBOOT_BOOTUTIL_LIB_FOR_DIRECT_XIP.

Note that testing/confirming DirectXIP images only works when DIRECT_XIP_REVERT is enabled.

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -459,6 +459,20 @@ boot_swap_type_multi(int image_index)
 }
 
 int
+boot_write_copy_done(const struct flash_area *fap)
+{
+    uint32_t off;
+
+    off = boot_copy_done_off(fap);
+    BOOT_LOG_DBG("writing copy_done; fa_id=%d off=0x%lx (0x%lx)",
+                 flash_area_get_id(fap), (unsigned long)off,
+                 (unsigned long)(flash_area_get_off(fap) + off));
+    return boot_write_trailer_flag(fap, off, BOOT_FLAG_SET);
+}
+
+
+#ifndef MCUBOOT_BOOTUTIL_LIB_FOR_DIRECT_XIP
+int
 boot_set_next(const struct flash_area *fa, bool active, bool confirm)
 {
     struct boot_swap_state slot_state;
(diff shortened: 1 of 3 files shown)


boot/bootutil/src/bootutil_public.c
```diff
@@ -528,6 +542,72 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
...
+int
+boot_set_next(const struct flash_area *fa, bool active, bool confirm)
+{
+    struct boot_swap_state slot_state;
+    int rc;
+
+    if (active) {
+        /* The only way to set active slot for next boot is to confirm it,
+         * as DirectXIP will conclude that, since slot has not been confirmed
+         * last boot, it is bad and will remove it.
```
nordicjm, line 555: Seems like this should be reworked, in swap using move you can have 2 applications constantly and flip between them, I don't see why this should be any different for XIP - might have an end application in one slot and e.g. a FOTA update receiver in the other, if mcuboot will always erase the slot which is not active, that seems like a bug. Outside the scope of this PR however.

boot/bootutil/src/bootutil_public.c
```diff
@@ -528,6 +542,72 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
...
+        }
+        /* Pass */
+
+    case BOOT_MAGIC_GOOD:
+        if (confirm) {
+            if (slot_state.copy_done == BOOT_FLAG_UNSET) {
+            /* Magic is needed for DirectXIP to even try to boot application.
+             * DirectXIP will set copy-done flag before attempting to boot
+             * application. Next boot, application that has copy-done flag
+             * is expected to alread have ok flag, otherwise it will be removed.
```
nordicjm, line 580: *already

boot/bootutil/src/bootutil_public.c
```diff
@@ -528,6 +542,72 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
...
+             */
```
nordicjm, line 581: Indent by extra 2 spaces

boot/bootutil/src/bootutil_public.c
de-nordic: That would require rework of how we set/reset flags for DIrectXIP.
Using retention API we could avoid writing any bits to flash just to test alternative image, so this is also an option here.

nvlsianpu (commented):
shouldn't this be fromtree?

de-nordic:
> shouldn't this be fromtree?

Yes, but I am waiting for reviews so that we can weed out basic mistakes.

boot/bootutil/src/bootutil_public.c
nordicjm: outstanding

nordicjm (changes requested):
2 minor issues need fixing

boot/bootutil/src/bootutil_public.c
de-nordic: Fixed

boot/bootutil/src/bootutil_public.c
de-nordic: Fixed

nordicjm (approved):

PR #1677, opened 2023-04-14
boot_serial: Fix include paths for zephyr builds

nordicjm:
Fixes include paths for zephyr builds to use the system zcbor_* files rather than the local files.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -26,7 +26,6 @@
 #include "sysflash/sysflash.h"
 
 #include "bootutil/bootutil_log.h"
-#include "zcbor_encode.h"
 
 #ifdef __ZEPHYR__
 #include <zephyr/sys/reboot.h>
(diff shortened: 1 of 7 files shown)


de-nordic:
I do not think this is correct. MCUboot compiles with included source for zcbor. Zephyr may have different version of zcbor and MCUboot should use the one it carries within.

nordicjm:
> I do not think this is correct. MCUboot compiles with included source for zcbor. Zephyr may have different version of zcbor and MCUboot should use the one it carries within.

Don't see why it should use the bundled (and very outdated with lack of features) version when there's a newer version available in the zephyr code base? Files in mcuboot tree should be for use by other OS's where they do not have access to the zcbor files (e.g. as a module)

nordicjm:
Note that this just fixes an issue with the wrong .h files being included, zephyr mcuboot builds already use the .c files from the zcbor repo as per https://github.com/mcu-tools/mcuboot/pull/1642

de-nordic (approved):
Looks OK.
Note: I would rather change how include paths for zcbor headers are defined and not use __ZEPHYR__ and other system specific ifdefs within common code, but there is already precedence for that anyway, so we may leave the paths option for later.

nordicjm:
I was thinking of adding an include path of `.` or maybe making a sub-directory to move the zcbor files to and adding that as an include path for non-zephyr builds but I don't currently have the know how for how other OS's set all that up but that would be a good future improvement, that way they can all use `#include <zcbor_....h>`

utzig:
Maybe another solution to have it up to date would be to add it a submodule at `/ext/zcbor`, and add a CI workflow that check Zephyr's `west.yml` to get use the same SHA.

de-nordic:
@utzig @nordicjm I have logged issue here: https://github.com/mcu-tools/mcuboot/issues/1687 where we can discuss how headers are picked into source code.

PR #1684, opened 2023-04-21
fix FIH return value for EC256 and add FIH for ED25519

Roolli:
There was an issue with bootutil_verify_sig for EC256 when the return type was not correct when FIH was enabled. ED25519 FIH support has also been added.


davidvincze (approved):
Thanks for the fix.

PR #1685, opened 2023-04-21
bootutil: Fix for ECDSA-P256 signature verification refactor

de-nordic:
Refactor removed code passing decoded ASN.1 signatures, without taking into account that some of bootutil_ecdsa_p256_verify may actually use it.
This has been breaking change and should have not happened without providing fix for bootutil_ecdsa_p256_verify implementations, which has not been done, or before giving developers time to fix them.


de-nordic:
Fixing it in implementations.

PR #1686, opened 2023-04-21
boot: Fix DOS line endings

d3zd3z:
This file made it into the repo with DOS line endings. Fix this to match everything else.


davidvincze (approved):
Thank you.

PR #1689, opened 2023-04-25
ci: use tomlib instead toml

marekmatej:


ci/get_features.py
```diff
@@ -16,7 +16,7 @@
 
 import argparse
 import os.path
-import toml
+import tomlib as toml
```
gustavonihei, line 19:
```suggestion
import tomllib as toml
```
Typo in the package name.

I am not sure if backwards compatibility is desirable for the CI. If so, then I believe this could also work:
```suggestion
try:
  import toml
except:
  import tomllib as toml
```

gustavonihei:
@marekmatej if the `toml` will no longer be supported, then it should be removed from the PIP requirements file

PR #1692, opened 2023-04-27
boot: bootutil: Add boot information

nordicjm:
Adds TLV defines and optional function for use with the bootloader shared data feature.

### DIFF
--- a/boot/bootutil/src/boot_record.c
+++ b/boot/bootutil/src/boot_record.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018-2021 Arm Limited
  * Copyright (c) 2020 Linaro Limited
+ * Copyright (c) 2023, Nordic Semiconductor ASA
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,12 +30,6 @@
 #include "bootutil_priv.h"
 #include "bootutil/image.h"
 #include "flash_map_backend/flash_map_backend.h"
-
-/* Error codes for using the shared memory area. */
-#define SHARED_MEMORY_OK            (0)
-#define SHARED_MEMORY_OVERFLOW      (1)
-#define SHARED_MEMORY_OVERWRITE     (2)
-#define SHARED_MEMORY_GEN_ERROR     (3)
 
 /**
  * @var shared_memory_init_done
@@ -191,7 +186,6 @@ boot_save_boot_status(uint8_t sw_module,
         }
     }
 
-
     if (!boot_record_found || !hash_found) {
         return -1;
     }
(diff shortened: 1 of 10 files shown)


nordicjm:
@utzig @de-nordic @nvlsianpu @d3zd3z can you review this PR please?

boot/bootutil/src/loader.c
```diff
@@ -146,7 +146,8 @@ boot_add_shared_data(struct boot_loader_state *state,
 
 #ifdef MCUBOOT_DATA_SHARING
     rc = boot_save_shared_data(boot_img_hdr(state, active_slot),
-                                BOOT_IMG_AREA(state, active_slot));
+                                BOOT_IMG_AREA(state, active_slot),
+                                active_slot);
```
de-nordic, line 150: I am not sure whether this is correct. I mean sharing `active_slot` may be correct because this is MCUboot/MCUmgr established slot numbering, but the BOOT_IMG_AREA(state, active_slot) pointer (`struct flash_area`) may differ between MCUboot and app builds, so sharing it does not help.

nordicjm: `active_slot` is what I added for XIP, the boot image area is part of the original API which presumably things like mynewt use so don't want to change/break it.

boot/bootutil/include/bootutil/zephyr/mcuboot_shared.h
```diff
@@ -0,0 +1,55 @@
...
+
+enum mcuboot_recovery_mode {
+    MCUBOOT_RECOVERY_MODE_NONE,
+    MCUBOOT_RECOVERY_MODE_SERIAL_RECOVERY,
+    MCUBOOT_RECOVERY_MODE_DFU,
+};
+
+struct mcuboot_configuration {
+    /** Version of this configuration struct #mcuboot_configuration_version */
+    uint8_t configuration_version;
```
de-nordic, line 39: Don't you want to have version + structure size header first?
This way the other side can read the header first and then decide whether it can process it, by the version, and how much data to read and how to cast it, by the size.

nordicjm: Wouldn't be able to use it like this, unless you mean the put the version and size at the end of the struct? Because the retention area size is fixed, the dts would need to be updated, so if the bootloader used version 1 and application used version 2 which were both different sizes, there would need to be a way to determine something like that from a common area

nordicjm: Read through the docs and updated as per requirements

nvlsianpu (approved):

Laczen:
@nordicjm, I am surprized that you have not taken any action to remove this PR.

The solution provided is not beneficial for mcuboot. The introduction of a zephyr dependent config `#ifndef __zephyr___` in bootutil code will start a deviation from common code to os specific code. The basics of the proposed PR: using the retention system to share data is wrong and unnecessary. A simple change to the mcuboot config allows to define the shared data memory as a part of a zephyr dts file.

d3zd3z:
My suggestions would be to perhaps split this into two PRs, as we are kind of doing two different things here. One is using the retention system in Zephyr as a mechanism to share data. The other is a new type of boot record. I think both of these are fine, but the new type of boot record should probably just be in the main boot code, and not Zephyr specific, and just the use of retention would be in the Zephyr code. These could be selected by configs.

d3zd3z (changes requested):
Small issue with a weird include that I think is redundant anyway.

boot/bootutil/src/boot_record.c
```diff
@@ -28,14 +29,9 @@
 #include "bootutil/boot_status.h"
 #include "bootutil_priv.h"
 #include "bootutil/image.h"
+#include "../../bootutil/src/bootutil_priv.h"
```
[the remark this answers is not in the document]
thedjnK: This include is needed to know if the mode is swap using scratch, as that is where the define is set, moving this out would probably be a big change and nothing else really needs it. https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/bootutil_priv.h#L64

boot/bootutil/src/boot_record.c
```diff
@@ -230,3 +225,105 @@ boot_save_boot_status(uint8_t sw_module,
...
+#elif defined(MCUBOOT_SIGN_ED25519)
+    uint8_t signature_type = MCUBOOT_SIGNATURE_TYPE_ED25519;
+#else
+    uint8_t signature_type = MCUBOOT_SIGNATURE_TYPE_NONE;
+#endif
+
+#if defined(MCUBOOT_SERIAL)
+    uint8_t recovery = MCUBOOT_RECOVERY_MODE_SERIAL_RECOVERY;
+#elif defined(MCUBOOT_SERIAL_WAIT_FOR_DFU) || (defined(__ZEPHYR__) && defined(CONFIG_BOOT_USB_DFU_GPIO))
+    uint8_t recovery = MCUBOOT_RECOVERY_MODE_DFU;
```
Laczen, line 264: The use of `__zephyr__` in a bootutil file should not be allowed, a generic `CONFIG_...` define should be created to add support for this case.

nvlsianpu: Honestly - even `CONFIG_...` shouldn't be used directly here. Rathe any Kconfig should be converted into Mcuboot-platform-agnostic configuration definition by `bootloader/mcuboot/boot/zephyr/include/mcuboot_config/mcuboot_config.h` and that one might be usede.

nordicjm: Fixed

Laczen:
IMO this PR does not follow the intended use of mcuboot shared data:
1. `boot_add_data_to_shared_area()` allows to add a tlv to the shared data area. This should be used directly at the start to save any configuration specific info to the shared data area.
2. `boot_save_shared_data()` allows image specific data to be added and has to be implemented by the os.

If the `boot_save_shared_data()` is to be default included in the library this might break existing os implementations.

d3zd3z (approved):

PR #1696, opened 2023-04-28
boot/zephyr: switch main return type to 'int'

keith-packard:
Adapt to Zephyr's change requiring main to return int.


stephanosio (approved):
For more details, see https://github.com/zephyrproject-rtos/zephyr/pull/54628

nordicjm (approved):

PR #1697, opened 2023-05-02
bootutil/crypto: Fix local variable name typo for PSA Crypto key parse

adeaarm:
Follow up on #1666 . A variable name has been changed during code review, but not in all the code branches, and was not caught by automated tests, which are not merged yet, so this was caught only when running the tests locally. This kind of issues can be covered by automation only when the simulator changes will be merged in with #1668 .


davidvincze (approved):

davidvincze:
Thank you for the fix.

PR #1701, opened 2023-05-05
sim: Increase test coverage for ram-load

Roolli:
This PR adds new test cases to the simulator to test the ram-load feature of MCUboot. New test variations have also been added for the simulator CI.

### DIFF
--- a/.github/workflows/sim.yaml
+++ b/.github/workflows/sim.yaml
@@ -44,6 +44,8 @@ jobs:
         - "sig-rsa validate-primary-slot ram-load multiimage"
         - "sig-rsa validate-primary-slot direct-xip multiimage"
         - "sig-ecdsa hw-rollback-protection multiimage"
+        - "ram-load enc-aes256-kw multiimage"
+        - "ram-load enc-aes256-kw sig-ecdsa-mbedtls multiimage"
     runs-on: ubuntu-latest
     env:
       MULTI_FEATURES: ${{ matrix.features }}
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -123,6 +123,19 @@ struct SlotPlace {
     size: u32,
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum ImageManipulation {
+    None,
+    BadSignature,
+    WrongOffset,
+    IgnoreRamLoadFlag,
+    /// True to use same address,
+    /// false to overlap by 1 byte
+    OverlapImages(bool),
+    CorruptHigherVersionImage,
+}
+
+
 impl ImagesBuilder {
     /// Construct a new image builder for the given device.  Returns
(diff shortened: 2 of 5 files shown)


sim/src/image.rs
```diff
@@ -1269,6 +1329,28 @@ impl Images {
...
     }
 
+    pub fn run_ram_load_boot_with_result(&self, expected_result: bool) -> bool {
+        if !Caps::RamLoad.present() {
+            return false;
+        }
+        // Clone the flash so we can tell if unchanged.
+        let mut flash = self.flash.clone();
+
+        // Create RAM config where the image would not fit.
```
davidvincze, line 1339: Is it still valid?

sim/src/image.rs
```diff
@@ -221,11 +234,58 @@ impl ImagesBuilder {
...
+        Images {
+            flash,
+            areadesc: self.areadesc,
+            images,
+            total_count: None,
+            ram: self.ram,
+        }
+    }
+
+    pub fn make_no_upgrade_image_manipulated(self, deps:&DepTest, img_manipulation: ImageManipulation) -> Images {
```
davidvincze, line 258: Could it be combined with make_no_upgrade_image()?

davidvincze (commented):
What's the impact of the new test cases on the execution time?

sim/src/image.rs
davidvincze: Thanks for the update.

sim/src/image.rs
davidvincze: Thanks.

.github/workflows/sim.yaml
```diff
@@ -43,6 +43,10 @@ jobs:
         - "sig-rsa validate-primary-slot direct-xip"
         - "sig-rsa validate-primary-slot ram-load multiimage"
         - "sig-rsa validate-primary-slot direct-xip multiimage"
+        - "ram-load enc-aes256-kw multiimage"
+        - "ram-load enc-aes256-kw sig-ecdsa-mbedtls multiimage"
+        - "ram-load enc-aes256-ec256 multiimage"
+        - "ram-load enc-aes256-x25519 multiimage"
```
davidvincze, line 49: I think these 2 can be removed. These image encryption algs are already tested separately and we have
other feature sets to test ram-load with multiple encrypted images. I believe these don't have too much added value.
What do you think?

davidvincze (commented):
Please rebase this patch stack.

.github/workflows/sim.yaml
Roolli: I have no objections.

Roolli:
> What's the impact of the new test cases on the execution time?

Insignificant, as the tests that are taking a long time aren't being ran.

davidvincze (approved):

PR #1703, opened 2023-05-09
boot: mynewt: add support for the mcumgr echo command in serial boot mode

kasjer:
This allows user to enable echo for mcumgr command in serial boot. Code was enabled in zephyr only so far.
Now mynewt build can also have this feature if enabled.

No code changes just option in mynewt build to enable existing mcuboot feature.


sjanc (approved):

utzig (approved):
LGTM

PR #1704, opened 2023-05-09
Fix CI after mynewt update

kasjer:
ci: mynewt: Download nrfx from separate repository

Mynewt used to have copy of nrfx.
Now nrfx is taken from original repository and since some CI targets want to build for NRF MCUs nrfx repository nees to be downloaded.


kasjer:
Without this change all subsequent PR will fail CI for mynewt and will be marked with read cross with no good reason.
There is no change mcuboot as such just CI is corrected to work with latest mynewt.

utzig (approved):
LGTM

PR #1705, opened 2023-05-09
boot_serial: Unify zcbor include paths

### LINKED ISSUE #1687: Move logic selecting where are headers taken from to Makefiles
Currently there are a few places, in common code, where included headers are #ifdefed for  various systems, for example in boot/boot_serial/src/boot_serial.c we use #ifdef __ZEPHYR__ to point where are some headers taken from.

Instead we should use the same methods of inclusion (<> or "") and the same header names but use Makefiles and build system settings to modify paths where the headers are searched for.

nordicjm:
Moves zcbor files to a subdirectory and replaces differing include path styles to be unified.

Fixes #1687


de-nordic (approved):

nordicjm:
@utzig can you review this please?

utzig (approved):

PR #1706, opened 2023-05-11
ci: mynewt: Use newt for downloading dependencies

sjanc:
This makes sure that Mynewt targets have all required dependencies downloaded.

### DIFF
--- a/ci/mynewt_install.sh
+++ b/ci/mynewt_install.sh
@@ -30,17 +30,18 @@ install_newt() {
     popd
 }
 
-shallow_clone_mynewt() {
-    mkdir -p repos/apache-mynewt-core
-    git clone --depth=1 https://github.com/apache/mynewt-core repos/apache-mynewt-core
+install_mynewt() {
+    build="$PWD/build"
+    newt new "$build"
     [[ $? -ne 0 ]] && exit 1
 
-    # nrfx is now taken from original repository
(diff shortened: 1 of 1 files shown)


ci/mynewt_install.sh
```diff
@@ -30,13 +30,17 @@ install_newt() {
...
 
-    # nrfx is now taken from original repository
-    git clone --depth=1 --branch v2.8.0 https://github.com/NordicSemiconductor/nrfx.git repos/nordic-nrfx
+    pushd /tmp/mynewt
+    newt upgrade --shallow=1
+    [[ $? -ne 0 ]] && exit 1
+
+    popd
+    rm /tmp/mynewt/repos/mcuboot -rf
+    cp -r /tmp/mynewt/repos .
```
utzig, line 43: What's the problem of using `mv`?

sjanc: no problem, I'll update it to use mv

d3zd3z (changes requested):

nordicjm:
Rebase needed

utzig:
I tried updating this PR with the suggested changes but it is broken because `newt new` creates a project that requires Mynewt 1.11.0 release which is not available yet. The new release is being made these week afaik. I suggest using #1777 right now and come back to this PR after Mynewt is fully released.

sjanc:
This should work now (you were unlucky to hit in the middle of me tagging repos for release;-)

sjanc:
Hi,

Why #1777 was merged instead of this?

PR #1709, opened 2023-05-16
zephyr: Add VERSION file

nordicjm:
Adds a file which contains the current MCUboot code version, which can be used by Zephyr builds.


nordicjm:
@d3zd3z there is now a zephyr version file as per this PR, please update this file when doing releases/tags.

boot/zephyr/VERSION
```diff
@@ -0,0 +1,5 @@
+VERSION_MAJOR = 1
+VERSION_MINOR = 11
```
de-nordic, line 2: We are still at minor 10. Why patchlevel is 99?

nordicjm: Currently at 1.11.0-dev according to a97f009d209deb20278a7e12234364bad02533f6, put it to 99 to copy the zephyr style release pattern of "not a proper release" since the "dev" text in the VERSION file is completely ignored

d3zd3z (changes requested):
Can you also add instructions on updating this to `docs/release.md` so that whoever is doing the release will remember to update this?

nordicjm:
> Can you also add instructions on updating this to `docs/release.md` so that whoever is doing the release will remember to update this?

Added

de-nordic (approved):

d3zd3z (approved):

PR #1715, opened 2023-05-23
zephyr: single_loader: fix wrong variable name

wag-thom:
Compilation failed if serial-recovery and encryption is enabled: wrong variable name for flash-area pointer.

Bug introduced with MR #1566


de-nordic (approved):

d3zd3z (changes requested):
The change itself looks fine, however, the commit itself will need work. In general, commits created through the github UI aren't going to be properly formatted. If you'd like to fix this, you'll probably need to apply these commits by cloning the tree, and pulling the changes to a local directory. You'll need to have a valid email address as the author of the commit, as well as have a Signed-off-by.

If instead, you'd prefer, one of the maintainers should be able to apply this commit for you. However, we'll still need an email address in order to give you credit in the commit.

wag-thom:
Sorry, I am not able to do that correctly at the moment. Someone else should do it, I don't need credit.

Next time I'll better just open an issue ;-)

PR #1727, opened 2023-06-21
nuttx: add support for swap without scratch area

michallenc:
Definition of MCUBOOT_SWAP_USING_MOVE in case swap without scratch area is configured in NuttX was missing from mcuboot_config.h file.

Also necessary function `flash_area_sector_from_off()` is defined and declared in order to support swap without scratch.


michallenc:
Hi @nordicjm, @d3zd3z, @utzig. I´ve noticed you are doing reviews on MCUboot pull requests, could you please take a look at this? Thanks!

utzig (approved):
LGTM

PR #1734, opened 2023-06-29
Add ECDSA signature support to PSA Crypto backend

Roolli:
Following up on the recent addition of the PSA Crypto API in MCUBoot, this PR adds the ability to verify ECDSA P256 and P384 signatures. If P384 is being used the hashing algorithm switches to SHA384.


boot/bootutil/include/bootutil/caps.h
```diff
@@ -35,22 +35,23 @@ uint32_t bootutil_get_caps(void);
...
+#define BOOTUTIL_CAP_ED25519                (1<<9)
+#define BOOTUTIL_CAP_ENC_EC256              (1<<10)
+#define BOOTUTIL_CAP_SWAP_USING_MOVE        (1<<11)
+#define BOOTUTIL_CAP_DOWNGRADE_PREVENTION   (1<<12)
+#define BOOTUTIL_CAP_ENC_X25519             (1<<13)
+#define BOOTUTIL_CAP_BOOTSTRAP              (1<<14)
+#define BOOTUTIL_CAP_AES256                 (1<<15)
+#define BOOTUTIL_CAP_RAM_LOAD               (1<<16)
+#define BOOTUTIL_CAP_DIRECT_XIP             (1<<17)
+#define BOOTUTIL_CAP_HW_ROLLBACK_PROT       (1<<18)
```
thedjnK, line 54: This breaks compatibility with everything

thedjnK (changes requested):

boot/bootutil/include/bootutil/caps.h
nordicjm: This could be used by downstream users to communicate capabilities with applications or as part of regression tests. It is common sense that once bitmask values are set, they are not changed or reordered, and new values are appended to the end of the list, not pushed into the middle of the list causing 16 other entries to be shifted around.

davidvincze: If used correctly, it should not cause any problems. But given its importance, the patch can be easily updated to append the new cap to the end instead. I believe the point here was to improve readability.
@Roolli may I ask you to update the patch accordingly?  Please also revert the modifications of 206b914 on this file.

d3zd3z: I don't see any particular reason to change this. True that it would need to be compiled. Please just add the new one at the end.

adeaarm (approved):

davidvincze:
@nordicjm could someone from Nordic have a look at the SHA-related modifications? Thanks for the help!

de-nordic:
> @nordicjm could someone from Nordic have a look at the SHA-related modifications? Thanks for the help!

I will take a look.

de-nordic (changes requested):
I have enough https://github.com/mcu-tools/mcuboot/issues/1754

davidvincze:
> I have enough #1754

Could you please elaborate on this in the focus of this change? You mention encryption in the issue, and I know there is at least one PR that I know of introducing some modifications to the encryption code, but in this case it's about the signature.
If this PR is breaking something I would like to know about it and I would like to understand how.

de-nordic:
> > I have enough #1754
> 
> Could you please elaborate on this in the focus of this change? You mention encryption in the issue, and I know there is at least one PR that I know of introducing some modifications to the encryption code, but in this case it's about the signature. If this PR is breaking something I would like to know about it and I would like to understand how.

Build me MCUboot for Zephyr with BOOT_ENCRYPT_EC256  enabled.

davidvincze:
@de-nordic may I ask you to review the changes around the hash functions, just to make sure these modifications won't cause you any problems with the integration work. Please request changes or approve because currently this PR is blocked for unrelated reasons thus delaying our work as well.

davidvincze:
@Roolli please check the ext/nrf/cc310_glue.h file and update it as required.

de-nordic:
> @Roolli please check the ext/nrf/cc310_glue.h file and update it as required.

Here is commit, on sdk-nrf, that fixes th cc310 for us: https://github.com/nrfconnect/sdk-mcuboot/commit/ae4344b68eb561f819f2f114da685c97cb3a8971

.github/workflows/sim.yaml
```diff
@@ -44,6 +44,8 @@ jobs:
         - "sig-rsa validate-primary-slot ram-load multiimage"
         - "sig-rsa validate-primary-slot direct-xip multiimage"
         - "sig-ecdsa hw-rollback-protection multiimage"
+        - "sig-ecdsa-psa psa-crypto-api"
```
oberon-sk, line 47: no need to specify `psa` twice
how about:

```
sig-ecdsa psa-crypto-api sig-p256
sig-ecdsa psa-crypto-api sig-p384
```

or

```
sig-ecdsa-psa sig-p256 
sig-ecdsa-psa sig-p384
```

the latter fits better with the previous scheme and the `psa-crypto-api` feature should then be removed

Roolli: Hi, thanks for the suggestions. I've implemented the second approach as it fits the current scheme more, but I opted not to add an algorithm specifier for p256 as that's the default.

sim/mcuboot-sys/build.rs
```diff
@@ -15,6 +15,8 @@ fn main() {
     let sig_rsa3072 = env::var("CARGO_FEATURE_SIG_RSA3072").is_ok();
     let sig_ecdsa = env::var("CARGO_FEATURE_SIG_ECDSA").is_ok();
     let sig_ecdsa_mbedtls = env::var("CARGO_FEATURE_SIG_ECDSA_MBEDTLS").is_ok();
+    let sig_ecdsa_psa = env::var("CARGO_FEATURE_SIG_ECDSA_PSA").is_ok();
+    let sig_use_p384 = env::var("CARGO_FEATURE_SIG_USE_P384").is_ok();
```
oberon-sk, line 19: remove "use" for consistency

nvlsianpu (commented):
I reviewed the code roughly,
looks sane to me. Don't see any regression.

nvlsianpu (commented):
@de-nordic Can you run your smoke tests suite on this PR?

adeaarm:
> So, one question that isn't clear to me. How are we distinguishing 256 and 384 bit ECDSA signatures in the TLV? Do we just rely on the size of the signature? What happens if there is a 256 bit signature in the TLV, but the build is configured for 384 bit signatures? Does it fail properly, or does something confusing happen.
> 
> I thought we were going to add an extra TLV to specify the signature size that should be given before the ECDSA signature, and would default to 256, but in the 384 case would be present to suggest a new signature length?

Signatures encoding in ASN-1 is

```
 SEQUENCE  {
     INTEGER r
     INTEGER s
 }
```
 
This means that the `length` information is part of the signature, in the length field of the ASN-1 TLV. When the signature is verified this structure is parsed and if the length does not match the length of (r,s) the verification fails. This is similar to how the curve information is encoded as part of the key encoding.

d3zd3z:
This does work to distinguish a given 256 bit key and a given 384 bit key. However, there are multiple curves possible at each of these sizes. Perhaps we are just saying in MCUboot that we will only ever support one key of a given size, but it seems like it might be a good idea to at least think of encoding the actual key type used, as the size isn't sufficient to determine which key is being used.

adeaarm:
> This does work to distinguish a given 256 bit key and a given 384 bit key. However, there are multiple curves possible at each of these sizes. Perhaps we are just saying in MCUboot that we will only ever support one key of a given size, but it seems like it might be a good idea to at least think of encoding the actual key type used, as the size isn't sufficient to determine which key is being used.

From `image.h`

```
/*
 * Image trailer TLV types.
 *
 * Signature is generated by computing signature over the image hash.
 * Currently the only image hash type is SHA256.
 *
 * Signature comes in the form of 2 TLVs.
 *   1st on identifies the public key which should be used to verify it.
 */
```

the public key for ECDSA encodes the OID of the curve in ASN-1:

```
 SEQUENCE {
    SEQUENCE {
        OBJECT idEcPublicKey
        OBJECT namedCurve
    }
    BIT STRING publicKey
 }
```

i.e. from `namedCurve` you have the precise information about the curve used and this TLV is always present with the actual signature TLV (encoded as shown in the other comment). Please correct me if I am making wrong assumptions here, but I believe you have all the information you need already in the ASN-1 encodings to precisely identify the curve used and the associated signature pair.

d3zd3z:
The public key has the OID of the algorithm used, but the signature itself is just r and s encoded in asn.1. We do reference a hash of the public key, so perhaps that is sufficient. I'm just going off other signature formats, such as COSE where the algorithm is always encoded as part of the signature.

de-nordic (approved):

PR #1735, opened 2023-06-30
boot_serial: fix image number handle in image upload request

pepe2k:
According to the [SMP protocol documentation](https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_1.html#image-upload-request), `image number` value in `image upload request` is optional and can be included only in packet with `off` (data offset) set to `0` (first packet in upload request).

In one of recent changes (commit cb07e888691d), initialization of the `img_num` variable was removed which, in extreme case (no image number provided in upload request), results in use of its uninitialized value in `flash_area_open()` call which then might lead to request abort.

This fixes above regression and also makes MCUboot implementation of the `image upload request` aligned with Zephyr documentation of the protocol by considering image number only from first (off == 0) `image upload request` SMP packet. In addition, `image number` value is set to `0` if the request doesn't provide this field.

Ref: https://docs.zephyrproject.org/latest/services/device_mgmt/smp_groups/smp_group_1.html#image-upload-request
Fixes: cb07e888691d ("boot_serial: Replace cbor auto-generated code with zcbor functions")

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -22,6 +22,7 @@
 #include <ctype.h>
 #include <stdio.h>
 #include <errno.h>
+#include <limits.h>
 
 #include "sysflash/sysflash.h"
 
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

nordicjm (commented):
needs `limits.h` include

carlescufi (approved):

carlescufi:
@davidvincze @d3zd3z Could you please take a look and approve/merge if happy?

d3zd3z (approved):

PR #1738, opened 2023-07-06
espressif: refactor after removing IDF submodule 

almir-okato:
This PR removes the IDF from gitmodules. 
The main reasons for the removal are avoiding to have unnecessary clones from IDF and avoiding system environment conflicts.

Instead of using IDF downloaded as submodule, the Espressif port uses a parameter passed on building as reference for HAL code:
- `ESP_HAL_PATH` - path where are the components used by Espressif port HAL.

An existing IDF installation also can be used, so if the parameters are not present when building, the build system will search for IDF_PATH from the environment variables.

Currently, the HAL code expected considers the IDF `release/v5.1` branch code basis. As the supported RTOSes get updated with the MCUboot Espressif port expected changes, a check for HAL versions should be add.


almir-okato:
@sylvioalves @utzig PTAL

utzig (approved):
LGTM

sylvioalves:
LGTM

PR #1739, opened 2023-07-07
boot: mynewt: Add build version comparison

kasjer:
For downgrade prevention it's possible to use build number for grater control.
So far only Zephyr has this option in Kconfig now
option is also available in mynewt.


utzig (approved):
LGTM

PR #1743, opened 2023-07-11
boot: mynewt: Respect minimum write size

kasjer:
For MCUs with restriction on minimum write size (STM32H7) unaligned writes resulted in flash write errors preventing any sort of update.

Now MCUBOOT_BOOT_MAX_ALIGN can be set accordingly to value that flash driver uses.
For alignment <= 8 default value provided by mcuboot config is still used.


utzig (approved):
LGTM

PR #1744, opened 2023-07-14
mynewt: Make sysinit() call configurable

kasjer:
In same cases (loging, hash, crypto) main function called newt tool generated sysinit() function to create
uart device and crypto.

Now user can specify that sysinit should be called for other cases if needed. This can be useful if some other package should be included in the build and it has package initialization function.


utzig (approved):
LGTM

Issue #1754, opened 2023-07-21
Encryption rework, that nobody asked for, keeps on breaking a lot of stuff


de-nordic:
All reworks to encryption that are not bug fixes will be nacked automatically unless developers that do them will learn to build and test them on some platforms, including Zephyr.
It seems that none of the changes are built or tested for Zephyr supported boards, as for some it is not even possible to build them.

I do not think that this project needs changes that are done for the sole purpose of making some changes and providing reports to employers on job well done.
This project is used by customers and breaking it does not help, specifically that we have a lot of areas that need improvement but instead extra hands are dropped into something that was working so far, before these hands got involved.
Try to be your own customer.

We struggle with complicated configuration, where it is hard to tell what encryption is supported by what lib and how to select options.
We struggle with lack of single linear main file, and MCUboot looks like several projects with different main.c files just accidentally gathered together under one directory tree.
We struggle with bootloader binary size.
Use of flash area as flash access does not seem optimal and cuts into name space of projects that try to link MCUboot as an app.
We have a lot of stuff to do, if somebody wants to help.

It does not help us that we have to re-test encryption configuration again and again.

CC @nordicjm @d3zd3z @davidvincze

davidvincze:
Hi,
Many changes were merged lately to support new functionalities, a new crypto library and a signature type. These modifications for example also included the refactoring (to a certain degree) of crypto modules with the aim of making the project easier to maintain/understand/develop by clearing the abstraction layers and boundaries. I think this is consistent with what you wrote above. I'm aware of the problems with the code structure and configuration, with the OS-specific files and code parts embedded into the bootloader's internal code, but I believe we want to achieve the same thing.

There were some problems with breaking compatibility (TLV, caps..), but they've been fixed quickly and understood, we should have been much more attentive with public interfaces. The goal is to make the next release acceptable to everyone.
Nevertheless, I'm not sure that blocking a PR is the right solution to solve an independent bug that's already present on `main`.
It is inevitable that new mistakes will be made as the project structure is far from ideal. Obviously, no one wants to cause trouble for others; If we find a bug all I want is cooperation so that we can fix it as soon as possible, later we can discuss and examine the reasons for not being caught earlier (new possibilities to improve the CI and processes?).

It cannot be expected a contributor will test the patch for all the OSs and platforms that use MCUboot. That's why we use CI and that's why the code is tested by all the interested parties before every release either by their own test environment or manually, focusing on the details/configs that are most important to them.

[...]

PR #1767, opened 2023-08-03
zephyr: Add shared data support

nordicjm:
Adds the ability to share mcuboot configuration with the
    application using Zephyr's retention subsystem.


boot/bootutil/src/boot_record.c
```diff
@@ -23,19 +24,19 @@
 
 #include "mcuboot_config/mcuboot_config.h"
 
+#if defined(__ZEPHYR__) && defined(MCUBOOT_DATA_SHARING)
```
d3zd3z, line 27: Shouldn't this ifdef (without the Zephyr part) just be in Zephyr's mcuboot_config.h? That's where the definitions are needed, anyway.

d3zd3z (changes requested):
Another little comment on includes.

boot/bootutil/src/boot_record.c
thedjnK: The definitions are only needed for this file (for the version), when included from the mcuboot config file, it causes a lot of build errors, presumably because parts of the code are being setup as "libraries" which would not have access to the application's generated files, which is where the version file would be placed

Laczen: @d3zd3z this is exactly my problem with the proposed solution: Instead of using existing mcuboot solutions the PR needs to overload the existing routines with its own set of routines to get a working solution. This results in os specific parts of the code inside bootutil. The construct is very fragile and will lead to difficult to catch errors. Given the issue #1754 solutions like the proposed should be avoided even if they have shown not to break mcuboot on other oses.
** on top of this the provided solution does not have any benefit over the existing shared memory solution that can easily be defined for zephyr. **

boot/bootutil/src/boot_record.c
```diff
@@ -23,19 +24,19 @@
...
+#include "../../bootutil/src/bootutil_priv.h"
 #include "flash_map_backend/flash_map_backend.h"
 
-/* Error codes for using the shared memory area. */
-#define SHARED_MEMORY_OK            (0)
-#define SHARED_MEMORY_OVERFLOW      (1)
-#define SHARED_MEMORY_OVERWRITE     (2)
-#define SHARED_MEMORY_GEN_ERROR     (3)
-
+#if !defined(MCUBOOT_CUSTOM_DATA_SHARING_FUNCTION)
```
Laczen, line 39: What is the mcuboot size implication of using a `CUSTOM_DATA_SHARING_FUNCTION` compared to the existing solution? Does this `CUSTOM_DATA_SHARING_FUNCTION` have any benefits ?

Laczen: The provided solution is cleaner and better than before, but it can still be avoided completely.

boot/bootutil/src/boot_record.c
d3zd3z: I don't really like having this here like this, but I can't really think of a better way of handling it.

d3zd3z (approved):

nordicjm:
@sylvioalves @almir-okato @gustavonihei seems mcuboot no longer builds for espressif platforms in CI due to a 5 byte RAM overflow (which is not caused by this PR), can you take a look at this?

Laczen:
The proposed solution using the retention subsystem will add about 2k of code (for the retention subsystem) to mcuboot. A simple shared memory based solution would add no code at all. The proposed retention system provides no benefit at all and is just a waste of space.

Laczen:
> Funny how you were disproven and yet continue to parrot the same invalid claims.
>
A shout from your side on a non-verifiable claim does not disprove anything. Wether it is 2k, 1.4k or 500 bytes is not important. Even adding 1 byte of code without adding anything should be avoided. So if you can show that your solution does not add anything to the code compared to a standard shared memory solution the proposed PR might be ok: take on the challenge...

> > On top of this the change made in #1692 modifies a public api as reported in #1775.
> 
> It adds 2 new parameters to a function, which is allowed by the project and is approved by the project maintainer

Even a maintainer can mis something. A bad decision has been made. Now two things can happen: stick by the decision or reverse the decision. Everyone with a little experience knows that reversing the decision is the right way forward.

boot/zephyr/shared_data.c
```diff
@@ -0,0 +1,124 @@
...
+    /* Check whether TLV entry is already added.
+     * Get the boundaries of TLV section
+     */
+    tlv_end = shared_data_size;
+    offset  = SHARED_DATA_HEADER_SIZE;
+
+    /* Iterates over the TLV section looks for the same entry if found then
+     * returns with error: SHARED_MEMORY_OVERWRITE
+     */
+    while (offset < tlv_end) {
```
carlescufi, line 66: @nordicjm does this preserve the actual on-RAM format that is used by the current  `MCUBOOT_DATA_SHARING` option?

nordicjm: I have not compared the two but both to ensure they are exactly the same but both store the data in TLV format, with retention having the optional header/checksum parts also

de-nordic (approved):

PR #1773, opened 2023-08-07
bootutil: Provide flash_area_to_image

de-nordic:
The flash_area_id_to_image has been replaced with flash_area_to_image that moves flash_area_get_id inside the flash_area_id_to_image and changes the logic so that it is now independent from image numbers and should not require updates when BOOT_IMAGE_NUMBER gets increased.


nordicjm (approved):

nordicjm:
Will need a rebase, at which point a release note should be added

boot/bootutil/src/bootutil_public.c
```diff
@@ -458,18 +458,18 @@ boot_swap_type_multi(int image_index)
     return BOOT_SWAP_TYPE_NONE;
 }
 
-static int flash_area_id_to_image(int id)
+static int flash_area_to_image(const struct flash_area *fa)
```
nvlsianpu, line 461: Looks, OK
Only one my concern is that what if `fa`  would out of range mcuboot's images.

de-nordic: Yeah, that is one of the things we should take care within code not to happen.

nvlsianpu: Ok, I'm fine with that statement.

nvlsianpu (approved):

PR #1777, opened 2023-08-09
ci: mynewt: Update to install upstream Mbed-TLS

utzig:
Update after recent removal of Mbed-TLS from Mynewt tree; now using upstream from Mbed-TLS github repo.


utzig:
This should temporarily fix the Mynewt CI build. #1706 is more correct, if it is not updated today I will take it over and re-submit with the suggested changes.

nordicjm (approved):

d3zd3z (approved):

PR #1788, opened 2023-08-29
zephyr: Allow user-defined boot serial extensions

nordicjm:
This allows for out-of-tree modules to define their own boot serial
    functions by using iterable sections

+106 byte flash increase with this change and both in-tree handlers enabled

### DIFF
--- a/boot/zephyr/boot_serial_extension_zephyr_basic.c
+++ b/boot/zephyr/boot_serial_extension_zephyr_basic.c
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2021-2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdio.h>
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/flash.h>
+#include <zephyr/mgmt/mcumgr/mgmt/mgmt_defines.h>
+#include <zephyr/mgmt/mcumgr/grp/zephyr/zephyr_basic.h>
+#include <../subsys/mgmt/mcumgr/transport/include/mgmt/mcumgr/transport/smp_internal.h>
+
+#include <flash_map_backend/flash_map_backend.h>
+#include <sysflash/sysflash.h>
+
+#include "bootutil/bootutil_log.h"
+#include "../boot_serial/src/boot_serial_priv.h"
+#include <zcbor_encode.h>
+
+#include "bootutil/image.h"
+#include "bootutil/bootutil_public.h"
+#include "bootutil/boot_hooks.h"
+
+#include <boot_serial/boot_serial_extensions.h>
+
+BOOT_LOG_MODULE_DECLARE(mcuboot);
+
+#ifdef CONFIG_BOOT_MGMT_CUSTOM_STORAGE_ERASE
+static int bs_custom_storage_erase(const struct nmgr_hdr *hdr,
+                                   const char *buffer, int len,
+                                   zcbor_state_t *cs)
+{
+    int rc;
+    const struct flash_area *fa;
+
+    (void)buffer;
+    (void)len;
+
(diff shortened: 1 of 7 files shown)


boot/zephyr/boot_serial_extension_zephyr_basic.c
```diff
@@ -0,0 +1,166 @@
...
+                zcbor_tstr_put_lit(cs, "");
+            }
+        }
+    }
+
+    zcbor_tstr_put_lit(cs, "rc");
+    zcbor_uint32_put(cs, MGMT_ERR_OK);
+    zcbor_map_end_encode(cs, 10);
+
+    return rc;
```
de-nordic, line 155: I am not sure whether `rc` here is stable: in case when it fails on the first slot in loop, then we have rc == 0 here finally and it seems that we return success even though we have failed in the first iteration, but if we fail in the last iteration then we return error here.
Does it make sense to return here something else than 0 if we assume that in case of error we just return empty string? Or should this be reflected by the "rc" returned by line 152?

boot/zephyr/boot_serial_extensions.c
```diff
@@ -1,161 +1,30 @@
...
-        } else if (hdr->nh_op == NMGR_OP_READ) {
-#ifdef MCUBOOT_MGMT_CUSTOM_IMG_LIST
-            if (hdr->nh_id == ZEPHYR_MGMT_GRP_BASIC_CMD_IMAGE_LIST) {
-                mgmt_rc = bs_custom_img_list(cs);
+    STRUCT_SECTION_FOREACH(mcuboot_bs_custom_handlers, function) {
+        if (function->handler) {
+            mgmt_rc = function->handler(hdr, buffer, len, cs);
+
+            if (mgmt_rc != MGMT_ERR_ENOTSUP) {
+                break;
```
de-nordic, line 26: Wouldn't this mean that there might be more than one handler called for the same buffer and mutiple responses being sent for one?

nordicjm: No, it would be a 1:1 mapping, if there are 3 handlers registered, the first one will be called, if it is not valid for that then it must return MGMT_ERR_ENOTSUP, it will then check the next handler, if the next handler is valid, it will return something else (i.e. usually EOK but could return EINVAL or anything) and then the loop here will be broken out, the final handler would not be called

boot/zephyr/boot_serial_extension_zephyr_basic.c
nordicjm: You know you are right in that it makes no sense, though it does appear that this code is actually what it does now: https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/boot_serial_extensions.c

Not really sure what this should return, I don't see this command documented in the zephyr docs?

nordicjm: In fact I would consider just removing this command entirely, what is the point in it? This is the hook it calls:
```
int boot_img_install_stat_hook(int image_index, int slot, int *img_install_stat)
{
    return BOOT_HOOK_REGULAR;
}
```
So why return this at all then img mgmt list can return actual information including status about the image

nordicjm: @nvlsianpu since you added this, what is the point in this command? Does anything use it? Can we just get rid of it?

nvlsianpu: This command was added in order to make possible (among other) fetch NET core image installation status on nRF53, which was never used anywhere.

nvlsianpu: Regard the return value - I would make it more consistent with failures this code could encounter.

nordicjm: I'm of the mind it should just be deleted, if something is needed downstream for this it can be added downstream, and this is really not at all in the style of MCUmgr commands by generating a string with multiple pieces of information - in CSV format, and just appending it. I will update the PR

de-nordic (approved):

PR #1791, opened 2023-09-04
boot: zephyr: use indication LED also in timeout based recovery

pepe2k:
This adds support for indication LED option (`MCUBOOT_INDICATION_LED`) in the timeout based recovery. Configured LED will be enabled when entering the recovery and disabled after selected timeout (if no mcumgr command was received).

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -620,6 +620,10 @@ int main(void)
     rc = boot_console_init();
     int timeout_in_ms = CONFIG_BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT;
     uint32_t start = k_uptime_get_32();
+
+#ifdef CONFIG_MCUBOOT_INDICATION_LED
+    gpio_pin_set_dt(&led0, 1);
+#endif
 #endif
 
     FIH_CALL(boot_go, fih_rc, &rsp);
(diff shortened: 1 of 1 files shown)


de-nordic:
This is Zephyr change. Is there Zephyr PR that adds the Kconfig?

pepe2k:
> This is Zephyr change. Is there Zephyr PR that adds the Kconfig?

Hi @de-nordic, the related Kconfig already exists (https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/Kconfig#L514). Do you recommend to cover this with additional Kconfig symbol?

This change only extends LED indication use case for the timeout-based serial recovery (https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/main.c#L616). Currently, if you enable timeout-based recovery (`BOOT_SERIAL_WAIT_FOR_DFU`), indication LED (`MCUBOOT_INDICATION_LED`) and "Stay in bootloader if no application" (`BOOT_SERIAL_NO_APPLICATION`), the LED will be set to on only after entering `boot_serial_enter()` (https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/main.c#L513). With this change, the LED will be _also_ set to on earlier, during the recovery timeout period, making it possible to tell whether the MCUboot is waiting for umcumgr command.

nordicjm (approved):

PR #1793, opened 2023-09-04
imgtool: dumpinfo: Fixed support for the custom alignment

mingulov:
Fixed support for the custom alignment (in case if BOOT_MAGIC_2 is provided). Current version fails to parse it with an error:

ValueError: invalid literal for int() with base 0: b'...

Using the same solution as 'swap_size' parsing as a fix.

### DIFF
--- a/scripts/imgtool/dumpinfo.py
+++ b/scripts/imgtool/dumpinfo.py
@@ -138,7 +138,7 @@ def dump_imginfo(imgfile, outfile=None, silent=False):
             max_align = 8
         elif trailer_magic[-len(BOOT_MAGIC_2):] == BOOT_MAGIC_2:
             # The alignment value is encoded in the magic field
-            max_align = int(trailer_magic[:2], 0)
+            max_align = int.from_bytes(trailer_magic[:2], "little")
         else:
             # Invalid magic: the rest of the image trailer cannot be processed.
             print("Warning: the trailer magic value is invalid!")


utzig (changes requested):
The proposal solution doesn't seem correct to me. For `swap_size` it was OK because it is always written in little endian, but max align is written according to the given endiannes, as can be seen here: https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/image.py#L179

`int(trailer_magic[:2], 0)` is obviously wrong because it is converting a string representing a number which is not what we have here.

One simple way to do it, is to convert in both little and big endian, maybe just get a `min` of those, and check the value against the list of valid max align values, eg:

```python
ma = trailer_magic[:2]
max_align = min(int.from_bytes(ma, "little"), int.from_bytes(ma, "big"))
assert max_align in [8, 16, 32], "Invalid alignment"
```

Well, something like that...

mingulov:
Unfortunately dumpinfo does not support big endian images at all (or to be more precise - images with non sys.byteorder endianness), it will fail much earlier due to the 1st data parsing:
https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/dumpinfo.py#L71

```
_header = struct.unpack('IIHHIIBBHI', b[:28])
```
(incorrect protected_tlv_size, img_size will trigger struct.error).

So there is no need to implement separate support for 'max_align' only.

Almost the same issue with 'verify' command: https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/image.py#L629

mingulov:
After more checks - it would be better to do it properly and not re-write again when add proper support for little/big endianness.

Issue #1792, opened 2023-09-04
Compilation error. 


tito97sp:
### Defect detected

__Static_assert_ defined in C11.

### Error log
Compilation raises the following compilation error in this line:

[https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L163-L164]([](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L163-L164))

```
mcuboot/boot/bootutil/include/bootutil/image.h:163:15: error: expected constructor, destructor, or type conversion before '(' token
  163 | _Static_assert(sizeof(struct image_header) == IMAGE_HEADER_SIZE,
      |               ^
compilation terminated due to -Wfatal-errors.

```

### Proposed Fix
Following modification fixes the issue.

```
static_assert((sizeof(struct image_header)) == IMAGE_HEADER_SIZE, 
                "struct image_header not required size\n");
```

### Change Log
Change added in this commit e3cbbece365d54514b1a909507820818a3125f3c

### Comments
Please @d3zd3z can you check this?

PR #1794, opened 2023-09-04
Fix error with deprecated _Static_assert macro.

tito97sp:
In C, when you don't include <assert.h>, the compiler treats static_assert as a keyword that maps to _Static_assert. Using static_assert is preferred because the same code will work in both C and C++.

Discussion open in #1792


denravonska:
um, request to reopen.

PR #1798, opened 2023-09-04
Update docs build releases; enable emoji rendering support

utzig:
* Update Gemfile.lock to match current github pages
* Add emoji support
* Add webrick as dependency; required for testing locally if using ruby versions > 3, and works with 2.7.4.


almir-okato (approved):
LGTM

PR #1801, opened 2023-09-07
boot: zephyr: don't allow GPIO entrance in timeout based serial recovery

pepe2k:
If a timeout based serial recovery is selected, there is no point in keeping support for GPIO based entrance method. These two ways for entering serial recovery should be mutually exclusive, same as it's done in USB DFU based recovery (Kconfig symbols: `BOOT_USB_DFU_GPIO` and `BOOT_USB_DFU_WAIT`).

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -141,7 +141,7 @@ menu "Entrance methods"
 menuconfig BOOT_SERIAL_ENTRANCE_GPIO
 	bool "GPIO"
 	default y
-	depends on GPIO
+	depends on GPIO && !BOOT_SERIAL_WAIT_FOR_DFU
 	help
 	  Use a GPIO to enter serial recovery mode.
 


pepe2k:
As an alternative solution I can think of here, `BOOT_SERIAL_ENTRANCE_GPIO` and `BOOT_SERIAL_WAIT_FOR_DFU` could be moved into Kconfig choice group, in similar way as it's done with the USB DFU (entrance can be one of: none, GPIO, timeout).

nordicjm:
They're not mutually exclusive though, there is no reason both of these modes cannot be supported in a single image, that would be limiting a user's choice without a good reason.

pepe2k:
> They're not mutually exclusive though, there is no reason both of these modes cannot be supported in a single image, that would be limiting a user's choice without a good reason.

Hi @nordicjm, ok, is there any reason why it's different in case of USB DFU, timeout and GPIO based entrances aren't available at the same time?

nordicjm:
Have not really looked at USB DFU support other than when playing around with some devices and abandoning it in favor of serial recovery and have improved serial recovery a lot in the past year, no reason USB DFU can't be reworked to support both as well

PR #1803, opened 2023-09-08
boot_serial: minor fixes in 'erase_range()' log output and for 'bs_list()'

pepe2k:
This short series includes 3 minor fixes for `boot_serial`:

1. drop redundant parameters from `bs_list()`
The `buf` and `len` parameters aren't used inside `bs_list()` function.

2. boot_serial: include flash area offset in log info
Currently, log info in `erase_range()` and `bs_upload()` shows range being erased/written relative to selected flash area which might be misleading. Include flash area offset in output so that the absolute addresses of the range being erased/written are shown.

3. don't use `%j` length modifier with Zephyr's minimal libc
The Zephyr's 'Minimal C library' doesn't support the `%j` length modifier when `cbprintf` is configured with minimum features.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -240,7 +240,7 @@ bs_list_img_ver(char *dst, int maxlen, struct image_version *ver)
  * List images.
  */
 static void
-bs_list(char *buf, int len)
+bs_list(void)
 {
     struct image_header hdr;
     uint32_t slot, area_id;
@@ -521,7 +521,7 @@ bs_set(char *buf, int len)
 out:
     if (rc == 0) {
         /* Success - return updated list of images */
-        bs_list(buf, len);
+        bs_list();
     } else {
         /* Error code, only return the error */
         zcbor_map_start_encode(cbor_state, 10);
@@ -551,7 +551,7 @@ static void
 bs_list_set(uint8_t op, char *buf, int len)
 {
     if (op == NMGR_OP_READ) {
-        bs_list(buf, len);
+        bs_list();
     } else {
 #ifdef MCUBOOT_SERIAL_IMG_GRP_IMAGE_STATE
         bs_set(buf, len);
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -596,8 +596,8 @@ static off_t erase_range(const struct flash_area *fap, off_t start, off_t end)
     }
 
     size = flash_sector_get_off(&sect) + flash_sector_get_size(&sect) - start;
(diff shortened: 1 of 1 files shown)


nordicjm (commented):
Seems like the "Writing at ..." parts should be updated also

pepe2k:
@nordicjm 

> Seems like the "Writing at ..." parts should be updated also

I've added that in second commit (`boot_serial: include flash area offset in log info`).

Thanks!

nordicjm:
Format needs updating for CI to pass

boot/boot_serial/src/boot_serial.c
```diff
@@ -596,8 +596,13 @@ static off_t erase_range(const struct flash_area *fap, off_t start, off_t end)
     }
 
     size = flash_sector_get_off(&sect) + flash_sector_get_size(&sect) - start;
-    BOOT_LOG_INF("Erasing range 0x%jx:0x%jx", (intmax_t)start,
-		 (intmax_t)(start + size - 1));
+#if defined(__ZEPHYR__) && defined(CONFIG_MINIMAL_LIBC) && defined(CONFIG_CBPRINTF_NANO)
```
de-nordic, line 599: No. Avoid Zephyr specific `CONFIG_` things in code.
You can define two different formatting strings on top of the unit and guard them with CONIFG.
Probably using `inttypes.h` definitions for PRI would be helpful here.
Maybe we need to change the string to avoid using intmax at all if possible, at all the pointer size should be known prior to compilation.

boot/boot_serial/src/boot_serial.c
```diff
@@ -240,7 +240,7 @@ bs_list_img_ver(char *dst, int maxlen, struct image_version *ver)
  * List images.
  */
 static void
-bs_list(char *buf, int len)
+bs_list(void)
```
de-nordic, line 243: There are conditionally compiled in calls, like bs_set, that still take both parameters.

pepe2k: @de-nordic yes, but the `bs_set()` makes use of these parameters, the `bs_list()` doesn't. Am I not seeing something obvious here?

boot/boot_serial/src/boot_serial.c
pepe2k: @de-nordic 

> No. Avoid Zephyr specific `CONFIG_` things in code.

OK. Let me think about alternative solution.

boot/boot_serial/src/boot_serial.c
de-nordic: Ah, sorry. I have looked at diffs quickly and missed that the bs_set is called from different function.

PR #1807, opened 2023-09-12
espressif: update secure boot and flash encryption

almir-okato:
Adjust secure boot and flash encryption after IDF v5.x updates. It also allows to enable secure boot on ESP32-C2.

Note: changes on secure_boot.c and some of flash_encryption.c functions were mostly copied "as is" from IDF.


almir-okato:
@marekmatej @sylvioalves PTAL

utzig (approved):

PR #1816, opened 2023-09-25
bootutii: Show error if flash area open fails (and panic)

nordicjm:
Shows an error if a particular flash area fails to open and panics the bootloader


de-nordic (approved):

d3zd3z (changes requested):
Code looks good, but there is a typo in the commit text: "bootutii".

Issue #1769, opened 2023-08-04
Implement imgtool sanity test in CI


davidvincze:
It has happened before that basic functionalities, commands of `imgtool` have been broken (e.g. #1746) and it went unnoticed for a while. This issue wants to focus on and track the effort to get a new test implemented as part of the existing imgtool GitHub [workflow](https://github.com/mcu-tools/mcuboot/blob/main/.github/workflows/imgtool.yaml).

This new **sanity test** can even be implemented as a Python unit test using one of the existing unit test frameworks or a simple basic bash script  that invokes all the available commands of imgtool with different parameters and then compare the output with a known data set (e.g. signed template image, public key... ) to detect regression.

mingulov:
It would be nice.
After a brief check - it is even worse now, as #1752 has broken some existing imgtool unittests - so even existing tests are already partially broken as they are not executed by CI.
I have created that #1819 as a 1st step to have better imgtool test.

davidvincze:
#1819 is in good state, should be merged soon.

PR #1819, opened 2023-09-26
imgtool: initial sanity test (for CI build)

mingulov:
imgtool tests are not used by CI (workflows), so even existing tests might be broken (or some other functionality, as described in #1769). 

Modified workflows to use existing imgtool unittests also trigger it on 'pull_request'.
An initial sanity test for imgtool commands is added.
As some already existing unittests are broken, fix them to be able to start to use this change.
Added missing dependencies to setup.py / requirements.txt (it also prevented sanity test execution).

An example for CI build can be found [here](https://github.com/mingulov/mcuboot/actions/workflows/imgtool.yaml).


scripts/tests/test_keys.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+@pytest.mark.parametrize("key_type", KEY_TYPES)
+@pytest.mark.parametrize("encoding", KEY_ENCODINGS)
+def test_getpub(key_type, encoding, tmp_path_persistent):
+    """Get public key"""
+    runner = CliRunner()
+
+    gen_key = tmp_name(tmp_path_persistent, key_type, GEN_KEY_EXT)
+    pub_key = tmp_name(tmp_path_persistent, key_type, PUB_KEY_EXT + "." + encoding)
```
davidvincze, line 113:
```suggestion
    pub_key = tmp_name(tmp_path_persistent, key_type, PUB_KEY_EXT
                       + "." + encoding)
```
minor comment (>80 chars)

scripts/tests/test_keys.py
```diff
@@ -0,0 +1,234 @@
...
+
+    result = runner.invoke(
+        imgtool,
+        [
+            "verify",
+            "--key",
+            str(wrong_key),
+            str(image_signed),
+        ],
+    )
```
davidvincze, line 232: Already listed in conftest.py, do we still need it?
For rsa-2048, rsa-3072, ecdsa-p256, ecdsa-p384, ed25519 we expect `assert result.exit_code == 0` to succeed and `assert result.exit_code != 0` to fail (which is always cross-checked with x25519).
For x25519 the `assert result.exit_code == 0` after the first verify will always fail and it's covered in conftest.py.\
What do you think?

mingulov: An idea was to do a verification with a valid key, just wrong one - not the same which had been used for signing, but using x25519 is definitely wrong there as a verification fails really due to unsupported key type there.

I think it will be fixed by generation another key with the same type.

mingulov: It is fixed in the forced update, 2 keys are generated now ('GEN_KEY' and 'GEN_ANOTHER_KEY') for every check and that check 'ensures' that another key can not be used for a verification.

.github/workflows/imgtool.yaml
```diff
@@ -11,6 +12,36 @@ concurrency:
...
+        pipenv run pytest --junitxml=../junit/pytest-results-${{ matrix.python-version }}.xml
+    - name: Upload test results
+      uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: pytest-results-${{ matrix.python-version }}
+        path: |
+          junit/pytest-results-${{ matrix.python-version }}*.xml
+        if-no-files-found: ignore
   environment:
```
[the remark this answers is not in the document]
mingulov: Yes I will check how it could be done.

scripts/imgtool/keys/general.py
```diff
@@ -11,13 +12,13 @@
 class KeyClass(object):
     def _emit(self, header, trailer, encoded_bytes, indent, file=sys.stdout,
               len_format=None):
-        if file and file is not sys.stdout:
+        if file and not isinstance(file, io.TextIOBase):
```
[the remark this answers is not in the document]
mingulov: It is tried to be handled for each case now in a general way, please check.

.github/workflows/imgtool.yaml
```diff
@@ -11,7 +12,38 @@ concurrency:
...
+    if: ${{ github.event_name == 'push' }}
```
mingulov, line 46: Updated, so the existing 'environment' job (for 'imgtool_run.sh') will run for 'push' events only, so it will be the same behavior as without this patch.

mingulov: Updated, all known comments must be handled now.

davidvincze (approved):
Thank you for the fixes and for your heroic parience.
Could you please do a rebase? Otherwise it looks good to me.

davidvincze:
I've tested the patch by manually triggering a CI build after rebase: https://github.com/davidvincze/mcuboot/actions/runs/9207789552

PR #1831, opened 2023-10-04
bootutil: Disable MCUBOOT_BOOT_MAX_ALIGN assert for non-swap modes

butok:
- Assert should be checked only for SWAP update modes.
- Allow platforms with page size >32 Bytes (e.g. LPC) to use MCUBoot, at least for non-SWAP update modes.

### DIFF
--- a/boot/bootutil/include/bootutil/bootutil_public.h
+++ b/boot/bootutil/include/bootutil/bootutil_public.h
@@ -85,8 +85,10 @@ extern "C" {
 
 #ifdef MCUBOOT_BOOT_MAX_ALIGN
 
+#if defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_SCRATCH)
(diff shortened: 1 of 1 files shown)


nordicjm (changes requested):
These are not MCUboot defines

butok:
> These are not MCUboot defines

You are right, Zephyr-specific. 
Fixed to MCUBoot-specific.

boot/bootutil/include/bootutil/bootutil_public.h
```diff
@@ -85,8 +85,10 @@ extern "C" {
 
 #ifdef MCUBOOT_BOOT_MAX_ALIGN
 
+#if MCUBOOT_SWAP_USING_MOVE || MCUBOOT_SWAP_USING_SCRATCH
```
nordicjm, line 88:
```suggestion
#if defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_SCRATCH)
```

nordicjm (changes requested):

nordicjm (approved):

de-nordic (approved):

PR #1833, opened 2023-10-05
imgtool: make "align" command line parameter optional

butok:
"Align" parameter should be optional:
- it has a default value.
- it is not used for non-swap update modes.


d3zd3z (approved):

mingulov:
Sorry could someone clarify how it was intended to work / where it works?

For me without 'align' option I receive an exception on 'align=int(align)':
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

butok:
> Sorry could someone clarify how it was intended to work / where it works?
> 
> For me without 'align' option I receive an exception on 'align=int(align)': TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

Interesting, why we are using  align=int(align). 
All other integer parameters do not have such approach ( slot_size=slot_size, max_sectors=max_sectors, max_align=max_align etc.).
Guess, we should delete int().  May we?

PR #1842, opened 2023-10-18
Update version files for 2.0.0

d3zd3z:
Update documentation and version tags for final 2.0.0 release.


nordicjm (commented):
The title is wrong, no changes in PR relate to zephyr (or changes for zephyr are missing from this PR)

d3zd3z:
I seem to have Zephyr-on-the-brain. I'll fix this.

d3zd3z:
@nordicjm I've fixed the stray Zephyr. We won't think too deeply about the commit already there that incorrectly mentions Zephyr.

thedjnK (changes requested):
Missing update of https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/VERSION

nordicjm (approved):

de-nordic (approved):

d3zd3z:
I'd like to pull in the mynewt version upgrade. It was unclear to me if this is supposed to happen before or after the actual release. But, shouldn't the mynewt version also be 2.0.0?

Issue #1845, opened 2023-10-26
Imgtool trailer size check incorrect for swap-using-move


heinwessels:
**What**
---

The calculation of the trailer size (`_trailer_size()`) is incorrect when the bootloader strategy is swap using move. It does not take the flash sector size into account, but rather the flash write size. For devices such as the STM32H7 the write size (32) is much smaller than the sector size (128k), which results in the calculated trailer size being much smaller than it will be in reality.

This will result in the image being signed successfully. But when this image ends up in the primary slot, and a swap is requested, the swap will fail and the bootloader will panic. 

```
I: Starting bootloader. optima-platform: v0.5.2-dev.49+C3F25B1AB
I: Primary image: magic=good, swap_type=0x2, copy_done=0x1, image_ok=0x1
I: Secondary image: magic=good, swap_type=0x2, copy_done=0x3, image_ok=0x3
I: Boot source: none
I: Swap type: test
I: Starting swap using move algorithm.
W: Not enough free space to run swap upgrade
W: required 1048576 bytes but only 917504 are available
E: panic!
```

In this example the available `917504` is correct (`7 * 128k = 917504`), but the image was not `1048576` and instead `787416`. Note: these logs are from multiple different trail and errors, so the actual image size might be off by a few bytes.

It fails on [this line](https://github.com/mcu-tools/mcuboot/blob/304fd41980ed929533b9f387dde1b463b0be5b90/boot/bootutil/src/image_validate.c#L379):
```
if (it.tlv_end > bootutil_max_image_size(fap)) { 
        rc = -1;
        goto out;
}
```
where:
```
it.tlv_end = 787752
bootutil_max_image_size(fap) = 786432
```
which then prints the values relative to the sector sizes.
```
            BOOT_LOG_WRN("required %d bytes but only %d are available",
                         (last_idx + 1) * sector_sz,
                         first_trailer_idx * sector_sz);
```

**Expected behaviour**
---

The imgtool take take the boot stategy into account, and fail while signing. And not fail in when an update is requested.

**Impact**
---

Quite bad, as it can brick your device in some circumstances. For example in our case where we only have a connection to the device through the application, and rely on the application to update. If the bootloader can't boot then the device is bricked.

**More notes:**
---

As far as I know the imgtool doesn't currently know what the boot strategy is, and can therefore not really know what size the trailer will be. For example, the code in the bootloader that determines the max image size depends on the boot strategy:
```C
[...]

PR #1849, opened 2023-10-30
zephyr: Add estimated image footer size to cache in sysbuild

nordicjm:
Adds MCUboot's estimated overhead footer size to the application's cache when using sysbuild, this allows that information to be propagated to applications which can use the information to reduce the available size for an application, preventing the MCUboot error of image too large to swap.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -1,6 +1,7 @@
 # CMakeLists.txt for building mcuboot as a Zephyr project
 #
 # Copyright (c) 2017 Open Source Foundries Limited
+# Copyright (c) 2023 Nordic Semiconductor ASA
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


heinwessels:
I'm not following exactly what this PR will achieve. Will this solve https://github.com/mcu-tools/mcuboot/issues/1845 by letting the application compilation fail? (Meaning imgtool doesn't have to be updated).

nordicjm:
> I'm not following exactly what this PR will achieve. Will this solve #1845 by letting the application compilation fail? (Meaning imgtool doesn't have to be updated).

This means that (when using sysbuild) the application slot size will take into account all overheads that the image has, so in the case of using swap using move this would include the swap data, the image footer (TLV) data, the swap status data. At present you can build images in zephyr that are below the size of a slot, then upload then and mark them for test, at that point MCUboot will reject them saying they are too large, this prevents that issue by preventing such images from compiling entirely until their flash usage is reduced.

de-nordic (approved):

Issue #221, opened 2018-02-05
Support for revoking/invalidating keys


aditihilbert:
Fabio Utzig added a comment - 22/May/17 11:25 AM
I haven't seen this ticket before, but one suggestion to implement this would be to have another partition to put keys. We could build mcuboot with the original key, and if it needs to be revoked we write a new key to some special partition. If the key is available on the partition we assume it is to be used instead of the one linked in mcuboot. This, of course, assumes we control the fw by using our key. 

One problematic situation would be if we loose control of the key, before we are able to update the fw someone can use the stolen key to update the fw first and assume control. 

But this should be enough to start a discussion...

mbolivar:
I think if we are going to support revoking keys, we should be sure that they can be permanently revoked.

In this proposal, though, we'd be storing keys in a location that can be updated later (say if the application were compromised and persuaded to "un-revoke").

d3zd3z:
Some thoughts.

First, key revocation becomes a lot more important when we allow key chains instead of just individual keys (this could be either x.509, or our own mechanism to allow a chain of signatures).

Revocation should happen immutably, but there isn't always very good hardware support it. Some devices have one-time programmable flash, which fits very well with this. Another option is to have a small partition to hold revocation (with, say two flash partition). This would need to be protected from modification by regular apps.

This (along with things like hardware anti-rollback) starts to make MCUboot more dependent on hardware-specific features, which means we should probably come up with a mechanism for optional APIs to the BSP code, possibly with defines to determine which to use.

joh06937:
I was just writing up a new issue before noticing this idea was already posted here.

I'll second this idea, though my thought was to keep it a little simpler and provide the ability for a developer to provide N public keys during compilation of MCUBoot and invalidate "older" ones during run-time when an application update uses (successfully) a "newer" one.

It looks like the boot loader has support for having multiple public keys available for validation applications (though, in an SDK I'm using that includes MCUBoot and Zephyr, there wasn't any sanctioned way I found to provide this; I had to make a few customizations to Kconfig/CMake to outsource the generation of the C array and manually put bootutil_keys[] and bootutil_key_cnt together in my application). I simply changed the access to the array to instead use an API (same for getting the count), and then added an API for "invalidating" any keys that need to be invalidated. Right now, the system could have maybe a __weak implementation that does nothing -- keeping functionality exactly as it exists now -- but then a developer on a particular platform could implement an invalidation function.

d3zd3z:
[...]

PR #1869, opened 2023-11-30
boot/bootutil: give the hash of the received key to boot_retrieve_public_key_hash

Olstyle:
MCUBOOT_HW_KEY already allows to put your own signature key management into the bootloader, independent of mcuboot. In the moment boot_retrieve_public_key_hash is limited to exactly one key hash, since it is only called once and there is no way for the called function to determine whether it will be returning an accepted hash.    
Adding the hash itself to the call enables the project specific code to look for a fitting hash instead of just returning one. This way two things can be accomplished: 

1. Multiple HW keys can be used since now iterating through a list of hashes is feasible
2. Automatic revocation (see #221 ) can be added since the function now knows which key was used and might decide to invalidate other keys based on that fact

The key revocation scheme I plan to use based on this patch is based on a sorted list of key hashes. Normally the first entry is expected to be used for signing. If an entry deeper down the list is used, this indicates that the private keys up to that entry have been compromised and they should be invalidated. 
Invalidation itself is HW specfic. For example on an STM32 it is possible to overwrite already written flash with zeros, which can be used to delete a non zero validity flag.

### DIFF
--- a/boot/bootutil/include/bootutil/sign_key.h
+++ b/boot/bootutil/include/bootutil/sign_key.h
@@ -51,6 +51,7 @@ extern struct bootutil_key bootutil_keys[];
  * Retrieve the hash of the corresponding public key for image authentication.
  *
  * @param[in]      image_index      Index of the image to be authenticated.
+ * @param[in]      actual_key_hash  hash of the key to test
  * @param[out]     public_key_hash  Buffer to store the key-hash in.
  * @param[in,out]  key_hash_size    As input the size of the buffer. As output
  *                                  the actual key-hash length.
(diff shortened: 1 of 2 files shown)


d3zd3z (commented):
I do think this is slightly better. But, I have a little bit of a concern that we are effectively breaking an API, and this would result in a major version change. It is hard to tell who out there uses this function, so we will need to make sure there are clear directions on how to update this. So, at a minimum, there need to be release notes on this.

But, I really wonder if this whole interface is actually right. Presumably a system with a hardware key wouldn't have the key available (perhaps the public key?), and it seems really weird to me how the key returned from this function is then written to a global. If anything, this makes this code untestable in the simulator (this is not introduced by this change, but already there). I'm not sure how to do this better.

Olstyle:
To be honest this is just the smallest change I could think of. If we accept this is a breaking change anyway, the interface could be changed to directly return a bool about whether the key hash is accepted.

From what I understood from the original MCUBOOT_HW_KEY setup, the argument for only using hashes was that those might be saved in OTP memory or something similarly expensive. That's not actually what I am doing. I just needed an interface which lets me look at some info about the key (or the key itself) and make a decision instead of giving a key directly. 

Maybe `bootutil_find_key` needs to be rethought for a third time. Originally I expected it to return a (pointer to a) key, but neither version of it does this.

PR #1870, opened 2023-11-30
add the option to use custom definitions for all crypto inline functions

Olstyle:
As previously discussed in #1478 , this adds a define named `MCUBOOT_USE_CUSTOM_CRYPT`. 
Using that define, `headername_custom.h` will be included in the crypto headers instead of enabling any pre defined code path.
Since the include calls throughout mcuboot are clean already, it is actually enough to implement only the used algorithms. So if e.g. RSA is not used, a definition of `rsa_custom.h` is NOT necessary.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr.h
@@ -15,8 +15,9 @@
 #include "mcuboot_config/mcuboot_config.h"
 
 #if (defined(MCUBOOT_USE_MBED_TLS) + \
-     defined(MCUBOOT_USE_TINYCRYPT)) != 1
-    #error "One crypto backend must be defined: either MBED_TLS or TINYCRYPT"
+     defined(MCUBOOT_USE_TINYCRYPT) + \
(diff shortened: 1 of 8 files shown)


Olstyle:
Please don't activate an auto close bot who is faster in handling issues than the maintainers.

d3zd3z (changes requested):
It isn't quite clear to me the benefit of adding a set of "custom" headers that aren't included, instead of just allowing what the header that should be used itself to be a define? Then the two existing ones wouldn't be special, other than being in tree, and custom ones can be named appropriately by whoever is making them.

Olstyle:
> It isn't quite clear to me the benefit of adding a set of "custom" headers that aren't included, instead of just allowing what the header that should be used itself to be a define? Then the two existing ones wouldn't be special, other than being in tree, and custom ones can be named appropriately by whoever is making them.

This is merely the smallest change I could think of without breaking the current way of configuring the crypto library, making this a minor change.

From a functional point of view one could get rid of the library define all together and just have multiple versions of the crypto headers in separate folders. Configuration would be done by including the "glue header" folder of the desired library.
But of course this is then a major version change.

PR #1878, opened 2023-12-19
imgtool: fix "align" script error

butok:
Fix "align" possible script error, caused by #1833

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -377,6 +377,7 @@ class BasedIntParamType(click.ParamType):
               'keyword to automatically generate it from the image version.')
 @click.option('-v', '--version', callback=validate_version,  required=True)
 @click.option('--align', type=click.Choice(['1', '2', '4', '8', '16', '32']),
+              default='1',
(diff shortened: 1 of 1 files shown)


scripts/imgtool/main.py
```diff
@@ -416,7 +416,7 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
         pad = True
     img = image.Image(version=decode_version(version), header_size=header_size,
                       pad_header=pad_header, pad=pad, confirm=confirm,
-                      align=int(align), slot_size=slot_size,
+                      align = int(align) if align else 1, slot_size=slot_size,
```
rustammendel, line 419: Could we revert this back to `align=int(align), slot_size=slot_size,`
and apply the change suggested on line 379?

rustammendel (approved):

d3zd3z (approved):
Not so sure about `1` being the default, but it is fine to give this a default.

The other option would be to make this actually option, not just a default, and require it to be specified in configurations that is needed.

PR #1888, opened 2024-01-16
scripts: imgtool: Fix img verify for hex file format

lzala:
Currently imgtool --verify fails for hex files with:

Invalid image magic; is this an MCUboot image?

Added support for hex files by converting hex to bin using IntelHex::tobinstr().

Reusing image.load() needs a bit of rework, maybe a common load method will be done in the future.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -623,8 +623,15 @@ class Image():
 
     @staticmethod
     def verify(imgfile, key):
-        with open(imgfile, "rb") as f:
-            b = f.read()
+        ext = os.path.splitext(imgfile)[1][1:].lower()
+        try:
+            if ext == INTEL_HEX_EXT:
+                b = IntelHex(imgfile).tobinstr()
+            else:
+                with open(imgfile, 'rb') as f:
+                    b = f.read()
+        except FileNotFoundError:
+            raise click.UsageError(f"Image file {imgfile} not found")
 
         magic, _, header_size, _, img_size = struct.unpack('IIHHI', b[:16])
         version = struct.unpack('BBHI', b[20:28])


lzala:
@rustammendel can you please help with the review ?

scripts/imgtool/image.py
```diff
@@ -623,8 +623,15 @@ def pad_to(self, size):
...
-            b = f.read()
+        ext = os.path.splitext(imgfile)[1][1:].lower()
+        try:
+            if ext == INTEL_HEX_EXT:
+                b = IntelHex(imgfile).tobinstr()
+            else:
+                with open(imgfile, 'rb') as f:
+                    b = f.read()
+        except FileNotFoundError:
+            raise click.UsageError("Input file not found")
```
rustammendel, line 634: nit: Would it worth to mention the input file that is not found is image file? Might be easier for the user to distinguish whether keyfile or the imgfile path is wrong.

rustammendel (commented):
Left a little comment, otherwise LGTM. However I'm not a maintainer on this repo.

lzala:
Hi @d3zd3z can you please help with the review?

scripts/imgtool/image.py
lzala: good point, I'll add the file name as well

nordicjm:
@lzala can you fix compliance issue?

lzala:
@nordicjm  I'm trying to fix the issue seems to be related to a wrong token configuration or something .. because the author and the "Sigend off by:" are the the same but it is looking for a different email address

d3zd3z (approved):

nordicjm:
@lzala this needs a rebase

lzala:
@nordicjm rebase done

PR #1890, opened 2024-01-24
boot/zephyr: Define SOC_FLASH_0_ID and SPI_FLASH_0_ID

de-nordic:
The defines have been taken from the Zephyr flash_map.h but as they are provided there for MCUboot only, they can be just defined here.


de-nordic:
Marked DNM because to avoid merging while important zcbor work is in progress.

nvlsianpu (approved):

PR #1897, opened 2024-02-01
boot: Fix swap-move algorithm failing to validate multi-image

### LINKED ISSUE #1768: swap_move: boot_read_image_header mixes slot numbers when more than one image is swapped in multi-image configuration
When multi-image configuration is used, for example two images, and swap is requested, either confirmed or test, on both images, the swap will happen but the boot will fail with error:
```
E: Image in the primary slot is not valid!
```
or 
```
Unable to find bootable image
```
depending on image size, as the `boot_read_image_header` may hit more or less proper data.

The problem comes from lines:
https://github.com/mcu-tools/mcuboot/blob/fefc398cc13ebbc527e297fe9df78cd98a359d75/boot/bootutil/src/swap_move.c#L99-L116
where it seems that logic by mistake swaps slot numbers, which results in slots being paired with incorrect headers.

When the same configuration is used but only swap of one image is requested then the problem does not occur.

de-nordic:
In multi image swap validation of images could fail due to headers being incorrectly re-read from storage.

Fixes #1768

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -4,6 +4,7 @@
  * Copyright (c) 2016-2020 Linaro LTD
  * Copyright (c) 2016-2019 JUUL Labs
  * Copyright (c) 2019-2023 Arm Limited
+ * Copyright (c) 2024 Nordic Semiconductor ASA
  *
  * Original license:
  *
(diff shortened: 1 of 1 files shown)


de-nordic:
@matt-wood-ct can you also take a look if this fixes your problem?

matt-wood-ct:
> @matt-wood-ct can you also take a look if this fixes your problem?

Looking into it, but don't wait for me as I'm not sure when I'm going to get this retested

nvlsianpu (commented):
Looks sane.
I tested that it actually works. After looking into the code I got to the conclusion that `boot_read_image_header()` was doing unwanted operation of seeking `slot` and `offset` for physical header reading (which is the case only for ongoing image move operation).

nvlsianpu (approved):
As I wrote boot_read_image_header() should read from the primary slot which according to iteration over images. This could be done achieved by resetting `bs->idx` as well or `bs->op` to a new state (I didn't analyse consequences of the second option).

de-nordic:
> As I wrote boot_read_image_header() should read from the primary slot which to iteration over images. This could be achieved by resetting `bs->idx` as well or `bs->op` to a new state (I didn't analyse consequences of the second option).

Added a comment to clarify why there is NULL instead of bs now.

PR #1903, opened 2024-02-22
Add BOOT_SERIAL_WAIT_FOR_DFU_ALWAYS Zephyr Kconfig option.

mondalaci:
This PR adds the `BOOT_SERIAL_WAIT_FOR_DFU_ALWAYS` Kconfig option. When set to `n`, under normal circumstances, the bootloader doesn't enter serial recovery mode, allowing the firmware to start up as quickly as possible. However, when physically resetting the device, the bootloader enters serial recovery mode, which comes in very handy when the firmware is broken.

This patch was actually created by Vidar Berg at https://devzone.nordicsemi.com/f/nordic-q-a/108213/mcuboot-should-only-timeout-after-hardware-reset/469112 who allowed me to submit it here as a PR.

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -150,12 +150,20 @@ config BOOT_SERIAL_DETECT_DELAY
 	  the one used to place the device in bootloader mode.
 
 menuconfig BOOT_SERIAL_WAIT_FOR_DFU
-	bool "Wait a prescribed duration to see if DFU is invoked by receiving a MCUmgr comand"
+	bool "Wait a prescribed duration to see if DFU is invoked by receiving a MCUmgr command"
 	depends on BOOT_SERIAL_UART || BOOT_SERIAL_CDC_ACM
 	help
 	  If y, MCUboot waits for a prescribed duration of time to allow
 	  for DFU to be invoked. The serial recovery can be entered by receiving any
 	  mcumgr command.
+
+config BOOT_SERIAL_WAIT_FOR_DFU_ALWAYS
+	bool "Always enter serial recovery mode upon startup"
+	depends on BOOT_SERIAL_WAIT_FOR_DFU
+	help
+	  If y, MCUboot always enters serial recovery mode upon startup for the
+	  duration of BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT. Otherwise, MCUboot only
+	  enters serial recovery mode if the board was physically reset.
 
 config BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT
 	int "Duration to wait for the serial DFU timeout in ms"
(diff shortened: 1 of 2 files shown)


boot/zephyr/Kconfig.serial_recovery
```diff
@@ -150,13 +150,21 @@ config BOOT_SERIAL_DETECT_DELAY
...
 	  for DFU to be invoked. The serial recovery can be entered by receiving any
 	  mcumgr command.
 
+config BOOT_SERIAL_WAIT_FOR_DFU_ALWAYS
+	bool "Always enter serial recovery mode upon startup"
+	depends on BOOT_SERIAL_WAIT_FOR_DFU
+	help
+	  If y, MCUboot always enters serial recovery mode upon startup for the
+	  duration of BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT. Otherwise, MCUboot only
+	  enters serial recovery mode if the board was physically reset.
```
nordicjm, line 166: This comment doesn't really make sense, what does resetting have to do with staying in serial recovery mode?

nordicjm (changes requested):
Comment does not really make sense, and there already is a mode of entering serial recovery when physical reset button is used `CONFIG_BOOT_SERIAL_PIN_RESET`

mondalaci:
Sorry, the Kconfig description is inaccurate. The following description is accurate:

> If y, MCUboot enters serial recovery for the duration of BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT upon physically resetting the board. Otherwise, MCUboot immediately executes the application.

The use case is that I want the board to boot up as quickly as possible under normal circumstances, but I also want to be able to manually trigger the bootloader even if the application firmware is broken.

It's also worth mentioning that my device is battery-powered without a dedicated on/off switch, but it has a RESET button, so the timeout allows restarting the application firmware in case of a dead loop without getting stuck in the bootloader.

CONFIG_BOOT_SERIAL_PIN_RESET doesn't fulfill the above use case because:
- It doesn't time out when hardware reset is used but gets stuck in serial recovery mode.
- It delays the execution of the application firmware when the board is powered under normal circumstances. (My device has an optional battery, and it can be powered via USB, too.)

mondalaci:
Please remove the stale label. This commit has merit, and I'm using this Kconfig option.

nordicjm:
> The use case is that I want the board to boot up as quickly as possible under normal circumstances, but I also want to be able to manually trigger the bootloader even if the application firmware is broken.

Then use `CONFIG_BOOT_SERIAL_NO_APPLICATION`

mondalaci:
Where can I find the documentation of `CONFIG_BOOT_SERIAL_NO_APPLICATION`?

It's unavailable in https://github.com/nrfconnect/sdk-nrf/blob/main/modules/mcuboot/boot/zephyr/Kconfig

PR #1909, opened 2024-02-29
boot: Refactoring image dependency functions to reduce code size

de-nordic:
There have been duplicate functions:
 boot_verify_dependencies
 boot_verify_slot_dependencies
 boot_verify_slot_dependency
with, very similar internals, scattered around unit. The commit have moved them on top and squashed where possible.


de-nordic:
I have somehow managed to break the PR and it will not update from my branch and it sees the mcuboot/main from the past.

PR #1916, opened 2024-03-12
boot: zephyr: Fix estimated size calculation

nordicjm:
Fixes an issue with the estimated size calculation which wrongly used the maximum align size for some multiplications, this would mean that in some instances the estimated maximum image size was smaller than the actual allowed size


de-nordic (approved):
Looks OK.

PR #1917, opened 2024-03-13
boot: zephyr: Add a custom boot banner

erwango:
Cutomize MCUBoot boot banner to ease identifiation of boot steps in console.

### DIFF
--- a/boot/zephyr/prj.conf
+++ b/boot/zephyr/prj.conf
@@ -35,3 +35,6 @@ CONFIG_MCUBOOT_LOG_LEVEL_INF=y
 CONFIG_CBPRINTF_NANO=y
 ### Use the minimal C library to reduce flash usage
 CONFIG_MINIMAL_LIBC=y
+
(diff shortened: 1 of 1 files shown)


nordicjm (changes requested):
The problem with this is it does not print the MCUboot version, it prints the zephyr version

erwango:
> The problem with this is it does not print the MCUboot version, it prints the zephyr version

Agreed. This was a quick attempt, printing the right version would require additional change in Zephyr. I need to find some time to address it.

nordicjm:
> > The problem with this is it does not print the MCUboot version, it prints the zephyr version
> 
> Agreed. This was a quick attempt, printing the right version would require additional change in Zephyr. I need to find some time to address it.

It should be able to be done in MCUboot itself as ncs does it

nordicjm:
Closing as superseded by https://github.com/mcu-tools/mcuboot/pull/1945

PR #1920, opened 2024-03-15
boot: zephyr: add `BOOT_MAX_IMG_SECTORS_OVERRIDE`

### LINKED ISSUE #1919: zephyr: `CONFIG_BOOT_MAX_IMG_SECTORS` is not configurable for downstream platforms
### Problem

The `BOOT_MAX_IMG_SECTORS` setup currently requires modifications to the mcuboot repo to support boards with non-default values.

https://github.com/mcu-tools/mcuboot/blob/5d067f0e0ca7639b8c1f9aa2804c8718caae2440/boot/zephyr/Kconfig#L383-L390

Because this file is the absolute first contents of the kconfig tree, there is no ability to update this default on a per-board basis.

https://github.com/mcu-tools/mcuboot/blob/5d067f0e0ca7639b8c1f9aa2804c8718caae2440/boot/zephyr/Kconfig#L758

As a result, there are board specific overlays in this application that for purposes of maintability really should be set wherever the board is defined, e.g:
https://github.com/mcu-tools/mcuboot/blob/5d067f0e0ca7639b8c1f9aa2804c8718caae2440/boot/zephyr/boards/lpcxpresso55s28.conf#L1-L6

The only way to set this value is explicitly via `CONFIG_BOOT_MAX_IMG_SECTORS=N`. This is only possible from the command line or by modifying the mcuboot application. Attempting to set it in a `board_defconfig` file results in compilation errors as the symbol only exists for the bootloader application.

### Expected behaviour

Defaults for the zephyr bootloader that reasonably could change on a per-board basis should be configurable in the same location that the board is defined.

JordanYates:
Add a symbol to allow the default value of `BOOT_MAX_IMG_SECTORS` to be updated by external kconfig files.

${BOARD_ROOT}/board/kconfig.default
```
// This board requires at least 206 sectors
config BOOT_MAX_IMG_SECTORS_OVERRIDE
	bool "Override default value"
	default y if MCUBOOT

config BOOT_MAX_IMG_SECTORS
	int "Override default value"
	depends on MCUBOOT
	default 256
```

Resolves #1919

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -365,9 +365,12 @@ config BOOT_ENCRYPTION_KEY_FILE
 	  with the public key information will be written in a format expected by
 	  MCUboot.
 
+config BOOT_MAX_IMG_SECTORS_OVERRIDE
+	bool "Use external default for BOOT_MAX_IMG_SECTORS"
+
 config BOOT_MAX_IMG_SECTORS
 	int "Maximum number of sectors per image slot"
-	default 128
+	default 128 if !BOOT_MAX_IMG_SECTORS_OVERRIDE
 	help
 	  This option controls the maximum number of sectors that each of
 	  the two image areas can contain. Smaller values reduce MCUboot's


nordicjm (changes requested):
This is not the right way to do this, you can use `APPLICATION_CONFIG_DIR` to specify a directory that contains the configuration, you would need to copy `app.overlay` and `prj.conf` from `boot/zephyr` then you can add a `boards` directory with the overlays for each board

JordanYates:
> This is not the right way to do this, you can use `APPLICATION_CONFIG_DIR` to specify a directory that contains the configuration, you would need to copy `app.overlay` and `prj.conf` from `boot/zephyr` then you can add a `boards` directory with the overlays for each board

So the right way is to copy-paste two files to a new location (requiring a manual re-sync each time mcuboot is updated) and then requiring the user to add `-DAPPLICATION_CONFIG_PATH=/some/path` on each bootloader build, where `/some/path` is likely to be different depending on where that particular board is implemented.

Compared to adding the MCUboot configs to the board `kconfig.defconfig` and everything working by default.
I'm not sure I want to do it the right way 😄.

PR #1929, opened 2024-03-28
boot/bootutil: Remove direct accesses to flash_area::fa_xyz

adri326:
I plan on working towards building a Rust API for bootutil, and I would like to pass an opaque `struct flash_area` to bootutil, so that this struct can be fully managed from the rust side, with only a few `extern "C"` functions that will need to be defined on the rust side before linking.

This PR removes the few direct accesses to `flash_area::fa_id`, `flash_area::fa_off` and `flash_area::fa_size` that are present in bootutil, and replaces them with the corresponding `flash_area_get_XYZ` functions.

Currently, however, one can only use an opaque `struct flash_area` if they enable the new flash sector API (`MCUBOOT_USE_FLASH_AREA_GET_SECTORS`), since otherwise `loader.c` will attempt to create an array of `struct flash_area`. I thus do not know how we could ensure that no additional mentions of `flash_area::fa_XYZ` will sneak back into the code.

One option would be to create a `bootutil_shared.h` file, which would contain the minimal expected subset of `flash_map_backend.h`, and progressively have files include it, until the `flash_map_backend.h` only needs to provide the `struct flash_area` for when `MCUBOOT_USE_FLASH_AREA_GET_SECTORS` isn't toggled on.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -173,7 +173,7 @@ boot_status_off(const struct flash_area *fap)
     elem_sz = flash_area_align(fap);
 
 #if MCUBOOT_SWAP_USING_SCRATCH
-    if (fap->fa_id == FLASH_AREA_IMAGE_SCRATCH) {
+    if (flash_area_get_id(fap) == FLASH_AREA_IMAGE_SCRATCH) {
         off_from_end = boot_scratch_trailer_sz(elem_sz);
     } else {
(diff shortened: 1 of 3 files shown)


de-nordic (approved):
Thanks.

de-nordic:
@adri326 Do not get used to flash_area in MCUboot: https://github.com/orgs/mcu-tools/projects/1?pane=issue&itemId=40095438

PR #1932, opened 2024-04-04
Allow the usage of builtin keys in verification

davidvincze:
Introduce a new MCUBOOT_BUILTIN_KEY option to enable the usage of
builtin keys for signature verification. This way the details of the key
handling mechanism are abstracted away from the boot code and this
responsibility is delegated to the given crypto library.
This is an alternative option to the existing MCUBOOT_HW_KEY feature,
however in this case we can entirely rely on key IDs and not only the
code, but also the image metadata does not contain any public key data.

Enable the usage of builtin keys in the ECDSA verification module with
the PSA Crypto API based cryptographic backend.
This way parsing and importing the verification keys can also be avoided.

### DIFF
--- a/boot/bootutil/src/image_ecdsa.c
+++ b/boot/bootutil/src/image_ecdsa.c
@@ -3,7 +3,7 @@
  *
  * Copyright (c) 2016-2019 JUUL Labs
  * Copyright (c) 2017 Linaro LTD
- * Copyright (C) 2021-2023 Arm Limited
+ * Copyright (C) 2021-2024 Arm Limited
  *
  * Original license:
  *
@@ -35,6 +35,7 @@
 #include "bootutil/fault_injection_hardening.h"
 #include "bootutil/crypto/ecdsa.h"
 
+#if !defined(MCUBOOT_BUILTIN_KEY)
 fih_ret
 bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
                     uint8_t key_id)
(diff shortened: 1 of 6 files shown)


docs/design.md
```diff
@@ -1182,8 +1182,10 @@ If you want to enable and use encrypted images, see:
 
 By default, the whole public key is embedded in the bootloader code and its
 hash is added to the image manifest as a KEYHASH TLV entry. As an alternative
-the bootloader can be made independent of the keys by setting the
-`MCUBOOT_HW_KEY` option. In this case the hash of the public key must be
+the bootloader can be made independent of the keys (avoiding the incorporation
+of the public key into the code) by using one of the following options:
```
nvlsianpu, line 1186: for legibility I would mention these options there by names, and adds note on its main gain immediately after, then explain them in details:
```  console
... by using one of the following options: `MCUBOOT_HW_KEY` or `MCUBOOT_BUILTIN_KEY`.

Using any of these options MCUboot is independent from the public key(s).
The key(s) can be provisioned any time and by different parties.

Hardware KEYs support options details:
```

boot/bootutil/src/image_ecdsa.c
```diff
@@ -42,6 +42,7 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
     int rc;
     bootutil_ecdsa_context ctx;
     FIH_DECLARE(fih_rc, FIH_FAILURE);
+#ifndef MCUBOOT_BUILTIN_KEY
```
nvlsianpu, line 45: I would do two implementation of a ~function for call it there, one for MCUBOOT_BUILTIN_KEY and second for  MCUBOOT_BUILTIN_KEY

tamasban:
LGTM as an initial support for builtin keys. However, in the long run, might be better to provide a user-facing solution to specify the key ID. In the current solution, it is the responsibility of the crypto library. It seems hidden to me, and might not fit to a general crypto library. 

Options could be:
- provide a HAL API
- add a TLV which holds the key-id

Could you open an issue to track this?

davidvincze:
> LGTM as an initial support for builtin keys. However, in the long run, might be better to provide a user-facing solution to specify the key ID. In the current solution, it is the responsibility of the crypto library. It seems hidden to me, and might not fit to a general crypto library.
> 
> Options could be:
> 
> * provide a HAL API
> * add a TLV which holds the key-id
> 
> Could you open an issue to track this?

I've just opened an issue to track: https://github.com/mcu-tools/mcuboot/issues/1935

tamasban (approved):

d3zd3z (approved):

PR #1937, opened 2024-04-10
bootutil/crypto: Add ECDSA P256 support using mbedTLS

stevenlee7189:
This pull request introduces the BOOT_ECDSA_MBEDTLS configuration, enabling the use of the mbedTLS library for the ECDSA P256 signature.


de-nordic:
Why there are not CI jobs running on this?

d3zd3z:
> Why there are not CI jobs running on this?

Usually it will ask about running from new contributors. But, it isn't asking now. I wonder if that is because there are now conflicts.

PR #1938, opened 2024-04-10
Fix check tlv

d3zd3z:
Add an additional check to the image validation to ensure that TLV entries that need to be protected are actually in the protected section.  This works with a whitelist of entries that are allowed to be unprotected.  There is a define `ALLOW_ROGUE_TLVS` that can be set to bypass this check.


boot/bootutil/src/tlv.c
```diff
@@ -132,3 +132,23 @@ bootutil_tlv_iter_next(struct image_tlv_iter *it, uint32_t *off, uint16_t *len,
...
+ */
+int
+bootutil_tlv_iter_is_prot(struct image_tlv_iter *it, uint32_t off)
+{
+	if (it == NULL || it->hdr == NULL || it->fap == NULL) {
+		return -1;
+	}
+
+	return off < it->prot_end;
+}
```
davidvincze, line 154: Tabs are used here for indentation, please use whitespace.

boot/bootutil/src/image_validate.c
```diff
@@ -420,6 +444,27 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
...
+		       found = true;
+		       break;
+		  }
+	     }
+	     if (!found) {
+		  FIH_SET(fih_rc, FIH_FAILURE);
+		  goto out;
+	     }
+	}
+#endif
```
davidvincze, line 466: Tabs are also used here for indentation, please use whitespace.

boot/bootutil/src/image_validate.c
```diff
@@ -349,6 +349,30 @@ bootutil_get_img_security_cnt(struct image_header *hdr,
...
+     IMAGE_TLV_SHA384,
+     IMAGE_TLV_RSA2048_PSS,
+     IMAGE_TLV_ECDSA224,
+     IMAGE_TLV_ECDSA_SIG,
+     IMAGE_TLV_RSA3072_PSS,
+     IMAGE_TLV_ED25519,
+     IMAGE_TLV_ENC_RSA2048,
+     IMAGE_TLV_ENC_KW,
+     IMAGE_TLV_ENC_EC256,
+     IMAGE_TLV_ENC_X25519,
```
davidvincze, line 370: This is the current situation - considering how `imgtool` works.
In theory the keyhash/pubkey and encryption TLVs could be in the protected area.
Do you think it would make sense to move these TLVs in the future?
As for the keyhash/pubkey TLV, I prefer to have it together with the signature TLV to make re-signing of an image more convenient.

davidvincze (commented):
Left a few minor comments, otherwise it looks good to me.

boot/bootutil/src/image_validate.c
d3zd3z: I think the keyhash/pubhash can stay in the unprotected. The things directly related to signature verification make sense to keep there. It probably makes sense to move the enc ones, but it isn't urgent.

I'll fix the whitespace.

davidvincze (approved):

PR #1939, opened 2024-04-11
bootutil: Fix memory leak in HKDF implementation

taltenbach:
The `bootutil_hmac_sha256_set_key` routine performs some dynamic memory allocations when mbedTLS is used. To properly free the allocated memory, `bootutil_hmac_sha256_drop` must be called before re-initializing the HMAC context using `bootutil_hmac_sha256_init`. However, in the `hkdf` routine, the HMAC context was freed only once even though it was initialized multiple times.

This PR fixes the issue.

### DIFF
--- a/boot/bootutil/src/encrypted.c
+++ b/boot/bootutil/src/encrypted.c
@@ -276,6 +276,8 @@ hkdf(uint8_t *ikm, uint16_t ikm_len, uint8_t *info, uint16_t info_len,
         goto error;
     }
 
+    bootutil_hmac_sha256_drop(&hmac);
+
     /*
      * Expand
      */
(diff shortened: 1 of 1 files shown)


taltenbach:
> This looks correct to me. When I first implemented the HKDF is was Tinycrypt only, and I don't remember when MBED-TLS support was added, but the API usage was probably "misinterpreted". Did you measure how much memory is leaked?

I didn't but from a quick glance at the mbedTLS and MCUboot code:
* All allocations seem to take place in `mbedtls_md_setup`, which is invoked by `bootutil_hmac_sha256_set_key`.
* Each time `mbedtls_md_setup` is invoked, two allocations are made for a total of 236 bytes of memory:
    * `mbedtls_calloc(1, sizeof(mbedtls_sha256_context))` => 108 bytes
    * `mbedtls_calloc(2, md_info->block_size)` => 2 * 64 = 128 bytes
* `bootutil_hmac_sha256_set_key` seems to called three times by `hkdf`, but the HMAC context is freed only once, so at each call to `hkdf`, 2 * 236 = 472 bytes are leaked.
* `hkdf` is invoked once per call to `boot_enc_decrypt`, so the total number of bytes leaked will depend on the MCUboot configuration I guess

davidvincze (approved):
Looks good to me too. Thank you for the fix!

d3zd3z:
This should have a release note stub, though.

PR #1946, opened 2024-04-18
mbedtls config: fix too early check_config.h includes

tomi-font:
check_config.h was included manually by custom configuration files. This caused compilation errors when updating MbedTLS to 3.6.0 because check_config.h was processed too early, before the whole configuration is defined, effectively causing configuration check errors.

MbedTLS already takes care of including check_config.h at the right time. Remove those erroneous manual check_config.h includes.


frkv (approved):
LGTM (Good improvement that makes the code work like it is intended in Mbed TLS)

de-nordic (approved):

nvlsianpu (approved):

PR #1956, opened 2024-05-07
boot/zephyr: use socs conf

marekmatej:
Use socs folder to configure Espressif targets.
Remove all boards confs which was created by copying other boards.


boot/zephyr/socs/esp32s3_procpu.conf
almir-okato, line 1: Why it was an ESP32-S2 conf renamed to a ESP32-S3 one? Same configuration options?

marekmatej: I renamed it by mistake, but the config for the S2 and S3 is the same in this case.

sylvioalves:
LGTM

almir-okato (approved):
LGTM

sylvioalves:
@d3zd3z PTAL.

PR #1957, opened 2024-05-08
zephyr: print version number before boot

toonst:
Before booting, It might be handy to see the version of the image we boot into.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -553,6 +553,10 @@ int main(void)
                  rsp.br_image_off);
 #endif
 
+    BOOT_LOG_INF("Image version: v%d.%d.%d", rsp.br_hdr->ih_ver.iv_major,
+                                                    rsp.br_hdr->ih_ver.iv_minor,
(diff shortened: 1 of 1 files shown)


boot/zephyr/main.c
```diff
@@ -536,6 +536,9 @@ int main(void)
 
     BOOT_LOG_INF("Bootloader chainload address offset: 0x%x",
                  rsp.br_image_off);
+    BOOT_LOG_INF("Image version: v%d.%d.%d", rsp.br_hdr->ih_ver.iv_major,
```
de-nordic, line 539: This should be configurable, because we start to take extra space with things that will be rarely viewed by anyone.

d3zd3z: I would imagine those concerned deeply about space will have logging off anyway. I'd hate to see yet another ifdef in the code just for this.

d3zd3z (approved):

nordicjm:
Rebase needed

toonst:
Rebased

nordicjm (approved):

PR #1961, opened 2024-05-16
zephyr: boards: Add files needed for NXP FRDM MCXN947 QSPI variant

mmahadevan108:
1. set BOOT_MAX_IMG_SECTORS value for frmd_mcxn947_qspi. W25Q64 flash on the board is very large (8MB), so we must increase the number of max sectors when targeting this board with MCUboot.
2. Set the zephyr,flash chosen node to point to internal flash as this board supports booting from internal flash only.


mmahadevan108:
@d3zd3z , could you help assign reviewers for this PR.

nordicjm (approved):

PR #1967, opened 2024-05-28
boot: SHA512 verification

michalek-no:
adds TLV and Kconfig to decouple verification from other options.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/sha.h
+++ b/boot/bootutil/include/bootutil/crypto/sha.h
@@ -34,13 +34,16 @@
     #error "One crypto backend must be defined: either CC310/MBED_TLS/TINYCRYPT/PSA_CRYPTO"
 #endif
 
-#if defined(MCUBOOT_SIGN_EC384)
+#if defined(MCUBOOT_SHA512)
+    #define IMAGE_HASH_SIZE (64)
+    #define EXPECTED_HASH_TLV IMAGE_TLV_SHA512
+#elif defined(MCUBOOT_SIGN_EC384)
     #define IMAGE_HASH_SIZE (48)
     #define EXPECTED_HASH_TLV IMAGE_TLV_SHA384
 #else
     #define IMAGE_HASH_SIZE (32)
     #define EXPECTED_HASH_TLV IMAGE_TLV_SHA256
-#endif /* MCUBOOT_SIGN_EC384 */
+#endif /* MCUBOOT_SIGN */
 
 /* Universal defines for SHA-256 */
 #define BOOTUTIL_CRYPTO_SHA256_BLOCK_SIZE  (64)
@@ -82,7 +85,9 @@ typedef psa_hash_operation_t bootutil_sha_context;
 static inline int bootutil_sha_init(bootutil_sha_context *ctx)
 {
     *ctx = psa_hash_operation_init();
-#if defined(MCUBOOT_SIGN_EC384)
+#if defined(MCUBOOT_SHA512)
+    psa_status_t status = psa_hash_setup(ctx, PSA_ALG_SHA_512);
+#elif defined(MCUBOOT_SIGN_EC384)
     psa_status_t status = psa_hash_setup(ctx, PSA_ALG_SHA_384);
 #else
     psa_status_t status = psa_hash_setup(ctx, PSA_ALG_SHA_256);
@@ -107,7 +112,9 @@ static inline int bootutil_sha_finish(bootutil_sha_context *ctx,
 {
     size_t hash_length = 0;
     /* Assumes the output buffer is at least the expected size of the hash */
-#if defined(MCUBOOT_SIGN_EC384)
+#if defined(MCUBOOT_SHA512)
+    return (int)psa_hash_finish(ctx, output, PSA_HASH_LENGTH(PSA_ALG_SHA_512), &hash_length);
+#elif defined(MCUBOOT_SIGN_EC384)
     return (int)psa_hash_finish(ctx, output, PSA_HASH_LENGTH(PSA_ALG_SHA_384), &hash_length);
 #else
     return (int)psa_hash_finish(ctx, output, PSA_HASH_LENGTH(PSA_ALG_SHA_256), &hash_length);
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -24,6 +24,12 @@ config BOOT_USE_MBEDTLS
 	help
 	  Use mbedTLS for crypto primitives.
 
+config BOOT_USE_PSA_CRYPTO
+	bool
+	# Hidden option
+	help
+	  Hidden option set if using PSA crypt for cryptography functionality
+
 config BOOT_USE_TINYCRYPT
 	bool
 	# Hidden option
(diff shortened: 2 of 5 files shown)


boot/bootutil/include/bootutil/crypto/sha.h
```diff
@@ -82,7 +85,9 @@ typedef psa_hash_operation_t bootutil_sha_context;
 static inline int bootutil_sha_init(bootutil_sha_context *ctx)
 {
     *ctx = psa_hash_operation_init();
-#if defined(MCUBOOT_SIGN_EC384)
+#if defined(MCUBOOT_SHA512)
+    psa_status_t status = psa_hash_setup(ctx, PSA_ALG_SHA_512);
```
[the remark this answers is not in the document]
michalek-no: 512 and 384 are using the same init/clone/update... calls.
256 is handled by different functions.

boot/zephyr/Kconfig
```diff
@@ -72,8 +72,10 @@ choice BOOT_SIGNATURE_TYPE
 	default BOOT_SIGNATURE_TYPE_RSA
 
 config BOOT_SIGNATURE_TYPE_NONE
-	bool "No signature; use only hash check"
-	select BOOT_USE_TINYCRYPT
+	bool "No signature; use only sha256 check"
+
+config BOOT_SIGNATURE_TYPE_NONE_SHA512
+	bool "No signature; use only sha512 check"
```
nordicjm, line 78: Not really sure what the point of this change is? The signature would only be for detecting bit flips, the chance of having a bit flip and a sha256 hash still matching is basically impossible, so not sure why allowing a sha512 for this would improve anything?

michalek-no: In general yes, but it's security requirement we got.

nvlsianpu: AFAIK  `BOOT_SIGNATURE_TYPE_NONE` was introduced for giving SW evaluation (for development testing) value. It is not intended to use in any real product. If we are not going to use `BOOT_SIGNATURE_TYPE_NONE_SHA512` for anything like this - then it is not needed.

de-nordic:
Sorry, got to add own sign off on the thing because I just added one line.

boot/zephyr/Kconfig
```diff
@@ -24,6 +24,13 @@ config BOOT_USE_MBEDTLS
 	help
 	  Use mbedTLS for crypto primitives.
 
+config BOOT_USE_PSA_CRYPTO
+	bool
+	# Hidden option
+	default n
```
nordicjm, line 30: `default n` is never needed

boot/zephyr/Kconfig
```diff
@@ -67,19 +74,68 @@ config SINGLE_APPLICATION_SLOT
...
+	help
+	  Hidden option to set by configurations that allow SHA512
+
+choice BOOT_IMG_HASH_ALG
+	prompt "Selected image hash algorithm"
+	help
+	  Hash algorithm used for image verification. Selection
+	  here may be limited by other configurations, like for
+	  example selected cryptographic signature.
+	default BOOT_IMG_HASH_ALG_SHA256 if BOOT_IMG_HASH_ALG_SHA256_ALLOW
```
nordicjm, line 98: default goes above help text, help text is always last. In addition the list of options seems reversed, this means if you supported sha512 and sha256, you would always default to sha256

boot/zephyr/Kconfig
```diff
@@ -67,19 +74,68 @@ config SINGLE_APPLICATION_SLOT
...
+config BOOT_IMG_HASH_ALG_SHA384
+	bool "SHA384"
+	depends on BOOT_IMG_HASH_ALG_SHA384_ALLOW
+	help
+	  SHA384 algorithm
+
+config BOOT_IMG_HASH_ALG_SHA512
+	bool "SHA512"
+	depends on BOOT_IMG_HASH_ALG_SHA512_ALLOW
+	depends on BOOT_USE_PSA_CRYPT
```
nordicjm, line 117: this should be part of the depends on for the allow part

de-nordic: ok will move to allow.

boot/zephyr/Kconfig
de-nordic: will fix.

boot/zephyr/Kconfig
```diff
@@ -67,19 +74,68 @@ config SINGLE_APPLICATION_SLOT
...
 choice BOOT_SIGNATURE_TYPE
 	prompt "Signature type"
 	default BOOT_SIGNATURE_TYPE_RSA
 
 config BOOT_SIGNATURE_TYPE_NONE
-	bool "No signature; use only hash check"
+	bool "No signature; use only sha check"
 	select BOOT_USE_TINYCRYPT
+	select BOOT_IMG_HASH_ALG_SHA256_ALLOW
+	select BOOT_IMG_HASH_ALG_SHA512_ALLOW
```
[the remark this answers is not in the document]
nordicjm: But that's the point, this is only used either for development where you want nil security or whereby you have a project that has nil security, it doesn't really have any real world use for a product, so even if it is slower, that doesn't matter, this is working fine on an nrf51 which is a cortex m0 from over a decade ago, I don't see the speed of verifying this ever being a problem for it's intended purposes.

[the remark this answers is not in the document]
de-nordic: imgtool support is here: https://github.com/mcu-tools/mcuboot/pull/2048, it is based on older version of this PR.
MCUmgr support could be added at some point, and it could also use the sha512 coprocessor on devices that have it.

Crypto engines often have better power characteristics than doing the same hashing on CPU, and the reduction of time may be significant.
In my experimentation with 32kB image, I had following results:
 - by chunks with read to mem by flash_read - 5 to 6ms
 - by chunks, but with direct access to storage, crypto engine doing the DMA - 3ms
 - by whole image at once, crypto doing DMA - 0ms

I probably should repeat the experiment with ticks or some smaller units, but this already show improvement.

Of course, we can restrict usage of sha512 to ed25519 kinda prehashing, but still we will have sha512 TLV added.

boot/zephyr/Kconfig
de-nordic: Moved it back, I have wasted too much time fighting with conflicts in Kconfig.

boot/zephyr/Kconfig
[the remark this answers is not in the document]
nordicjm: Yes, sha512 TLV support is fine but to me no signature mode is just for development or "I don't care about security" and a few extra ms speed there is not of concern, it should be limited to sha256, then use sha512 when using a signature that needs it. The sha256 is baseline "no thrills" and everything else is good security

boot/zephyr/Kconfig
nordicjm: Not addressed

boot/zephyr/Kconfig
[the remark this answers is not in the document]
de-nordic: Fine, I will move the line to some limbo PR.

boot/bootutil/include/bootutil/crypto/sha.h
```diff
@@ -131,7 +138,6 @@ static inline int bootutil_sha_drop(bootutil_sha_context *ctx)
     (void)ctx;
     return 0;
 }
-
```
nordicjm, line 134: unrelated? Can we add it back?

boot/zephyr/Kconfig
```diff
@@ -24,6 +24,12 @@ config BOOT_USE_MBEDTLS
 	help
 	  Use mbedTLS for crypto primitives.
 
+config BOOT_USE_PSA_CRYPTO
+	bool
+	# Hidden option
+	help
+	  Use PSA crypt for supporting cryptography functions
```
nordicjm, line 31: Use same style as below e.g.
```
	help
	  Hidden option set if using PSA crypt for cryptography functionality
```

boot/zephyr/Kconfig
```diff
@@ -67,19 +73,67 @@ config SINGLE_APPLICATION_SLOT
...
+config BOOT_IMG_HASH_ALG_SHA384_ALLOW
+	bool
+	help
+	  Hidden option to set by configurations that allow SHA384
+
+config BOOT_IMG_HASH_ALG_SHA512_ALLOW
+	bool
+	depends on BOOT_USE_PSA_CRYPTO
+	help
+	  Hidden option to set by configurations that allow SHA512
```
nordicjm, line 90: Hidden option ~~to~~ set by

boot/zephyr/Kconfig
```diff
@@ -67,19 +73,67 @@ config SINGLE_APPLICATION_SLOT
...
+	  SHA512 algorithm
+
+endchoice # BOOT_IMG_HASH_ALG
+
 choice BOOT_SIGNATURE_TYPE
 	prompt "Signature type"
 	default BOOT_SIGNATURE_TYPE_RSA
 
 config BOOT_SIGNATURE_TYPE_NONE
-	bool "No signature; use only hash check"
```
nordicjm, line 75: do we need to change this? sha256 is a hash

nordicjm (commented):
Minor nit.

d3zd3z:
Do we want this or #2048?

de-nordic:
> Do we want this or #2048?

Yes, this is the verification of SHA512 the #2048 adds.

nvlsianpu (approved):

PR #1978, opened 2024-06-07
bootutil: Fix signature length passed to ED25519_verify

de-nordic:
The signature length passed to ED25519_verify, from bootutil_verify_sig, has been accidentally hardcoded, which prevented vertification of signatures different than 32 byte in length.


de-nordic:
Eh, sorry messed up here.

Issue #1976, opened 2024-06-03
imgtool.py dumpinfo crash with struct.error


dhMat:
During testing imgtool I recognized that `python imgtool.py dumpinfo ...` crashes with following error:

````
Traceback (most recent call last):
  File "C:\temp\mcuboot\scripts\imgtool.py", line 22, in <module>
    main.imgtool()
  File "C:\Program Files\Python310\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Program Files\Python310\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Program Files\Python310\lib\site-packages\click\core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Program Files\Python310\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Program Files\Python310\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\temp\mcuboot\scripts\imgtool\main.py", line 244, in dumpinfo
    dump_imginfo(imgfile, outfile, silent)
  File "C:\temp\mcuboot\scripts\imgtool\dumpinfo.py", line 90, in dump_imginfo
    _tlv_prot_head = struct.unpack(
struct.error: unpack requires a buffer of 4 bytes
````

The key used for testing was generated with following command:
`python imgtool.py keygen -k test_keys.pem -t ecdsa-p256`

The image was signed with:
` python imgtool.py sign -k test_keys.pem -v 1.0 -H 256 --pad-header -S 0x0000 "C:\temp\mcuboot_fw.hex" "C:\temp\mcuboot_fw_signed.hex"`

When checking the file with `python verify` everthing looks fine. Only when the dumpinfo command is used the program crash.

### Short investigation
When adding following code before the command that crash in dumpinfo.py", line 90:
````python
        print(b[tlv_off:(tlv_off + image.TLV_INFO_SIZE)])
        print(f"data len: {len(b)}")
        print(f"tlv_off: {tlv_off}, TLV_INFO_SIZE: {image.TLV_INFO_SIZE}, protected_tlv_size: {protected_tlv_size}")
````
following results are printed
````
b''
data len: 278946
tlv_off: 221542010, TLV_INFO_SIZE: 4, protected_tlv_size: 17975
[...]

Issue #1846, opened 2023-10-27
imgtool custom TLV disregards endian and always uses big endian


nordicjm:
This was originally (a long time ago) reported to the wrong repository: https://github.com/apache/mynewt-mcumgr/issues/152

The issue with imgtool is that the value of `--custom-tlv` is always encoded in big-endian format, one might assume that selecting the endian of the output would determine this but it does not, and according to https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/boot_status.h#L156 actually should not because all fields **must** be in little endian, therefore imgtool has a bug in this functionality.

Example:
```
imgtool sign -k /tmp/aa/bootloader/mcuboot/root-rsa-2048.pem --align 4 --version 1.2.3 -H 0x200 -S 0x78000 --custom-tlv 0xaa 0x22550011 hello_world/zephyr/zephyr.hex test.hex
```

The following is the raw bytes from the hex file (taken from JFlash): `00 AA 00 04 00  22 55 00 11`, as can be seen 0x22550011 is wrongly added in big endian format.

PR #1983, opened 2024-06-21
Add tests for scripts/imgtool

rustammendel:
Add various imgtool commands


davidvincze:
Please also update the `cryptography` dependency in scripts/setup.py:17

scripts/tests/test_keys.py
```diff
@@ -1,3 +1,6 @@
+# Copyright 2024 Denis Mingulov
```
mingulov, line 1: I am agree to add the missing copyright notice on my behalf.

davidvincze: Thank you for confirming.

rustammendel:
Commit [ded02c9](https://github.com/mcu-tools/mcuboot/pull/1983/commits/ded02c9b08d7874f303f8ed9c5fd48e6273aa679) fixes issue [1846](https://github.com/mcu-tools/mcuboot/issues/1846) and might be breaking for some users that use integer type custom TLVs.

scripts/imgtool/keys/__init__.py
```diff
@@ -58,14 +58,24 @@ def load(path, passwd=None):
     except TypeError as e:
         msg = str(e)
         if "private key is encrypted" in msg:
+            print(msg)
             return None
         raise e
-    except ValueError:
+    except ValueError as e:
+        msg1 = str(e)
         # This seems to happen if the key is a public key, let's try
```
kszromek-nordic, line 68: Please consider verifying if password is correct at first since `msg1` contains "Invalid passphrase" in it.
- cryptography 40.0.2
`ValueError: Bad decrypt. Incorrect password?` 
- cryptography 45.0.4
`ValueError: Incorrect password, could not decrypt key`

In case of public key there is no such substring in `msg1`:
- cryptography 40.0.2
`ValueError: Could not deserialize key data. The data may be in an incorrect format or it may be encrypted with an unsupported algorithm.` 
- cryptography 45.0.4
`ValueError: Valid PEM but no BEGIN/END delimiters for a private key found. Are you sure this is a private key?`

scripts/imgtool/keys/__init__.py
```diff
@@ -58,14 +58,24 @@ def load(path, passwd=None):
...
+            pk = serialization.load_pem_public_key(
                 raw_pem,
                 backend=default_backend())
+        except ValueError as e:
+            # If loading as public key also fails, that indicates wrong
+            # passphrase input
+            msg2 = str(e)
+            if ("password may be incorrect" in msg1 and
+                    "Are you sure this is a public key" in msg2):
+                raise Exception("Invalid passphrase")
```
kszromek-nordic: It can be helpful to throw some exception when loading failed with ValueError but condition in line 76 is not met.

kszromek-nordic (changes requested):
Impressive and valuable work that is a solid foundation for ensuring imgtools realiability!

Only change that is important it related to improvements in loading of keys with passphrase since it impacts tests outcomes.
Consider also adding more descriptive docstrings to test functions to better document what each test is validating.

Other comments can be addressed outside of this PR:
- extending verification of images signed with encrypted key
- adding test cases using `--sha` parameter for `sign` command (https://github.com/mcu-tools/mcuboot/commit/c894d0479442e064e3a04b27e0e1548c48ba1c9d)

scripts/tests/test_keys.py
```diff
@@ -12,242 +15,588 @@
...
+            [
+                "getpubhash",
+                "--key",
+                str(gen_key),
+                "--output",
+                str(pub_key_hash),
+                "--encoding",
+                encoding,
+            ],
+            input="invalid"
```
kszromek-nordic: Since a monkeypatch for getpass is used, no stdin simulation seems to be needed.

timonburkard:
@rustammendel thanks for your effort!

This PR is open for a long time, what do you think, would it make sense to create a separate PR for the fix in `imgtool dumpinfo` for HEX files? (44e14a31f660cbf68a70ca508c2624d361e12675)

From my understanding this resolves #1976 

This one is important for us. Like this I hope we'll be able to speed it up?

@kszromek-nordic

PR #1990, opened 2024-07-01
boot: main: avoid unused build warning

sylvioalves:
In case ESP32 SoC is used, *start will get
build warning as it is not used.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -277,7 +277,9 @@ done:
  */
 static void do_boot(struct boot_rsp *rsp)
 {
+#ifndef CONFIG_SOC_FAMILY_ESPRESSIF_ESP32
     void *start;
+#endif /* CONFIG_SOC_FAMILY_ESPRESSIF_ESP32 */
 
     BOOT_LOG_INF("br_image_off = 0x%x\n", rsp->br_image_off);
     BOOT_LOG_INF("ih_hdr_size = 0x%x\n", rsp->br_hdr->ih_hdr_size);


boot/zephyr/main.c
```diff
@@ -292,7 +290,7 @@ static void do_boot(struct boot_rsp *rsp)
     copy_img_to_SRAM(0, rsp->br_hdr->ih_hdr_size);
 
     /* Jump to entry point */
-    start = (void *)(SRAM_BASE_ADDRESS + rsp->br_hdr->ih_hdr_size);
+    void *start = (void *)(SRAM_BASE_ADDRESS + rsp->br_hdr->ih_hdr_size);
```
sylvioalves, line 293: The other approach would be using `UNUSED(start)` in ESP32 #ifdef section.

sylvioalves:
@nordicjm @almir-okato PTAL.

almir-okato (approved):
LGTM

boot/zephyr/main.c
```diff
@@ -277,8 +277,6 @@ static void copy_img_to_SRAM(int slot, unsigned int hdr_offset)
  */
 static void do_boot(struct boot_rsp *rsp)
 {
-    void *start;
```
nordicjm, line 280: variables should be declared at start of function so this needs to stay, you can #ifdef around it here

nordicjm (changes requested):

sylvioalves:
@nordicjm I swiped the call order instead of creating an additional #ifdef just because of a single variable. Let me know if it fits you.

nordicjm (commented):
this is not fixed, functions are being called before variables have been defined

sylvioalves:
@nordicjm Ok, taking your suggestion and adding #ifdef around it.

nordicjm (approved):

PR #1996, opened 2024-07-09
Revert "boot: Add MCUBOOT_HW_KEY support for image encryption"

d3zd3z:
This reverts commit 0fa46270c0b52b371a094042731c69e100478cbb.

This breaks:

    samples/synchronization/sample.kernel.synchronization on
    b_u585i_iot02a/stm32u585xx/ns error

as this TF-M configuration uses its own keys.  This change is an API change that needs to be coordinated with TF-M changes.

Before this revert, compiling this test results in:

    .../encrypted.c:447: undefined reference to
    `boot_enc_retrieve_private_key`

### DIFF
--- a/boot/bootutil/include/bootutil/enc_key.h
+++ b/boot/bootutil/include/bootutil/enc_key.h
@@ -32,7 +32,6 @@
 #include <flash_map_backend/flash_map_backend.h>
 #include "bootutil/crypto/aes_ctr.h"
 #include "bootutil/image.h"
-#include "bootutil/sign_key.h"
 #include "bootutil/enc_key_public.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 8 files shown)


nordicjm (approved):

de-nordic (approved):

d3zd3z:
@DineshDK03 just a heads up that we had to revert this, as it was breaking TF-M. If we need to get this in, we will have to figure out some way to coordinate the API change with TF-M.

d3zd3z:
BTW, a simple fix for this is to re-apply the patch, but move the definition `boot_enc_retrieve_private_key` into `encrypted.c` instead of each platform. That way, any external platforms that don't define hardware keys will just work without modification. It also reduces duplication of the code.

davidvincze:
Reapplying commit https://github.com/mcu-tools/mcuboot/commit/0fa46270c0b52b371a094042731c69e100478cbb in #2022 with the requested modifications. @d3zd3z could you please review?

PR #2002, opened 2024-07-12
boot: Add missing boot_enc_init

de-nordic:
The boot_enc_set_key in boot_swap_image, when recovering from reset, has been caled on AES context that has not been initialized.


nvlsianpu (approved):

nvlsianpu (commented):
looks like that this was missing from the beginning.

PR #2003, opened 2024-07-12
boot: Rename boot_enc_decrypt to boot_decrypt_key

de-nordic:
All of boot_enc_ function follow the same pattern where they take encryption context as the first parameter, and the boot_enc_decrypt stands out here as it does not work around the encryption context, but is rather single-part decryption function only used for decrypting of the image encryption key.

### DIFF
--- a/boot/bootutil/include/bootutil/enc_key.h
+++ b/boot/bootutil/include/bootutil/enc_key.h
@@ -48,6 +48,9 @@ struct enc_key_data {
 extern const struct bootutil_key bootutil_enc_key;
 struct boot_status;
 
+/* Decrypt random, symmetric encryption key */
+int boot_decrypt_key(const uint8_t *buf, uint8_t *enckey);
+
 int boot_enc_init(struct enc_key_data *enc_state, uint8_t slot);
 int boot_enc_drop(struct enc_key_data *enc_state, uint8_t slot);
 int boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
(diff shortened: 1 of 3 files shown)


de-nordic:
> I think this is arguably worse; as part of the encrypted images "api" all functions have names that start with `boot_enc_`, and you are breaking this "standard" naming. Either it should be called `boot_enc_decrypt_key`, which I don't see a reason for either, or you want to reuse it for something else, in which case it's better to just rename as part of the new use PR.

Putting api and standard in quotes  seems to mock them rather then making them an argument in discussion.

Worse? Nah. All except this one function take `enc_key_data` objects. That one function is completely independent from encryption context setup by all the other functions and completely setups, decrypts, and drops encryption context within its body.

And the standard somehow names the encryption function `boot_encrypt`, even though it is one taking the context, and yet stands out in naming.

For me pattern was clear: everything `boot_enc_` operates around `enc_key`.

I do not want to make everything in one PR and I am not going to reuse `boot_decrypt`, actually I have to implement `boot_enc_decrypt` and `boot_enc_encrypt` , and various others encryption/decryption only functions, as the PSA, which I am using, uses different set of functions for decryption/encryption, which also includes the process setup.

nvlsianpu (approved):
I'm fine with the change which makes function name descriptive.

> I do not want to make everything in one PR and I am not going to reuse boot_decrypt, actually I have to implement boot_enc_decrypt and boot_enc_encrypt...

Yes - this will fit to the pattern.

de-nordic:
I have marked it as DNM, because while reworking `boot_enc_load` I am now considering making both `boot_enc_load` and `boot_decrypt_key`  part of `boot_status` API.

de-nordic:
> I have marked it as DNM, because while reworking `boot_enc_load` I am now considering making both `boot_enc_load` and `boot_decrypt_key` part of `boot_status` API.

Nope, not the good idea. But the boot_enc_load is gone.

IndigoPapi85 (approved):

nvlsianpu:
@d3zd3z, @utzig last call. This is just mcuboot internal function rename l I'm going to merge this PR.

PR #2004, opened 2024-07-12
boot: Change boot_enc_load to take slot number instead of image

de-nordic:
In all cases where boot_enc_load is called it is known what slot is addressed, so it is better to just pass the slot number instead of making the boot_enc_load figure out slot number from image index and provided flash area object.

### DIFF
--- a/boot/boot_serial/src/boot_serial_encryption.c
+++ b/boot/boot_serial/src/boot_serial_encryption.c
@@ -36,7 +36,7 @@ boot_image_validate_encrypted(const struct flash_area *fa_p,
     memset(&boot_data, 0, sizeof(struct boot_loader_state));
     image_index = BOOT_CURR_IMG(state);
     if(IS_ENCRYPTED(hdr)) {
-        rc = boot_enc_load(BOOT_CURR_ENC(state), image_index, hdr, fa_p, bs);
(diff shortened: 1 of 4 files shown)


de-nordic:
Setting DNM as the function is replaced with `boot_status_load_keys` and also drops the `struct enc_key_data *enc_state` parameter and moves the `boot_status *bs` param as the primary one.
The encryption context validation and init has been moved outside.

boot/boot_serial/src/boot_serial_encryption.c
```diff
@@ -36,7 +36,7 @@ boot_image_validate_encrypted(const struct flash_area *fa_p,
     memset(&boot_data, 0, sizeof(struct boot_loader_state));
     image_index = BOOT_CURR_IMG(state);
     if(IS_ENCRYPTED(hdr)) {
-        rc = boot_enc_load(BOOT_CURR_ENC(state), image_index, hdr, fa_p, bs);
+        rc = boot_enc_load(BOOT_CURR_ENC(state), 1, hdr, fa_p, bs);
```
nvlsianpu, line 39: Note for me: I have to check whether this line is fine

nvlsianpu: is fine

nvlsianpu (approved):

PR #2005, opened 2024-07-12
boot: Remove pointless slot identification

de-nordic:
In the boot_image_validate_encrypted there was call to flash_area_id_to_multi_image_slot, which tries to figure out slot index from flash area and image index, and the result of the call was not used for anything as slot index is hardcoded in the next call to be 1 (secondary).


nvlsianpu (approved):
Serial decryption feature allow to decrypt image

PR #2006, opened 2024-07-12
boot: Make boot_enc_valid take slot instead of image index

de-nordic:
There is no point for boot_enc_valid to take image index and flash area and use these to figure out slot number.


nvlsianpu (approved):

nvlsianpu:
@de-nordic Can you fix the conflict?

d3zd3z (approved):

PR #2013, opened 2024-07-18
boot: Remove image_index from boot_encrypt

de-nordic:
boot_encrypt required the image_index paired with flash area pointer to be able to figure out which slot it will operate on. Since in most calls the slot is known in advance it can be just passed to the function directly.
The commit replaces both parameters with slot number.


nvlsianpu (approved):

nvlsianpu:
@de-nordic  Conflict,

boot/bootutil/src/image_validate.c
```diff
@@ -150,8 +150,8 @@ bootutil_img_hash(struct enc_key_data *enc_state, int image_index,
             /* Only payload is encrypted (area between header and TLVs) */
             if (off >= hdr_size && off < tlv_off) {
                 blk_off = (off - hdr_size) & 0xf;
-                boot_encrypt(enc_state, image_index, flash_area_get_id(fap), off - hdr_size,
-                        blk_sz, blk_off, tmp_buf);
+                boot_encrypt(enc_state, 1, off - hdr_size,
```
davidvincze, line 153: Could you please replace it with a variable?
`int slot = flash_area_id_to_multi_image_slot(...);` to prevent future failures if the source slot is not the BOOT_SECONDARY_SLOT?

davidvincze (changes requested):

boot/bootutil/src/image_validate.c
de-nordic: Yes.
Although it is pointless here as the MUST_DECRYPT warrants that this will be secondary image.

davidvincze: It is at the moment, but it might change in the future and that's why I think it might be better to calculate it from the input parameters rather than using a hardcoded value. Not crucial but either way it should be `BOOT_SECONDARY_SLOT` instead of `1`. Could you please update it?

davidvincze: Thank you.

de-nordic: Another thing we should do: somehow embed slot/image info into flash area objects so that we would not have this stupid translation functions for something we know while creating the objects.

davidvincze (approved):

de-nordic:
@davidvincze Got some glitch in github where your ack appeared as request for changes.

davidvincze (approved):

davidvincze:
> @davidvincze Got some glitch in github where your ack appeared as request for changes.

I hope it's okay now :)

PR #2019, opened 2024-07-25
zephyr: Add support for automatically calculcating max sectors

nordicjm:


nordicjm:
How it looks on a device with large (different) sector sizes (stm32f411e_disco):
```
CMake Warning at CMakeLists.txt:392 (message):
  Unable to determine erase size of slot0 partition, cannot calculate minimum
  sector usage


CMake Warning at CMakeLists.txt:407 (message):
  Unable to determine erase size of slot1 partition, cannot calculate minimum
  sector usage


CMake Warning at CMakeLists.txt:421 (message):
  Unable to calculate minimum number of sector sizes, falling back to 128
  sector default.  Please disable CONFIG_BOOT_MAX_IMG_SECTORS_AUTO and set
  CONFIG_BOOT_MAX_IMG_SECTORS to the required value
```
It builds and default remains at 128

nordicjm:
@butok interestingly it looks like the nxp ones were a bit optimistic...
`boards/mimxrt595_evk_mimxrt595s_cm33.conf:CONFIG_BOOT_MAX_IMG_SECTORS=8192`
`cmake -GNinja -DBOARD=mimxrt595_evk/mimxrt595s/cm33 ..`:
> Calculated maximum number of sectors: 866

butok:
> @butok interestingly it looks like the nxp ones were a bit optimistic... `boards/mimxrt595_evk_mimxrt595s_cm33.conf:CONFIG_BOOT_MAX_IMG_SECTORS=8192` `cmake -GNinja -DBOARD=mimxrt595_evk/mimxrt595s/cm33 ..`:
> 
> > Calculated maximum number of sectors: 866

Is it bad or good?

butok:
> > > @butok interestingly it looks like the nxp ones were a bit optimistic... `boards/mimxrt595_evk_mimxrt595s_cm33.conf:CONFIG_BOOT_MAX_IMG_SECTORS=8192` `cmake -GNinja -DBOARD=mimxrt595_evk/mimxrt595s/cm33 ..`:
> > > > Calculated maximum number of sectors: 866
> > 
> > 
> > Is it bad or good?
> 
> Updated the comment just as you posted:
> 
> > This means 0xc000 is needed to store the update data but you have actually reserved 0x62000...
> 
> That's 344KiB of wasted space

The board has 64MB, so the configuration value was set to have a bigger coverage.
BUT when you will add this automatic calculation, we can remove all CONFIG_BOOT_MAX_IMG_SECTORS from board configs.

boot/zephyr/CMakeLists.txt
```diff
@@ -363,23 +363,66 @@ if(CONFIG_MCUBOOT_BOOT_BANNER)
...
+  string(SUBSTRING "${${node}}" 0 ${pos} ${node})
+  set(${node} "${${node}}" PARENT_SCOPE)
+endfunction()
+
+if(CONFIG_BOOT_MAX_IMG_SECTORS_AUTO)
+  dt_nodelabel(slot0_flash NODELABEL "slot0_partition")
+  dt_prop(slot0_size PATH "${slot0_flash}" PROPERTY "reg" INDEX 1)
+  dt_get_parent(slot0_flash)
+  dt_get_parent(slot0_flash)
+  dt_prop(erase_size_slot0 PATH "${slot0_flash}" PROPERTY "erase-block-size")
```
de-nordic, line 387: I kinda do not like that we get to the Flash device layer to get this information as I think that we start to pick information that we should not access directly from scripts, but unless the DTS parsing script does not start definitions for info like number of pages per partitions this is the only way. Still this makes changes to dts harder in the future.

de-nordic (approved):

boot/zephyr/CMakeLists.txt
d3zd3z: The other problem is that I don't believe the code cares what this property is even set to. The flash api directly asks the driver about the erase block size. And that API is needlessly complicated to allow for devices with variable sized sectors, even though the only reasonable thing to do with them in mcuboot is treat the device as if all of the sectors were the size of the largest one.

d3zd3z: What this means is that if this property is at a mismatch with the underlying device, the calculation will be wrong. For example, people have reported trying to increase the erase size in the DT, to discover it didn't work, as the code still asked the device what the real sector size is.

I think we should look into changing things so that mcuboot uses this property _only_ to determine the erase-block-size. This will make it easier to support multiple devices by just setting the value to the largest size.

nordicjm: For variably sized sectors, like some stm32 parts, that property is not set and this will give the warning then use the existing default of 128 instead, I did give it a try

PR #2028, opened 2024-07-31
bootutil: Add compressed image flags and TLV, fix misc. bugs in bootutil

nordicjm:
Adds some flags to indicate if the data of an image is compressed (lzma1 and lzma2) and adds a new TLV for a sha256 hash of the decompressed/decrypted image data, this would (optionally) be used for images where one or both options are enabled to be able to verify the data is correct and that is can be written/booted successfully

Note: no code additions relating to actual compression are present here, just addition of TLVs and flags

Also includes some bug fixes to bootutil's image header checks.

Requires https://github.com/mcu-tools/mcuboot/pull/2045 be merged


michalek-no (approved):

boot/bootutil/include/bootutil/image.h
```diff
@@ -89,6 +95,11 @@ struct flash_area;
 #define IMAGE_TLV_PUBKEY            0x02   /* public key */
 #define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
 #define IMAGE_TLV_SHA384            0x11   /* SHA384 of image hdr and body */
+#define IMAGE_TLV_RAW_SLOT_SHA256   0x18   /*
+                                            * SHA256 of compressed/encrypted image data in the slot
+                                            * which should be checked to ensure image integrity
+                                            * before decrypting/decompressing data
```
de-nordic, line 101: Isn't this little confusing? In case when image is encrypted actually of the other SHA TLV are used but the image is first decrypted and then sha is calculated on the raw image.
So if only the existing encryption is used the other TLV serve as the RAW sha.
Shouldn't we have opposite?

Currently what happens is the `bootutil_img_validate` will first call `bootutil_img_hash` that will decrypt image into RAM and calculate the sha with pre-selected, during compilation, sha algorithm and then compare the resulting sha with expected TLV (EXPECTED_HASH_TLV ), which would be the IMAGE_TLV_SHA256 by default.

This also means that if we validate slot, that has image already decrypted and decompressed we will check it against EXPECTED_HASH_TLV  stored sha, which currently serves as the "RAW" sha.

nordicjm:
Flags:
`IMAGE_F_COMPRESSED_LZMA1` = image is compressed, using lzma1 algorithm
`IMAGE_F_COMPRESSED_LZMA2` = image is compressed, using lzma2 algorithm

TLVs:
`IMAGE_TLV_RAW_SLOT_SHA256` = sha256 of the data in the slot, required for compressed data, optional for encrypted data, this allows for verifying that the image data is correct and not corrupted. Should be placed in the protected image region.
`IMAGE_TLV_COMP_SIZE` = size of the compressed data in the slot

The image header field for the image size should be the full decompressed size, and the hash there should be the hash of the decompressed data, since the only current supported route will be to load a compressed update to the secondary slot and decompress it into the primary slot, the alternative fields can be checked when there is a compressed image present, then ignored once that has been decompressed into the primary slot as the original fields will be correct

boot/bootutil/include/bootutil/image.h
```diff
@@ -101,6 +112,7 @@ struct flash_area;
 #define IMAGE_TLV_DEPENDENCY        0x40   /* Image depends on other image */
 #define IMAGE_TLV_SEC_CNT           0x50   /* security counter */
 #define IMAGE_TLV_BOOT_RECORD       0x60   /* measured boot record */
+#define IMAGE_TLV_COMP_SIZE         0x70   /* Compressed image size */
```
d3zd3z, line 115: Additionally, it is a little unclear what the sizes mean. Does this mean the has the uncompressed size? How would we find the TLV if the actual data stored is different than the size in the header?

nordicjm: For some reason I thought it went through looking for the magic bytes, apparently not... Will have a rethink

nordicjm: Have updated the PR to instead have 3 TLVs, the hash and signature would be equivalent to the compressed image fields but on the decompressed image data instead (this is so that the data can be validated before decompressing it)

nvlsianpu (approved):

de-nordic (approved):

nvlsianpu (approved):
I'm fine with additional commits.
@d3zd3z Can you give us a review?

nvlsianpu:
@davidvincze Can you review this?

d3zd3z (approved):

PR #2029, opened 2024-08-02
RFC: bootutil: Add TLV for  Pure (over image) signature

de-nordic:
The commit reserves IMAGE_TLV_SIG_PURE == 0x25 as indicator that the signature attached to image is calculated over image rather than its digest (pure, by the name of PureEdDSA). The indicator/flag is provided so that no special TLV is designated to one signature type, because, in theory, any signature algorithm that can be used over image rather than digest can be marked as "Pure".

The PureEdDSA ED25519 signature should be done directly over message, so image in our case. The current implementation is not really PureEdDSA nor HashEdDSA, as we basically pass sha256 as a message to PureEdDSA. ~~Not perfect.~~

**Why is this added?**
This PR adds TLV that indicates that PureEdDSA has been run over image rather than sha has been calculated and passed to ed25519 for signature/verification.

**Problems**
This only works with devices that are mapping flash to RAM address space, so that image signature verification can directly access whole image as message.
This means that configuration with external devices may not work.

Note also that signatures PureEdDSA(image) != PureEdDSA(sha512(image)) != HashEdDSA(sha512(image)), so there is no realu way to use multi-part signature verification with PureEdDSA.

The https://github.com/mcu-tools/mcuboot/pull/2048 is implementation where sha512 is used with PureEdDSA ~~, not perfect.~~

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -101,6 +101,9 @@ struct flash_area;
 #define IMAGE_TLV_ECDSA_SIG         0x22   /* ECDSA of hash output */
 #define IMAGE_TLV_RSA3072_PSS       0x23   /* RSA3072 of hash output */
 #define IMAGE_TLV_ED25519           0x24   /* ed25519 of hash output */
+#define IMAGE_TLV_SIG_PURE          0x25   /* Indicator that attached signature has been prepared
+                                            * over image rather than its digest.
+                                            */
 #define IMAGE_TLV_ENC_RSA2048       0x30   /* Key encrypted with RSA-OAEP-2048 */
 #define IMAGE_TLV_ENC_KW            0x31   /* Key encrypted with AES-KW 128 or 256*/
 #define IMAGE_TLV_ENC_EC256         0x32   /* Key encrypted with ECIES-EC256 */
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -69,6 +69,7 @@ TLV_VALUES = {
         'ECDSASIG': 0x22,
         'RSA3072': 0x23,
         'ED25519': 0x24,
+        'SIG_PURE': 0x25,
         'ENCRSA2048': 0x30,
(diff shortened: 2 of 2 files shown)


nvlsianpu (commented):
I'm fine with `_PURE` ,
think `_SHA512` ending would be fine as well.

d3zd3z:
The point at the beginning of the description is unclear to me. You state that we are computing a hash of the image, and then using PureEdDSA, and then state that is somehow different from HashEdDSA. I believe that _is_ exactly what HashEdDSA is. You hash the image first, and then use PureEdDSA to sign that hash.

What is your motivation for implementing PureEdDSA. Specifically what cryptographic advantage do we gain here. As far as I can tell, this is to avoid collision resistance of the hash function causing problems with the signature. But we need to be clear here, this is not about natural collision resistance of the hash function (SHA512 would reach a significant collision probability after `2^256` different signatures, this is of a similar order to the number of hydrogen atoms in the visible universe). This is about protecting against vulnerabilities found in the hash function where it becomes possible to generate two documents with the same hash without being able to generate an arbitrary hash. This, for example, is how MD5 failed. However, it wasn't long after than the the robustness of the algorithm collapsed to the point that it was completely vulnerable to even making arbitrary documents with a given hash.

It's arguable that this isn't really the realm of the signature algorithm, and that if a hash function does become vulnerable in this way that we should move away from it as quickly as possible, since if MD5 is any indicator, it is likely to collapse entirely.

But given that most implementations of EdDSA don't seem to be able to work without containing the entire image to be hashed in memory, I see little reason to add this complexity to MCUboot. Even the two-images with the same hash attack this protects against would be difficult to apply here (the same party has to generate both images). It would allow say an independent developer to submit an image for signing to a signing authority, and then deploy a different one with the same signature. However, there isn't any indicator that SHA512 is vulnerable to this, and if it were, there are enough other things that fall apart, it is likely that numerous other means could be used to exploit it as well.

nordicjm (approved):

nvlsianpu (approved):

PR #2041, opened 2024-08-20
boot_serial: Fix `incompatible-pointer-types` and `format` warnings

lluiscampos:
Clears a `format` and an `incompatible-pointer-types` warnings when building on 64 bits.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -1283,7 +1283,7 @@ boot_serial_output(void)
 static int
 boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 {
-    int rc;
+    size_t rc;
     uint16_t crc;
     uint16_t len;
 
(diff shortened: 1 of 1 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -1289,7 +1289,7 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 
 #ifdef __ZEPHYR__
     int err;
-    err = base64_decode( &out[*out_off], maxout - *out_off, &rc, in, inlen - 2);
+    err = base64_decode( &out[*out_off], maxout - *out_off, (size_t *)&rc, in, inlen - 2);
```
nordicjm, line 1292: no... absolutely not, if you get a warning here that's because the size of the variables do not match, and here you are pretending your 4 byte variable is actually 8 bytes, trumping whatever stack memory is around it.

nordicjm (changes requested):

boot/boot_serial/src/boot_serial.c
lluiscampos: But `base64_decode` expects a `size_t * `, and it is done the exact same way few lines below for `__ESPRESSIF__` code. Should we instead change the declaration of `rc` to be `size_t`?

nordicjm: yes, it needs to match. If the code below pretends a 4 byte variable is an 8 byte variable then the code below is horribly wrong

lluiscampos: Thank you, I will rework.

lluiscampos:
Hi @nordicjm please have a second look.

I now have split the work in two different commits, and for the critical one that we discussed about I added more context into the commit message.

nordicjm (approved):

PR #2051, opened 2024-08-28
ci: Update TF-M version to eb8ff0d

davidvincze:
Update TF-M version to keep up with code development. Fixes and new features have been introduced on both sides, using a more recent TF-M version helps to detect regression.


nordicjm:
@davidvincze any clue as to what's going on with the FIH failures? Seeing them on main CI too and other PRs

nordicjm (approved):

PR #2055, opened 2024-09-05
boot: Support bootstrapping for multiple images

KarlKTorp:
Bootstrapping is currently not allowed for more than one image. But that seems to be unintentional according to https://github.com/mcu-tools/mcuboot/pull/826#issuecomment-707939624

I ran into the same issue trying to boot two images (for M7 and M4) on a i.MX RT1170 using zephyr. We solved this using the suggested change and hope it could be added to main.

With this it's possible to enable MCUmgr SMP within MCUBoot and update a device that only has the bootloader in flash.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1960,8 +1960,7 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
         rc = boot_read_image_headers(state, !boot_status_is_reset(bs), bs);
 #ifdef MCUBOOT_BOOTSTRAP
         /* When bootstrapping it's OK to not have image magic in the primary slot */
-        if (rc != 0 && (BOOT_CURR_IMG(state) != BOOT_PRIMARY_SLOT ||
-                boot_check_header_erased(state, BOOT_PRIMARY_SLOT) != 0)) {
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

nordicjm:
@KarlKTorp can you fix the commit sign off?

KarlKTorp:
> @KarlKTorp can you fix the commit sign off?

@nordicjm I'm not sure why it does not pass. I do have the sign-off line added. Is something else missing in the commit message?

KarlKTorp:
@nordicjm I have tried to add a description line in the commit message as I can see this is common for other commits. Can we check if that solves the issue?

KarlKTorp:
I have tried to change the sign-off from 
Signed-off-by: Karl Kristian Dyrholm Torp <kdt@polytech.com>
to
Signed-off-by: KarlKTorp <kdt@polytech.com>
locally and running check-signed-off-by.sh. That seems to satisfy the script, but I'm not sure if that is the correct way to fix this?

nordicjm:
No that is not right, follow https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git to set your name correct in git then do `git commit --amend --reset-author`

KarlKTorp:
@nordicjm I think it should be working now. The local check was OK.

de-nordic (approved):

PR #2059, opened 2024-09-10
boot: zephyr: Add check for unexpected flash sector size

nordicjm:
Prints a debug log message if the device has a write block size for a flash device in DTS that is not the same as what the flash driver reports at run-time, this can be used to see if there is a faulty configuration as these compile-time values are used for various calculations

@butok


de-nordic (approved):

butok:
It should also check the program size (not only the sector size).

nordicjm:
> It should also check the program size (not only the sector size).

That's already done at runtime

nordicjm:
> This does not support configurations with external mem, right? Because with external mem there is no `write-block-size`/`erase-block-size`.

It does, but it is driver dependent, for example one of the nxp ones had a write block size of 8 in dts which was wrong, and something like spi-nor might have a value in dts if it uses static configuration or nothing if it uses sfdp

de-nordic:
> > This does not support configurations with external mem, right? Because with external mem there is no `write-block-size`/`erase-block-size`.
> 
> It does, but it is driver dependent, for example one of the nxp ones had a write block size of 8 in dts which was wrong, and something like spi-nor might have a value in dts if it uses static configuration or nothing if it uses sfdp

Bindings for spi-nor and qspi-nor does not provide definition of write-block-size, it is provided via Kconfig option, so only via simplified or run-time sfdp. There is though `CONFIG_SPI_NOR_FLASH_LAYOUT_PAGE_SIZE` that specifies erase page size for SPI... does not pair well with multi-instance spi-nor driver.

nordicjm:
> > > This does not support configurations with external mem, right? Because with external mem there is no `write-block-size`/`erase-block-size`.
> > 
> > 
> > It does, but it is driver dependent, for example one of the nxp ones had a write block size of 8 in dts which was wrong, and something like spi-nor might have a value in dts if it uses static configuration or nothing if it uses sfdp
> 
> Bindings for spi-nor and qspi-nor does not provide definition of write-block-size, it is provided via Kconfig option, so only via simplified or run-time sfdp. There is though `CONFIG_SPI_NOR_FLASH_LAYOUT_PAGE_SIZE` that specifies erase page size for SPI... does not pair well with multi-instance spi-nor driver.

Bad name, but this spi nor driver is what I mean: https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/nxp/mimxrt1020_evk/mimxrt1020_evk.dts#L104

de-nordic:
> 
> Bad name, but this spi nor driver is what I mean: https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/nxp/mimxrt1020_evk/mimxrt1020_evk.dts#L104

OK, that is new to me as there is not such definition in dts bindings for the compatible.

erwango:
This change broke mcuboot usage on STM32F2, STM32F4 and STM32F7 series in which use sectors of variable sizes. On these series, erase-block-size depends on sector size and then cannot be defined in flash controller parent.

erwango:
> This change specifically?

Ok, fair enough. Let me reword: This change is not compatible with _bla bla bla_

butok:
Probably, CONFIG_BOOT_MAX_IMG_SECTORS_AUTO=n should be set in STM32F2, STM32F4 and STM32F7 configuration files.

PR #2062, opened 2024-09-12
Make RAM loading available for single loaders

edersondisouza:
Code for loading the image to RAM is currently under `bootutil/loader.c`. However, single loaders (used when one sets `MCUBOOT_SINGLE_APPLICATION_SLOT`) cannot use this file - a single loader in fact re-implements some of the methods of `bootutil/loader.c`.
This PR splits the bits of code related to RAM loading to another file inside bootutil (`ram_load.c`) and make them available on `bootutil_priv.h`, so that single loaders can simply reuse this code.

Note: these are basically the first four patches from #2044 - it seems that the multiple source bits of that PR are a bit controversial, so this PR focus on the RAM loading availability, which is hopefully less controversial.


boot/zephyr/Kconfig
```diff
@@ -269,20 +261,13 @@ config BOOT_FIRMWARE_LOADER
...
-
-if BOOT_RAM_LOAD
-config BOOT_IMAGE_EXECUTABLE_RAM_START
-	hex "Boot image executable ram start"
-	default $(dt_chosen_reg_addr_hex,$(DT_CHOSEN_Z_SRAM))
+config BOOT_NO_UPGRADE
+	bool "No upgrade"
+	help
+	  No upgrade is performed - usually, this option will be used in
+	  conjunction with BOOT_RAM_LOAD.
```
nordicjm, line 268: I don't know what you're doing here but no, this Kconfig is not needed and put the ram load Kconfig back where it was

nordicjm (changes requested):

nordicjm:
> Code for loading the image to RAM is currently under bootutil/loader.c. However, single loaders (used when one sets MCUBOOT_SINGLE_APPLICATION_SLOT) cannot use this file - a single loader in fact re-implements some of the methods of bootutil/loader.c.

Because single application slot has nothing to do with RAM load mode, they are two completely different modes, RAM load is it's own dual-slot system, single application image is a single flash image. If you want to add a single RAM slot mode then that's something that can be added as a distinct mode, not as modifying single application slot

edersondisouza:
v2:
  - Have a separate Kconfig for RAM loading on single application slot

edersondisouza:
> > Code for loading the image to RAM is currently under bootutil/loader.c. However, single loaders (used when one sets MCUBOOT_SINGLE_APPLICATION_SLOT) cannot use this file - a single loader in fact re-implements some of the methods of bootutil/loader.c.
> 
> Because single application slot has nothing to do with RAM load mode, they are two completely different modes, RAM load is it's own dual-slot system, single application image is a single flash image. If you want to add a single RAM slot mode then that's something that can be added as a distinct mode, not as modifying single application slot

Ok - using a separate Kconfig instead of reusing the choice under !SINGLE_APPLICATION_SLOT.

boot/zephyr/Kconfig
```diff
@@ -320,6 +307,31 @@ config BOOT_SWAP_SAVE_ENCTLV
...
+
+config SINGLE_APPLICATION_SLOT_RAM_LOAD
+	bool "RAM load for single application slot"
+	help
+	  If y, the image is loaded to RAM and executed from there. For this reason,
+	  the image has to be linked to be executed from RAM. The address that the
+	  image is copied to is specified using the load-addr argument to the
+	  imgtool.py script which writes it to the image header.
+
+endif # SINGLE_APPLICATION_SLOT
```
nordicjm, line 320: this needs to be a distinct mode, not one that uses `SINGLE_APPLICATION_SLOT`, that Kconfig is as I said in the previous review for single flash application only, there needs to be a new Kconfig, one that does not use an existing Kconfig

edersondisouza: Ok - moved the Kconfig outside `if SINGLE_APPLICATION_SLOT`, so one doesn't need something like: 

```
CONFIG_SINGLE_APPLICATION_SLOT=y
CONFIG_SINGLE_APPLICATION_SLOT_RAM_LOAD=y
```

Only `CONFIG_SINGLE_APPLICATION_SLOT_RAM_LOAD=y` is needed. Implementation-wise though, it will select `SINGLE_APPLICATION_SLOT` as the code path is basically the same.

edersondisouza:
v3:
  - `CONFIG_SINGLE_APPLICATION_SLOT_RAM_LOAD` is not behind `CONFIG_SINGLE_APPLICATION_SLOT`.

boot/zephyr/Kconfig
```diff
@@ -320,6 +307,28 @@ config BOOT_SWAP_SAVE_ENCTLV
 
 endif # !SINGLE_APPLICATION_SLOT
 
+config SINGLE_APPLICATION_SLOT_RAM_LOAD
+	bool "RAM load for single application slot"
+	select SINGLE_APPLICATION_SLOT
```
nordicjm, line 312:
```suggestion
```

This needs to be a distinct mode, it can use the same .c file and check the ifdefs but this Kconfig needs to be its own Kconfig and not do anything with other MCUboot mode Kconfigs

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -108,6 +108,12 @@
 
 #endif /* CONFIG_SINGLE_APPLICATION_SLOT */
 
+#ifdef CONFIG_SINGLE_APPLICATION_SLOT_RAM_LOAD
+#define MCUBOOT_RAM_LOAD 1
```
nordicjm, line 112: And here, add a new define for this mode and add it to the base .c files that needs it to enable the required functions

edersondisouza:
v4:
  - `CONFIG_SINGLE_APPLICATION_SLOT_RAM_LOAD` doesn't select `CONFIG_SINGLE_APPLICATION_SLOT` anymore. This means that there a bit more widespread changes, though.

boot/bootutil/src/boot_record.c
```diff
@@ -249,6 +249,8 @@ int boot_save_shared_data(const struct image_header *hdr, const struct flash_are
 #else
     uint8_t mode = MCUBOOT_MODE_DIRECT_XIP;
 #endif
+#elif defined(MCUBOOT_SINGLE_APPLICATION_SLOT_RAM_LOAD)
+    uint8_t mode = MCUBOOT_MODE_SINGLE_SLOT_RAM_LOAD;
```
nordicjm, line 253: keep order and move after firmware loader

nordicjm (commented):
Minor nit., changes are OK

boot/bootutil/src/boot_record.c
teburd: @edersondisouza is on his vacation so I've taken over, fixed this.

nordicjm (approved):

nordicjm:
@teburd Needs a signed off line from you

teburd:
> @teburd Needs a signed off line from you

Done

nordicjm:
@teburd CI still failing on sign off

nvlsianpu (changes requested):
CI issues need to be fixed

teburd:
@nvlsianpu any updates? Seems I'll need a rebase now but would like feedback that this is good to go

teburd:
Latest rebase should fix the conflict that was introduced, the boot_enc_decrypt rename cause the rebase to show a conflict and that was fixed.

teburd:
> CI issues need to be fixed

They’ve been fixed again, please do review again when you xan

nvlsianpu (approved):

PR #2063, opened 2024-09-13
imgtool: Add --pure switch to allow calculating signatures over entire images

de-nordic:
The PR adds --pure switch to imgtool that allows to run ED25519 over entire image calculating the PureEdDSA signature.

The imgtool will add SIG_PURE TLV as indicator that whatever signature is used is calculated directly over image.
The PR also provides different output for signature verification, where there is signature reported on verification, rather than digest ~~as digest is no longer available~~, but will provide digest if key is not provided during verification.

Depends on https://github.com/mcu-tools/mcuboot/pull/2029 -- that reserves the SIG_PURE in MCUboot code
Based on https://github.com/mcu-tools/mcuboot/pull/2048

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -190,7 +190,15 @@ ALLOWED_KEY_SHA = {
     keys.X25519             : ['256', '512']
 }
 
-def key_and_user_sha_to_alg_and_tlv(key, user_sha):
+ALLOWED_PURE_KEY_SHA = {
+    keys.Ed25519            : ['512']
+}
+
+ALLOWED_PURE_SIG_TLVS = [
+    TLV_VALUES['ED25519']
+]
+
+def key_and_user_sha_to_alg_and_tlv(key, user_sha, is_pure = False):
     """Matches key and user requested sha to sha alogrithm and TLV name.
 
        The returned tuple will contain hash functions and TVL name.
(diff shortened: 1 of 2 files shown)


de-nordic:
May be interesting for you @taltenbach

nvlsianpu (approved):

scripts/imgtool/image.py
```diff
@@ -585,12 +600,18 @@ def create(self, key, public_key_format, enckey, dependencies=None,
...
-        tlv.add(hash_tlv, digest)
-        self.image_hash = digest
+        if not is_pure:
+            sha = hash_algorithm()
+            sha.update(self.payload)
+            digest = sha.digest()
+            message = digest
+            tlv.add(hash_tlv, digest)
+        else:
+            # Note that when Pure signature is used, hash TLV is not present.
```
de-nordic, line 610: @nordicjm @nvlsianpu I am bypassing addition of hash TLV in case when signature is pure; pure signature is not using it, but maybe the hash may be useful for something else is worth leaving it as it is?

nvlsianpu: @de-nordic Yes, it might be usefull: Hash is for check the integrity of the image, while the signature is for checking authenticity of the image. Usually HASH calculation is faster and doesn't require any key. I can expect that some cases might require just integrity calculation.

de-nordic: Now the TLV sha is there, planning, for the future - not here, to add switch to not include it.

nvlsianpu:
@de-nordic I would merge that. Any obstacles?

scripts/imgtool/image.py
```diff
@@ -588,9 +601,16 @@ def create(self, key, public_key_format, enckey, dependencies=None,
         sha = hash_algorithm()
         sha.update(self.payload)
         digest = sha.digest()
-        message = digest;
         tlv.add(hash_tlv, digest)
-        self.image_hash = digest
```
nvlsianpu, line 593: self.image_hash is used externally

nvlsianpu:
@de-nordic Should we proceed with this patch?

nordicjm:
@d3zd3z can you review?

scripts/imgtool/image.py
nvlsianpu: ^^

nvlsianpu (changes requested):

nvlsianpu (approved):

de-nordic (commented):
Wrong PR.

PR #2070, opened 2024-09-21
boot: zephyr: Allow disabling NRFX_WDT on nRF devices

joerchan:
Allow disabling feeding the watchdog on nrf devices directly using the nrfx_wdt functions.
The imply on NRFX_WDT cannot be disabled since NRFX_WDT is a promptless symbol and cannot be assigned in a config fragment. Not even deleting the wdt0 node in a devicetree overlay works since the configuration takes predence and ends in a #error without any nrfx_wdt instances.

To fix this create new Kconfig option with prompt and change the imply to this option. That way the option can be assigned to off by the user.

This is needed to support the use case of an external watchdog controlled using the zephyr watchdog API instead of the wdt peripheral.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -736,15 +736,20 @@ config BOOT_WATCHDOG_FEED
 	bool "Feed the watchdog while doing swap"
 	default y if WATCHDOG
 	default y if SOC_FAMILY_NORDIC_NRF
-	# for nRF nrfx based implementation is available
-	imply NRFX_WDT if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT0 if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT1 if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT30 if SOC_FAMILY_NORDIC_NRF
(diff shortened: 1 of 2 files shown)


boot/zephyr/Kconfig
```diff
@@ -736,15 +736,17 @@ config BOOT_WATCHDOG_FEED
...
+	depends on SOC_FAMILY_NORDIC_NRF
 	# for nRF nrfx based implementation is available
-	imply NRFX_WDT if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT0 if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT1 if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT30 if SOC_FAMILY_NORDIC_NRF
-	imply NRFX_WDT31 if SOC_FAMILY_NORDIC_NRF
-	help
-	  Enables implementation of MCUBOOT_WATCHDOG_FEED() macro which is
-	  used to feed watchdog while doing time consuming operations.
```
nordicjm, line 747: would move the comment up to the above Kconfig symbol

nordicjm (approved):

de-nordic (approved):

PR #2072, opened 2024-09-24
ci: Add issue and PR assigner

nordicjm:


nordicjm:
@utzig getting a different error here now saying bad credentials, was write access given to issue and PR on the fine grained access token?

nordicjm:
> > @utzig getting a different error here now saying bad credentials, was write access given to issue and PR on the fine grained access token?
> 
> I thought it could be because you're not an admin, but it failed the same way for me (#2074). I am somewhat suspicious that it will just work once merged. Anyway you should fix up the last commit. It looks OK to me, I think it's fine to merge for testing, and disable if it still not working. If I remember correctly when I added the `TWINE_TOKEN` for pushing `imgtool` I had the same issue.

Removed test commit

PR #2073, opened 2024-09-24
zephyr/Kconfig.serial_recovery: limit Slot info command

nvlsianpu:
BOOT_SERIAL_IMG_GRP_SLOT_INFO should be not available for direct-xip and ram-load modes.

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -204,7 +204,8 @@ config BOOT_SERIAL_IMG_GRP_IMAGE_STATE
 
 config BOOT_SERIAL_IMG_GRP_SLOT_INFO
 	bool "Slot info"
-	default y if UPDATEABLE_IMAGE_NUMBER > 1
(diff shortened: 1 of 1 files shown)


boot/zephyr/Kconfig.serial_recovery
```diff
@@ -204,7 +204,8 @@ config BOOT_SERIAL_IMG_GRP_IMAGE_STATE
 
 config BOOT_SERIAL_IMG_GRP_SLOT_INFO
 	bool "Slot info"
-	default y if UPDATEABLE_IMAGE_NUMBER > 1
+	depends on !BOOT_DIRECT_XIP && !BOOT_RAM_LOAD
+	default y if (UPDATEABLE_IMAGE_NUMBER > 1)
```
nordicjm, line 208: remove brackets so it matches the upstream data without adding a new change

nordicjm (changes requested):
Proper fix: https://github.com/mcu-tools/mcuboot/pull/2075

PR #2075, opened 2024-09-25
boot: bootutil: loader: Fix slot info for directXIP/RAM load

nordicjm:
Fixes an issue when either of these modes is used with serial recovery slot info enabled


de-nordic (approved):

boot/boot_serial/src/boot_serial.c
```diff
@@ -275,7 +275,7 @@ bs_list(char *buf, int len)
         int swap_status = boot_swap_type_multi(image_index);
 #endif
 
-        for (slot = 0; slot < MCUBOOT_IMAGE_NUMBER; slot++) {
+        for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
```
nvlsianpu, line 278: This is proper fix.
I just checked that BOOT_NUM_SLOTS is defined to be `2`. Looks like this is not true if we came to single bank.

nvlsianpu (approved):

PR #2077, opened 2024-09-27
boot: bootutil: loader: Fix issue with using pointers

nordicjm:
Fixes an issue whereby static buffers were changed into pointers, whereby they are then assumed to be the size of a pointer rather than the size of the actual buffers


de-nordic (approved):

nvlsianpu (approved):
fix and simplification at once.

PR #2080, opened 2024-10-02
PureEdDSA and support for calculating SHA directly on device

de-nordic:
Note on latest change: I have divided commits into bootutil and Zephyr themed so that any problem with Zephyr would not immediately break the feature PR.

Six commits:
 1. Adds support for sha512 when TinyCrypt is used
 1. adds PureEdDSA where image signature is calculated over image directly, rather than hash
 1. adds support for calculating sha directly on storage, rather than loading chunks to memory
 1. Zephyr Kconfig change that enables SHA512 with tinycrypt
 1. Zephyr Kconfig change that enables SHA directly on storage
 1. Zephyr Kconfig change that enables PureEdDSA

Note that mentioned below Kconfig options have proper MCUBOOT_ identifiers in code, but it is easier to communicate change by using them in examples.

Both commits provide support only on devices that are within address space.
There is possibility to extend both commits to work with external devices via features like XIP, where external device can be mapped to internal address space, but this is not approached here.

The PureEdDSA change adds TLV that marks the image as signed with Pure signature and logic in MCUboot that rejects images without the TLV, when expected by MCUboot. The change also slightly improves security of devices, specifically if the verification can be done by dedicated coprocessor.

Both solutions slightly improve performance as they cut on additional processing.

The PureEdDSA commit requires imgtool change: https://github.com/mcu-tools/mcuboot/pull/2063

Notes on stats:
 - In case when the sha512 is enabled with ed25519, the size of ROM, as build for the nrf52840dk, drops from 40486 to 39414, but there is visible slowdown in boot; for the sha256 the command line was:
 ```
west build -d builds/sha_on_board_256 -b nrf52840dk/nrf52840 bootloader/mcuboot/boot/zephyr/ -DCONFIG_BOOT_SIGNATURE_TYPE_ED25519=y -DCONFIG_BOOT_ED25519_TINYCRYPT=y -DCONFIG_BOOT_IMG_HASH_ALG_SHA256=y
```
for the 512, the change was in the last config:
```
west build -d builds/sha_on_board_512 -b nrf52840dk/nrf52840 bootloader/mcuboot/boot/zephyr/ -DCONFIG_BOOT_SIGNATURE_TYPE_ED25519=y -DCONFIG_BOOT_ED25519_TINYCRYPT=y -DCONFIG_BOOT_IMG_HASH_ALG_SHA512=y
```

Enabling the option `CONFIG_MCUBOOT_HASH_STORAGE_DIRECTLY`, that comes with second commit, reduces code by another ~50 bytes.

Enabling the `CONFIG_BOOT_SIGNATURE_TYPE_PURE`, provided by the third commit, shaves off additional ~150 bytes.

While SHA directly on device and in RAM works with the same image signed for specified sha, the PureEdDSA requires special signature processing via imgtool and lacks HASH TLV.

### CODE REVIEW

**Comment by de-nordic** [comment]
@taltenbach FYI

**Comment by nordicjm** [comment]
rebase needed

**Comment by de-nordic** [comment]
> rebase needed

Done

File: boot/zephyr/Kconfig [hunk 1]
```diff
@@ -88,6 +88,22 @@ config BOOT_IMG_HASH_ALG_SHA512_ALLOW
...
+	depends on !BOOT_ENCRYPT_IMAGE
+	help
+	  When possible to map storage device, at least for read operations,
+	  to address space or RAM area, enabling this option allows hash
+	  calculation functions to directly access the storage through that address
+	  space or using its own DMA. This reduces flash read overhead done
+	  by the MCUboot.
+	  Notes:
+	    - not supported when encrypted images are in use, because calculating
+	      SHA requires image to be decrypted first, which is done to RAM.
```
nordicjm, line 102:
```suggestion
	      SHA requires image to be decrypted first, which is done in RAM.
```

boot/zephyr/Kconfig
```diff
@@ -88,6 +88,22 @@ config BOOT_IMG_HASH_ALG_SHA512_ALLOW
...
+	    - currently only supported on internal storage of devices; this
+	      option will not work with devices that use external storage for
+	      either of image slots.
```
nordicjm, line 105:
```suggestion
	      either of the image slots.
```

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -141,6 +141,13 @@
 #define MCUBOOT_DECOMPRESS_IMAGES
 #endif
 
+/* Invoke hashing functions directly on storage. This requires for device
```
nordicjm, line 144:
```suggestion
/* Invoke hashing functions directly on storage device. This requires the device
```

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -141,6 +141,13 @@
...
+ * to be able to map storage to address space or RAM.
```
nordicjm, line 145:
```suggestion
 * be able to map storage to address space or RAM.
```

boot/bootutil/src/image_validate.c
```diff
@@ -361,6 +363,35 @@ bootutil_get_img_security_cnt(struct image_header *hdr,
...
+    int32_t rc;
+
+    rc = bootutil_tlv_iter_begin(&it, hdr, fap, IMAGE_TLV_SIG_PURE, false);
+    if (rc) {
+        return rc;
+    }
+
+    /* Search for the TLV */
+    rc = bootutil_tlv_iter_next(&it, &off, &len, NULL);
+
```
nordicjm, line 389: this should check the value too and it should only have one valid value

boot/bootutil/src/image_validate.c
```diff
@@ -425,6 +461,67 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
     FIH_DECLARE(security_counter_valid, FIH_FAILURE);
 #endif
 
+#ifdef MCUBOOT_DECOMPRESS_IMAGES
```
nordicjm, line 464: this should not be here

nordicjm (changes requested):

boot/bootutil/src/image_validate.c
de-nordic: You mean that the TLV should have a value?

boot/bootutil/src/image_validate.c
```diff
@@ -569,14 +646,21 @@ bootutil_img_validate(struct enc_key_data *enc_state, int image_index,
...
     }
 
+#if defined(EXPECTED_HASH_TLV) && !defined(MCUBOOT_SIGN_PURE)
     rc = !image_hash_valid;
     if (rc) {
         goto out;
     }
+#elif defined(MCUBOOT_SIGN_PURE)
+    /* This returns true on EQ, rc is err on non-0 */
+    rc = !FIH_EQ(valid_signature, FIH_SUCCESS);
```
nvlsianpu, line 662: use FIH_NOT_EQ()

boot/bootutil/src/bootutil_priv.h
```diff
@@ -268,6 +268,9 @@ struct boot_loader_state {
 fih_ret bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig,
                             size_t slen, uint8_t key_id);
 
+fih_ret bootutil_verify_img(const uint8_t *img, uint32_t size,
+                            uint8_t *sig, size_t slen, uint8_t key_id);
+
```
nvlsianpu, line 273: nit: I would appreciate code comment on difference to above function.

boot/zephyr/Kconfig
```diff
@@ -168,16 +191,38 @@ endif
...
+	  Select BOOT_SIGNATURE_TYPE_PURE to have a PureEdDSA calculating image
+	  signature directly on image, rather than hash of the image.
 
 if BOOT_SIGNATURE_TYPE_ED25519
+
+config BOOT_SIGNATURE_TYPE_PURE
+	bool "Use Pure signature of image"
+	depends on BOOT_SIGNATURE_TYPE_PURE_ALLOW
+	help
+	  The Pure signature is calculated directly over image rather than
```
d3zd3z, line 211: I still have some problems with the wording here. I would argue "This is a more secure signature" is a little misleading. It is unclear to me how this affects the security when doing hardware encryption, because we can still do the signature of the hash by the hardware. The only real thing that PureEdDSA does is make the algorithm more resilient to weaknesses found in the hash function used. It doesn't affect the security of the signature itself.

d3zd3z: There is actually a bit of a performance hit to Pure mode as well, as it requires hashing the image twice, whereas Hash EdDSA hashes the image once, and then hashes that has twice.

d3zd3z (changes requested):

boot/zephyr/Kconfig
de-nordic: > I still have some problems with the wording here. I would argue "This is a more secure signature" is a little misleading. It is unclear to me how this affects the security when doing hardware encryption, because we can still do the signature of the hash by the hardware. The only real thing that PureEdDSA does is make the algorithm more resilient to weaknesses found in the hash function used. It doesn't affect the security of the signature itself.

I can remove the "more secure" info.

nvlsianpu (approved):

boot/zephyr/Kconfig
de-nordic: Removed the "more secure" or confusing, I think, comments.

boot/bootutil/src/image_validate.c
de-nordic: Fixed

boot/bootutil/src/bootutil_priv.h
de-nordic: Added comment.

de-nordic:
@d3zd3z Got time to revisit?

de-nordic:
@d3zd3z @nordicjm Review?

nvlsianpu:
@d3zd3z Dominik has change the wording which I do believe applies review request you have made.
I would merge this PR in a day. Let me know If you see any other obstacles.

nvlsianpu:
@nordicjm revisit

boot/bootutil/src/image_validate.c
nordicjm: value of TLV should be 1 only and checked, a value of e.g. 2 should not be allowed but here it would be

docs/design.md
```diff
@@ -111,6 +111,8 @@ struct image_tlv {
 #define IMAGE_TLV_ECDSA_SIG         0x22   /* ECDSA of hash output */
 #define IMAGE_TLV_RSA3072_PSS       0x23   /* RSA3072 of hash output */
 #define IMAGE_TLV_ED25519           0x24   /* ED25519 of hash output */
+#define IMAGE_TLV_SIG_PURE          0x25   /* If true then any signature found has been
+                                            * calculated over image directly. */
```
nordicjm, line 115: looks like the first * on this line should not be here

de-nordic: Done

nordicjm (approved):

PR #2085, opened 2024-10-07
boot: bootutil: swap_scratch: Fix compressed image sector size check

nordicjm:
Fixes an issue with compressed update support whereby it would wrong continue to check all sector sizes and error due to the sector sizes of the secondary slot being 0 until overflow

### DIFF
--- a/boot/bootutil/src/swap_scratch.c
+++ b/boot/bootutil/src/swap_scratch.c
@@ -191,7 +191,17 @@ boot_slots_compatible(struct boot_loader_state *state)
             smaller = 1;
             i++;
         } else {
-            sz1 += boot_img_sector_size(state, BOOT_SECONDARY_SLOT, j);
+            size_t sector_size = boot_img_sector_size(state, BOOT_SECONDARY_SLOT, j);
+
+#ifdef MCUBOOT_DECOMPRESS_IMAGES
+            if (sector_size == 0) {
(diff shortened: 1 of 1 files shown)


nordicjm:
@d3zd3z @utzig TFM looks to be broken again?

boot/bootutil/src/swap_scratch.c
```diff
@@ -191,7 +191,17 @@ boot_slots_compatible(struct boot_loader_state *state)
...
+
+#ifdef MCUBOOT_DECOMPRESS_IMAGES
+            if (sector_size == 0) {
+                /* Since this supports decompressed images, we can safely exit if slot1 is
+                 * smaller than slot0.
+                 */
+                break;
+            }
+#endif
+            sz1 += sector_size;
```
de-nordic, line 204: Should we even have a full run of the function `boot_slots_compatible` when decompression is enabled? It seems that we are going to ignore the difference anyway, so we could just skip the check at all.

nordicjm: Slot 0 should still be larger than slot 1, plus it can be possibly enhanced in future if only some slots support compression

de-nordic: OK.
Note for the future: for devices that have uniform erase block size and block sizes that equal for partitions, if the are placed on different devices, we should figure out compile time check for this.

de-nordic (approved):

PR #2089, opened 2024-10-08
Allow bypassing ASN.1 processing of public key for ED25519

de-nordic:
Add conditional compilation of ASN.1 decoding of ED25519 key.

This allows to cut out the ASN.1 encoding, which results in reduced flash size.

Comparison on nrf52840dk/nrf52840 build of MCUboot with the ED25519 enabled 
```
west build -p -d builds/nrf52_mcuboot_ed25519_no_asn -b nrf52840dk/nrf52840 bootloader/mcuboot/boot/zephyr/ -DCONFIG_BOOT_SIGNATURE_TYPE_ED25519=y
```

Reduces code from 40422 bytes to 39918 bytes, when  `-DCONFIG_BOOT_KEY_IMPORT_BYPASS_ASN=y` is added.

Another benefit of the option is that it is no longer needed to have portion of the mbedTLS, that does the ASN.1 support, enabled when compiling MCUboot with ED25519 for TinyCrypt.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -240,6 +240,15 @@ config BOOT_ED25519_MBEDTLS
 	select BOOT_AES_MBEDTLS_DEPENDENCIES if MBEDTLS_BUILTIN && BOOT_ENCRYPT_IMAGE
 
 endchoice
+
+config BOOT_KEY_IMPORT_BYPASS_ASN
+	bool "Directly access key value without ASN.1 parsing"
+	help
+	  Originally, public keys compiled into MCUboot were
+	  stored in ASN.1 encoded format. Enabling this option
+	  bypasses the ASN.1 decoding and directly accesses the key
+	  in ASN.1 bitstream; this reduces MCUboot code by removing
+	  the ASN.1 processing.
 endif
 
 endchoice
--- a/boot/bootutil/src/image_ed25519.c
+++ b/boot/bootutil/src/image_ed25519.c
@@ -12,27 +12,30 @@
 #ifdef MCUBOOT_SIGN_ED25519
 #include "bootutil/sign_key.h"
 
+#if !defined(MCUBOOT_KEY_IMPORT_BYPASS_ASN)
 /* We are not really using the MBEDTLS but need the ASN.1 parsing functions */
 #define MBEDTLS_ASN1_PARSE_C
 #include "mbedtls/oid.h"
 #include "mbedtls/asn1.h"
+#endif
 
 #include "bootutil_priv.h"
 #include "bootutil/crypto/common.h"
 #include "bootutil/crypto/sha.h"
 
 #define EDDSA_SIGNATURE_LENGTH 64
-
-static const uint8_t ed25519_pubkey_oid[] = MBEDTLS_OID_ISO_IDENTIFIED_ORG "\x65\x70";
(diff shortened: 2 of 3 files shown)


boot/zephyr/Kconfig
```diff
@@ -183,6 +183,15 @@ config BOOT_ED25519_MBEDTLS
...
 endchoice
+
+config BOOT_KEY_IMPORT_BYPASS_ASN
+	bool "Directly access key value without ASN.1 parsing"
+	help
+	  Originally public keys compiled into MCUboot are
+	  stored in ASN.1 encoded format, Enabling the options
+	  bypasses the ASN.1 decoding and directly accesss the key
+	  in ASN.1 bitstream; this reduces MCUboot code by removing
+	  the ASN.1 processing.
```
nordicjm, line 194:
```suggestion
	  Originally, public keys compiled into MCUboot were
	  stored in ASN.1 encoded format. Enabling this option
	  bypasses the ASN.1 decoding and directly accesses the key
	  in ASN.1 bitstream; this reduces MCUboot code by removing
	  the ASN.1 processing.
```

nordicjm (approved):

de-nordic:
@d3zd3z hej, got time to review?

davidvincze:
It looks good to me, I have only have:

- Comment:  #include "mbedtls/oid.h" and #include "mbedtls/asn1.h" could be conditional includes if ASN.1 is bypassed.

- Note: at https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/crypto/ecdsa.h#L427 it does some manual parsing and checks the type of the key and curve. We could move it to a common ecdsa_helper file for example if needed, but bypassing ASN.1 moves us towards the usage of raw keys and signatures (instead of ASN.1) which has already been mentioned/suggested (considering the fact that currently MCUboot configuration clearly determines the expected signature and key type).

davidvincze (approved):

de-nordic:
> It looks good to me, I only have:
> 
>     * Comment:  #include "mbedtls/oid.h" and #include "mbedtls/asn1.h" could be conditional includes if ASN.1 is bypassed.
> 
>     * Note: at https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/crypto/ecdsa.h#L427 it does some manual parsing and checks the type of the key and curve. We could move it to a common ecdsa_helper file for example if needed, but bypassing ASN.1 moves us towards the usage of raw keys and signatures (instead of ASN.1) which has already been mentioned/suggested (considering the fact that currently MCUboot configuration clearly determines the expected signature and key type).

I will implement the changes.

davidvincze:
As I already mentioned, the parsing of ECDSA is handled differently in the existing code, but I think your solution is also reliable as the variability in size is only present for the signatures (encoding unsigned values as signed integers). The key is encoded as a bit string with 2 extra bytes (num of unused bits and compression) and its size is deterministic.

de-nordic:
> As I already mentioned, the parsing of ECDSA is handled differently in the existing code, but I think your solution is also reliable as the variability in size is only present for the signatures (encoding unsigned values as signed integers). The key is encoded as a bit string with 2 extra bytes (num of unused bits and compression) and its size is deterministic.

I have applied the comment regarding ifdefing includes.
I will try to do the edcsa, once I figure out the `bootutil_read_bigint`.

PR #2095, opened 2024-10-14
boot_serial: Fix serial recovery for LPC55x and MCXNx

butok:
- Increases the unaligned buffer default value for LPC55x and MCXNx.
- Fixes the serial recovery for LPC55x and MCXNx.

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -65,8 +65,10 @@ config MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD
 
 config BOOT_SERIAL_UNALIGNED_BUFFER_SIZE
 	int "Stack buffer for unaligned memory writes"
+	default 512 if SOC_SERIES_LPC55XXX
+	default 128 if SOC_SERIES_MCXN
 	default 64
-	range 0 128
+	range 0 1024
 	help
 	  Specifies the stack usage for a buffer which is used for unaligned
 	  memory access when data is written to a device with memory alignment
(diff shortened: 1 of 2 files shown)


boot/zephyr/Kconfig.serial_recovery
```diff
@@ -65,8 +65,8 @@ config MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD
 
 config BOOT_SERIAL_UNALIGNED_BUFFER_SIZE
 	int "Stack buffer for unaligned memory writes"
-	default 64
-	range 0 128
+	default 512
+	range 0 1024
```
nordicjm, line 69: the default should be changed for your boards/socs, not every board. There is no reason every board with serial recovery should suddenly have an extra 448 bytes of RAM taken up for literally nothing

butok: Hi @nordicjm 
The goal to have the same user experience as for Zephyr MCUMgr that works for all cases out of the box.
Having the higher value bring only benefits.
It works for all boards, faster, and without any resource penalty as the MCUBoot stack is too high - set to 10KB (used <20%).

de-nordic: Sure, but you bring surprise to people who have tuned down stack and build of MCUboot, after this is brought in, topples into stack fault.
So now they have to go to their config files and bring that down.
I understand what you are trying to fix but you are not fixing that for everyone, quite an opposite.
The right approach is to have tuning done for device/soc/board.

Most people are not aware of the config or just briefly know of existence, because they have never had to deal with it as default worked.

butok: > Sure, but you bring surprise to people who have tuned down stack and build of MCUboot, after this is brought in, topples into stack fault. So now they have to go to their config files and bring that down. I understand what you are trying to fix but you are not fixing that for everyone, quite an opposite. The right approach is to have tuning done for device/soc/board.
> 
> Most people are not aware of the config or just briefly know of existence, because they have never had to deal with it as default worked.

Updated. 
Default value increased only for LPC55 and MCXN, for all other platforms unchanged.

nordicjm (approved):

de-nordic (approved):

PR #2098, opened 2024-10-15
imgtool: bugfix #2096

### LINKED ISSUE #2096: Imgtool: broken image encryption
# Problem

Since commit 63fa7e45a213a4cf80c9f3b0e24c69090eff8130, imgtool cannot properly encrypt images.

# Explanation

It turns that a wrong number of arguments is passed to `Image.create()` when it is called here https://github.com/mcu-tools/mcuboot/blob/b9d69dd2a2d6df32da6608d549138288bb7d7aa5/scripts/imgtool/main.py#L512

Actually, 13 arguments are passed instead of 14: the argument `compression_type` is missing.

michalek-no:
fixing broken encryption caused by shift in function parameters.

Fixes #2096

### DIFF
--- a/scripts/imgtool/main.py
+++ b/scripts/imgtool/main.py
@@ -510,7 +510,7 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
         }
 
     img.create(key, public_key_format, enckey, dependencies, boot_record,
-               custom_tlvs, compression_tlvs, int(encrypt_keylen), clear,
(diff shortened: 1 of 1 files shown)


nordicjm:
@mimok can you try this and check the issue is resolved?

mimok:
Actually, I implemented the same thing as a work around. I believe it works (according to the content of the binary file) but i need to do additional tests these friday to see whether mcuboot is ok with binaries generated this way.

nvlsianpu (approved):

PR #2100, opened 2024-10-16
bootutil: Fix device bricked after power failure during swap-move revert

### LINKED ISSUE #1966: swap-move: Power failure during the writing of the magic could brick the device
Description
---
While reading through the implementation of the swap using move, I noticed a corner case that could, if I'm not mistaken, brick the device in case a power failure occurs during a revert process.

Let's suppose after an upgrade you have a non-functional image in the primary slot. The image won't be confirmed, leading to a revert at next boot. At the beginning of the revert process, `fixup_revert` is invoked, which rewrites the trailer in the secondary slot so that the revert looks like a permanent upgrade. Normally, after the execution of this routine, the secondary slot has a valid trailer, in particular with a valid magic number.

Let's imagine a power failure occurs during the writing of the trailer's magic, i.e. in `boot_write_magic` (see [here](https://github.com/mcu-tools/mcuboot/blob/45d379e715d82ae71ce2bb42882769a4cf4428e4/boot/bootutil/src/swap_move.c#L492)). The magic number in the secondary slot is in an undefined state and might be partially written, which implies at next boot it will be considered in `BOOT_MAGIC_BAD` state.

So, at next boot, we have the following state:
* Primary slot: magic=good, copy-done=set, image-ok=unset
* Secondary slot: magic=bad, copy-done=unset, image-ok=set

This doesn't match any state leading to an upgrade or revert process to be initiated, so `boot_swap_type_multi` will return `BOOT_SWAP_TYPE_NONE`. This means MCUboot will not perform the revert and attempt to boot from the primary slot, containing a non-functional image. Hence, the device is bricked unless it is possible to reflash the secondary slot without a functional image.

Note at the time `fixup_revert` is called, the boot status is virgin, meaning `boot_status_is_reset(bs)` is true at next boot so MCUboot won't detect a revert process was interrupted and won't try to resume it.

How to reproduce
---
To easily reproduce the issue, it is possible to modify the `boot_write_magic` routine to write the magic in two calls to `flash_area_write`. For instance, it is possible to replace:
https://github.com/mcu-tools/mcuboot/blob/45d379e715d82ae71ce2bb42882769a4cf4428e4/boot/bootutil/src/bootutil_public.c#L307

With:
```
rc = flash_area_write(fap, pad_off, &magic[0], BOOT_MAGIC_ALIGN_SIZE / 2);
if (rc != 0) {
    return BOOT_EFLASH;
}

rc = flash_area_write(fap, pad_off + BOOT_MAGIC_ALIGN_SIZE / 2, &magic[BOOT_MAGIC_ALIGN_SIZE / 2],
                      BOOT_MAGIC_ALIGN_SIZE - BOOT_MAGIC_ALIGN_SIZE / 2);
```

And to to run the tests for swap-move on the simulator:
```
cargo test --features "swap-move"
```
During the `revert_with_fails` test, a power failure is simulated during every modification to the flash memory, so in particular between the two `flash_area_write` in `boot_write_magic`. This will lead to an error:
```
[ERROR bootsim::image] Revert failed at interruption 5
[...]
[ERROR bootsim::image] Revert failed at interruption 236
test revert_with_fails ... FAILED
[...]
---- revert_with_fails stdout ----
thread 'revert_with_fails' panicked at sim/tests/core.rs:57:1:
assertion failed: !image.run_revert_with_fails()
```

I didn't try on real hardware, but I'm happy to do so if you consider that useful.

taltenbach:
This PR proposes a fix to https://github.com/mcu-tools/mcuboot/issues/1966, which describes a scenario where a device can be bricked if a revert process is interrupted when using swap-move.

As suggested in [this message](https://github.com/mcu-tools/mcuboot/issues/1966#issuecomment-2200541322), a very straightforward fix might be enough. The latter is implemented in this PR.

The idea is to perform a revert no matter the state of the magic number in the secondary slot's trailer, provided the `copy-done`
flag is set in the primary slot but the `image-ok` flag is not. The `copy-done` flag is set only after having completed an upgrade or
revert process so if the `copy-done` flag is set but the `image-ok` is unset, it is guaranteed an upgrade has been performed but the new image has not been confirmed, which implies a revert is needed.

That looks good to me but perhaps I missed some corner cases that would justify that `BOOT_MAGIC_UNSET` was used instead of `BOOT_MAGIC_ANY`. @utzig @d3zd3z do you have any input on that point?

Fixes #1966

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -118,7 +118,7 @@ static const struct boot_swap_table boot_swap_tables[] = {
     },
     {
         .magic_primary_slot =       BOOT_MAGIC_GOOD,
-        .magic_secondary_slot =     BOOT_MAGIC_UNSET,
+        .magic_secondary_slot =     BOOT_MAGIC_ANY,
         .image_ok_primary_slot =    BOOT_FLAG_UNSET,
         .image_ok_secondary_slot =  BOOT_FLAG_ANY,
         .copy_done_primary_slot =   BOOT_FLAG_SET,
--- a/docs/design.md
+++ b/docs/design.md
@@ -656,7 +656,7 @@ types described above via a set of tables.  These tables are reproduced below.
     State III
                      | primary slot | secondary slot |
     -----------------+--------------+----------------|
-               magic | Good         | Unset          |
+               magic | Good         | Any            |
             image-ok | 0xff         | Any            |
            copy-done | 0x01         | Any            |
     -----------------+--------------+----------------'


d3zd3z:
My only request would be to see if we can come up with a test for the simulator that provokes this failure. The simulator should be interrupting between each flash operation, although because that is really slow, it might only just be doing it randomly.

Otherwise, the fix seems reasonable to me.

d3zd3z (approved):

de-nordic:
We need documentation update for that, because state description there will not longer match the code.

taltenbach:
> My only request would be to see if we can come up with a test for the simulator that provokes this failure. The simulator should be interrupting between each flash operation, although because that is really slow, it might only just be doing it randomly.

@d3zd3z I created an issue (#2108) regarding that point with some ideas that we can discuss. I will try to find some time to implement a solution in the following weeks.

> We need documentation update for that, because state description there will not longer match the code.

@de-nordic You're right, this should be better now :)

PR #2114, opened 2024-11-11
bootutil: Add support for devices without erase

de-nordic:
Commits:
 - bootutil: add support for not calling erase when device does not require it and  MCUBOOT_DEV_WITHOUT_ERASE_THEN_NO_ERASE option that controls this; add MCUBOOT_MINIMAL_SCRAMBLE that allows to reduce amount of removed data, where applicable;
 - Zephyr: Kconfig options and supporting code
 - Zephyr: switch to flash_area_flatten in bs_custom_storage_erase
~- Zephyr: switch to boot_set_next, from flash_area_erase, for Zephyr builds of bootutil_public.~
 
The first MCUboot option allows to improve life time of devices for which erase is emulated by write, and is not really required by the device design and the second MCUboot option allows to just remove enough of data, where needed, by either reduced range of erase or reduced write; for example by scrambling only image header, rather than removing the entire image.

Tested on nrf52840dk and nrf54l15. In case of nrf54l it reduces swap time by half, in comparison to code using erase.

Note: Even though, finally, `swap_erase_trailer_sectors` is no longer used, I have left it there because it may be used in the future.
I have left a Note around `swap_status_init` function, that partially corresponds to this; it seems that in many places call to `swap_scramble_trailer_sectors`, whee followed by `swap_status_init`,  should in reality be `swap_erase_trailer_sectors` fulfilling hardware requirement for preparing device for write; the problem is that `swap_status_init` initializes only part of swap information, which means that if the region of entire swap info is not scrambled first, leftovers may lead to improper behaviour, for example test will not revert.
There is re-design of `swap_status_init` required to mitigate this, as in some paths the whole "init" may just mean removal of everything except information that `swap_status_init` would put there anyway, or combined init operation that would first remove no longer needed information than overwrite what should be re-initialized - but this requires further analysis.

So far this PR works fine on devices that do not require erase as well as these that require erase before write.


boot/bootutil/src/bootutil_priv.h
```diff
@@ -302,7 +302,15 @@ int boot_copy_region(struct boot_loader_state *state,
                      const struct flash_area *fap_src,
                      const struct flash_area *fap_dst,
                      uint32_t off_src, uint32_t off_dst, uint32_t sz);
+/* Prepare for write device that requires erase prior to write. This will
+ * do nothing on devices without erase requirement.
+ */
 int boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz);
+/* SImilar to boot_erase_region but will alwasy remove data */
```
nordicjm, line 309: -I+i, *always

boot/bootutil/src/loader.c
```diff
@@ -1174,20 +1174,108 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+
+        memset(buf, flash_area_erased_val(fa), sizeof(buf));
+
+        while (size_done < size) {
+            ret = flash_area_write(fa, size_done + off, buf, sizeof(buf));
+            if (ret != 0) {
+                break;
+            }
+            size_done += sizeof(buf);
+        }
```
nordicjm, line 1239: this just assumes max align will not overflow the size but makes no checks to ensure that, it should reduces buffer on final write to ensure it does not overflow

boot/bootutil/src/loader.c
```diff
@@ -1174,20 +1174,108 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+        ret = flash_area_get_sector(fa, 0, &header);
+        size = flash_sector_get_size(&header);
+
+        if (ret != 0) {
+            return ret;
+        }
+    } else {
+        size = MAX(sizeof(((struct image_header *)0)->ih_magic),
+                   BOOT_MAX_ALIGN);
+        size = (size + BOOT_MAX_ALIGN - 1) & ~(BOOT_MAX_ALIGN - 1);
```
nordicjm, line 1275: This should do the same for footer data because it might contain swap status or mcuboot flags for swapping images and those should always be cleared

boot/zephyr/Kconfig
```diff
@@ -832,6 +832,36 @@ endif # BOOT_DECOMPRESSION_SUPPORT
...
+config MCUBOOT_STORAGE_MINIMAL_SCRAMBLE
+	bool "Do minimal required work to remove data"
+	help
+	  In some cases MCUboot has to remove data, which usually means make
+	  it non-viable for MCUboot rather then completely destroyed.
+	  For example when MCUboot does not want to bother with broken image
+	  in some slot it will remove it.
+	  The same can be achieved with just removal of header, leaving the
+	  rest of image untouched, as without header MCUboot will not be able
+	  to recognize image in slot as bootable.
```
nordicjm, line 858: include footer

boot/bootutil/src/bootutil_public.c
```diff
@@ -548,7 +548,11 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
             /* The image slot is corrupt.  There is no way to recover, so erase the
              * slot to allow future upgrades.
              */
+#if !defined(CONFIG_MCUBOOT) && defined(__ZEPHYR__)
+            flash_area_flatten(fa, 0, flash_area_get_size(fa));
+#else
```
nordicjm, line 553: think it was mentioned in a previous PR that OS specific code should not be present in here

boot/bootutil/src/loader.c
de-nordic: Yes, that was intentional as I have assumes that MCUboot should write at BOOT_MAX_ALIGN pace anyway, but maybe that should be fixed.

boot/bootutil/src/bootutil_public.c
de-nordic: I can fix that although MCUboot has significant problem where we are coupled with Flash Map API which means that this has to be replaced with a function that is implemented for every supported system. I will figure way around.

de-nordic:
Converting to draft; after significant changes I need to re-test this on devices.

boot/zephyr/Kconfig
```diff
@@ -838,6 +838,45 @@ endif # BOOT_DECOMPRESSION_SUPPORT
...
+	  may emulate erase by write, as a way to scramble data rather then
+	  by hardware requirement. This unfortunately means that code that
+	  does erase before write, when not needed, will write device twice
+	  which not only reduces device life time but also doubles time
+	  of any write operation (one write for erase and one write for actual
+	  write).
+	  When this option is enabled, MCUboot will check for type of device
+	  and will avoid erase where not needed.
+
+# The depends on FLASH_HAS_EXPLICIT_ERASE is intentionlally missing for now
```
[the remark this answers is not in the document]
de-nordic: You are not able to tell whether external device, that zephyr has configured driver for, is actually used by MCUboot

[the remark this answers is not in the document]
nordicjm: it really should have been a dts property, making it kconfig was a bad choice

boot/bootutil/src/bootutil_misc.c
```diff
@@ -135,6 +135,58 @@ boot_trailer_sz(uint32_t min_write_sz)
...
+
+        ret = flash_area_get_sector(fa, 0, &sector);
+        if (ret < 0) {
+            return ret;
+        }
+
+        *size = flash_sector_get_size(&sector);
+    } else {
+        /* For device not requiring erase align to write block */
+        *size = ALIGN_UP(sizeof(((struct image_header *)0)->ih_magic), write_block);
```
nordicjm, line 184: this does not work for swap using offset

boot/bootutil/src/loader.c
```diff
@@ -1242,19 +1242,209 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+            if (ret != 0) {
+                break;
+            }
+            size_done += write_block;
+        }
+    }
+    return ret;
+}
+
+
```
nordicjm, line 1307:
```suggestion
```

boot/bootutil/src/loader.c
```diff
@@ -1242,19 +1242,209 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+{
+    size_t size;
+    int ret = 0;
+
+    /* Without minimal entire area needs to be scrambled */
+#if !defined(MCUBOOT_MINIMAL_SCRAMBLE)
+    size = flash_area_get_size(fa);
+    ret = boot_scramble_region(fa, 0, size);
+#else
+
```
nordicjm, line 1428:
```suggestion
```

nordicjm (commented):
noticed swap using offset file update is missing

boot/zephyr/Kconfig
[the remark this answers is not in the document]
de-nordic: No. It was right choice.
Making it DTS property would require to make it Kconfig at some point anyway, otherwise you would not be able to tune your subsystems according to what hardware has to offer, or rather what drivers are supporting.
Also, you can have, as I do, drivers that have no DTS entries but can still set Kconfigs to tune subsystems that use them.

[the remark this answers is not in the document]
nordicjm: dts means it can be used from inside of Kconfig and cmake and it is known **per driver**, having it at Kconfig means you have no clue about per driver. It is a hardware feature, not a software feature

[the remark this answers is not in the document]
de-nordic: I have a clue, because I can look at the Kconfig symbol and see what is selecting it, except things picked by ${dt_} macros, which just evaluate to value and you can not tell where they come from, until you open Kconfig, identify the DT, and then search for generated DTS file to figure out. That happens per driver.
```
Symbols currently y-selecting this symbol:
- NORDIC_QSPI_NOR
- SOC_FLASH_NRF
```
It is true that erase is hardware feature (or rather problem), but Kconfig exposes here that certain feature exists, is exposed by software/driver, and should be potentially addressed.
Following the example you have provided we can always do this:
```
config FLASH_HAS_HAS_SOME_FEATURE
        default $(dt_compat_any_has_prop, some-prop)
```
as well as we can just avoid it and go with directly setting the Kconfig by driver, because driver itself will be filtered out if there is no DTS option for it (usually, if driver has proper depends).

Problem with selecting things via ${dt_} macros is that you either have to add a lot of different compat, one per feature, or introduce device classes, because properties have common name space and "some-prop" existing in flash device can not be easily distinguished from "some-prop" in, for example, modem.

And still after having all the information you only know whether there is any device that has a feature but it is up to you to figure out if anything uses it and disable path that don't.

nordicjm:
Rebase needed

boot/bootutil/src/loader.c
```diff
@@ -1267,19 +1267,212 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
 boot_erase_region(const struct flash_area *fap, uint32_t off, uint32_t sz)
 {
-    return flash_area_erase(fap, off, sz);
+    if (device_requires_erase(fap)) {
+        return flash_area_erase(fap, off, sz);
+    }
+    return 0;
+}
+
+
```
nordicjm, line 1289:
```suggestion
```

boot/bootutil/src/loader.c
nordicjm: Not addressed

boot/bootutil/src/loader.c
```diff
@@ -1267,19 +1267,212 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+            if (first_offset >= off) {
+                /* Reached the first offsset in range and already erased it */
+                break;
+            }
+
+            /* Move down to previous sector, the flash_area_get_sector will
+             * correct the value to real page offset
+             */
+            off -= 1;
```
nordicjm, line 1401: `off -= csize;` ? with `-1` will run 4096 times for a single sector on nrf52

nordicjm: never mind, not an issue

nordicjm (approved):
nit. can be addressed later

nvlsianpu (approved):

PR #2115, opened 2024-11-12
codeowners: Move wildcard entry to top

nordicjm:
The wildcard entry needs moving to the top so other entries override it

@utzig Now that this seems to sort of work (the action fails, I have no idea why but I've given up trying to work that out), should any folders/files be added to you?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,4 @@
+* @d3zd3z
 boot/boot_serial/ @nordicjm @de-nordic
 boot/bootutil/ @davidvincze
 boot/cypress/ @romanjoe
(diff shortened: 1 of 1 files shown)


de-nordic (approved):

nordicjm:
> > @utzig Now that this seems to sort of work (the action fails, I have no idea why but I've given up trying to work that out)
> 
> What action fails?

The PR assignment one https://github.com/mcu-tools/mcuboot/actions/runs/11776180255/job/32798062220 have removed it. Though I can see the issue one is failing too. Something very off about this repo because they work fine on other repos

nordicjm:
> > Something very off about this repo because they work fine on other repos
> 
> Do you mind sending some of those other repos that work? I'm not interested in personal repos, only organization repos, because they are essentially different in the API. I can take a look and try to find out what's wrong.

The others are on personal repos

PR #2116, opened 2024-11-12
boot_serial: Support sha256, sha384 and sha512

de-nordic:
Add support for reporting various sha in image list. There is always only one sha compiled in, but serial recovery has been previously hardcoded to support sha256 only.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -81,6 +81,17 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #define ARRAY_SIZE ZCBOR_ARRAY_SIZE
 #endif
 
+#if defined(MCUBOOT_SHA512)
+    #define IMAGE_HASH_SIZE (64)
+    #define IMAGE_SHA_TLV   IMAGE_TLV_SHA512
+#elif defined(MCUBOOT_SIGN_EC384)
+    #define IMAGE_HASH_SIZE (48)
+    #define IMAGE_SHA_TLV   IMAGE_TLV_SHA384
+#else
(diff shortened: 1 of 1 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -336,7 +347,7 @@ bs_list(char *buf, int len)
             }
 
 #ifdef MCUBOOT_SERIAL_IMG_GRP_HASH
-            /* Retrieve SHA256 hash of image for identification */
+            /* Retrieve hash of image for identification */
             rc = boot_serial_get_hash(&hdr, fap, hash);
```
nordicjm, line 351: this will overflow, there is a `hash[32]` above

boot/boot_serial/src/boot_serial.c
```diff
@@ -1467,7 +1478,7 @@ static int boot_serial_get_hash(const struct image_header *hdr,
             break;
         }
 
-        if (type == IMAGE_TLV_SHA256) {
+        if (type == IMAGE_SHA_TLV) {
             /* Get the image's hash value from the manifest section. */
             if (len != 32) {
```
nordicjm, line 1483: length not updated

nordicjm (changes requested):

boot/boot_serial/src/boot_serial.c
de-nordic: Funny how it worked with this bugs, as I was able to get image list and mark image for test.

nordicjm (commented):
line 454 has `hash[32]`

nordicjm (approved):

gchwier (approved):

PR #2121, opened 2024-11-13
boot: zephyr: Add fallback for overhead calculation when auto fails

nordicjm:
Adds a fallback of 128 sectors for calculating the image overhead size if auto sectors is enabled but cannot calculate the value


nordicjm:
@erwango give this a try and see if it resolves the issue

nordicjm:
128 sectors is definitely wrong (which would have been wrong before the change to enable auto but would not have had a warning) you can add soc overlay files to MCUboot for the SoC's to set a lesser amount and disable the auto, that will get rid of 2 of the warnings. Don't have a solution for the erase part without an erase size, it uses it to align the overhead size up to the nearest sector size, so that value might not be wholly accurate

de-nordic (approved):

nordicjm:
> > 128 sectors is definitely wrong (which would have been wrong before the change to enable auto but would not have had a warning) you can add soc overlay files to MCUboot for the SoC's to set a lesser amount and disable the auto, that will get rid of 2 of the warnings. Don't have a solution for the erase part without an erase size, it uses it to align the overhead size up to the nearest sector size, so that value might not be wholly accurate
> 
> Thanks, I'll work on this. Are soc overlays names coming from HWMv2 ?

Yes, a combination of the SOC name and CPU cluster (if present) from the `soc.yml` file, in a `socs` folder instead of `boards`

erwango (approved):

PR #2126, opened 2024-11-19
IAR compiler compatibility break in image.h

Karambite:
Updated image.h to avoid breaking for IAR compiler. 
```
#ifndef __packed
#define __packed __attribute__((__packed__))
#endif
```
This is a compatibility break with the IAR compiler. __packed is already a keyword that eliminates padding between members. This is not a definition, meaning it will always fall under the #ifndef. Any additions that use __packed will be replaced by the definition given.

I created a macro that will pack the structures based on the compiler used.


nordicjm:
Commit needs a sign off and message text adding to commit i.e. `git commit --amend -s` then add description

nordicjm:
@Karambite would you be able to fix this today or on Monday? Would like to pull latest changes into Zephyr with a manifest update soon

Karambite:
> @Karambite would you be able to fix this today or on Monday? Would like to pull latest changes into Zephyr with a manifest update soon

Sorry I haven't had time to get back to this. I will try to close this over the weekend

Karambite:
I tried to rebase and accidentally closed this PR. A new one is opened here https://github.com/mcu-tools/mcuboot/pull/2149.

PR #2127, opened 2024-11-20
zephyr: Add CONFIG_MCUBOOT_CLEANUP_RAM

de-nordic:
Add Kconfig option to cleanup RAM in MCUboot before passing control to an application.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -219,6 +219,12 @@ config MCUBOOT_CLEANUP_ARM_CORE
 	  start-up code which can cause a module fault and potentially make the
 	  module irrecoverable.
 
+config MCUBOOT_CLEANUP_RAM
+	bool "Perform RAM cleanup"
+	depends on CPU_CORTEX_M4 || CPU_CORTEX_M33
+	help
+	  Sets contents of memory to 0 before jumping to application.
+
 config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -236,7 +236,34 @@ static void do_boot(struct boot_rsp *rsp)
     __set_CONTROL(0x00); /* application will configures core on its own */
     __ISB();
 #endif
+#if CONFIG_MCUBOOT_CLEANUP_RAM
+    __asm__ volatile (
+        /* vt->reset -> r0 */
+        "   mov     r0, %0\n"
+        /* base to write -> r1 */
+        "   mov     r1, %1\n"
+        /* size to write -> r2 */
+        "   mov     r2, %2\n"
+        /* value to write -> r3 */
+        "   mov     r3, %3\n"
+        "clear:\n"
+        "   str     r3, [r1]\n"
+        "   add     r1, r1, #4\n"
+        "   sub     r2, r2, #4\n"
+        "   cbz     r2, out\n"
+        "   b       clear\n"
+        "out:\n"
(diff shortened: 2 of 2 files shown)


boot/zephyr/main.c
```diff
@@ -236,7 +236,34 @@ static void do_boot(struct boot_rsp *rsp)
     __set_CONTROL(0x00); /* application will configures core on its own */
     __ISB();
 #endif
+#if CONFIG_MCUBOOT_CLEANUP_RAM
+    __asm__ volatile (
+        /* rt->reset -> r0 */
```
nvlsianpu, line 241:
```suggestion
        /* vt->reset -> r0 */
```

boot/zephyr/main.c
```diff
@@ -236,7 +236,34 @@ static void do_boot(struct boot_rsp *rsp)
...
+        "   sub     r2, r2, %4\n"
+        "   cbz     r2, out\n"
+        "   b       clear\n"
+        "out:\n"
+        "   dsb\n"
+        /* jump to reset vector of an app */
+        "   bx      r0\n"
+        :
+        : "r" (vt->reset), "i" (CONFIG_SRAM_BASE_ADDRESS),
+          "i" (CONFIG_SRAM_SIZE * 1024), "i" (0), "i" (sizeof(uint32_t))
```
nvlsianpu, line 261: ` "i" (sizeof(uint32_t))` operand is not used, assembly above used immediate value - which is just perfect.

nvlsianpu (changes requested):

nvlsianpu (approved):

boot/zephyr/main.c
de-nordic: Fixed

boot/zephyr/main.c
de-nordic: Yeah, so I have hardcoded this to #4 and remove the parameter.

nordicjm (changes requested):
Does not work with ARMv6:
```
/tmp/ccmsO9XM.s: Assembler messages:
/tmp/ccmsO9XM.s:495: Error: selected processor does not support `cbz r2,out' in Thumb mode
/tmp/ccmsO9XM.s:488: Error: invalid immediate: 536870912 is out of range
/tmp/ccmsO9XM.s:488: Error: value of 20000000 too large for field of 2 bytes at 000000c2
/tmp/ccmsO9XM.s:489: Error: invalid immediate: 32768 is out of range
```

nordicjm (commented):
Bit confused here, different symbol has been changed and depends on m7, we don't have any cortex m7's

de-nordic:
> Bit confused here, different symbol has been changed and depends on m7, we don't have any cortex m7's

I am now confused too, because it compiled and worked and the code is actually picked up.

de-nordic:
> > Bit confused here, different symbol has been changed and depends on m7, we don't have any cortex m7's
> 
> I am now confused too, because it compiled and worked and the code is actually picked up.

Yeah, sure, because I did the change to wrong Kconfig.

boot/zephyr/Kconfig
```diff
@@ -219,6 +219,12 @@ config MCUBOOT_CLEANUP_ARM_CORE
 	  start-up code which can cause a module fault and potentially make the
 	  module irrecoverable.
 
+config MCUBOOT_CLEANUP_RAM
+	bool "Perform RAM cleanup"
+	depends on CPU_CORTEX_M4
```
nordicjm, line 224: `|| CPU_CORTEX_M33`?

nordicjm (approved):

Issue #2093, opened 2024-10-11
Incorrect initialisation of hardware watchdog


danielb-tkjcy3rq:
[MCUBOOT_WATCHDOG_SETUP()](https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/include/mcuboot_config/mcuboot_config.h#L370) is missing a call to `wdt_install_timeout`. At least on my hardware (imxrt1170), the unit wasn't rebooting if I deliberately put an infinite loop to run down the watchdog timer.

There is the question of how to get users to set their desired timeout, but I guess that could be solved using a config option?

PR #2131, opened 2024-11-26
boot: Allow watchdog timeout to be configured

KarlKTorp:
This change adds a new Kconfig option MCUBOOT_WATCHDOG_TIMEOUT to specify the watchdog timeout value. Several devices (NXP among others) require a timeout to be installed in order to enable the watchdog. On e.g. imxrt1170 enabling the watchdog without a timeout will print the error "No valid timeouts installed" and fail to setup the watchdog.

Fixes: https://github.com/mcu-tools/mcuboot/issues/2093

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -750,6 +750,14 @@ config BOOT_WATCHDOG_FEED_NRFX_WDT
 	imply NRFX_WDT30
 	imply NRFX_WDT31
 
+config MCUBOOT_WATCHDOG_TIMEOUT
+	int "Watchdog timeout in milliseconds"
+	default 0
+	depends on BOOT_WATCHDOG_FEED
+	help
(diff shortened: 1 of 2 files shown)


nordicjm (changes requested):
The watchdog code in mcuboot is a bit of a red herring, it does not set up a watchdog, it only feeds it. And it does not make sense to set the watchdog up every time it is fed

KarlKTorp:
Right. That was my bad. It was of cause during setup the timeout should be installed - good catch. 

This is just a suggestion for the watchdog issues on NXP, we have used this method for some time. But I can see that watchdog support is limited in mcuboot and this might not be the optimal solution. That's up to you guys :)

KarlKTorp:
Sorry for letting this PR hang. The setup issue was fixed. Is this feature something you would consider adding?

PR #2138, opened 2024-12-05
boot: zephyr: boot_record: Save boot data with single image

dpw13:
Ensure that bootloader info and measurements are saved to retained memory when using mcuboot in single-image mode.

### DIFF
--- a/boot/bootutil/src/boot_record.c
+++ b/boot/bootutil/src/boot_record.c
@@ -237,7 +237,9 @@ int boot_save_shared_data(const struct image_header *hdr, const struct flash_are
                           const uint8_t slot, const struct image_max_size *max_app_sizes)
 {
     int rc;
+#if !defined(MCUBOOT_SINGLE_APPLICATION_SLOT)
     uint8_t image = 0;
+#endif
 
 #if defined(MCUBOOT_SINGLE_APPLICATION_SLOT)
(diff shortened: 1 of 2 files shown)


nordicjm (commented):
Some fields should be excluded from single application mode, can you update `boot/bootutil/src/boot_record.c` and add ifdefs like so:
```
#if !defined(MCUBOOT_SINGLE_APPLICATION_SLOT)
    if (!rc) {
        rc = boot_add_data_to_shared_area(TLV_MAJOR_BLINFO,
                                          BLINFO_RUNNING_SLOT,
                                          sizeof(slot), (void *)&slot);
    }
#endif
...
#if !defined(MCUBOOT_SINGLE_APPLICATION_SLOT)
    while (image < BOOT_IMAGE_NUMBER && !rc) {
        if (max_app_sizes[image].calculated == true) {
            rc = boot_add_data_to_shared_area(TLV_MAJOR_BLINFO,
                                              (BLINFO_MAX_APPLICATION_SIZE + image),
                                              sizeof(max_app_sizes[image].max_size),
                                              (void *)&max_app_sizes[image].max_size);

        }

        ++image;
    }
#endif
```

nordicjm (commented):
nit. and will be good to go

nordicjm (approved):

PR #2141, opened 2024-12-10
fix imgtool's sha512 for compression

michalek-no:
fixup! scripts: imgtool: compression

passing user_sha parameters while creating compressed image


nordicjm (commented):
no fixup lines in upstream

nordicjm (approved):

Issue #2129, opened 2024-11-23
Main branch unable to build for Mbed OS


zhiyong-ft:
Hi,

I am trying to integrate MCUBoot into our project, and noticed that the main branch as of commit: 8a2e2edba596de7471139ec9d9011d2dd8cbd86e
cannot be compiled for mbed-os projects.

I get the following error when compiling in Mbed-Studio with both GCC_ARM and ARMC6:
```
[Warning] loader.c@630,13: variable 'image_index' set but not used [-Wunused-but-set-variable]
Compile [ 98.3%]: ram_load.c
[Error] ram_load.c@48,2: #error "Platform MUST define executable RAM bounds in case of RAM_LOAD"
[Error] ram_load.c@80,22: 'IMAGE_EXECUTABLE_RAM_START' undeclared (first use in this function)
[Error] ram_load.c@81,21: 'IMAGE_EXECUTABLE_RAM_SIZE' undeclared (first use in this function)
[Error] ram_load.c@84,20: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@85,19: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@326,24: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@338,14: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@339,14: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@381,14: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@382,14: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@402,23: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@404,42: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@405,20: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@407,10: 'struct boot_loader_state' has no member named 'slot_usage'
[Error] ram_load.c@408,10: 'struct boot_loader_state' has no member named 'slot_usage'
[ERROR] .\mcuboot\boot\bootutil\src\ram_load.c:48:2: error: #error "Platform MUST define executable RAM bounds in case of RAM_LOAD"
```
[...]

PR #2152, opened 2024-12-18
Update Mbed OS CMake build file for Mbed CE

multiplemonomials:
Hi, I'm Jamie, maintainer of [Mbed CE](https://github.com/mbed-ce), the community fork of Mbed. As Mbed OS 6 has been end-of-lifed by ARM, we hope to keep support for the OS alive in our fork for a long time coming!

Several members of the project have expressed interest in getting mcuboot running with Mbed CE, so I took some time and put together a [test project](https://github.com/mbed-ce/mbed-mcuboot-bootloader) based off of George Beckstein's original work.  I was able to get it working more or less out of the box, but I did have to make some build system changes to get things running.

In particular, I wasn't happy with how the original design required running manual commands after compiling the code.  That was an unfortunate necessity with the original Mbed 5/6 build system, but with Mbed CE, we use standard CMake, so we can improve this situation!  I added a CMake script which automates calling imgtool to generate initial images, update images, and the public key file needed to build the bootloader.

With these changes, I am able to get a [rudimentary demo application](https://github.com/mbed-ce/mbed-mcuboot-demo-app) booting and updating.  I and the others on the mbed-ce project (including @lefebvresam and @zhiyong-ft) hope to test out (and document) more features in time!

On another note, we heard from [this issue](https://github.com/mcu-tools/mcuboot/issues/2129) that there is discussion about removing the Mbed port of mcuboot.  We at mbed-ce are happy to help maintain and test the port going forward, and really want it to stay in the main branch of mcuboot!

**Compatibility note:** This PR will maintain compatibility with Mbed OS 6 using the Mbed CLI build system and/or Mbed Studio.  However, it will break compatibility with the Mbed CLI 2 CMake-based build system.  This build system never really reached maturity before ARM Mbed shut down, so I doubt that too many people will be using it who do not use Mbed CE.


boot/bootutil/src/swap_scratch.c
```diff
@@ -174,6 +174,13 @@ boot_slots_compatible(struct boot_loader_state *state)
     j = sz1 = secondary_slot_sz = 0;
     smaller = 0;
     while (i < num_sectors_primary || j < num_sectors_secondary) {
+        // Make sure that the boot_img_sector_size() calls below don't access undefined memory.
```
multiplemonomials, line 177: I ran into a rather nasty issue when testing where I didn't set BOOT_MAX_IMG_SECTORS large enough, so this for loop exhibited undefined behavior.  Throwing in a check against this.

nordicjm: this doesn't even make sense, this is already checked on line 156 above

boot/bootutil/src/swap_scratch.c
```diff
@@ -801,6 +808,13 @@ int app_max_size(struct boot_loader_state *state)
...
+        if(i >= BOOT_MAX_IMG_SECTORS || j >= BOOT_MAX_IMG_SECTORS)
+        {
+            BOOT_LOG_WRN("Cannot upgrade: BOOT_MAX_IMG_SECTORS too small");
```
nordicjm, line 814: as above

nordicjm (changes requested):

boot/bootutil/src/swap_scratch.c
multiplemonomials: Hmm, I checked in the debugger, and if I set BOOT_MAX_IMG_SECTORS to 1000, on a device that uses 1536, I see: 
image.

Looking into it a bit more, it seems this may be an issue due to the Mbed version of flash_area_get_sectors() never returning a `count` value greater than `MCUBOOT_MAX_IMG_SECTORS`.  That causes the existing warning (which I totally didn't notice X_X) to trigger.

multiplemonomials: ok pushed a fix!

boot/mbed/CMakeLists.txt
```diff
@@ -2,9 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Mbed-MCUboot Port
-
-cmake_minimum_required(VERSION 3.19.0 FATAL_ERROR)
```
nordicjm, line 6: this should still be here specifying the minimum version needed

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -0,0 +1,115 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+check_python_package(imgtool.main MCUBOOT_IMGTOOL_FOUND)
+
+get_filename_component(IMGTOOL_SCRIPTS_DIR ${CMAKE_CURRENT_LIST_DIR}/../../scripts REALPATH)
+
+# Find or install imgtool ------------------------------------------------------------------------
+
+if(NOT MCUBOOT_IMGTOOL_FOUND)
+
```
nordicjm, line 11:
```suggestion
```

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2024 Jamie Smith
+# SPDX-License-Identifier: Apache-2.0
+
+check_python_package(imgtool.main MCUBOOT_IMGTOOL_FOUND)
+
+get_filename_component(IMGTOOL_SCRIPTS_DIR ${CMAKE_CURRENT_LIST_DIR}/../../scripts REALPATH)
+
+# Find or install imgtool ------------------------------------------------------------------------
```
nordicjm, line 8: dashes not needed on comment

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -0,0 +1,115 @@
...
+
+if(NOT MCUBOOT_IMGTOOL_FOUND)
+
+    # If we are using the Mbed venv, we can install asn1tools automatically
```
nordicjm, line 12: cmake indent is 2 spaces not 4

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -0,0 +1,115 @@
...
+# This initial image is what should be flashed to a blank device (along with the bootloader).
+#
+# NOTE: This function must be called *after* mbed_set_post_build() for the target!
+#
+#
+# If you wish to specify the output path of the initial image, pass that as the second argument to
+# this function.  Otherwise, it will default to ${CMAKE_CURRENT_BINARY_DIR}/<target name>-initial-image.hex
+#
+function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: file name
+
```
nordicjm, line 42:
```suggestion
```

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -0,0 +1,115 @@
...
+    endif()
+    set(SLOT_SIZE_HEX ${CMAKE_MATCH_1})
+
+    add_custom_command(
+        TARGET ${TARGET}
+        POST_BUILD
+        COMMAND
+            ${Python3_EXECUTABLE} -m imgtool.main
+            sign
+
```
nordicjm, line 88: empty lines here should be removed, this is all one command

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -76,6 +66,18 @@ function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: file name
     endif()
     set(SLOT_SIZE_HEX ${CMAKE_MATCH_1})
 
+    get_property(objcopy GLOBAL PROPERTY ELF2BIN)
+
+    if(${IMAGE_TYPE} STREQUAL "initial")
+        set(IMGTOOL_EXTRA_ARGS
+            # this should only be used for binaries you plan on initially flashing to your target at the factory.
+            # It pads the resulting binary to the slot size and adds initialized trailer TLVs. This is not needed
+            # for update binaries.
```
nordicjm, line 75: this isn't needed for swap modes by default, they should boot without it, it's needed for directXIP

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -98,18 +100,72 @@ function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: file name
...
+function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: initial image base filename
+    # Figure out file name
+    if("${ARGN}" STREQUAL "")
+        set(INITIAL_IMAGE_BASENAME $<TARGET_FILE_BASE_NAME:${TARGET}>-initial-image)
+    else()
+        set(INITIAL_IMAGE_BASE_PATH ${ARGN})
+    endif()
+
+    _mcuboot_generate_image(${TARGET} initial ${INITIAL_IMAGE_BASENAME})
+
```
nordicjm, line 153:
```suggestion
```

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -98,18 +100,72 @@ function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: file name
...
\ No newline at end of file
+    # Create a flash target.
+    # We need to be slightly creative here -- Mbed thinks that the application start address
+    # is <primary slot address> + <header size>, but we actually want to upload to <primary slot address>.
+    # So we need to temporarily override MBED_UPLOAD_BASE_ADDR with an offset value
+    math(EXPR MBED_UPLOAD_BASE_ADDR "${MBED_UPLOAD_BASE_ADDR} - ${HEADER_SIZE_HEX}" OUTPUT_FORMAT HEXADECIMAL)
+
+    gen_upload_target(${TARGET}-${IMAGE_TYPE}-image ${IMAGE_BASE_PATH}.bin)
+    add_dependencies(flash-${TARGET}-${IMAGE_TYPE}-image ${TARGET})
+
```
nordicjm, line 132:
```suggestion
```

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -98,18 +100,72 @@ function(mcuboot_generate_initial_image TARGET) # optional 2nd arg: file name
...
+function(mcuboot_generate_update_image TARGET) # optional 2nd arg: update image base filename
+    # Figure out file name
+    if("${ARGN}" STREQUAL "")
+        set(UPDATE_IMAGE_BASENAME $<TARGET_FILE_BASE_NAME:${TARGET}>-update-image)
+    else()
+        set(UPDATE_IMAGE_BASENAME ${ARGN})
+    endif()
+
+    _mcuboot_generate_image(${TARGET} update ${UPDATE_IMAGE_BASENAME})
+
```
nordicjm, line 175:
```suggestion
```

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -107,10 +107,14 @@ function(_mcuboot_generate_image TARGET IMAGE_TYPE IMAGE_BASENAME)
 
             ${TARGET_HEX_FILE} ${IMAGE_BASE_PATH}.hex
 
+        COMMAND ${CMAKE_COMMAND} -E echo "-- built: ${IMAGE_BASE_PATH}.hex"
+
         # Also generate bin file
         COMMAND
             ${objcopy} -I ihex -O binary ${IMAGE_BASE_PATH}.hex ${IMAGE_BASE_PATH}.bin
 
+        COMMAND ${CMAKE_COMMAND} -E echo "-- built: ${IMAGE_BASE_PATH}.bin"
```
nordicjm: why do you need echos here?

boot/mbed/mcuboot_imgtool.cmake
```diff
@@ -172,4 +173,25 @@ function(mcuboot_generate_update_image TARGET) # optional 2nd arg: update image
...
+            getpub
+            --key ${MCUBOOT_SIGNING_KEY_ABSPATH}
+            --output ${SIGNING_KEYS_C_PATH}
+
+        DEPENDS ${MCUBOOT_SIGNING_KEY_ABSPATH}
+        COMMENT "Converting signing key to C source..."
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        VERBATIM
+    )
+
```
nordicjm, line 196: as above

boot/mbed/src/flash_map_backend.cpp
```diff
@@ -215,11 +215,15 @@ int flash_area_get_sectors(int fa_id, uint32_t* count, struct flash_sector* sect
...
     bd_addr_t offset = 0;
     *count = 0;
-    while (*count < MCUBOOT_MAX_IMG_SECTORS && bd->is_valid_read(offset, bd->get_read_size())) {
+    while (bd->is_valid_read(offset, bd->get_read_size())) {
 
-        sectors[*count].fs_off = offset;
         bd_size_t erase_size = bd->get_erase_size(offset);
-        sectors[*count].fs_size = erase_size;
+
+        if(*count < MCUBOOT_MAX_IMG_SECTORS)
```
nordicjm, line 222: spaces needed in c code i.e. `if (`

nordicjm (commented):
changes need to be squashed

boot/mbed/src/flash_map_backend.cpp
```diff
@@ -215,11 +215,15 @@ int flash_area_get_sectors(int fa_id, uint32_t* count, struct flash_sector* sect
...
+        {
```
nordicjm, line 223: also as code above, bracket goes on same line as if

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: The empty lines here are important for grouping the comments with the lines they are commenting

boot/mbed/CMakeLists.txt
multiplemonomials: Technically this line is only needed in CMake scripts that can be executed as a top-level project.  But it's not doing any harm so I will re-add it.

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: removed

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: Hmm, interesting.  I got this from George's docs, so I guess he was wrong about that?  Does that mean that one can use the same image for the initial flash and for the update?

boot/mbed/src/flash_map_backend.cpp
multiplemonomials: done

boot/mbed/src/flash_map_backend.cpp
multiplemonomials: done

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: converted to 2

boot/mbed/mcuboot_imgtool.cmake
zhiyong-ft: I am not sure if Mr. Beckstein is wrong here. At least as of v1.8.0, if a padded image is downloaded into secondary slot, bootloader will upgrade from it even primary application hasn't called `boot_set_pending()` yet. Could you check what `boot_set_pending()` does? I assume it adds a bunch of TLVs so bootloader can recognize and accept the image in secondary slot.

multiplemonomials: I think what @nordicjm is saying is that we should remove this line for everything and just use the "update" image for all use cases

zhiyong-ft:
@multiplemonomials I tested encrypted image with 0fa393ad641ce651f7380c10dbb6def26e5ffe9e as bootloader, and mcuboot v1.8.0 included in primary application on DISCO_F746NG. This feature seems working now. I guess there was a bug in v1.8.0 to upgrade from encrypted image. I will wait until you are done with this batch of revision to update documentation.

A minor issue: is `target.restrict_size` no longer honored? I have this in mbed_app.json5:
```
 "target.restrict_size": "0x1C000",
```
The build bootloader hex file has content up till 0x1D034.

Other than this I don't see other issues as far as bootloader is concerned.

lefebvresam:
If you use the new methodology with flash bank settings, then:

```
            // This replaces "target.restrict_size":
            "target.memory_bank_config": {
                "FLASH": {
                    "size": 0xBE000,
                    "start": 0x8021000
                }
            }
```

zhiyong-ft:
> @zhiyong-ft If you use the new methodology with flash bank settings, then:
> 
> ```
>             // This replaces "target.restrict_size":
>             "target.memory_bank_config": {
>                 "FLASH": {
>                     "size": 0xBE000,
>                     "start": 0x8021000
>                 }
>             }
> ```
> 
> Size without Application header (0x1000) and Trailer TLV's (0x1000), total slot size 0xC0000 in my case. Start at the end of Application Header, which is at 0x8020000.

Could you point me the documentation on this? I tried to build the bootloader for NUCLEO_F767ZI, starting address in hex file is: `0x200000`, while the correct starting address should be `0x8000000`. I suspect this has something to do with this setting not being set.

lefebvresam:
@zhiyong-ft https://github.com/mbed-ce/mbed-os/issues/401#issuecomment-2526407974

boot/mbed/mcuboot_imgtool.cmake
nordicjm: use the cmake `message()` function https://cmake.org/cmake/help/latest/command/message.html#general-messages with `STATUS`

boot/mbed/mcuboot_imgtool.cmake
nordicjm: It's not in-fitting with the project e.g. https://github.com/mcu-tools/mcuboot/blob/main/boot/espressif/CMakeLists.txt#L407 and https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/CMakeLists.txt#L372 so should be updated to match

boot/mbed/mcuboot_imgtool.cmake
nordicjm: As it's done in zephyr, you build once without confirmation and flash the hex file, there is a bin file equivalent which you use for firmware updates (that you must mark for update using test or confirmed) - this applies to swap modes since they use the flags. DirectXIP does not use any flags, it will just boot the newest image. The exception is directXIP with revert, this requires the confirmed flag even on the initial image as it uses that to determine if the image is valid and to be booted

nordicjm:
Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: Ah, unfortunately message() can only print stuff at configure time.  I want this message to print at build time after the hex and bin files have been built.

lefebvresam:
> Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used

v1.8.0 from what?

nordicjm:
> > Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used
> 
> v1.8.0 from what?

MCUboot: https://github.com/mcu-tools/mcuboot/releases/tag/v1.8.0

lefebvresam:
> > > Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used
> > 
> > 
> > v1.8.0 from what?
> 
> MCUboot: https://github.com/mcu-tools/mcuboot/releases/tag/v1.8.0

I am using the fork:

```
[submodule "mcuboot"]
	path = mcuboot
	url = https://github.com/multiplemonomials/mcuboot.git
	branch = dev/get-version-number
```

zhiyong-ft:
> Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used

The unfortunate fact is that before @multiplemonomials started to work on it, v1.8.0 was the last working version that can build in Mbed Studio and upgrade unencrypted image from secondary slot, yet crashes when trying to upgrade from encrypted image. But this is behind us now.

lefebvresam:
> > > > Just to point out that v1.8.0 is **3 years old** at this point, it should not be being used
> > > 
> > > 
> > > The unfortunate fact is that before @multiplemonomials started to work on it, v1.8.0 was the last working version that can build in Mbed Studio and upgrade unencrypted image from secondary slot, yet crashes when trying to upgrade from encrypted image. But this is behind us now.
> > 
> > 
> > Using encrypted images will be my next step. And returning from using mbed-ce is no option anymore. My firmware is almost finished and I spent two years on it.
> 
> I outlined steps needed to get encrypted image to work in another thread. Give it a try. It is rather straightforward once we know it is working. It took me a while to map out the kinks because it was not mentioned at all in Mr. Beckstein's original port. He too seems have had some trouble.

Can you paste a link to that?

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: gotcha, removed.

boot/mbed/mcuboot_imgtool.cmake
multiplemonomials: removed blanks

multiplemonomials:
@zhiyong-ft @nordicjm I have updated this PR in response to comments and to add a way to generate encrypted images.  Should be ready for another look!

zhiyong-ft:
@multiplemonomials I can confirm that mcuboot as of ce340d5722ee4898137a1e8106988399db20bc53 can boot primary application and upgrade encrypted images from secondary slot.

boot/mbed/CMakeLists.txt
```diff
@@ -33,18 +33,16 @@ target_link_libraries(${LIB_TARGET}
...
-    PUBLIC
-        mbed-baremetal
-    )
+# Add signing key generated source file
+mcuboot_generate_signing_keys_file(${CMAKE_CURRENT_BINARY_DIR}/signing_keys.c)
+target_sources(${LIB_TARGET} PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/signing_keys.c)
+
+if("MCUBOOT_ENCRYPT_RSA=1" IN_LIST MBED_CONFIG_DEFINITIONS)
+	mcuboot_generate_encryption_key_file(${CMAKE_CURRENT_BINARY_DIR}/enc_keys.c)
+	target_sources(${LIB_TARGET} PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/enc_keys.c)
```
nordicjm, line 45: 2 space indent for cmake

nordicjm (approved):
minor nit. looks ok

boot/mbed/CMakeLists.txt
multiplemonomials: Updated.  Also could you approve the workflows if you have a sec?

multiplemonomials:
@nordicjm updated to add signed-off-by, could you trigger CI again?

multiplemonomials:
Any idea why the Zephyr check is failing?  I don't believe it's related to anything I did...

nordicjm:
> Any idea why the Zephyr check is failing? I don't believe it's related to anything I did...

You need to rebase your branch, that issue has been fixed

multiplemonomials:
Oh, gotcha.  Done!

multiplemonomials:
@nordicjm Would you be able to run the workflows?  Also I pushed a minor update to fix a CMake issue and a multiple macro def error.

lefebvresam:
I also like clear instructions with no ambiguities. cfr my earlier questions: why giving key if prim image is not encryped? Why giving pem file which contains priv and public key? etc.

multiplemonomials:
Well, the pem file is used by the cmake scripts to export the puclic key.  If you really wanted, you could just pass the public key as a C source file for the signing key.

lefebvresam:
The unlogic thing is:
For signing: `imgtool.py sign -k signing_keys.pem` -> both keys
For encryption: `-E enc_key_pub.pem` -> only a pub key

zhiyong-ft:
> Also, quick question. George Beckstein's original documentation said that the header size had to be set (via the argument to imgtool) to be one entire flash sector.
> 
> However, there are some devices, such as STM32H7, where the flash sector size is so large (128kB) that imgtool cannot encode a header that large. Also, I tried not setting the header size just to see what would happen, and it seems like it works fine without making the header a complete flash sector.
> 
> So, is this actually a requirement? It seems like it is not and I should remove this from our documentation.

I am a bit annoyed by this too. I think it all boils down to if bootloader needs to make any changes to header section after initial writing. If bootloader every needs to write something into header, then it has to erase at least an entire erasable sector. I don't know the source code of mcuboot enough to make a judgement call here.

nordicjm:
> Also, quick question. George Beckstein's original documentation said that the header size had to be set (via the argument to imgtool) to be one entire flash sector.
> 
> However, there are some devices, such as STM32H7, where the flash sector size is so large (128kB) that imgtool cannot encode a header that large. Also, I tried not setting the header size just to see what would happen, and it seems like it works fine without making the header a complete flash sector.
> 
> So, is this actually a requirement? It seems like it is not and I should remove this from our documentation.

That is not correct, the header size should be a multiple of the vector table size, which on a cortex M4 default design is 512 bytes. The header is never rewritten

nordicjm:
> The unlogic thing is: For signing: `imgtool.py sign -k signing_keys.pem` -> both keys For encryption: `-E enc_key_pub.pem` -> only a pub key If only pub is needed for signing why not encfore this?
> 
> And signing_keys.pem or enc_keys.pem is the source for priv and public key but in these files I only see:
> 
> ```
> -----BEGIN PRIVATE KEY-----
> blabla
> -----END PRIVATE KEY-----
> ```

Bootloader has public signing key and full key for encryption (encryption is symmetrical, there is no public key). Private signing key is needed to sign images, public key can only verify them, thus what is needed is correct

zhiyong-ft:
> That is not correct, the header size should be a multiple of the vector table size, which on a cortex M4 default design is 512 bytes. The header is never rewritten

Thanks for your clarification, this is such a relief. Some of MCUs I work with have 256KB sectors. 

Could you also comment on if we should use a separate key to encrypt images or we can safely use the public derived from the signing key for encryption as well? I have tried, just want to know which is better practice from security's perspective.

nordicjm:
> > That is not correct, the header size should be a multiple of the vector table size, which on a cortex M4 default design is 512 bytes. The header is never rewritten
> 
> Thanks for your clarification, this is such a relief. Some of MCUs I work with have 256KB sectors.
> 
> Could you also comment on if we should use a separate key to encrypt images or we can safely use the public derived from the signing key for encryption as well? I have tried, just want to know which is better practice from security's perspective.

You would need to use overwrite only mode for those devices since they cannot store the swap status information, so you have no ability to revert (or alternatively direct-xip). You can't use the same key no, signing is an asymmetrical key of RSA/ED25519/ECDSA-P256, encryption is a symmetrical AES key

zhiyong-ft:
@multiplemonomials I can confirm that header doesn't have to occupy an entire erasable sector. On STM32F767ZI, the following parameters in 'mbed_app.json` of primary application works just fine:
```
"mcuboot.primary-slot-address": "0x8018000",
 "mcuboot.slot-size": "0xA8000",
 "target.mbed_app_start": "0x8019000",
 "target.mbed_app_size": "0xA7000",
```
First 4KB of that 32KB sector is reserved for header. With such setup, bootloader can boot primary application, primary application can call `boot_set_pending()` to trigger firmware upgrade, and rollback to previous version. I think you can safely update documentation to remove restriction on header.

multiplemonomials:
@nordicjm Thanks for the info, I have updated our docs. I'm good to merge if you are!

nordicjm:
Keys should be different for signing and encryption

PR #2161, opened 2024-12-24
[RFC] Hooks to support choosing image to boot in runtime

edersondisouza:
[An alternative to https://github.com/mcu-tools/mcuboot/pull/2044]

In order to allow an MCUBoot user to have more control on where to get image to boot - eventually bypassing all of MCUBoot compare and swap mechanism - but still preserving integrity (and other) checks provided by MCUBoot, this PR introduces a few hooks that an application can implement to achieve such control.

The main hooks introduced allow one to modify `boot_go` - the method that effectively selects an image to boot and some methods related to Flash Area ID/Device ID - so that applications are no constrained to statically defined flash partitions, but have some runtime leeway - enabling, for instance, images to decide which flash partition is going to be used based on GPIO state.

New configurations are used to enable these hooks, so they can be enabled independently from currently available hooks.

This PR also provides a sample of how those hooks could be used.

Finally, there are some patches exposing "RAM loading" functions so they can be used by an external module - while not strictly needed on this PR, they are used in the sample. If deemed better, I can extract those patches to another PR.

As always, comments and suggestions on how to better achieve this are highly appreciated =D


samples/runtime-source/zephyr/app/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(non_flash_backend_app)
+
+if (NOT DEFINED FROM_WHO)
```
nordicjm, line 8:
```suggestion
if(NOT DEFINED FROM_WHO)
```

samples/runtime-source/zephyr/app/boards/frdm_k64f.overlay
```diff
@@ -0,0 +1,5 @@
+/ {
+    chosen {
+        zephyr,code-partition = &slot1_partition;
+    };
```
nordicjm, line 4: tab indents for dts

samples/runtime-source/zephyr/app/prj.conf
```diff
@@ -0,0 +1,3 @@
+CONFIG_BOOTLOADER_MCUBOOT=y
+
+CONFIG_MCUBOOT_SIGNATURE_KEY_FILE="./bootloader/mcuboot/root-rsa-2048.pem"
```
nordicjm, line 3: this will be the default so can be dropped

samples/runtime-source/zephyr/app/src/main.c
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/sys/printk.h>
+
+int main(void)
+{
+	printk("Hello World from %s on %s, slot %s!\n",
+	       MCUBOOT_HELLO_WORLD_FROM, CONFIG_BOARD,
+           DT_PROP(DT_CHOSEN(zephyr_code_partition), label));
```
nordicjm, line 14: tab for main indent then spaces to align

samples/runtime-source/zephyr/app/src/main.c
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2017 Linaro, Ltd.
```
nordicjm, line 2: might want to update copyright

samples/runtime-source/zephyr/hooks/hooks.c
```diff
@@ -0,0 +1,116 @@
...
+static int current_id;
+
+#define SW1_NODE        DT_ALIAS(sw1)
+#if DT_NODE_HAS_STATUS(SW1_NODE, okay)
+static struct gpio_dt_spec sw1_spec = GPIO_DT_SPEC_GET(SW1_NODE, gpios);
+#endif
+
+fih_ret boot_go_hook(struct boot_rsp *rsp)
+{
+    int rc;
```
nordicjm, line 34: tabs

samples/runtime-source/zephyr/hooks/hooks.c
```diff
@@ -0,0 +1,116 @@
...
+
+        flash_area_close(_fa_p);
+        break;
+    }
+
+    FIH_RET(fih_rc);
+}
+
+int flash_area_id_from_multi_image_slot_hook(int image_index, int slot,
+        int *area_id)
```
nordicjm, line 104: align to `int image_index`

samples/runtime-source/zephyr/sample.conf
```diff
@@ -0,0 +1,8 @@
+CONFIG_FLASH_RUNTIME_SOURCES=y
+CONFIG_SINGLE_APPLICATION_SLOT=y
+CONFIG_BOOT_SIGNATURE_TYPE_RSA=y
```
nordicjm, line 3: needs a sample.yml file so that twister will build this

nordicjm (approved):
Some nits. in the sample with styling, other than that looks good

nordicjm:
rebase will fix the CI error too

samples/runtime-source/zephyr/sample.conf
nordicjm: would also need adding here https://github.com/mcu-tools/mcuboot/blob/main/.github/workflows/zephyr_build.yaml#L86

edersondisouza:
v2:
  - Fixed style issues;
  - Made sample build-able by twister.

samples/runtime-source/zephyr/app/prj.conf
edersondisouza: Hmmm... got this when removed this line:

```
CMake Warning at /home/ederson/work/zephyr/zephyr-west/zephyr/cmake/mcuboot.cmake:28 (message):
  Neither CONFIG_MCUBOOT_GENERATE_UNSIGNED_IMAGE or
  CONFIG_MCUBOOT_SIGNATURE_KEY_FILE are set, the generated build will not be
  bootable by MCUboot unless it is signed manually/externally.
```

So I kept it.

samples/runtime-source/zephyr/hooks/hooks.c
```diff
@@ -0,0 +1,116 @@
...
+#include <zephyr/drivers/gpio.h>
+#include <zephyr/kernel.h>
+
+#include "bootutil/bootutil.h"
+#include "bootutil/bootutil_public.h"
+#include "bootutil/fault_injection_hardening.h"
+
+#define BOOT_TMPBUF_SZ 256
+
+static const struct flash_area *_fa_p;
```
de-nordic, line 16: why is this not a function local variable?

de-nordic (commented):
Looks ok, some minor problem.
Just one note though: I am working on removing usage of flash area and flash area id within MCUboot, so in the long term the hooks will be affected, although there will be compatibility mote. Just letting you know.

The replacement will actually require systems to provide own implementation of open function that will have to map from MCUboot ids to system specific, which means that proposed here hook `flash_area_id_from_multi_image_slot_hook` will loose its importance, but you will be able to write your open function to override ids however you want.

This will take a moment to get completed so I think that for now we are good with the `flash_area_id_from_multi_image_slot_hook`,

samples/runtime-source/zephyr/hooks/hooks.c
```diff
@@ -0,0 +1,116 @@
...
+											 int *area_id)
+{
+	*area_id = known_ids[current_id];
+
+	return 0;
+}
+
+int flash_area_get_device_id_hook(const struct flash_area *fa, uint8_t *dev_id)
+{
+	*dev_id = fa->fa_id;
```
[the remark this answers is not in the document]
edersondisouza: Hmmm... the value returned by `flash_area_get_device_id` ends up being used (at least for Zephyr) as parameter for `flash_area_open` or `flash_device_base`. For the later, yes, it is the flash device. However, for the former, it ends up being the fixed partition id - kinda circular, but IIUC, that's how it goes.

Indeed, this is the same pattern found at https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/flash_map_extended.c#L145 for ARM. In this sample, one could even ask if there's a point for the `flash_area_get_device_id_hook`, as it's for ARM so it would just get `fa_id` anyway, but I'm also looking at other platforms (that may or may not need more patches in the future).

edersondisouza:
v3:
  - Make some variables local;
  - Use getter to access object fields.

samples/runtime-source/zephyr/hooks/hooks.c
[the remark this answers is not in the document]
de-nordic: I need to check that, but as far as I can tell the fa_id is used with flash_area_open but the flash_area_get_device_id is not. The getter for fa_id should be flash_area_get_id. The function is here provided because Zephyr does not have one, does not have dev id l.
There might be a bug here as I you can actually see in the code, that does the ifdef on arm, that the id for non-arm devices would be hardcoded to single value; additionally Zephyr devices are identified by flash API by device pointer, and the id is only for the partition.

The current implementation of flash_area_get_device_id has inconsistent behavior here between architectures: why would ARM have these ids but others not? That is not even described in api.

Give me a few moments to take another look at that.

boot/bootutil/src/loader.c
```diff
@@ -110,6 +110,11 @@ static struct sector_buffer_t sector_buffers;
 #define BUF_SZ 1024
 #endif
 
+struct boot_loader_state *boot_get_loader_state(void)
+{
+    return &boot_data;
+}
```
nordicjm, line 116: actually after having a talk with David about other issues I don't think this will work, @d3zd3z

edersondisouza: Hmm... why? My goal here is to have some state that can be prepared and passed to `boot_load_image_to_sram` without needing to expose `struct boot_loader_state` itself. I went with getting a static one instead of creating a new one to avoid allocation. Note that the one that matters to me is the `boot_get_loader_state` at `boot/zephyr/single_loader.c` - which is there just to be shared - instead of the one here at bootutil, which I only added for completeness.

samples/runtime-source/zephyr/hooks/hooks.c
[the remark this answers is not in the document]
edersondisouza: Hi @de-nordic! Any news here?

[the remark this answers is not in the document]
de-nordic: OK, sorry for lagging. We have a bug. Here is only function, Zephyr specific implementation, that takes __device id__:
https://github.com/mcu-tools/mcuboot/blob/06747985eee7c27d891f2529c58e18f4a84ea6a3/boot/zephyr/flash_map_extended.c#L43-L52
It is supposed to get device base == where flash address space starts.
As you can see it is hardcoded to be able to process only internal flash (at this point).
The device id is not present in Zephyr variant of flash_area structure, but is provided by cypress, embed and otherr flash area backends provided as part of support for these systems.
If you look at any of these you will notice, that this is set to a device id not flash area id:
https://github.com/mcu-tools/mcuboot/blob/06747985eee7c27d891f2529c58e18f4a84ea6a3/boot/cypress/cy_flash_pal/cy_flash_map.c#L76-L82

This can not be flash area device `flash_device_base` would not work properly for Zephyr. Frankly the only usage, I can see, is in do_boot that is for RISC V and therefore the `CONFIG_ARM` ifdef in the flash_area_get_device_id does not break anything.

edersondisouza:
v4:
 - Keep default behaviour for `flash_area_get_device_id_hook`

samples/runtime-source/zephyr/hooks/hooks.c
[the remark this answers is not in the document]
edersondisouza: Got it. For the sample in this PR, I can simply use default behaviour, so it won't "empower" more confusion on flash area id vs device id.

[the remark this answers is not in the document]
de-nordic: Logged an issue here https://github.com/mcu-tools/mcuboot/issues/2188

de-nordic (approved):

edersondisouza:
v5:
  - Rebased (and fixed conflicts/issues from it).

de-nordic (approved):

boot/bootutil/src/loader.c
d3zd3z: The use of a static variable will prevent this code from being run in the simulation environment, where the code is invoked by multiple threads simultaneously.

de-nordic: > The use of a static variable will prevent this code from being run in the simulation environment, where the code is invoked by multiple threads simultaneously.

Unfortunately I am not that fluent with the simulator, so unable to comment here.
@d3zd3z Do you find this to be a blocker? How should we deal with this, have you any insight?

d3zd3z: So the issue with the statics is that we want the state to be static on target, but only gotten at the top level and passed down. This is effectively just moving this definition from the zephyr-specific code, although I question then why the declaration remains within the Zephyr code as well.

I'm not quite sure I understand why this gets added here as well. If it does get moved to loader.c, we should remove the one in the Zephyr-specific code. I'm not quite I follow it yet, but it almost looks at this point as if we now have two copies of the state, one in a zephyr-specific file, and now one here.

boot/zephyr/Kconfig
```diff
@@ -472,6 +472,17 @@ config BOOT_IMAGE_EXECUTABLE_RAM_SIZE
 	default $(dt_chosen_reg_size_int,$(DT_CHOSEN_Z_SRAM),0)
 endif
 
+config FLASH_RUNTIME_SOURCES
+	bool "Images are read from flash partitions defined at runtime"
+	select SINGLE_APPLICATION_SLOT
```
nordicjm, line 477:
```suggestion
	depends on SINGLE_APPLICATION_SLOT
```

samples/runtime-source/zephyr/app/prj.conf
nordicjm: right because not using sysbuild

nordicjm (commented):
Changes OK, just needs Kconfig changing

edersondisouza:
v6:
  - Use `depends on` instead of `select` for `SINGLE_APPLICATION_SLOT` at `FLASH_RUNTIME_SOURCES` config.

PR #2162, opened 2025-01-02
Add swap using offset mode

### LINKED ISSUE #2134: RFC: Swap using move but using extra sector in slot 1
The idea for this suggestion is to move the extra sector used for moving, to be part of slot 1 instead of slot 0, and to have a disjoint write location for MCUboot and the application which loads the firmware update: MCUboot deals with the first sector of this slot but the application will load a firmware update to the second slot of this slot. This means that when the update is initiated, the second slot image is already shifted up by a sector and no additional move needs to take place.

* Slot 0 sector X will be put into slot 1 sector X
* Slot 0 sector X will be erased
* Slot 1 sector (X + 1) will be put into slot 0 sector X
* X is incremented and loop runs until end of the image

This should speed up update times due to not having to shift the slot 0 image up by a sector, erasing a lot of erase and write processes. This would however mean that MCUboot would likely need a lot of changes to support this, slot 1 would need an additional sector instead of slot 0, and MCUmgr would need to be updated to write the update starting at the second sector for secondary slots, not the first sector

nordicjm:
Fixes #2134

165KiB update using swap using move on nrf5340dk: 18 seconds
182KiB update using swap using offset on nrf5340dk: 13 seconds


Laczen:
@nordicjm this is a very nice improvement. There are some minor documentation nits.

A further possible improvement could be to place the swap state in the update image slot (this will not be erased during the swap, and thus the image could go up to the slot size - trailer size).

Laczen:
Some food for further thoughts: I have created a similar approach outside of mcuboot, by using a different terminology:
A "backup slot" is defined that starts one sector before the "update slot" and has the same size of the "update slot" and the "image slot", so the backup slot covers a part of the "update slot" (ends one sector sooner).

The image update moves one sector from image to backup followed by a move from update to image, when the update is finished the old image is in the backup slot and the new image is in the image slot. A restore is a move from the backup slot to the image slot.

The method also allows removing the "swap with scratch" as the "scratch" sector can be defined as the first sector of the backup image.

For devices that have plenty of flash the backup slot can also be defined separate from the update slot to enhance the life and keep flash wear equal for all slots (update erases all flash only once).

nordicjm:
> The method also allows removing the "swap with scratch" as the "scratch" sector can be defined as the first sector of the backup image.

The plan is that swap using scratch will be removed in the coming future, but as it stands right now this would not work, e.g. on an stm32 where the sector sizes are 32KiB, 32KiB, 64KiB, 128KiB, 128KiB - making it the first sector give a scratch size of 32KiB which means you can't complete a single swap. The idea is that MCUboot's "sector size" will not be the flash sector size in future, it will be able to be configured so in the case of the stm32 will be 128KiB, in the case of a device with internal and external flash whereby internal flash has 2KiB sectors and external flash has 8KiB sectors, both will be 8KiB - once this change is made it means that swap using scratch can be removed entirely because despite the device have different sector sizes, the move is done in multiples of the largest sector size. The added benefit of this is that the code is smaller and swap status area is smaller too

nordicjm:
> Hi @nordicjm SWAP_USING_OFFSET is better than SWAP_USING_MOVE. Are there any cases where SWAP_USING_MOVE is preferable over OFFSET?

When building for existing products where the bootloader is configured in swap using move mode, other than that swap using offset should become the new default

docs/design.md
```diff
@@ -235,7 +235,48 @@ Where:
...
+used only for allowing sectors to move down during a swap. Therefore the most
+memory-size-effective slot layout is when the secondary slot is larger than
+the primary slot by exactly one sector, although same-sized slots are allowed
+as well. The algorithm is limited to support sectors of the same sector layout.
+All slot's sectors should be of the same size. This algorithm uses 2 flags per
+sector update rather than the 3 flags used by swap using move, which requires
+a smaller swap status area size.
+
+When using this algorithm the maximum image size available for the application
+will be the smallest of the following:
```
nvlsianpu, line 262: just: `will be:`?

docs/release-notes.d/swap-using-offset.md
```diff
@@ -0,0 +1,8 @@
+- Added a new swap using offset algorithm which is set with
+  `MCUBOOT_SWAP_USING_OFFSET`. This algorithm is similar to swap
+  using move but avoids moving the sectors in the primary slot
+  up by having the update image written in the second sector in
+  the update slot, which offers a faster update process and
+  requires a smaller swap status area
+- Made swap using offset the default algorithm for Zephyr builds
```
nvlsianpu, line 7: This is set as default if preferred... as it was with the move alg before

nvlsianpu (approved):
New swap alg. is introduced as it should be - new swap code is placed in dedicate source, changes are made in logical chunks and legible.
This implements what I had in my minds a few years ago.

boot/boot_serial/include/boot_serial/boot_serial_encryption.h
```diff
@@ -15,13 +15,18 @@
...
  * @return                FIH_SUCCESS on success, error code otherwise
  */
 fih_ret
 boot_image_validate_encrypted(const struct flash_area *fa_p,
                               struct image_header *hdr, uint8_t *buf,
-                              uint16_t buf_size);
+                              uint16_t buf_size
+#ifdef MCUBOOT_SWAP_USING_OFFSET
+                              , uint32_t start_off
+#endif
```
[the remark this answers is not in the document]
nordicjm: I added it with ifdef to not increase memory consumption for other modes, though it is indeed possible that the compiler might be smart enough to optimise it out automatically if it's always 0

boot/bootutil/src/swap_offset.c
```diff
@@ -0,0 +1,785 @@
...
+
+    if (check_other_sector == true && out_hdr->ih_magic != IMAGE_MAGIC &&
+        slot == BOOT_SECONDARY_SLOT) {
+        if (boot_swap_type_multi(BOOT_CURR_IMG(state)) != BOOT_SWAP_TYPE_REVERT) {
+            off = 0;
+        } else {
+            off = boot_img_sector_size(state, BOOT_SECONDARY_SLOT, 0);
+        }
+
+        rc = flash_area_read(fap, off, out_hdr, sizeof *out_hdr);
```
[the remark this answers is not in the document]
nordicjm: Leftover from the other swap file, fixed

docs/design.md
nordicjm: Fixed

docs/release-notes.d/swap-using-offset.md
nordicjm: Removed this line as it is not being made the default, can submit another PR after if we want to make it default

PR #2165, opened 2025-01-06
boot: espressif: add support for upgrade modes w/o scratch

martinjaeger:
Upgrade using move decreases flash wear and is now the default mode used by Zephyr.

The implementation is done by MCUboot itself, so it is just a matter of setting the correct define to enable it.

Swap using move was successfully tested with ESP32C6.


sylvioalves:
LGTM

boot/espressif/hal/include/mcuboot_config/mcuboot_config.h
```diff
@@ -45,23 +45,43 @@
  * the default upgrade mode.
  */
 
-/* Uncomment to enable the overwrite-only code path. */
-/* #define MCUBOOT_OVERWRITE_ONLY */
+/* Define to enable the swap-using-move code path. */
+#if defined(CONFIG_BOOT_SWAP_USING_MOVE)
+#define MCUBOOT_SWAP_USING_MOVE 1
+#endif
```
almir-okato, line 51: Since this is the Espressif Port `mcuboot_config.h` file, I suggest to keep the naming pattern, like `CONFIG_ESP_BOOT_SWAP_USING_MOVE`, same for the others below.
```suggestion
/* Define to enable the swap-using-move code path. */
#if defined(CONFIG_ESP_BOOT_SWAP_USING_MOVE)
#define MCUBOOT_SWAP_USING_MOVE 1
#endif
```

boot/espressif/hal/include/mcuboot_config/mcuboot_config.h
```diff
@@ -45,23 +45,43 @@
...
-/* #define MCUBOOT_DIRECT_XIP */
+/* Define to enable the direct-xip code path. */
+#if defined(CONFIG_BOOT_DIRECT_XIP)
+#define MCUBOOT_DIRECT_XIP
+#endif
 
 /* Define to enable the ram-load code path. */
 #if defined(CONFIG_BOOT_RAM_LOAD)
 #define MCUBOOT_RAM_LOAD
 #endif
```
almir-okato, line 69: These are currently undefined modes for Espressif Port, I would like to ask you to add a comment for both and avoid adding them to the .conf files.

boot/espressif/port/esp32/bootloader.conf
```diff
@@ -2,6 +2,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+# Define upgrade mode (default is CONFIG_BOOT_SWAP_USING_SCRATCH)
+# CONFIG_BOOT_SWAP_USING_MOVE=y
+# CONFIG_BOOT_UPGRADE_ONLY=y
+# CONFIG_BOOT_DIRECT_XIP=y
+# CONFIG_BOOT_RAM_LOAD=y
```
almir-okato, line 9:
```suggestion
```
Remove the undefined modes, same for the other .conf files.

boot/espressif/hal/include/mcuboot_config/mcuboot_config.h
martinjaeger: Ok, makes sense. Done.

almir-okato (approved):
LGTM

PR #2195, opened 2025-01-31
mbed: update cmake script to sign and encrypt images post build

zhiyong-ft:
Updated cmake script to automatically generate signing, encryption keys from user provided key-pairs then use these keys to sign and encrypt images. Also updated documentation on encrypted images and how to generate initial clear text images that support encryption during firmware upgrade. Referenced relevant discussion.


multiplemonomials:
@nordicjm @d3zd3z  Would you have a chance to take a look at this soon? This PR only touches docs and the CMake scripts used by Mbed, so it should be pretty quick I hope.

nordicjm (commented):
first commit is OK, second commit?? Drop, and merge commit, drop

PR #2196, opened 2025-02-03
boot: zephyr: add ram_load overlays for nrf52840dk and mimxrt1050_evk boards

danieldegrasse:
Add ram load overlays for nrf52840dk and mimxrt1050_evk boards. These ram load overlays are moved from the Zephyr in tree smp_svr sample, to enable ram load support for multiple platforms.

### DIFF
--- a/boot/zephyr/boards/nrf52840dk_nrf52840_ram_load.overlay
+++ b/boot/zephyr/boards/nrf52840dk_nrf52840_ram_load.overlay
@@ -1,0 +1,26 @@
+/ {
+	sram@2003FF00 {
+		compatible = "zephyr,memory-region", "mmio-sram";
+		reg = <0x2003FF00 0x100>;
+		zephyr,memory-region = "RetainedMem";
+		status = "okay";
+
+		retainedmem {
+			compatible = "zephyr,retained-ram";
+			status = "okay";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			boot_info0: boot_info@0 {
+				compatible = "zephyr,retention";
+				status = "okay";
+				reg = <0x0 0x100>;
+			};
+		};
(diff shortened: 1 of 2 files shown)


boot/zephyr/boards/nrf52840dk_nrf52840_ram_load.overlay
```diff
@@ -0,0 +1,26 @@
+/ {
+	sram@2003FC00 {
+		compatible = "zephyr,memory-region", "mmio-sram";
+		reg = <0x2003FC00 DT_SIZE_K(1)>;
```
nordicjm, line 4:
```suggestion
	sram@2003FF00 {
		compatible = "zephyr,memory-region", "mmio-sram";
		reg = <0x2003FF00 0x100>;
```

de-nordic: @danieldegrasse Can you respond or make changes, so that we can go through with this PR?

danieldegrasse: Sure, should be fixed now

danieldegrasse:
Is the CI failure here expected, because a new revision of Zephyr would be needed to make this overlay work correctly?

nordicjm:
> Is the CI failure here expected, because a new revision of Zephyr would be needed to make this overlay work correctly?

Yes failure here is fine since it's trying to build zephyr's `main` smp_svr which fails without changes in your PR

nordicjm (approved):

danieldegrasse:
@nordicjm what is needed to move this PR forwards? I'm not sure what makes the most sense to sequence this PR since it requires changes on the Zephyr side

PR #2197, opened 2025-02-05
Add support for RAMLOAD mode with revert

danieldegrasse:
This PR adds support for ram loading mode with revert to MCUBoot. This mode functions similar to direct XIP mode with revert- MCUBoot will now refuse to load an image to RAM unless the image trailer indicates that the image is confirmed or marked pending.

If multiple images are valid, the one with the highest version will be selected for ramloading, so downgrade is still not possible.

This PR also adds the `BLINFO_FLASH_AREA` TLV tag to the bootloader boot record, which records the flash area ID in the boot record. This can be used by systems like Zephyr to determine which slot was booted, which is useful for marking an image as confirmed or querying status

Note that this PR includes the commit from #2196, as I validated support for this feature using the RT1050-EVK. If desired, I can rebase this to drop that commit


boot/bootutil/include/bootutil/boot_status.h
```diff
@@ -119,6 +119,7 @@ extern "C" {
 #define BLINFO_MAX_APPLICATION_SIZE_IMAGE_2 0x07
 #define BLINFO_MAX_APPLICATION_SIZE_IMAGE_3 0x08
 #define BLINFO_MAX_APPLICATION_SIZE_IMAGE_4 0x09
+#define BLINFO_FLASH_AREA           0x0A
```
de-nordic, line 122: Isn't this already done vis BLINFI_RUNNING_SLOT?

boot/bootutil/src/boot_record.c
```diff
@@ -330,6 +330,12 @@ int boot_save_shared_data(const struct image_header *hdr, const struct flash_are
                                           BLINFO_RUNNING_SLOT,
                                           sizeof(slot), (void *)&slot);
     }
+
+    if (!rc) {
+        rc = boot_add_data_to_shared_area(TLV_MAJOR_BLINFO,
+                                          BLINFO_FLASH_AREA,
+                                          sizeof(fap->fa_id), (void *)&fap->fa_id);
+    }
```
de-nordic, line 338: Three issues:
 1) i am trying to disentangle MCUboot from Flash Map API as this entanglement blocks us from freely modifying both Zephyr and MCUboot, and so the usage of flash area id and flash_area within MCUboot is going to be phased out. There is also problem with flash area functions having different prototypes and flash_area objects having different members, depending on a system.
2) area id in MCUboot and Zephyr app that got boot does not have to be the same; it is enough if there is storage before images that gets removed with overlay in mcuboot and numbering will differ
3) mcuboot should not directly access flash_area objects, rather use getter functions like flash_area_get_id.

boot/bootutil/include/bootutil/boot_status.h
danieldegrasse: `RUNNING_SLOT` gives the index of the active slot, but for the MCUBoot subsystem code in Zephyr we need access to the flash area ID. For example, Slot 0 on the RT1050 has flash area ID 1 (because the MCUBoot partition has area ID 0)

We could change the value set by `RUNNING_SLOT`, but I wanted to avoid breaking any APIs so I added this. It was needed in order to get features like confirming a running image working with MCUBoot

boot/bootutil/src/boot_record.c
danieldegrasse: Ok, makes sense. I have no issue changing this, really all I want is a workable solution for indicating the flash area that Zephyr's MCUBoot subsystem should use when marking an image as confirmed. Is there an API for mapping the boot slot number back to a flash area we can use?

boot/bootutil/include/bootutil/boot_status.h
nordicjm: You have to convert from running slot to flash area ID in your application itself, as img mgmt does e.g. https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/zephyr_img_mgmt.c#L128 the flash area ID has no meaning between different applications e.g. application could have bootloader slot split into 2 in future, now you open the wrong thing

nordicjm (commented):
can use retention for running image slot so that commit should be dropped, can you also add a section to the documentation similar to direct-xip with revert https://github.com/mcu-tools/mcuboot/blob/main/docs/design.md?plain=1#L452 and add a file here https://github.com/mcu-tools/mcuboot/tree/main/docs/release-notes.d with a small note on adding this (see other files in directory for examples and for maximum line length and style, think it's 64 characters or so

boot/bootutil/include/bootutil/boot_status.h
danieldegrasse: Ahh, ok. I was testing this with the DFU subsystem- so would it make sense to call `img_mgmt_flash_area_id` here in `boot_fetch_active_slot`: https://github.com/zephyrproject-rtos/zephyr/pull/85254/files#diff-696b0049cfb981ee66f0191ccaf755f25a632c35b86da7dc1d739c3de8a60ed3L78? Otherwise, functions like marking the image as pending weren't working, since `boot_fetch_active_slot` is really fetching the flash area ID of the active slot, not the slot number.

nordicjm: That returns either:
```
#define BOOT_PRIMARY_SLOT               0
#define BOOT_SECONDARY_SLOT             1
```
Which should then be used to check if it's slot0 if 0 or slot1 if 1 (it doesn't care about additional images since they are not "running"),

djr-spectrummfg:
I would like it to be possible to downgrade images while using the RAM load mode

danieldegrasse:
@nordicjm PR should be updated- commit adding `BLINFO_FLASH_AREA` has been dropped, and the DFU subsystem has been updated in the Zephyr side PR to map the running slot number to a flash ID per your suggestion. Docs have also been added

danieldegrasse:
> @nordicjm PR should be updated- commit adding `BLINFO_FLASH_AREA` has been dropped, and the DFU subsystem has been updated in the Zephyr side PR to map the running slot number to a flash ID per your suggestion. Docs have also been added

@nordicjm friendly ping- hoping to try and move this forwards now that the Zephyr release is complete and merge window is open

nordicjm:
https://github.com/zephyrproject-rtos/zephyr/pull/85254 needs a rebase and this could do with one too. Have given this a try on an nrf52840dk and it does not work using the same from the zephyr PR:
```
*** Booting MCUboot v2.1.0-rc1-262-g2fc7c0ebcce7 ***
*** Using Zephyr OS build v4.0.0-4467-g59da4986e437 ***
I: Starting bootloader
I: Primary   slot: version=0.0.0+0
I: Image 0 Secondary slot: Image not found
D: Erasing faulty image in the primary slot.
I: No slot to load for image 0
E: Unable to find bootable image
```
Switching back to normal RAM load mode works:
```
*** Booting MCUboot v2.1.0-rc1-262-g2fc7c0ebcce7 ***
*** Using Zephyr OS build v4.0.0-4467-g59da4986e437 ***
I: Starting bootloader
I: Primary   slot: version=0.0.0+0
I: Image 0 Secondary slot: Image not found
I: Image 0 RAM loading to 0x20006000 is succeeded.
I: Image 0 loaded from the primary slot
I: Bootloader chainload address offset: 0x20006000
I: Image version: v0.0.0
I: Jumping to the first image slot
*** Booting Zephyr OS build v4.0.0-4467-g59da4986e437 ***
[00:00:00.000,457] <inf> smp_sample: build time: Mar 12 2025 09:01:31
```

nordicjm:
Seems to be `./smp_svr/zephyr/zephyr.signed.confirmed.hex` that is faulty, the `./smp_svr/zephyr/zephyr.slot1.signed.confirmed.hex` build file works

nordicjm:
Right I see the issue, the nrf52840dk overlay here would need to be changed to delete slot0 and slot1 then re-create them with the same slot sizes

danieldegrasse:
> Right I see the issue, the nrf52840dk overlay here would need to be changed to delete slot0 and slot1 then re-create them with the same slot sizes

Makes sense- I'll make that change here so that the sample can be tested on the `nrf52840dk`

danieldegrasse:
Updated this PR and the Zephyr side. Just a note- I tested this using the following sequence:
- flash `smp_svr` sample built with sysbuild (west build -p always -b mimxrt1050_evk//hyperflash -T samples/subsys/mgmt/mcumgr/smp_svr/sample.mcumgr.smp_svr.ram_load_revert.serial)
- rebuild, and load the new `zephyr.signed.bin` created for the `smp_svr` application using `mcumgr-client upload`
- mark the new application as testable using `mcumgr-client test`
- reset the target, and verify the new image is loaded in test mode (and will revert if it does not confirm)

danieldegrasse:
As far as sequencing goes, would it make sense to just combine this PR and #2196? Not sure if this will ease the process of getting both in, since they both need sequencing with Zephyr

nordicjm (approved):

d3zd3z (approved):

PR #2198, opened 2025-02-08
RFC: boot: bootutil: Swap offset: Do not fail on header in wrong sector

de-nordic:
Swap offset does not expect header in first sector/page of secondary slot, rather in secondary, and only attempts to swap an image that starts at second sector.
This commit demotes condition when header is found in first page to just a warning and allows boot to proceed, as long as there is a header is second slot.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1069,9 +1069,7 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
         }
 
         if (first_sector_hdr.ih_magic == IMAGE_MAGIC) {
-            BOOT_LOG_ERR("Secondary header magic detected in first sector, wrong upload address?");
-            fih_rc = FIH_NO_BOOTABLE_IMAGE;
-            goto check_validity;
+            BOOT_LOG_WRN("Secondary header magic detected in first sector, wrong upload address?");
         }
     }
 #endif


nordicjm (commented):
this should delete it as it currently does because the what has been uploaded is invalid, no different than if you upload a corrupt image to a slot then mark it for upgrade

de-nordic:
> this should delete it as it currently does because the what has been uploaded is invalid, no different than if you upload a corrupt image to a slot then mark it for upgrade

OK, but wouldn't it suffice to just remove the first page then?
I have stumbled to this issue when I had leftover header in slot from previous usage and the mcuboot refused to boot image programmed to the next page. I know it should never happen, but it is like you have stuck something irrelevant to the first page, that will not be used for anything, but MCUboot refuses to work.

nordicjm:
> > this should delete it as it currently does because the what has been uploaded is invalid, no different than if you upload a corrupt image to a slot then mark it for upgrade
> 
> OK, but wouldn't it suffice to just remove the first page then? I have stumbled to this issue when I had leftover header in slot from previous usage and the mcuboot refused to boot image programmed to the next page. I know it should never happen, but it is like you have stuck something irrelevant to the first page, that will not be used for anything, but MCUboot refuses to work.

needs to clear first 2 sectors and trailer sector at least, in original implementation just used the existing clear image erase code that was already present in the file. Hmm do you mean like with https://github.com/mcu-tools/mcuboot/pull/2199 ?

de-nordic:
> > > this should delete it as it currently does because the what has been uploaded is invalid, no different than if you upload a corrupt image to a slot then mark it for upgrade
> > 
> > 
> > OK, but wouldn't it suffice to just remove the first page then? I have stumbled to this issue when I had leftover header in slot from previous usage and the mcuboot refused to boot image programmed to the next page. I know it should never happen, but it is like you have stuck something irrelevant to the first page, that will not be used for anything, but MCUboot refuses to work.
> 
> needs to clear first 2 sectors and trailer sector at least, in original implementation just used the existing clear image erase code that was already present in the file. 

So, in my case there was a sequence on the same board:
1) I have tested offset
2) I had to do some work on swam-move
3) I have nrfjproged the same set I have used in 1), at once
4) reset and I got the error, even though the set from 1) worked fine, just that this time I have leftover header from 2).

So why to clear two sectors? The header in the second sector is what I have expected

> Hmm do you mean like with #2199 ?

Yeah, maybe. But that had to be a lot of bad things happening for the magic to appear properly.

PR #2201, opened 2025-02-11
scripts: imgtool: pretended encryption flag

michalek-no:
allows to set encryption flags and TLV's without applying encryption.


nordicjm (approved):

scripts/imgtool/image.py
nvlsianpu, line 1: Isn't --clear serves the same?

Issue #2052, opened 2024-08-29
Swap move should only need one extra sector


d3zd3z:
When using swap-move with large sectors, where the code is a single sector, it appears that there needs to be two extra sectors, in the first slot, rather than just one. This shouldn't need to be the case, and is especially wasteful when sectors are large.

As an example, the 128k sector ST device in the sim can be re-configured for swap-move, and it will fail.

nordicjm:
Well 2 extra sectors are always going to be required, not 2 complete sectors, but one complete sector for the swap and the image overhead, and since you **should** be able to clear the status of an image or set it as confirmed, then that area should be eraseable, thus 2 sectors

de-nordic:
> Well 2 extra sectors are always going to be required, not 2 complete sectors, but one complete sector for the swap and the image overhead, and since you **should** be able to clear the status of an image or set it as confirmed, then that area should be eraseable, thus 2 sectors

We do not erase status. What really happens is that the status is not copied when image is swapped that is why there is such complicated logic deciding on what should be executed; for example you may have two slots with images that have no flags set, which means you boot slot1, or primary slot has some flags set but secondary not - now you decide by bits in primary. If you have some flags set in primary and secondary then what happens depends on flags.
What we actually need is one sector + all the space needed for trailer part with flags.

nordicjm:
> > Well 2 extra sectors are always going to be required, not 2 complete sectors, but one complete sector for the swap and the image overhead, and since you **should** be able to clear the status of an image or set it as confirmed, then that area should be eraseable, thus 2 sectors
> 
> We do not erase status. What really happens is that the status is not copied when image is swapped that is why there is such complicated logic deciding on what should be executed; for example you may have two slots with images that have no flags set, which means you boot slot1, or primary slot has some flags set but secondary not - now you decide by bits in primary. If you have some flags set in primary and secondary then what happens depends on flags. What we actually need is one sector + all the space needed for trailer part with flags.

There is an option Kconfig for MCUmgr to delete updates marked for test, if you do not erase the status this means that the flags will remain the next time an image is loaded to the secondary slot if you cannot erase the flags part.

gschwaer:
[...]

PR #2206, opened 2025-02-16
boot: bootutil: Fix swap-scratch when the trailer is larger than the last sector

taltenbach:
When swap-scratch is used and the trailer is larger than the last sector of the primary and/or the secondary slot, the upgrade would fail and a potential flash memory corruption could occur due to an underflow in case a sector contains both part of the firmware image and part of the trailer.

This issue was not shown by the simulator since when `max-align-32` is not selected, the trailer fit in a single sector in all tested configuration, and when `max-align-32` is selected, the firmware image size is chosen so that no sector contain both part of the firmware image and part of the trailer. So, perhaps that's a known limitation and it is expected the users ensure their firmware images are small enough so that this issue doesn't occur, but since I didn't find any documentation mentioning that point and that using a too large image could lead to potentially dangerous underflow, I've considered that as being a bug and attempted to fix it.

The first commit modifies the simulator to compute the maximum firmware image size the same way, irrespective of whether or not `max-align-32` is selected, allowing in both case the sector containing the first part of the trailer (or the whole trailer if it is small enough) to also contain part of the firmware image. As expected, this leads to the tests to fail in the configuration where the trailer doesn't fit in a single sector.

The following commits fixes the issue in the swap-scratch implementation.

---

Example of configuration where the issue occurs:
* Primary slot: 680 KiB -> 170 * 4096-byte sectors
* Secondary slot: 680 KiB -> 85 * 8192-byte sectors
* Scratch area: 8 KiB -> 1 * 8192-byte sector
* Flash write size: 32 bytes

Therefore:
* Slot trailer size: (3 * 85 * 32) + (32 * 5) = 8320 bytes
* Scratch trailer size: (3 * 1 * 32) + (32 * 5) = 256 bytes

This leads to:
```
         PRIMARY                               SECONDARY                              SCRATCH
|          ...          |              |          ...          |
+-----------------------+              +-----------------------+             +-----------------------+
| Firmware (4096 bytes) | Sector 166   | Firmware (8064 bytes) | Sector 83   | Firmware (7936 bytes) |
|           .           |              |           .           |             |           .           |
+-----------------------+              |           .           |             |           .           |
| Firmware (3968 bytes) | Sector 167   |           .           |             |  Trailer (256 bytes)  |
|  Trailer (128 bytes)  |              | Trailer  (128 bytes)  |             |           .           |
+-----------------------+              +-----------------------+             +-----------------------+
|  Trailer (4096 bytes) | Sector 168   | Trailer  (8192 bytes) | Sector 84
|           .           |              |           .           |
+-----------------------+              |           .           |
|  Trailer (4096 bytes) | Sector 169   |           .           |
|           .           |              |           .           |
+-----------------------+              +-----------------------+
```

The first operation performed by the swap-scratch algorithm is to swap the last sector containing part of the firmware image. Assuming the image is larger than 672 KiB, this means swapping sectors 166+167 (primary) and 83 (secondary) in this example. This operation is performed by the `boot_swap_sectors` routine with `sz == 8192`. Before the fixes introduced by this MR:
* `copy_sz` was computed as `sz - trailer_sz = 8192 - 8320 = -128`, so `copy_sz == 4294967167` after wrap around.
* `copy_sz` was computed without taking the size of the scratch area into account. In this example, it is clear that if the whole 8064 bytes of firmware data are copied from sector 83 (secondary) to the scratch area, the trailer of the scratch area would be overwritten. `copy_sz` must therefore be adjusted to ensure the scratch trailer won't be corrupted.
* Only sectors 166+167 (primary) and 83 (secondary) were erased, causing in particular issues when attempting to rewrite the trailer in the primary slot.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -331,12 +331,87 @@ boot_write_enc_key(const struct flash_area *fap, uint8_t slot,
 }
 #endif
 
-uint32_t bootutil_max_image_size(const struct flash_area *fap)
+#ifdef MCUBOOT_SWAP_USING_SCRATCH
+size_t
+boot_get_first_trailer_sector(struct boot_loader_state *state, size_t slot, size_t trailer_sz)
 {
-#if defined(MCUBOOT_SWAP_USING_SCRATCH) || defined(MCUBOOT_SINGLE_APPLICATION_SLOT) || \
-    defined(MCUBOOT_FIRMWARE_LOADER) || defined(MCUBOOT_SINGLE_APPLICATION_SLOT_RAM_LOAD)
+    size_t first_trailer_sector = boot_img_num_sectors(state, slot) - 1;
+    size_t sector_sz = boot_img_sector_size(state, slot, first_trailer_sector);
+    size_t trailer_sector_sz = sector_sz;
+
+    while (trailer_sector_sz < trailer_sz) {
+        /* Consider that the image trailer may span across sectors of different sizes */
+        --first_trailer_sector;
+        sector_sz = boot_img_sector_size(state, slot, first_trailer_sector);
+
+        trailer_sector_sz += sector_sz;
+    }
+
+    return first_trailer_sector;
+}
+
+/**
+ * Returns the offset to the end of the first sector of a given slot that holds image trailer data.
+ *
+ * @param state      Current bootloader's state.
+ * @param slot       The index of the slot to consider.
+ * @param trailer_sz The size of the trailer, in bytes.
+ *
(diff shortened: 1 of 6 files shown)


d3zd3z (approved):
Thanks for putting the work into figuring this out. I'm curious if this is enough to fix swap-scratch or swap-offset for things like the STM32F4, with just 3 128k sectors available.

taltenbach:
> Thanks for putting the work into figuring this out. I'm curious if this is enough to fix swap-scratch or swap-offset for things like the STM32F4, with just 3 128k sectors available.

You're welcome :) Are you referring to https://github.com/mcu-tools/mcuboot/issues/2052? I'm might be able to put some time into this issue in the following weeks.

PR #2207, opened 2025-02-17
boot: zephyr: kconfig: Add new defaults option for FIH

nordicjm:
Adds a new menu with options which can be used by e.g. sysbuild to
    select which default options are set in a build, adds options for
    fault injection hardening modes

also

    boot: zephyr: kconfig: Fix leaking Kconfigs
    
    Fixes some leaking Kconfig options which were out of the
    MCUboot menu


boot/zephyr/Kconfig
```diff
@@ -1069,6 +1086,8 @@ config MCUBOOT_DEVICE_CPU_CORTEX_M0
 	default n
 	select SW_VECTOR_RELAY if !CPU_CORTEX_M0_HAS_VECTOR_TABLE_REMAP
 
+endmenu
```
nvlsianpu, line 1089: move `endmenu` before `config MCUBOOT_DEVICE_SETTINGS`

nvlsianpu (approved):

boot/zephyr/Kconfig
nordicjm: Fixed

de-nordic (approved):

PR #2210, opened 2025-02-18
bootutil: Do not compile boot_trailer_scramble_off when not needed

de-nordic:
Add conditional compilation, dependent on MCUBOOT_MINIMAL_SCRAMBLE, for:
 boot_trailer_scramble_offset
 boot_header_scramble_off_sz

Faster than fixing TF-M, that does not implement flash_area_get_sector.


de-nordic:
No, can't fix that easily.

PR #2211, opened 2025-02-18
Revert support for devices without erase due to problem with TF-M

de-nordic:
TF-M lacks implementation of flash_area_get_sector and getting the function in will take longer than acceptable for the main to be blocked from compiling on Zephyr CI.


nordicjm:
*devices

nordicjm (approved):

PR #2215, opened 2025-02-23
sim: Ensure the estimated TLV area size is always equal to the actual size

taltenbach:
To generate largest possible and oversized images, the simulator needs to determine first the maximum size of an image. When signing the images using ECDSA, the actual size of the generated image was not exactly always the desired size. Indeed, in that case, the size of the TLV area estimated during the computation of the maximum image size might not be always equal to the actual TLV area size. This is because the ECDSA signatures are encoded as ASN.1 and the size of the ASN.1 representation can vary depending on the value of the two integers the signature is composed of.

This was not a big deal but was adding a bit of randomness in the simulation and was for example making difficult to generate reliably an oversized image, possibly leading to occasional false failures of the tests using oversized images.

To avoid the issue, a very simple workaround is implemented: when using ECDSA, each image is signed repeatedly until a signature with the maximum size is obtained. This is possible without modifying the image because ECDSA signing is not a deterministic process and gives a completely different signature at each try. On average, four attempts should be required to get the desired signature size.

### DIFF
--- a/sim/src/image.rs
+++ b/sim/src/image.rs
@@ -1827,9 +1827,8 @@ fn install_image(flash: &mut SimMultiFlash, slot: &SlotInfo, len: ImageSize,
 
             info!("slot: 0x{:x}, HDR: 0x{:x}, trailer: 0x{:x}",
                 slot_len, HDR_SIZE, trailer);
-            // the overflow size is rougly estimated to work for all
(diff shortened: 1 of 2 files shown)


d3zd3z (approved):
This looks good. Given that the max size should be encountered in 1 out of 4 signatures, 100 tries should only fail about every 10^12 times, so that should be more than adequate.

PR #2228, opened 2025-03-07
docs: table formatting fix

michalek-no:
add missing markers.


nvlsianpu (approved):

nordicjm (changes requested):
The table is valid syntax as seen on https://docs.mcuboot.com/design.html

PR #2236, opened 2025-03-17
imgtool: dumpinfo avoid exit

gzzi:
Hi,

I'm using `dump_imginfo` to export image info in a yaml file from another python source code. I don't want that this script stop when calling this function with `silent=True`.

Regards,

GG

### DIFF
--- a/scripts/imgtool/dumpinfo.py
+++ b/scripts/imgtool/dumpinfo.py
@@ -251,7 +251,7 @@ def dump_imginfo(imgfile, outfile=None, silent=False):
     ###############################################################################
 
     if silent:
-        sys.exit(0)
+        return
 
     print("Printing content of signed image:", os.path.basename(imgfile), "\n")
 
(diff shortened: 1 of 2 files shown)


scripts/imgtool/dumpinfo.py
```diff
@@ -251,7 +251,7 @@ def dump_imginfo(imgfile, outfile=None, silent=False):
     ###############################################################################
 
     if silent:
-        sys.exit(0)
+        return
```
gchwier, line 254: for backward compatibility, you can skip printing message `dumpinfo has run successfully` in main.py:dumpinfo method when silent option is used (just add `if not silent:`)

nvlsianpu (changes requested):
As above & please provide a commit message which will following rules (see an oldest commit for example).

gchwier (approved):

nordicjm:
@gzzi can you use a proper contactable email for sign-off?

nvlsianpu (approved):
Patch content is fine.
Email address in sig-off-by line need to match with the one in you git.

PR #2259, opened 2025-04-10
zephyr: Disable SPI_NOR by default for nrf7002dk/nrf5340/cpuapp

de-nordic:
The change allows to build MCUboot for nrf7002dk/nrf5340/cpuapp with default configuration.


nordicjm (approved):

JordanYates:
What was the point of this?

> The change allows to build MCUboot for nrf7002dk/nrf5340/cpuapp with default configuration

The application compiles fine without this change (as of Zephyr v4.2), so the above isn't accurate.

Instead, the SPI_NOR is force disabled on a board that is a Wi-Fi dev kit, that literally cannot fit a second image anywhere other than SPI_NOR when the bootloader and Wi-Fi is enabled.

PR #2261, opened 2025-04-10
Add Cortex-R support

m-braunschweig:
This PR adds support for Booting on a Cortex-R core. It has been tested with the following branch which also includes some instructions what Kconfig options need to be set: https://github.com/siemens/zephyr/tree/mika/ti/ti-am2434-native-boot


nordicjm (changes requested):

m-braunschweig:
Thanks for the review. I addressed the changes in the last force-push and added a comment to the MPU

m-braunschweig:
@nordicjm can you please take a look again? I changed the MPU code to be more readable via macros but if you insist I can also add a `#include <zephyr/../../arch/arm/core/mpu/cortex_a_r/arm_mpu_internal.h>` to use the internal Zephyr functions.

Also sorry but I didn't notice that I removed the Copyright notices for the small changes in the release note commit instead of the original commit while rebasing

boot/zephyr/arm_cleanup.c
```diff
@@ -1,21 +1,50 @@
...
+
+#define READ_COPROCESSOR_REGISTER(out, coproc, opc1, crn, crm, opc2) \
+	__asm__ volatile("mrc " #coproc ", " #opc1 ", %0, " #crn ", " #crm ", " #opc2 "\n" : "=r" (out) ::);
+
+#define WRITE_COPROCESSOR_REGISTER(in, coproc, opc1, crn, crm, opc2) \
+	__asm__ volatile("mcr " #coproc ", " #opc1 ", %0, " #crn ", " #crm ", " #opc2 "\n" :: "r" (in) :)
+
+#endif /* CONFIG_CPU_CORTEX_R5 */
+
+void cleanup_arm_interrupts(void) {
```
nordicjm: newline for `{`

boot/zephyr/arm_cleanup.c
```diff
@@ -24,11 +53,27 @@ void cleanup_arm_nvic(void) {
 	for (uint8_t i = 0; i < ARRAY_SIZE(NVIC->ICPR); i++) {
 		NVIC->ICPR[i] = 0xFFFFFFFF;
 	}
+#else
+
```
nordicjm:
```suggestion
```

nordicjm (commented):
Is OK to add, just needs nit. fixes

m-braunschweig:
First force-push is a capitalize first letter in the comment fix that I previously missed; second force-push are your suggestions.

Do you mean it's ok to add the internal header or the inline ASM to read/write coprocessor registers via the macros?

nordicjm:
@m-braunschweig seemingly this branch has conflicts that must be resolved? Not sure what those are but can you check and fix?

boot/zephyr/arm_cleanup.c
```diff
@@ -1,21 +1,51 @@
...
+#ifdef CONFIG_CPU_CORTEX_M
 #include <cmsis_core.h>
+#endif
+
 #if CONFIG_CPU_HAS_NXP_SYSMPU
 #include <fsl_sysmpu.h>
 #endif
 
-void cleanup_arm_nvic(void) {
+#ifndef CONFIG_CPU_CORTEX_M
```
nvlsianpu: is this only for CONFIG_CPU_CORTEX_R5?

nvlsianpu (changes requested):
Can we have explicate ifdefry
`#ifdef CONFIG_CPU_CORTEX_R5`
for code for supporting Cortex-R in that patch.
This would help wit maintenance and improve legibility of Cortex-R relevant code.

m-braunschweig:
> @m-braunschweig seemingly this branch has conflicts that must be resolved? Not sure what those are but can you check and fix?

Resolved. Was due to the Copyright line in the `main.c`

> Can we have explicate ifdefry `#ifdef CONFIG_CPU_CORTEX_R5` for code for supporting Cortex-R in that patch. This would help wit maintenance and improve legibility of Cortex-R relevant code.

Yeah. Are you fine with me splitting the `arm_cleanup.c` into `cleanup/arm_cortex_m.c` and `cleanup/arm_cortex_r.c` with both providing the same functions and only one being compiled via CMake? For the `main.c` I don't have an idea currently on how to effectivly remove/decrease the `#ifdef`s

boot/zephyr/arm_cleanup.c
m-braunschweig: Currently yes. The GIC is also ARM specific while `z_soc_irq_eoi` should is architecture agnostic

m-braunschweig:
Updated it. One problem is that we can't use `CONFIG_CPU_CORTEX_R` since that Kconfig option doesn't exist in Zephyr. For now I only use `CONFIG_CPU_CORTEX_R5` since that's the exact core I'm using to test the code

m-braunschweig:
Updated it to use `ARMV7_R` and also the comment for Cortex-R to be more accurate

boot/zephyr/main.c
```diff
@@ -226,10 +238,31 @@ static void do_boot(struct boot_rsp *rsp)
...
 
+#ifdef CONFIG_CPU_CORTEX_M
     __set_MSP(vt->msp);
+#endif
+
 #if CONFIG_MCUBOOT_CLEANUP_ARM_CORE
+#ifdef CONFIG_CPU_CORTEX_M
     __set_CONTROL(0x00); /* application will configures core on its own */
     __ISB();
+#else
```
natto1784, line 249: I am good with it as it is but can we do `#ifdef CONFIG_ARMV7_R` here just to make the use case clearer. Same with other `#else`s as well?

natto1784 (approved):
Great addition, aside from my other comment, this looks good enough as it is

boot/zephyr/main.c
nvlsianpu: can be improved in this PR or in a next one.

nvlsianpu (approved):

m-braunschweig:
I will try to remember to do the changes when submitting a PR that provides default configs for the AM243x LaunchPad (which isn't supported by upstream Zephyr yet)

PR #2263, opened 2025-04-14
Add checks to crypto operation calls in MCUboot

Karambite:


boot/bootutil/src/image_rsa.c
```diff
@@ -94,9 +94,15 @@ pss_mgf1(uint8_t *mask, const uint8_t *hash)
...
-        bootutil_sha_finish(&ctx, htmp);
+        if (bootutil_sha_update(&ctx, hash, PSS_HLEN)) {
+            goto out;
+        }
+        if (bootutil_sha_update(&ctx, counter, 4)) {
+            goto out;
+        }
+        if(bootutil_sha_finish(&ctx, htmp)){
+            goto out;
+        }
```
de-nordic, line 105: No reason to do so? What is the point, fail quicker?

Karambite: Yes. If the any of the sha operation fails, then why should it even attempt the rest? It stays consistent with other crypto implementations.

de-nordic: I see no reason. If it fails it fails. There is no info here for reason nor any way to recover. But the code gets increased for adding condition checks and jumps.

PR #2269, opened 2025-04-17
boot: bootutil: swap_offset: Fix maximum application size

nordicjm:
Fixes this functionality to have the maximum allowable application size

### DIFF
--- a/boot/bootutil/src/swap_offset.c
+++ b/boot/bootutil/src/swap_offset.c
@@ -307,23 +307,26 @@ static int app_max_sectors(struct boot_loader_state *state)
     uint32_t sz = 0;
     uint32_t sector_sz;
     uint32_t trailer_sz;
-    uint32_t first_trailer_idx;
+    uint32_t available_sectors_pri;
+    uint32_t available_sectors_sec;
+    uint32_t trailer_sectors = 0;
 
     sector_sz = boot_img_sector_size(state, BOOT_PRIMARY_SLOT, 0);
     trailer_sz = boot_trailer_sz(BOOT_WRITE_SZ(state));
-    first_trailer_idx = boot_img_num_sectors(state, BOOT_PRIMARY_SLOT);
 
     while (1) {
         sz += sector_sz;
+        ++trailer_sectors;
 
         if  (sz >= trailer_sz) {
             break;
         }
-
-        first_trailer_idx--;
     }
 
-    return first_trailer_idx;
+    available_sectors_pri = boot_img_num_sectors(state, BOOT_PRIMARY_SLOT) - trailer_sectors;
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/swap_offset.c
```diff
@@ -307,23 +307,26 @@ static int app_max_sectors(struct boot_loader_state *state)
...
 
         if  (sz >= trailer_sz) {
             break;
         }
-
-        first_trailer_idx--;
     }
 
-    return first_trailer_idx;
+    available_sectors_pri = boot_img_num_sectors(state, BOOT_PRIMARY_SLOT) - trailer_sectors;
```
de-nordic, line 326: To make sure, because this is swap_offset, the swap page is already included into this or should be -1 here to the equation for the swap page?

boot/bootutil/src/swap_offset.c
```diff
@@ -333,34 +336,35 @@ int boot_slots_compatible(struct boot_loader_state *state)
...
     num_sectors_sec = boot_img_num_sectors(state, BOOT_SECONDARY_SLOT);
-    num_usable_sectors_pri = app_max_sectors(state);
+    num_usable_sectors = app_max_sectors(state);
 
-    if ((num_sectors_pri != num_sectors_sec) &&
-            ((num_sectors_pri + 1) != num_sectors_sec) &&
-            ((num_usable_sectors_pri + 1) != (num_sectors_sec))) {
+    if (num_sectors_pri != num_sectors_sec &&
+        (num_sectors_pri + 1) != num_sectors_sec &&
+        num_usable_sectors != (num_sectors_sec - 1)) {
```
de-nordic, line 347: Hm. The allowed primary will be always at least one page shorter than secondary, because that page is used for swapping, so:
```
num_sectors_pri - 1 >= num_sectors_sec
```
is the first condition.
From that we have sectors really available for image as:
```
num_usable_sectors = num_sectors_sec - int((trailer_sz + PAGE_SIZE  - 1) / PAGE_SIZE)
```
That is how much could be uploaded if `num_sectors_pri - 1 >= num_sectors_sec`. Secondary slot sets the limit.

If we have
```
num_sectors_pri - 1 < num_sectors_sec
```
then:
```
num_usable_sectors = (num_sectors_pri - 1) - int((trailer_sz + PAGE_SIZE  - 1) / PAGE_SIZE)
```
That is how much could be swapped, the primary slot sets the limit.

boot/bootutil/src/swap_offset.c
nordicjm: there is no swap page in primary, only secondary, but the upgrade trailer data resides only in the primary, this is why secondary -1's for swap sector only and primary -'s the trailer size only

de-nordic (approved):

PR #2276, opened 2025-04-23
Revert "zephyr: arm: Update reading the flash image reset vector"

nordicjm:
This reverts commit 453096b17ddc3aac7bf6afb97c40591d5ea3aa9c.

Comment from @de-nordic 

> This is revert of commit which was supposed to allow picking interrupt vector table from flash area but the whole modification unfortunately misunderstood difference between flash device ID and flash area ID.


de-nordic (approved):

nordicjm:
@mmahadevan108 ping that this commit is being reverted due to the above misunderstanding of what the ID means in this instance, please submit a fixed PR with the above in mind if you require this feature

PR #2278, opened 2025-04-23
boot: zephyr: boards: nrf54lm20apdk

michalek-no:
adding default configs.

### DIFF
--- a/boot/zephyr/boards/nrf54lm20apdk_nrf54lm20a_cpuapp.conf
+++ b/boot/zephyr/boards/nrf54lm20apdk_nrf54lm20a_cpuapp.conf
@@ -1,0 +1,16 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+CONFIG_BOOT_MAX_IMG_SECTORS=256
+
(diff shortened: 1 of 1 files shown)


boot/zephyr/boards/nrf54lm20apdk_nrf54lm20a_cpuapp.conf
```diff
@@ -0,0 +1,16 @@
...
+# Ensure that the SPI NOR driver is disabled by default
+CONFIG_SPI_NOR=n
+
+# TODO: below are not yet supported and need fixing
+CONFIG_FPROTECT=n
+
+CONFIG_BOOT_WATCHDOG_FEED=n
+
+CONFIG_PSA_CRYPTO_DRIVER_CRACEN=n
+CONFIG_PSA_CRYPTO_DRIVER_OBERON=y
```
nordicjm, line 16: ignoring `CONFIG_BOOT_WATCHDOG_FEED=n` the rest are not upstream symbols so need to be removed (if needed they can be added as a noup). And likely it would make more sense to add it as a soc Kconfig fragment rather than a board

nordicjm (changes requested):

nordicjm:
moved downstream

PR #2282, opened 2025-04-26
boot: bootutil: Fix security counter updated before revert after resuming upgrade

taltenbach:
This PR fixes an issue that was preventing an interrupted then resumed upgrade to be reverted when the hardware rollback protection is enabled and the swap-scratch, swap-move or swap-offset strategy is used.

After a "test" upgrade has been performed, the security counter of every image must not be updated immediately to be able to roll back in case the upgrade is not confirmed. To that end, the `boot_update_hw_rollback_protection` routine, called just before exiting MCUboot, was checking the swap type of the image and updating the security counter only if the swap type was none, with the assumption that it implies no upgrade has been performed by MCUboot. This was working properly in the nominal case, when the upgrades are not interrupted, however this assumption is false if MCUboot has just resumed an upgrade after a power cycle because in that case the swap type is set to none after the end of the upgrade process:
https://github.com/mcu-tools/mcuboot/blob/81315483fcbdf1f1524c2b34a1fd4de6c77cd0f4/boot/bootutil/src/loader.c#L2251-L2252

So, let's suppose an image A with a security counter N is upgraded, in "test" mode, to an image B with a security counter M > N:
1. MCUboot starts the upgrade from A to B.
2. During the upgrade, the device is power cycled.
3. MCUboot resumes and completes the upgrade.
4. Before booting image B, MCUboot updates the security counter value to M.
5. The image B is faulty and is therefore not confirmed, the device is rebooted.
6. MCUboot wants to revert the upgrade but refuses to roll back to image A since its security counter is lower than M. Therefore no revert is performed and the device might be bricked since B is faulty.

This was not detected by the tests since in the simulator the upgrades were performed with images having the same security counter value. In addition to fixing the issue, this MR also updates the simulator to use images with different security counter values, enabling to catch such kind of bugs.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2246,18 +2246,26 @@ boot_update_hw_rollback_protection(struct boot_loader_state *state)
 {
 #ifdef MCUBOOT_HW_ROLLBACK_PROT
     int rc;
+    uint8_t image_index;
+    struct boot_swap_state swap_state;
+
+    image_index = BOOT_CURR_IMG(state);
+
+    rc = boot_read_swap_state_by_id(FLASH_AREA_IMAGE_PRIMARY(image_index), &swap_state);
+    if (rc != 0) {
+        return rc;
+    }
 
     /* Update the stored security counter with the active image's security
-    * counter value. It will only be updated if the new security counter is
-    * greater than the stored value.
-    *
-    * In case of a successful image swapping when the swap type is TEST the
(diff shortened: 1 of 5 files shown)


d3zd3z (approved):

de-nordic (approved):

d3zd3z:
@taltenbach This now has merge conflicts. Please rebase, and we'll be able to merge.

taltenbach:
@d3zd3z Done :)

PR #2285, opened 2025-04-27
sim: Enable oversized_secondary_slot test with max-align-32

taltenbach:
The test case `oversized_secondary_slot`, introduced by https://github.com/mcu-tools/mcuboot/pull/1286 was not enabled when the `max-align-32` feature was selected because the test case was failing with some configurations.

From the PR description:
>  added oversized_secondary_slot test for ensure that too big image upgrade attempt will be rejected. I doesn't work for max-align-32 feature due to not precise operations on flash image in sim code. Fixing of this problem is out of scope for this PR.

Since 88294be188c25e4b7d02cc9c55fabfc9286c83b7, the `oversized_secondary_slot` test is now passing with `max-align-32` in all configurations and can therefore be enabled.


d3zd3z (approved):
Yay

de-nordic (approved):

PR #2305, opened 2025-05-14
Add support for multiple same-type signatures with key ID parsing

maulik-arm:
This PR adds support for signing and verifying images with multiple signatures of the same type (e.g., multiple EC256 signatures), enhancing flexibility in secure boot scenarios. It also introduces Key ID TLV parsing to enable the bootloader to select the correct key from a set of built-in keys.

#### 1. **bootutil: Parse key ID TLV for built-in keys**
- Adds support for parsing Key ID TLVs in the image when `MCUBOOT_BUILTIN_KEY` is enabled.
- This enables selection of the correct built-in key for verification.

#### 2. **imgtool: Add support for multiple signatures and key ID TLVs**
- Enables signing an image with multiple keys of the same type.
- Adds CLI support for passing multiple `--key` arguments.
- Allows optional specification of a key ID per signature, which gets encoded in the TLV.
- Updates imgtool test suite to verify both multiple signatures and key ID inclusion.

#### 3. **bootutil: Add support for verifying multiple same-type signatures**
- Updates signature verification logic to loop through all same-type signature TLVs.
- When `MCUBOOT_BUILTIN_KEY` or `MCUBOOT_HW_KEY` is enabled, the key ID is used to select the appropriate key for verification.

---
### Notes

- Backwards compatible: Images with a single signature continue to work as before.
- Only takes effect when config 'MCUBOOT_IMAGE_MULTI_SIG_SUPPORT'  is enabled.
- Designed to work with existing MCUboot signature verification flow with minimal disruption.

### DIFF
--- a/boot/bootutil/CMakeLists.txt
+++ b/boot/bootutil/CMakeLists.txt
@@ -14,28 +14,32 @@ target_include_directories(bootutil
         src
 )
 
-target_sources(bootutil
-    PRIVATE
-        src/boot_record.c
-        src/bootutil_misc.c
-        src/bootutil_public.c
-        src/caps.c
-        src/encrypted.c
-        src/fault_injection_hardening.c
-        src/fault_injection_hardening_delay_rng_mbedtls.c
-        src/image_ecdsa.c
-        src/image_ed25519.c
-        src/image_rsa.c
-        src/image_validate.c
-        src/loader.c
-        src/swap_misc.c
-        src/swap_move.c
-        src/swap_scratch.c
-        src/tlv.c
+set(BOOTUTIL_SOURCES
+    src/boot_record.c
+    src/bootutil_misc.c
+    src/bootutil_public.c
+    src/caps.c
+    src/encrypted.c
+    src/fault_injection_hardening.c
+    src/fault_injection_hardening_delay_rng_mbedtls.c
+    src/image_ecdsa.c
+    src/image_ed25519.c
+    src/image_rsa.c
+    src/loader.c
+    src/swap_misc.c
+    src/swap_move.c
+    src/swap_scratch.c
+    src/tlv.c
 )
+
 if(CONFIG_BOOT_RAM_LOAD)
-    target_sources(bootutil
-        PRIVATE
-            src/ram_load.c
-    )
+    list(APPEND BOOTUTIL_SOURCES src/ram_load.c)
 endif()
+
+if(MCUBOOT_IMAGE_MULTI_SIG_SUPPORT)
+    list(APPEND BOOTUTIL_SOURCES src/image_multi_sig.c)
+else()
+    list(APPEND BOOTUTIL_SOURCES src/image_validate.c)
+endif()
+
+target_sources(bootutil PRIVATE ${BOOTUTIL_SOURCES})
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -100,6 +100,7 @@ struct flash_area;
  */
 #define IMAGE_TLV_KEYHASH           0x01   /* hash of the public key */
 #define IMAGE_TLV_PUBKEY            0x02   /* public key */
+#define IMAGE_TLV_KEYID             0x03   /* Key ID */
 #define IMAGE_TLV_SHA256            0x10   /* SHA256 of image hdr and body */
 #define IMAGE_TLV_SHA384            0x11   /* SHA384 of image hdr and body */
 #define IMAGE_TLV_SHA512            0x12   /* SHA512 of image hdr and body */
--- a/boot/bootutil/include/bootutil/sign_key.h
+++ b/boot/bootutil/include/bootutil/sign_key.h
@@ -51,6 +51,7 @@ extern struct bootutil_key bootutil_keys[];
  * Retrieve the hash of the corresponding public key for image authentication.
  *
  * @param[in]      image_index      Index of the image to be authenticated.
+ * @param[in]      key_index        Index of the key to be used.
  * @param[out]     public_key_hash  Buffer to store the key-hash in.
  * @param[in,out]  key_hash_size    As input the size of the buffer. As output
  *                                  the actual key-hash length.
@@ -58,8 +59,10 @@ extern struct bootutil_key bootutil_keys[];
  * @return                          0 on success; nonzero on failure.
  */
 int boot_retrieve_public_key_hash(uint8_t image_index,
+                                  uint8_t key_index,
                                   uint8_t *public_key_hash,
                                   size_t *key_hash_size);
+
 #endif /* !MCUBOOT_HW_KEY */
 
 extern const int bootutil_key_cnt;
--- a/boot/bootutil/src/image_multi_sig.c
+++ b/boot/bootutil/src/image_multi_sig.c
@@ -34,31 +34,20 @@
 
 #include <flash_map_backend/flash_map_backend.h>
 
+#include "bootutil/bootutil_log.h"
+#include "bootutil/crypto/ecdsa.h"
 #include "bootutil/image.h"
 #include "bootutil/crypto/sha.h"
 #include "bootutil/sign_key.h"
 #include "bootutil/security_cnt.h"
 #include "bootutil/fault_injection_hardening.h"
+#include "bootutil_priv.h"
 
 #include "mcuboot_config/mcuboot_config.h"
 
-#ifdef MCUBOOT_ENC_IMAGES
-#include "bootutil/enc_key.h"
-#endif
-#if defined(MCUBOOT_SIGN_RSA)
-#include "mbedtls/rsa.h"
-#endif
-#if defined(MCUBOOT_SIGN_EC256)
 #include "mbedtls/ecdsa.h"
-#endif
-#if defined(MCUBOOT_ENC_IMAGES) || defined(MCUBOOT_SIGN_RSA) || \
-    defined(MCUBOOT_SIGN_EC256)
 #include "mbedtls/asn1.h"
-#endif
 
-#include "bootutil_priv.h"
-
-#ifndef MCUBOOT_SIGN_PURE
 /*
  * Compute SHA hash over the image.
  * (SHA384 if ECDSA-P384 is being used,
@@ -68,32 +57,18 @@ static int
 bootutil_img_hash(struct boot_loader_state *state,
                   struct image_header *hdr, const struct flash_area *fap,
                   uint8_t *tmp_buf, uint32_t tmp_buf_sz, uint8_t *hash_result,
-                  uint8_t *seed, int seed_len
-#if defined(MCUBOOT_SWAP_USING_OFFSET) && defined(MCUBOOT_SERIAL_RECOVERY)
-                  , uint32_t start_offset
-#endif
-                 )
+                  uint8_t *seed, int seed_len)
 {
     bootutil_sha_context sha_ctx;
     uint32_t size;
     uint16_t hdr_size;
     uint32_t blk_off;
     uint32_t tlv_off;
-#if !defined(MCUBOOT_HASH_STORAGE_DIRECTLY)
     int rc;
     uint32_t off;
     uint32_t blk_sz;
-#endif
-#if defined(MCUBOOT_ENC_IMAGES)
-    struct enc_key_data *enc_state;
-    int image_index;
-#endif
-#if defined(MCUBOOT_SWAP_USING_OFFSET)
-    uint32_t sector_off = 0;
-#endif
 
-#if (BOOT_IMAGE_NUMBER == 1) || !defined(MCUBOOT_ENC_IMAGES) || \
-    defined(MCUBOOT_RAM_LOAD)
+#if (BOOT_IMAGE_NUMBER == 1) || defined(MCUBOOT_RAM_LOAD)
     (void)state;
     (void)hdr_size;
     (void)blk_off;
@@ -105,33 +80,6 @@ bootutil_img_hash(struct boot_loader_state *state,
     (void)fap;
     (void)tmp_buf;
     (void)tmp_buf_sz;
-#endif
-#endif
-
-#ifdef MCUBOOT_ENC_IMAGES
-    if (state == NULL) {
-        enc_state = NULL;
-        image_index = 0;
-    } else {
-        enc_state = BOOT_CURR_ENC(state);
-        image_index = BOOT_CURR_IMG(state);
-    }
-
-    /* Encrypted images only exist in the secondary slot */
-    if (MUST_DECRYPT(fap, image_index, hdr) &&
-            !boot_enc_valid(enc_state, 1)) {
-        return -1;
-    }
-#endif
-
-#if defined(MCUBOOT_SWAP_USING_OFFSET)
-    /* For swap using offset mode, the image starts in the second sector of the upgrade slot, so
-     * apply the offset when this is needed
-     */
-#if defined(MCUBOOT_SERIAL_RECOVERY)
-    sector_off = boot_get_state_secondary_offset(state, fap) + start_offset;
-#else
-    sector_off = boot_get_state_secondary_offset(state, fap);
 #endif
 #endif
 
@@ -151,12 +99,6 @@ bootutil_img_hash(struct boot_loader_state *state,
     /* If protected TLVs are present they are also hashed. */
     size += hdr->ih_protect_tlv_size;
 
-#ifdef MCUBOOT_HASH_STORAGE_DIRECTLY
-    /* No chunk loading, storage is mapped to address space and can
-     * be directly given to hashing function.
-     */
-    bootutil_sha_update(&sha_ctx, (void *)flash_area_get_off(fap), size);
-#else /* MCUBOOT_HASH_STORAGE_DIRECTLY */
 #ifdef MCUBOOT_RAM_LOAD
     bootutil_sha_update(&sha_ctx,
                         (void*)(IMAGE_RAM_BASE + hdr->ih_load_addr),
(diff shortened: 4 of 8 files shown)


scripts/imgtool/image.py
```diff
@@ -814,72 +843,107 @@ def verify(imgfile, key):
         if magic != IMAGE_MAGIC:
             return VerifyResult.INVALID_MAGIC, None, None, None
 
-        tlv_off = header_size + img_size
+        # Locate the first TLV info header
+        base_tlv_off = header_size + img_size
+        tlv_off = base_tlv_off
```
davidvincze, line 848: What's the reason behind it? Leftover code from development?

scripts/imgtool/main.py
```diff
@@ -450,7 +464,7 @@ def sign(key, public_key_format, align, version, pad_sig, header_size,
          dependencies, load_addr, hex_addr, erased_val, save_enctlv,
          security_counter, boot_record, custom_tlv, rom_fixed, max_align,
          clear, fix_sig, fix_sig_pubkey, sig_out, user_sha, is_pure,
-         vector_to_sign, non_bootable):
+         vector_to_sign, non_bootable, psa_key_ids):
```
davidvincze, line 467: It should be passed to the Image constructor in line 473.
Currently it's only initialized to None which is then used by the `create` method.

scripts/imgtool/main.py
```diff
@@ -441,6 +453,8 @@ def convert(self, value, param, ctx):
               help='send to OUTFILE the payload or payload''s digest instead '
               'of complied image. These data can be used for external image '
               'signing')
+@click.option('--psa-key-ids', multiple=True, type=int, required=False,
```
davidvincze, line 456: I believe `multiple=True` is not needed if it takes a list of IDs.

scripts/imgtool/main.py
```diff
@@ -97,6 +97,15 @@ def save_signature(sigfile, sig):
         signature = base64.b64encode(sig)
```
davidvincze, line 97: `save_signature` can be deleted if unused

scripts/imgtool/image.py
```diff
@@ -135,13 +136,19 @@ def add(self, kind, payload):
         """
         e = STRUCT_ENDIAN_DICT[self.endian]
         if isinstance(kind, int):
-            if not TLV_VENDOR_RES_MIN <= kind <= TLV_VENDOR_RES_MAX:
+            if kind in TLV_VALUES.values():
+                buf = struct.pack(e + 'I', kind)
```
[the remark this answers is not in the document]
maulik-arm: Ack. updated

boot/bootutil/include/bootutil/sign_key.h
```diff
@@ -39,6 +39,17 @@ struct bootutil_key {
...
+#ifdef MCUBOOT_BUILTIN_KEY
+/**
+ * Verify that the specified key ID is valid for authenticating the given image.
+ *
+ * @param[in]  image_index   Index of the image to be verified.
+ * @param[in]  key_id        Identifier of the key to be verified against the image.
+ *
+ * @return                   0 if the key ID is valid for the image; nonzero on failure.
+ */
+int boot_verify_key_id_for_image(uint8_t image_index, int32_t key_id);
```
RcColes: I think we should avoid using `int32_t` for the key ID, and should instead use `uint32_t` to be compatible with TF-PSA-CRYPTO

Relatedly, I think we should standardize the "error" key ID to be 0, to be in line with `PSA_KEY_ID_NULL` (https://arm-software.github.io/psa-api/crypto/1.1/api/keys/ids.html)

boot/bootutil/src/image_multi_sig.c
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+#if defined(EXPECTED_HASH_TLV)
+    rc = !image_hash_valid;
+    if (rc) {
+        goto out;
+    }
+#endif
+    if (FIH_NOT_EQ(key_must_sign, true) || FIH_NOT_EQ(key_might_sign, true) ||
+        FIH_EQ(key_must_sign_count, 0)) {
```
RcColes: I think potentially this should be removed, or behind an option. There's a reasonable argument to say that we can enforce the image being must_signed at least once via the key policy, and removing this gives some interesting options (for example, images being might_sign only on a dev platform which is provisioned differently)

boot/bootutil/src/image_multi_sig.c
```diff
@@ -0,0 +1,536 @@
...
+    size_t key_hash_size = sizeof(key_hash);
+    int rc;
+    FIH_DECLARE(fih_rc, FIH_FAILURE);
+
+    bootutil_sha_init(&sha_ctx);
+    bootutil_sha_update(&sha_ctx, key, key_len);
+    bootutil_sha_finish(&sha_ctx, hash);
+    bootutil_sha_drop(&sha_ctx);
+
+    for(key_index = 0; key_index < 2; key_index++) {
```
RcColes: This feels like it should be a config define

boot/bootutil/src/image_multi_sig.c
```diff
@@ -0,0 +1,536 @@
...
+ * @param state         Pointer to the boot state object.
+ * @param slot          Slot of the current image to get the security counter of.
+ * @param fap           Pointer to a description structure of the image's
+ *                      flash area.
+ * @param security_cnt  Pointer to store the security counter value.
+ *
+ * @return              0 on success; nonzero on failure.
+ */
+int32_t
+bootutil_get_img_security_cnt(struct boot_loader_state *state, int slot,
```
RcColes: I wouldn't have expected this to need to be in this file, as it shouldn't be affected by the multiple signatures. I'd instead think we should have some common validation logic, and then some minimal duplicated code (even duplicating `bootutil_img_validate` seems like it's going to introduce some major maintenance effort

maulik-arm: So I have refactored the image_validate.c file such that it contains the common validation logic and rest moved to their own respective file. Hopefully should be much easier to follow as well.

maulik-arm: Update done in bootutil_find_key.c file

scripts/imgtool/main.py
maulik-arm: I have updated so that psa_key_ids is applied to the image using the set function in the new PS.

scripts/imgtool/main.py
maulik-arm: With multiple =true, Click automatically parses  all cases where 1. No key ids are passed. 2 single key id is passed and 3 multiple key ids are passed

boot/bootutil/CMakeLists.txt
```diff
@@ -14,28 +14,40 @@ target_include_directories(bootutil
...
-        src/image_ecdsa.c
-        src/image_ed25519.c
-        src/image_rsa.c
-        src/image_validate.c
-        src/loader.c
-        src/swap_misc.c
-        src/swap_move.c
-        src/swap_scratch.c
-        src/tlv.c
+set(BOOTUTIL_SOURCES
```
Anton-TF: Is there a reason to move back to legacy CMake style with lists? It possible to call **target_sources()** multiple times conditionally.

maulik-arm: No specific reason, just a preference so that everything sourced at a central place. But will update to move to modern way.

RcColes (approved):
LGTM

boot/bootutil/src/bootutil_find_key.c
```diff
@@ -0,0 +1,137 @@
...
+    int rc;
+    uint32_t key_id;
+    FIH_DECLARE(fih_rc, FIH_FAILURE);
+
+    /* Key id is passed */
+    assert(key_id_buf_len == sizeof(uint32_t));
+    key_id = (((uint32_t)key_id_buf[0] << 24) |
+                      ((uint32_t)key_id_buf[1] << 16)  |
+                      ((uint32_t)key_id_buf[2] << 8)   |
+                      ((uint32_t)key_id_buf[3]));
```
davidvincze: As I saw the key ID is now added with the set endianness (not hardcoded `big`). In this case this reordering is not required anymore, right? Thanks!

boot/bootutil/src/bootutil_image_hash.c
```diff
@@ -0,0 +1,193 @@
+/*
```
davidvincze: Could you please rename the file to `bootutil_img_hash.c`? 
This naming better aligns with other files, please update the CMake file(s) as well.

boot/bootutil/CMakeLists.txt
```diff
@@ -33,9 +35,24 @@ target_sources(bootutil
         src/swap_scratch.c
         src/tlv.c
 )
+
+if(MCUBOOT_HW_ROLLBACK_PROT)
```
davidvincze: MCUboot is configured through the mcuboot_config/mcuboot_config.h file, it has no effect on the CMake system. Please include all files unconditionally in the target_sources for bootutil.

boot/bootutil/CMakeLists.txt
```diff
@@ -33,9 +35,24 @@ target_sources(bootutil
...
         PRIVATE
             src/ram_load.c
     )
 endif()
+
+target_compile_definitions(bootutil
+    PRIVATE
+        $<$<BOOL:${MCUBOOT_IMAGE_MULTI_SIG_SUPPORT}>:MCUBOOT_IMAGE_MULTI_SIG_SUPPORT>
+        $<$<BOOL:${MCUBOOT_IMAGE_MULTI_SIG_SUPPORT}>:MCUBOOT_ROTPK_MAX_KEYS_PER_IMAGE=${MCUBOOT_ROTPK_MAX_KEYS_PER_IMAGE}>
+)
```
davidvincze: Please use the mcuboot_config.h here as well:
1. Update the `samples/mcuboot_config/mcuboot_config.template.h` file describing this new option,
2. Update the config file (which is part of the porting layer - provided by the "target OS"), for TF-M it should be: https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/refs/heads/main/bl2/ext/mcuboot/include/mcuboot_config/mcuboot_config.h.in (which can be configured from CMake).

boot/bootutil/src/image_ecdsa.c
```diff
@@ -38,7 +38,7 @@
 #if !defined(MCUBOOT_BUILTIN_KEY)
 fih_ret
 bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
-                    uint8_t key_id)
+                    uint32_t key_id)
```
davidvincze, line 44: Please also update: `boot/bootutil/src/image_ed25519.c:141`.

boot/bootutil/src/bootutil_image_hash.c
```diff
@@ -0,0 +1,193 @@
...
+ */
+
+#include <stdint.h>
+#include <flash_map_backend/flash_map_backend.h>
+
+#include "bootutil/bootutil_log.h"
+#include "bootutil/crypto/sha.h"
+#include "bootutil/fault_injection_hardening.h"
+#include "bootutil/image.h"
+#include "bootutil/sign_key.h"
```
davidvincze: As I see this include is not needed.
```suggestion

```

boot/bootutil/src/bootutil_find_key.c
```diff
@@ -0,0 +1,137 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <stdint.h>
+#include <flash_map_backend/flash_map_backend.h>
```
davidvincze: As I see this include is not needed.
```suggestion
```

boot/bootutil/src/image_validate.c
```diff
@@ -671,6 +389,15 @@ bootutil_img_validate(struct boot_loader_state *state,
 #ifndef MCUBOOT_SIGN_PURE
             FIH_CALL(bootutil_verify_sig, valid_signature, hash, sizeof(hash),
                                                            buf, len, key_id);
+#ifdef MCUBOOT_IMAGE_MULTI_SIG_SUPPORT
+            rc = boot_plat_check_key_policy((valid_signature == 0), key_id,
```
davidvincze, line 397: Please add its declaration to `boot/bootutil/include/bootutil/sign_key.h` with some documentation.

boot/bootutil/src/image_validate.c
```diff
@@ -737,7 +464,15 @@ bootutil_img_validate(struct boot_loader_state *state,
     rc = FIH_NOT_EQ(valid_signature, FIH_SUCCESS);
 #endif
 #ifdef EXPECTED_SIG_TLV
+#ifdef MCUBOOT_IMAGE_MULTI_SIG_SUPPORT
+    if (FIH_NOT_EQ(key_must_sign, true) || FIH_NOT_EQ(key_might_sign, true)) {
+        FIH_RET(FIH_FAILURE);
+    } else {
+        FIH_RET(FIH_SUCCESS);
```
davidvincze: By returning here we are skipping the `security_counter_valid` check in line 478.

boot/bootutil/src/bootutil_find_key.c
maulik-arm: Thank you for spotting this.

boot/bootutil/CMakeLists.txt
maulik-arm: mcuboot_config.template.h updated and from tf-m side patch raised for review : https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/39746

samples/mcuboot_config/mcuboot_config.template.h
```diff
@@ -33,6 +33,20 @@
...
+/*
+ * Multi-signature support
+ *
+ * Currently, only ECDSA signatures using curve P-256 and P-384
+ * are supported for multi-signature images.
+ */
+
+/* Uncomment to enable verification of images with multiple signatures */
+/* #define MCUBOOT_IMAGE_MULTI_SIG_SUPPORT */
+#ifdef MCUBOOT_ROTPK_MAX_KEYS_PER_IMAGE
```
davidvincze:
```suggestion
#ifndef MCUBOOT_ROTPK_MAX_KEYS_PER_IMAGE
```

boot/bootutil/include/bootutil/sign_key.h
```diff
@@ -24,6 +24,9 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#ifdef MCUBOOT_IMAGE_MULTI_SIG_SUPPORT
+#include <stdbool.h>
+#endif /* MCUBOOT_IMAGE_MULTI_SIG_SUPPORT */
```
davidvincze: Please move it below `#include "mcuboot_config/mcuboot_config.h"`

samples/mcuboot_config/mcuboot_config.template.h
maulik-arm: Sorry I  meant to define number of keys if MCUBOOT_IMAGE_MULTI_SIG_SUPPORT is defined.

boot/bootutil/include/bootutil/sign_key.h
```diff
@@ -51,17 +68,41 @@ extern struct bootutil_key bootutil_keys[];
...
+ * @param[out] key_might_sign  Set to true if the key might be used to sign.
+ * @param[out] key_must_sign   Set to true if the key must be used to sign.
+ * @param[out] key_must_sign_count  Set to the number of keys that must sign.
+ *
+ * @return 0 on success, or a negative error code on failure.
+ */
+int boot_plat_check_key_policy(bool valid_sig, uint32_t key,
+                               bool *key_might_sign, bool *key_must_sign,
+                               uint8_t *key_must_sign_count);
+#endif /* MCUBOOT_IMAGE_MULTI_SIG_SUPPORT */
```
de-nordic, line 104: Neither used nor implemented.

boot/bootutil/src/bootutil_find_key.c
```diff
@@ -0,0 +1,151 @@
...
+
+    for(key_index = 0; key_index < NUM_OF_KEYS; key_index++) {
+        rc = boot_retrieve_public_key_hash(image_index, key_index, key_hash, &key_hash_size);
+        if (rc) {
+            return -1;
+        }
+
+        /* Adding hardening to avoid this potential attack:
+         *  - Image is signed with an arbitrary key and the corresponding public
+         *    key is added as a TLV field.
```
de-nordic, line 106: Formatting.

boot/bootutil/src/bootutil_find_key.c
```diff
@@ -0,0 +1,151 @@
...
+         * - During public key validation (comparing against key-hash read from
+         *   HW) a fault is injected to accept the public key as valid one.
+         */
+        FIH_CALL(boot_fih_memequal, fih_rc, hash, key_hash, key_hash_size);
```
de-nordic, line 110: What is the point of finding a key by key id to then verify it with hash? If you have a hash, you can literally map it to desired key id, without exposing api specific key id to anyone. Hey, you can even change internal key mapping in further release without need to re-sign old images, because key id changes, as knowledge on key id is not needed at the build time.

boot/bootutil/include/bootutil/image.h
```diff
@@ -98,6 +98,7 @@ extern "C" {
  */
 #define IMAGE_TLV_KEYHASH           0x01    /* hash of the public key */
 #define IMAGE_TLV_PUBKEY            0x02    /* public key */
+#define IMAGE_TLV_KEYID             0x03    /* Key ID */
```
de-nordic, line 101: I do not think it is good idea. Why not just encode pointer to a key in memory, or data section, if we are already exposing internal api identifiers to user.
Note also that this basically means that you are locked with certain id and have to re-sign images if you find a problem later and figure out that, for some reason, you can not use ids (for example key storage is broken for id [n,m]). Now you not only have to re-build MCUboot but re-issue all app images, even if they do not change internally, just to make them work with mcuboot.

This increases  coupling between MCUboot internal implementation and user release process and decrease information hiding, which is not good for coherent and modular design of anything.

de-nordic:
I am quite sure neither of 
image
is present in code.

boot/bootutil/include/bootutil/image.h
maulik-arm: Hey, thanks for the feedback! So,

1. **PSA key IDs aren’t “internal” magic**  
   They’re part of the *public* Crypto API—`psa_key_id_t` is exactly the handle you pass into every `psa_*` call. It’s not a hidden pointer or private index, but the documented way to name, persist, and reference keys.  
   *Note:* Key IDs are only required when using `MCUBOOT_BUILTIN_KEY`; other build options don’t touch them.

2. **Raw pointers are brittle & unsafe**  becuase of hard coded flash addresses, 
   - **Build fragility:** Hard-coded flash addresses break as soon as you relink.  
   - **Expanded attack surface:** Fault-injection or pointer replay attacks can redirect you to *any* memory.  
   - **Relocation headaches:** Juggling build-time offsets is far worse than using a stable numeric handle.

3. **Key IDs give you safe indirection**  
   The PSA layer enforces your key’s access policy on every handle. Flip bits in the ID? You get a clean “invalid handle” error, not arbitrary memory corruption.

4. **What this PR actually does**  
   Right now, `MCUBOOT_BUILTIN_KEY` still hard-bakes your image-ID → key-ID 1:1, so changing it forces a full re-sign of all blobs. This PR pulls that magic number into one override point—no more hunting through every header.  
   That lays the groundwork for a future enhancement - manifest based approach: your image carries only a symbolic tag, and at boot we’ll read a small signed table to resolve “secure-slot-A → key ID 42.” Future key rotations or remaps will then be a one-line table update—not a re-signing marathon.


Hope that clarifies why we picked PSA key IDs over raw pointers—and how we’re already stepping toward a more flexible manifest approach down the road.

boot/bootutil/src/bootutil_find_key.c
maulik-arm: Looks like there’s a mix-up here:

1. MCUBOOT_BUILTIN_KEY
    Uses a PSA key ID baked into the image header.
    That’s the one mode where an integer handle is attached.
2. MCUBOOT_HW_KEY
    Doesn’t use a numeric key ID in the image.
    Instead you ship a hash, which at boot we compare against the platform’s provisioned key.

Hope that clears it up—let me know if you have any other questions!

boot/bootutil/include/bootutil/sign_key.h
maulik-arm: To be implemented by the platform to configure respective policy. Example TF-M implementation: https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/refs/heads/main/bl2/src/thin_psa_crypto_core.c#226

maulik-arm:
> Passing PSA API key id as TLV is bad idea.
> 
> This PR moves a lot of code around and modifies it at the same time, making it barely reviewable.
> 
> * bootutil: imgtool: Fix CI failures  -- touches a crap load of code here an there and only one change is related  to the topic
> * imgtool: Rename key-ids to psa-key-ids -- just bloats PR for no reason and could be tackled separately
> * imgtool: Add support for multiple signatures -- regarding commit message: as far as I can tell multiple signatures do not improve security, quite the opposite , although they increase flexibility; you basically now have more locks and it is enough to unlock only one. And I do not think that we implement any key revocation at this point.
> * bootutil: Parse key id for built in keys  -- there is no need, or even desire, to pass psa key id, which is internal API id (equivalent of a pointer, if you will) via TLV
> 
> There is no documentation for the new feature, I mean the multiple key thing.
> 
> I do not understand why we need to pass this psa ID with image.
> 
> I do not think you even need to identify a key required for a signature. All you need to have is a way to run your image signature through all the keys, preferably, available for that image index; once you have one key verify signature you are good to go. User does not have to know a key id, nor we have to verify that this is the right key - signature verification process will do that for us.

Hey, Thank you for comments. Let me try to address the concerns:

1. “Passing PSA API key id as TLV is a bad idea.”
TLV here is just a flexible, metadata container—we’re not exposing hidden pointers, but a public psa_key_id_t handle. Embedding the key-ID lets us pick the right key immediately (no trial-and-error), and it’s fully sanctioned by the PSA spec.

2. “This PR moves a lot of code around… makes it barely reviewable.”
Fair point—there’s a lot of churn. Here’s the breakdown:

- **Incremental commits**  
  I split CI‐fixes, refactors, and feature changes into separate commits so you can review each slice in isolation.

- **Why the refactor?**  
  `image_validate.c` was overloaded with `#ifdef`s and mixed concerns. I pulled out:  
  - `bootutil_find_key.c` (key lookup)  
  - `bootutil_img_hash.c` (image hashing)  
  - `bootutil_img_security_cnt.c` (security-counter logic)

- **Feature scope is tiny**  
  The only behavioral changes here are:  
  1. Parsing/writing the new `psa-key-id` TLV  
  2. Multi-sig support (gated under `#ifdef MCUBOOT_IMAGE_MULTI_SIG_SUPPORT`)

3. “Renaming key-ids to psa-key-ids bloats PR.”
We wanted alignment with the PSA terminology so users aren’t confused by “key-id” vs “psa_key_id_t.

4. “Multiple signatures don’t improve security…”

**Policy-driven multi-sig security**  
   Multi-signature isn’t just about slapping on extra locks—it gives you **flexible policy**:

   - **Require any one signature** (“OR”): quick recovery if one signer is offline.  
   - **Require all signatures** (“AND”): maximum assurance that every stakeholder signed off.  

   You implement your desired policy in `boot_plat_check_key_policy()`. Internally you get the list of signatures, inspect which keys signed, and accept or reject the boot based on your “M-of-N” rules.

   For a real-world example, see TF-M’s thin PSA core at line 226 of `thin_psa_crypto_core.c`—that’s where the platform checks your configured policy before returning “OK.”  
   <https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/refs/heads/main/bl2/src/thin_psa_crypto_core.c#226>


5. “Why identify a key at all? You could just try all available keys until one works.”
In resource and time constrained boot environments, iterating over every key (and every curve/algorithm combo) can be costly. A direct ID lookup is faster, unambiguous, and avoids DRM-style “blackboxing” of which key slot is in use.

Why scanning all slots isn’t enough?

Imagine you’ve got two images (secure vs non-secure) and three provisioned keys. If you naively loop over every key slot and accept the first one that “fits,” an attacker could re-sign the secure image with a key that’s valid in PSA but wasn’t meant for that image. You’d happily verify it and boot “secure” code with the wrong signer.

By embedding the exact psa_key_id_t in the TLV, we force the bootloader to call psa_verify_hash(handle, …) on that handle only. No guessing, no stray-slot wins—only the intended key can pass.

I hope that clarifies many things, but review is highly appreciated so please let me know if any comments. 
Thank you

boot/bootutil/include/bootutil/image.h
de-nordic: > Hey, thanks for the feedback! So,
> 
>     1. **PSA key IDs aren’t “internal” magic**
>        They’re part of the _public_ Crypto API—`psa_key_id_t` is exactly the handle you pass into every `psa_*` call. It’s not a hidden pointer or private index, but the documented way to name, persist, and reference keys.
>        _Note:_ Key IDs are only required when using `MCUBOOT_BUILTIN_KEY`; other build options don’t touch them.

Yes they are, because as far as I can tell when you use `psa_import_key`, unless you set required key id as an attribute, that function is able to allocate any value as key id.

> 
>     2. **Raw pointers are brittle & unsafe**  becuase of hard coded flash addresses,
>        
>        * **Build fragility:** Hard-coded flash addresses break as soon as you relink.
>        * **Expanded attack surface:** Fault-injection or pointer replay attacks can redirect you to _any_ memory.
>        * **Relocation headaches:** Juggling build-time offsets is far worse than using a stable numeric handle.

I, literally, was not proposing using pointers. 

> 
>     3. **Key IDs give you safe indirection**
>        The PSA layer enforces your key’s access policy on every handle. Flip bits in the ID? You get a clean “invalid handle” error, not arbitrary memory corruption.

That does not mean I have to have them glued to an image, I can do the same stuff internally. The same thing will happen with "bit flip" there.

> 
>     4. **What this PR actually does**
>        Right now, `MCUBOOT_BUILTIN_KEY` still hard-bakes your image-ID → key-ID 1:1, so changing it forces a full re-sign of all blobs. This PR pulls that magic number into one override point—no more hunting through every header.
>        That lays the groundwork for a future enhancement - manifest based approach: your image carries only a symbolic tag, and at boot we’ll read a small signed table to resolve “secure-slot-A → key ID 42.” Future key rotations or remaps will then be a one-line table update—not a re-signing marathon.

Right. When a key leaks we have to re-issue applications anyway.
But you still do not have to glue API specific ID to the image. PSA is not going to be the only supported backend ever and we should avoid directly exposing api specific stuff to signing or encryption.

> 
> 
> Hope that clarifies why we picked PSA key IDs over raw pointers—and how we’re already stepping toward a more flexible manifest approach down the road.

Oh...

maulik-arm: > Yes they are, because as far as I can tell when you use psa_import_key, unless you set required key id as an attribute, that function is able to allocate any value as key id.

Yes, but we can set deterministic repeatable handle/id using _psa_set_key_id_ so that shouldn't be any issue. right?

maulik-arm: > But you still do not have to glue API specific ID to the image. PSA is not going to be the only supported backend ever and we should avoid directly exposing api specific stuff to signing or encryption.

But there is already 1:1 image to key id mapping in current code ([https://github.com/mcu-tools/mcuboot/blob/9f209100336455c21f9f127f5bb9f2b5a394bb4f/boot/bootutil/include/bootutil/crypto/ecdsa.h#L397]). So wouldn't this PR add flexibility?

de-nordic: > > But you still do not have to glue API specific ID to the image. PSA is not going to be the only supported backend ever and we should avoid directly exposing api specific stuff to signing or encryption.
> 
> But there is already 1:1 image to key id mapping in current code ([https://github.com/mcu-tools/mcuboot/blob/9f209100336455c21f9f127f5bb9f2b5a394bb4f/boot/bootutil/include/bootutil/crypto/ecdsa.h#L397]). So wouldn't this PR add flexibility?

Yes, and I want out of that not only for ecdsa.

You do not need to have the key_id glued to the image to have flexibility.
You only need pair <image, purpose> to request a psa_key_id_t id (or other type, depending on backend) to get a key id that is usable for your backend.
You can have as many keys as you want. At this point we are not discussing policies applied by backend regarding key rejection or revocation, or how to inform backend on key status.
Lets assume the function is identified as `boot map_me_the_key(uint32_t image_id, int purpose, key_type *t)` , purpose is optional I just try to make the thing work with encryption, that returns true and available key id if there is, another,  one available.
Now you have something like:
```
   verified = 0;
   while (map_me_the_key(image, SIGNATURE. &key)) {
       ret = psa_verify_me_the_signature(key, stuff_to_verify);
       if (ret == PSA_SUCCESS) { // weakest point probably
             verified++; // this can not be simple int, this is just an example
      }
   }
   // here check verified and do stuff
```

Assuming that `map_me_the_key` gives back another key available for a given image, you are able to make this work:
- if you expect only one key to verify image
- if you expect at least 1 of N keys to verify an image
- if you expect M of N keys to verify the image, as long as you check that `verified` afterwards against your expected 1 or M.

`map_me_the_key` can have different lists of key ids for every image, for some images or just have one and ignore image number at all.

You do not need to specify PSA key id in image.
Drawback and advantage is that every key is run against the image, so if you are mapping 1000 keys, that may take time, but also you do not know which key worked. You can also just exit loop the first time you are satisfied with number of verification attempts.
(you still need to second check the verification "counter" later...).

de-nordic: You literally want a function that will check whether given key ID may be used with given image; you may as well just have a function that will give you a key id for a given image.

boot/bootutil/include/bootutil/sign_key.h
de-nordic: That is not called nor used by MCUboot. That is TF-M thing and goes over there.

maulik-arm: It is a platform specific hook which is called in image_validate.c in same commit after sign verification  (under #ifdef  MCUBOOT_IMAGE_MULTI_SIG_SUPPORT). There is already a similar platform specific  function 'boot_retrieve_public_key_hash' which is not implemented by MCUboot.  So may I know the difference between the two?

boot/bootutil/include/bootutil/image.h
ahasztag: @de-nordic So if I understand correctly what you would suggest is mapping simply a constant mapping of key ID to image number instead of passing it to TLV-s?

One issue is how would key generations/key revocation be handled in such case

PR #2308, opened 2025-05-15
Additional dbg logging in MCUboot to allow easier finding of reason for image validation failure

de-nordic:


boot/bootutil/src/bootutil_public.c
```diff
@@ -783,6 +799,7 @@ boot_image_load_header(const struct flash_area *fa_p,
 
     if (!boot_u32_safe_add(&size, hdr->ih_img_size, hdr->ih_hdr_size) ||
         size >= flash_area_get_size(fa_p)) {
+        BOOT_LOG_ERR("Incorrect image size");
```
nordicjm, line 802: should log them probably

boot/bootutil/src/encrypted.c
```diff
@@ -411,6 +414,8 @@ boot_decrypt_key(const uint8_t *buf, uint8_t *enckey)
     uint8_t *cpend;
     size_t olen;
 #endif
+
+    BOOT_LOG_DBG("boot_decrypt_key:");
```
nordicjm, line 418: missing some debug?

boot/bootutil/src/image_validate.c
```diff
@@ -280,6 +287,8 @@ bootutil_find_key(uint8_t *keyhash, uint8_t keyhash_len)
     const struct bootutil_key *key;
     uint8_t hash[IMAGE_HASH_SIZE];
 
+    BOOT_LOG_DBG("bootutil_find_key:");
```
nordicjm, line 290: likewise here, the : makes it look like there is some expected output but other ones above do not have it when there is nothing to output

boot/bootutil/src/loader.c
```diff
@@ -283,6 +283,10 @@ static int
 boot_version_cmp(const struct image_version *ver1,
                  const struct image_version *ver2)
 {
+    BOOT_LOG_DBG("boot_version_cmp: ver1 %d.%d.%d vs ver2 %d.%d.%d",
+                 ver1->iv_major, ver1->iv_minor, ver1->iv_revision,
+                 ver2->iv_major, ver2->iv_minor, ver2->iv_revision);
+
```
nordicjm, line 289: this should just log the full versions part if the Kconfig is enabled, having it output in 2 parts is ugly

boot/bootutil/src/loader.c
```diff
@@ -2340,6 +2357,8 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
     bool has_upgrade;
     volatile int fih_cnt;
 
+    BOOT_LOG_DBG("context_boot_go:");
```
nordicjm, line 2360: and here

boot/bootutil/src/image_validate.c
```diff
@@ -120,6 +124,7 @@ bootutil_img_hash(struct boot_loader_state *state,
     /* Encrypted images only exist in the secondary slot */
     if (MUST_DECRYPT(fap, image_index, hdr) &&
             !boot_enc_valid(enc_state, 1)) {
+        BOOT_LOG_DBG("bootutil_img_hash Error: encrypted image found in primary slot");
```
nordicjm, line 127: : should be before error and error should be lowercase like other instances

boot/bootutil/src/bootutil_public.c
```diff
@@ -262,6 +264,7 @@ boot_read_swap_state(const struct flash_area *fap,
     off = boot_swap_info_off(fap);
     rc = flash_area_read(fap, off, &swap_info, sizeof swap_info);
     if (rc < 0) {
+        BOOT_LOG_DBG("boot_read_swap_state: error %d reading state", rc);
```
nvlsianpu, line 267: For all places where code exits with BOOT_EFLASH i would go with BOOT_LOG_ERR()

de-nordic: I have removed the logging here. At this point I am trying to avoid logging errors on read/write operations; they increase the code size and actually do nothing when flash is breaking, which is most probable case for these failures.

nvlsianpu (approved):

nordicjm (approved):

PR #2309, opened 2025-05-15
bootutil: Cut down allowed TLV list

de-nordic:
For some of TLV types we exactly know which are expected.


boot/bootutil/src/image_validate.c
```diff
@@ -448,16 +448,15 @@ static int bootutil_check_for_pure(const struct image_header *hdr,
...
-     IMAGE_TLV_ECDSA224,
-     IMAGE_TLV_ECDSA_SIG,
-     IMAGE_TLV_RSA3072_PSS,
-     IMAGE_TLV_ED25519,
+#if defined(EXPECTED_KEY_TLV)
+     EXPECTED_KEY_TLV,
+#endif
+#if defined(EXPECTED_HASH_TLV)
+     EXPECTED_HASH_TLV,
+#endif
```
nordicjm, line 456: would say if this is not set then it should allow them all

de-nordic: Good point

PR #2310, opened 2025-05-15
Allow bypasing signature hash key matching

de-nordic:
Currently MCUboot checks whether key hash in TLV matches any known builtin public keys.
When only one key in compiled in, most cases, then this match is pointless as image signature verification will just verify validity of the key anyway.
The PR adds MCUBOOT_BYPASS_KEY_MATCH option that allows to turn of the key matching and slightly reduces the binary size.

### DIFF
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -278,6 +278,8 @@ bootutil_img_hash(struct boot_loader_state *state,
 #   define KEY_BUF_SIZE         (SIG_BUF_SIZE + 24)
 #endif /* !MCUBOOT_HW_KEY */
 
+#if !defined(MCUBOOT_BYPASS_KEY_MATCH)
+/* Find functions are only needed when key is checked first */
 #if !defined(MCUBOOT_HW_KEY)
 static int
 bootutil_find_key(uint8_t *keyhash, uint8_t keyhash_len)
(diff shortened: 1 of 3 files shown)


boot/zephyr/Kconfig
```diff
@@ -325,6 +325,18 @@ endif
 
 endchoice
 
+config BOOT_BYPASS_KEY_MATCH
+	bool "Do not match TLV key hash against built in key"
+	depends on !BOOT_SIGNATURE_TYPE_NONE
```
nordicjm, line 330: depends on !hw key?

boot/bootutil/src/image_validate.c
```diff
@@ -631,7 +634,12 @@ bootutil_img_validate(struct boot_loader_state *state,
             if (rc) {
                 goto out;
             }
+#if !defined(MCUBOOT_BYPASS_KEY_MATCH)
             key_id = bootutil_find_key(buf, len);
+#else
+            /* There is only one key */
+            key_id = 0;
```
nvlsianpu: consider function returning 0 for MCUBOOT_BYPASS_KEY_MATCH:
``` C
static int bootutil_find_key(buf, len)
{
  (void) buf;
   (void) len;
   return 0;
}
```

rationale: less #ifdefs cutting function.

de-nordic:
[DNM] Till Zephyr release completes.

nvlsianpu (approved):

PR #2316, opened 2025-05-22
boot: bootutil: avoid full erase on BOOT_MAGIC_BAD in boot_set_next

ofirshe:
Previously, boot_set_next() would perform a full slot erase when BOOT_MAGIC_BAD was detected. This is unnecessary, as boot_validate_slot() will later handle the invalid image appropriately by invoking boot_scramble_slot().

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -573,10 +573,7 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
         if (active) {
             rc = BOOT_EBADVECT;
         } else {
-            /* The image slot is corrupt.  There is no way to recover, so erase the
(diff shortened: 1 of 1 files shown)


ofirshe:
@de-nordic

nordicjm (changes requested):
this is from the user application side, the image would not be scrambled because this function does not run from MCUboot

ofirshe:
> this is from the user application side, the image would not be scrambled because this function does not run from MCUboot

@nordicjm You're right - this function runs from the user application, so the image won't be scrambled at that point. What I meant is that the invalid image will be detected on the next reboot, and MCUboot will take the appropriate action.
I agree with @de-nordic that the decision on how to handle this should be left to the user, which is the motivation behind this PR. I've changed the commit message accordingly.

PR #2317, opened 2025-05-22
Updates for 2.2.0-rc2 release

d3zd3z:
Update release notes and the various release version files.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -76,6 +76,16 @@
 - Fixed issue with watchdog not being fed during flash erase
   operations, which could cause the watchdog to time out on long
   erase operations and prevent firmware updates from being possible.
+- Fix issues related to calculating the maximum image size for a given
+  configuration.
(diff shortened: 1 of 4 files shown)


nordicjm (approved):

docs/release-notes.md
```diff
@@ -76,6 +76,12 @@
 - Fixed issue with watchdog not being fed during flash erase
   operations, which could cause the watchdog to time out on long
   erase operations and prevent firmware updates from being possible.
+- Fix issues related to calculating the maximum image size for a given
+  configuration.
+- Fix an issue with sha hash calculations in a loop.
+- Fix an issue with the security counter being updated before an image is
+  confirmed.
+- Added a contributing guideline.
```
nvlsianpu, line 84: Can you add this one (#2306):
Fix an issues related to referencing arm-vector table of the application which causes jumping to wrong address instead of the application reset vector for some builds when zephyr LTO (linking time optimization) was enabled.

PR #2318, opened 2025-05-23
bootutil: Unify app_max_size() implementations

tomchy:
Remove redundand application size calculations in favour of common function, implemented inside bootutil_misc.c.
That way slot sizes use the same restrictions as image validation.


boot/bootutil/src/swap_scratch.c
```diff
@@ -892,17 +1007,31 @@ int app_max_size(struct boot_loader_state *state)
...
+#if MCUBOOT_SWAP_USING_SCRATCH
+    size_t primary_max_app_sz = app_max_size_adjust_to_trailer(state,
+                                                               BOOT_PRIMARY_SLOT,
+                                                               primary_sz);
+    size_t secondary_max_app_sz = app_max_size_adjust_to_trailer(state,
+                                                                BOOT_SECONDARY_SLOT,
+                                                                secondary_sz);
+    return (primary_max_app_sz < secondary_max_app_sz ?
+            primary_max_app_sz : secondary_max_app_sz);
+#else /* MCUBOOT_SWAP_USING_SCRATCH */
```
taltenbach: Might have been confused by the numerous #ifdef in this file, but can this code really be compiled when MCUBOOT_SWAP_USING_SCRATCH is used? I would expect the first implementation of `app_max_size` in this file to be always used when swap-scratch is selected

boot/bootutil/src/swap_scratch.c
```diff
@@ -47,6 +47,119 @@ int boot_status_fails = 0;
 #define BOOT_STATUS_ASSERT(x) ASSERT(x)
 #endif
 
+#if MCUBOOT_SWAP_USING_SCRATCH
+size_t
+boot_get_first_trailer_sector(struct boot_loader_state *state, size_t slot, size_t trailer_sz)
```
taltenbach, line 61: nitpick: This routine could be static since it is not exposed anymore outside this file

taltenbach, line 61: Could be useful to keep the doxygen documentation that was in `bootutil_priv.h`

boot/bootutil/src/swap_scratch.c
```diff
@@ -47,6 +47,119 @@ int boot_status_fails = 0;
...
+     * |              sector               |
+     * +-----------------------------------+
+     * |Image->|             |<--trailer---|
+     * +-----------------------------------+
+     * |         |<----scratch trailer---->|
+     * +-----------------------------------+
+     *            <-padding->
+     *
+     * The value of the padding depends on the amount of trailer data that is contained in the first
+     * trailer containing part of the trailer in the primary and secondary slot.
```
taltenbach:
```suggestion
     * sector containing part of the trailer in the primary and secondary slot.
```
Whoops, typo of mine

boot/bootutil/src/swap_scratch.c
```diff
@@ -47,6 +47,119 @@ int boot_status_fails = 0;
...
+
+    trailer_sz_in_first_sector = trailer_sector_end_off - slot_trailer_off;
+
+    size_t trailer_padding = 0;
+    size_t scratch_trailer_sz = boot_scratch_trailer_sz(BOOT_WRITE_SZ(state));
+
+    /* If the scratch trailer doesn't fit in the last sector of the primary or secondary slot,
+     * some padding might have to be inserted between the end of the firmware image and the
+     * beginning of the trailer to ensure there is enough space for the trailer in the scratch area
+     * when the last sector of the secondary will be copied to the scratch area.
```
taltenbach:
```suggestion
    /* Some padding might have to be inserted between the end of the firmware image and the
     * beginning of the trailer to ensure there is enough space for the trailer in the scratch area
     * when the last sector of the secondary will be copied to the scratch area.
```
Padding can be needed even if the scratch trailer is smaller than the last sector of both the primary and slot (and in practice that's usually the case since the scratch trailer is very small). See the example on https://github.com/mcu-tools/mcuboot/pull/2206#issue-2855865110.

For the context: the initial comment was saying "If the trailer doesn't fit in the last sector", where "trailer" was denoting the slot trailer and not the scratch trailer, since I wanted to explain that padding is necessary when `slot_trailer_sz` is greater than the size of the last sector of both the primary and secondary slots. Indeed, before the fix adding this padding, the code was only considering the case where `slot_trailer_sz` is lower than the size of the last sector of either the primary or secondary slot, in which case padding is never needed since `slot_trailer_sz >= scratch_trailer_sz` so there is necessarily enough space for the scratch trailer at the end of the scratch area when swapping the sectors that contain the trailers.

taltenbach (commented):
Changes regarding swap-scratch LGTM, only minor comments. Thanks for the detailed explanation in `app_max_size_adjust_to_trailer`, this is now a lot clearer than before!

boot/bootutil/src/swap_scratch.c
tomchy: It seems I was confused - you're right, it is for `MCUBOOT_DIRECT_XIP` and `MCUBOOT_RAM_LOAD`, so the `MCUBOOT_SWAP_USING_SCRATCH` can be removed.

I think it might be beneficial (at least for the reader :slightly_smiling_face: ) to isolate non-scratch code from this file in the future.

boot/bootutil/src/swap_scratch.c
tomchy: You're right - using the term `last sector` may cause more confusion than expected.
I've applied your suggestion and added an additional sector inside the graph to make it clear that scratch trailer is smaller and placed differently than the regular trailer. As a bonus - scratch area size is also added.

boot/bootutil/src/swap_scratch.c
tomchy: Added missing docs for the remaining, new static functions.

boot/bootutil/src/swap_scratch.c
taltenbach: > I think it might be beneficial (at least for the reader 🙂 ) to isolate non-scratch code from this file in the future.

I definitely agree! I guess many upgrade strategies are in this `swap_scratch.c` file for historical reasons and should be move in their dedicated file

boot/bootutil/src/swap_scratch.c
taltenbach: Very clear now, thanks!

taltenbach:
LGTM, @tomchy feel free resolve all open threads (I can't do it myself)

PR #2319, opened 2025-05-23
boot: zephyr: boards: frdm_mcxn947_qspi and mcx_n9xx_evk_qspi: remove overlays

DerekSnell:
Zephyr support for MCXN947 QSPI variants boots directly from QSPI flash.  Remove overlays to keep boot_partition in QSPI flash.  Also removes CONFIG_BOOT_UPGRADE_ONLY=y, because swap mode is supported when image slots are in QSPI.


butok:
Thanks for removing the broken frdm_mcxn947_qspi.

PR #2154, opened 2024-12-18
mcuboot supports the stm32 external flash in XiP mode

FRASTM:
New version of the old PR https://github.com/mcu-tools/mcuboot/pull/1891

Adding the stm32h7s3l8 nucleo board

Following https://github.com/zephyrproject-rtos/zephyr/discussions/76305 

Build the application with sysbuild  for the nucleo_h7S3L8 : 
1. `west build -p auto -b nucleo_h7s3l8 samples/boards/st/hello_world_xip --sysbuild -- -DSB_CONFIG_BOOTLOADER_MCUBOOT=y `
3. then, download mcuboot at address 0x8000000 and download the  signed hello_world_xip application at 0x70000000 (where i can read the magic nb)

PR #2323, opened 2025-05-26
Xip mcuboot

FRASTM:
New version of the https://github.com/mcu-tools/mcuboot/pull/2154

Adding the stm32h7s3l8 nucleo board

Following https://github.com/zephyrproject-rtos/zephyr/discussions/76305

Build the application with sysbuild for the nucleo_h7S3L8 :

west build -p auto -b nucleo_h7s3l8 samples/boards/st/hello_world_xip --sysbuild -- -DSB_CONFIG_BOOTLOADER_MCUBOOT=y 
then, download mcuboot at address 0x8000000 and download the signed hello_world_xip application at 0x70000000 (where i can read the magic nb)


boot/zephyr/flash_map_extended.c
```diff
@@ -20,7 +20,23 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
-#if (!defined(CONFIG_XTENSA) && DT_HAS_CHOSEN(zephyr_flash_controller))
+#if defined(CONFIG_STM32_MEMMAP)
+/* MEMORY MAPPED for XiP on external NOR flash takes the sspi-nor or ospi-nor or qspi-nor device */
+#define FLASH_DEVICE_ID SPI_FLASH_0_ID
+#if DT_NODE_HAS_STATUS(DT_INST(0, st_stm32_xspi_nor), okay)
+#define FLASH_DEVICE_NODE DT_INST(0, st_stm32_xspi_nor)
+#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST(0, st_stm32_xspi),1)
```
erwango, line 28: Instance 0 of `st_stm32_xspi` could be used for other purposes (PSRAM on N6-DK for instance).
We should ensure we get the parent of the `st_stm32_xspi_nor` node instead.
Something like:
```suggestion
#define DT_DRV_COMPAT st_stm32_xspi_nor
#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)
```

erwango (changes requested):

boot/zephyr/boards/b_u585i_iot02a.overlay
```diff
@@ -0,0 +1,40 @@
...
+/delete-node/ &scratch_partition;
+/delete-node/ &storage_partition;
+
+&mx25lm51245 {
+	partitions {
+		compatible = "fixed-partitions";
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		slot0_partition: partition@0 {
```
nordicjm, line 23: as I said in the other PR, adding this here is pointless, this should go in your base board file in zephyr otherwise you will have completely different partitions for mcuboot vs every application, and the applications will not boot because of that

boot/zephyr/boards/b_u585i_iot02a.overlay
```diff
@@ -0,0 +1,40 @@
...
+			reg = <0x00000000 DT_SIZE_K(416)>;
+		};
+		slot1_partition: partition@68000 {
+			label = "image-1";
+			reg = <0x00068000 DT_SIZE_K(416)>;
+		};
+		scratch_partition: partition@d0000 {
+			label = "image-scratch";
+			reg = <0x000d00000 DT_SIZE_K(64)>;
+		};
```
nordicjm, line 34: and unless you have different sector sizes or variable sector sizes, scratch should not be present

boot/zephyr/flash_map_extended.c
titouanc: I just ran into this exact issue on my STM32H7S78-DK, see https://github.com/zephyrproject-rtos/zephyr/pull/88052#issuecomment-2990216467

This is because the flash for XiP is on XSPI2, and there is a PSRAM on XSPI1 which is therefore enabled in the device tree. Consequently, the macro being reviewed here is mistakenly using the XSPI1 (PSRAM) address space instead of XSPI2.

I solved the issue with the following change:
```suggestion
#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_PARENT(FLASH_DEVICE_NODE),1)
```

I think we could do the same for ospi/qspi as well

erwango:
Please rebase, following merge of #2348 .

boot/zephyr/flash_map_extended.c
FRASTM: ok, done

boot/zephyr/boards/b_u585i_iot02a.overlay
FRASTM: I understand, that the slot 0 & 1 partitions are kept in the flash0
However, when building the samples/sysbuild/with_mcuboot/ --sysbuild, is there any  b_u585i_iot02a.overlay,   to declare the `zephyr,code-partition` to be placed in the external flash, not in the flash0 ?
The slot0_partition is usually in the flash0, but for that samples, we want to place the slot0 in the external flash. The code must then be compiled, linked and downloaded in the external flash. 
I see that ./build/with_mcuboot/zephyr/zephyr.signed.bin  is linked at address  0 instead of 0x70000000
But this could be another problem : _CONFIG_FLASH_BASE_ADDRESS 0x0_

nordicjm: no, a board variant should be used for that (either in tree or out of tree with board extensions)

boot/zephyr/flash_map_extended.c
```diff
@@ -24,15 +24,14 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 /* MEMORY MAPPED for XiP on external NOR flash takes the sspi-nor or ospi-nor or qspi-nor device */
 #define FLASH_DEVICE_ID SPI_FLASH_0_ID
 #if DT_NODE_HAS_STATUS(DT_INST(0, st_stm32_xspi_nor), okay)
-#define DT_DRV_COMPAT st_stm32_xspi_nor
 #define FLASH_DEVICE_NODE DT_INST(0, st_stm32_xspi_nor)
-#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)
+#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_PARENT(FLASH_DEVICE_NODE),1)
```
erwango: The initial line is correct. Instance should actually be taken into account (in case another xspi controller is dealing with another memory as in N6 case.)
This should rather be extended to other IPs IMO

erwango (changes requested):

erwango:
> next step : (end of july) : going to add a variant for the stm32U5 series

I think we should get at least the `flash_extended.c` update merged right now since it is broken for QSPI/OSPI atm due to the bindings change.

FRASTM:
> > next step : (end of july) : going to add a variant for the stm32U5 series
> 
> I think we should get at least the `flash_extended.c` update merged right now since it is broken for QSPI/OSPI atm due to the bindings change.

OK, so :
- [ ]  use this PR for flash_map extended.c
- [ ] set another PR for the stm32U5 series (st_stm32_ospi_nor compatible) defining a variant

NB Still CI failure due to strange error on signature

erwango:
@de-nordic Any idea why the `./ci/check-signed-off-by.sh` is complaining about ?
Supposedly guilty commit is signed ....

erwango (approved):

boot/zephyr/flash_map_extended.c
FRASTM: keeping DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)

titouanc (approved):
For what it's worth, successfully tested on my STM32H7Z78-DK (with latest Zephyr main at b3ccee473ed64ac9ba7260beafb4fd175eee13ab)

erwango:
> [zephyrproject-rtos/zephyr#88404](https://github.com/zephyrproject-rtos/zephyr/issues/88404)

https://github.com/zephyrproject-rtos/zephyr/issues/88404 has been integrated in v4.2 to align ST bindings with dtc recommendations (and other vendors)
This involves a different way to access size information, which current PR is addressing. Otherwise MCUBoot can't access size of the flash. This is already fixed fir XSPI, but missing for O/QSPI

nordicjm (approved):

PR #2337, opened 2025-06-11
bootutil: Fix boot_scramble_region escaping flash area

de-nordic:
Incorrect range check fix.


boot/bootutil/src/loader.c
```diff
@@ -1291,12 +1291,12 @@ boot_scramble_region(const struct flash_area *fa, uint32_t off, uint32_t size, b
 
                 off -= write_block;
```
nordicjm, line 1292: does this need changing also?

nordicjm (approved):
Needs doc release note adding

boot/bootutil/src/loader.c
nvlsianpu: Its backwards, so last chunk to be erase has `off== end_offset`.

PR #2338, opened 2025-06-12
boot: bootutil: Fix encryption/decryption size during copy

Stephane-LeRoy:
During region copy, when chunk contains both header and TLVs, the encryption/decryption size is too large as it takes into account TLVs only.
The fix modifies the encryption/decryption size to take into account both header and TLVs.


nvlsianpu:
@Stephane-LeRoy How you encountered this issue? What is your BUF_SZ?

Stephane-LeRoy:
Issue is visible when image is smaller than chunk size (ie buf size)

PR #2342, opened 2025-06-16
zephyr: Fix pinreset trigger

tomchy:
The hfwinfo returns bitmask, not single values.

### DIFF
--- a/boot/zephyr/io.c
+++ b/boot/zephyr/io.c
@@ -181,7 +181,7 @@ bool io_detect_pin_reset(void)
 
     rc = hwinfo_get_reset_cause(&reset_cause);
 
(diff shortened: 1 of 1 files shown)


boot/zephyr/io.c
```diff
@@ -181,7 +181,7 @@ bool io_detect_pin_reset(void)
 
     rc = hwinfo_get_reset_cause(&reset_cause);
 
-    if (rc == 0 && reset_cause == RESET_PIN) {
+    if (rc == 0 && (reset_cause & RESET_PIN) == RESET_PIN) {
```
nvlsianpu: `reset_cause & RESET_PIN` is evaluated to True if flag is set.

nordicjm:
@tomchy can you set PR as ready so it can be merged?

PR #2343, opened 2025-06-16
loader: Allow to specify slot number in version comparison

tomchy:
Allow to depend on a specific slot while specifying the version number. This functionality is useful when the Direct XIP mode is used and the booting process of other images is done by the next stage, not the MCUboot itself.


de-nordic (approved):

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -120,6 +120,10 @@
 #define MCUBOOT_VERSION_CMP_USE_BUILD_NUMBER
 #endif
 
+#ifdef CONFIG_BOOT_VERSION_CMP_USE_SLOT_NUMBER
+#define MCUBOOT_VERSION_CMP_USE_SLOT_NUMBER
+#endif
```
de-nordic, line 129: Some more verbose description on `MCUBOOT_VERSION_CMP_USE_SLOT_NUMBER` please. I know that this, in this case, is translation from Kconfig, but it affects generic code and we are, at this point, very bad at documenting all the crap we add to the mcuboot_config.h files.

boot/bootutil/src/loader.c
```diff
@@ -411,6 +411,24 @@ boot_verify_slot_dependency(struct boot_loader_state *state,
     uint8_t swap_type = state->swap_type[dep->image_id];
     dep_slot = BOOT_IS_UPGRADE(swap_type) ? BOOT_SECONDARY_SLOT
                                           : BOOT_PRIMARY_SLOT;
+#elif defined(MCUBOOT_VERSION_CMP_USE_SLOT_NUMBER)
+    switch(dep->slot) {
```
de-nordic, line 321: boot_verify_slot_dependencies does range check for image number, on the dependency TLV it read, it should do the same with slot number.

scripts/imgtool/main.py
```diff
@@ -404,7 +424,7 @@ def convert(self, value, param, ctx):
                    '(for mcuboot <1.5)')
 @click.option('-d', '--dependencies', callback=get_dependencies,
               required=False, help='''Add dependence on another image, format:
-              "(<image_ID>,<image_version>), ... "''')
+              "(<image_ID>,[<slot_ID>],<image_version>), ... "''')
```
de-nordic: I guess that these
image
are valid ID I can use here?
Bug there is no user input filter here. I know that MCUboot will reject the image, at least by the logic of `boot_verify_slot_dependency` (which should actually be in `boot_verify_slot_dependencies`), but it is not even clear what values user should pass here and if 0 is special.

tomchy: There is only a slight validation, based on the regular expression.
Since there is close to none justification for using more than those three predefined values, I'll try to use words instead of mysterious integers.

tomchy: I'm not sure about the change of placement - this logic must affect the local `dep_slot` variable, that's why it is there.
I guess you have some ideas, so we can discuss possible solutions tomorrow.

de-nordic (approved):

de-nordic:
@robertstypa Please take a look at the snake language code.

docs/design.md
```diff
@@ -890,6 +890,23 @@ process is presented below.
 + Boot into image in the primary slot of the 0th image position\
   (other image in the boot chain is started by another image).
 
+By enabling the `MCUBOOT_VERSION_CMP_USE_SLOT_NUMBER` configuration option,
+the dependency check may be extended to match for a spacified slot of a specific
```
kszromek-nordic:
```suggestion
the dependency check may be extended to match for a specified slot of a specific
```

docs/imgtool.md
```diff
@@ -91,7 +91,8 @@ primary slot and adds a header and trailer that the bootloader is expecting:
                                       the `auto` keyword to automatically generate
                                       it from the image version.
       -d, --dependencies TEXT         Add dependence on another image, format:
-                                      "(<image_ID>,<image_version>), ... "
+                                      "(<image_ID>,[<slot:active|primary|secondary>],
```
kszromek-nordic: Taking into account regex in https://github.com/mcu-tools/mcuboot/pull/2343/files#diff-a483337de0b1dd62c231e528c8c3a613542333df34af495467510d8db6b6b35bR313 shouldn't a comma end up inside square brackets?

PR #2349, opened 2025-06-23
boot_serial: Fix null access when validating images from bs_list or bs_set

### LINKED ISSUE #2336: Serial recovery with image state has null pointer usage
Commit d00b11dc234caee267eb12376e2279ed57c35d5d has introduced a regression, this added using the state to the maximum image size function, in serial recovery mode however this variable is NULL, which then causes a null pointer reference and undefined operation in swap using scratch, swap using move and swap using offset modes

taltenbach:
This PR fixes a regression in `boot_serial` caused by https://github.com/mcu-tools/mcuboot/commit/d00b11dc234caee267eb12376e2279ed57c35d5d. Until that commit it was possible to call `bootutil_img_validate` with a null state, provided the image is not encrypted. This is currently done by `bs_list` and `bs_set`.

However, this is no longer possible since `bootutil_max_image_size`, called from `bootutil_img_validate`, now needs a valid bootloader state for all swap strategies.

The issue wasn't noticed immediately and calling `bs_list` or `bs_set` currently lead to a null pointer access.

This MR proposes to fix the issue by initializing a valid bootloader state each time `bs_list` or `bs_set` is invoked. As a side effect, this allow to remove some code duplication and to simplify the code in some places.

The code was tested on an STM32H747 MCU using Zephyr 4.1.0. 

Fixes #2336.


nordicjm:
rebase on `main` to fix python error

boot/boot_serial/src/boot_serial.c
```diff
@@ -282,10 +282,10 @@ bs_list_img_ver(char *dst, int maxlen, struct image_version *ver)
  * List images.
  */
 static void
-bs_list(char *buf, int len)
+bs_list(char *buf, int len, struct boot_loader_state *state)
```
de-nordic: I think that state should be first; this is state of the system so it is the most important object here, I think.

boot/boot_serial/src/boot_serial.c
```diff
@@ -495,7 +475,7 @@ bs_list(char *buf, int len)
  * Set image state.
  */
 static void
-bs_set(char *buf, int len)
+bs_set(char *buf, int len, struct boot_loader_state *state)
```
de-nordic: Same here, state is most important object.

boot/boot_serial/src/boot_serial.c
```diff
@@ -295,8 +295,11 @@ bs_list(char *buf, int len)
...
-    image_index = 0;
-    IMAGES_ITER(image_index) {
+
+    IMAGES_ITER(BOOT_CURR_IMG(state)) {
+        image_index = BOOT_CURR_IMG(state);
+        (void) image_index; /* Might be unused depending on the configuration */
+
 #if defined(MCUBOOT_SERIAL_IMG_GRP_IMAGE_STATE) || defined(MCUBOOT_SWAP_USING_OFFSET)
         int swap_status = boot_swap_type_multi(image_index);
 #endif
```
nordicjm: variables are declared first, so this needs re-ordering (and can just reference `BOOT_CURR_IMG(state)` instead)

boot/boot_serial/src/boot_serial.c
```diff
@@ -544,40 +524,32 @@ bs_set(char *buf, int len)
     }
 
     if (img_hash.len != 0) {
-        IMAGES_ITER(image_index) {
+        IMAGES_ITER(BOOT_CURR_IMG(state)) {
+            image_index = BOOT_CURR_IMG(state);
+            (void) image_index; /* Might be unused depending on the configuration */
+
 #ifdef MCUBOOT_SWAP_USING_OFFSET
             int swap_status = boot_swap_type_multi(image_index);
```
nordicjm: as above

boot/boot_serial/src/boot_serial.c
```diff
@@ -691,16 +651,52 @@ bs_rc_rsp(int rc_code)
...
+
+#if !defined(MCUBOOT_DIRECT_XIP) && !defined(MCUBOOT_RAM_LOAD)
+    IMAGES_ITER(BOOT_CURR_IMG(state)) {
+        rc = boot_read_sectors(state, NULL);
+        if (rc != 0) {
+            BOOT_LOG_ERR("Failed to read sectors: %d", rc);
+            rc = MGMT_ERR_EUNKNOWN;
+            goto out;
+        }
+    }
```
nordicjm, line 676: shouldn't be needed to have all images present for this I don't think

boot/boot_serial/src/boot_serial_encryption.c
```diff
@@ -241,18 +228,14 @@ decrypt_image_inplace(const struct flash_area *fa_p,
 #if 0 //Skip this step?, the image will just not boot if it's not decrypted properly
         static uint8_t tmpbuf[BOOT_TMPBUF_SZ];
          /* First check if the encrypted image is a good image before decrypting */
-        FIH_CALL(boot_image_validate_encrypted,fih_rc,fa_p,&_hdr,tmpbuf,BOOT_TMPBUF_SZ);
+        FIH_CALL(boot_image_validate_encrypted,fih_rc,state,fa_p,&_hdr,tmpbuf,BOOT_TMPBUF_SZ);
```
nordicjm: add spaces to this

boot/bootutil/src/loader.c
```diff
@@ -2339,26 +2262,16 @@ check_downgrade_prevention(struct boot_loader_state *state)
...
-     * that they don't get allocated for non-boot-loader apps.  This is
-     * necessary because the gcc option "-fdata-sections" doesn't seem to have
-     * any effect in older gcc versions (e.g., 4.8.4).
-     */
-    TARGET_STATIC boot_sector_t primary_slot_sectors[BOOT_IMAGE_NUMBER][BOOT_MAX_IMG_SECTORS];
-    TARGET_STATIC boot_sector_t secondary_slot_sectors[BOOT_IMAGE_NUMBER][BOOT_MAX_IMG_SECTORS];
-#if MCUBOOT_SWAP_USING_SCRATCH
-    TARGET_STATIC boot_sector_t scratch_sectors[BOOT_MAX_IMG_SECTORS];
-#endif
+    struct boot_sector_buffer _sectors;
```
nordicjm: let's not declare variables with undescores at the first character

boot/boot_serial/src/boot_serial.c
taltenbach: Makes sense, fixed!

boot/boot_serial/src/boot_serial.c
taltenbach: Fixed

boot/boot_serial/src/boot_serial_encryption.c
taltenbach: Done

boot/bootutil/src/loader.c
taltenbach: Name changed

boot/boot_serial/src/boot_serial.c
taltenbach: Even if some images are not present, I guess `boot_read_sectors` should succeed, isn't it? If my understanding is correct this routine is simply fetching the layout of the flash areas from the flash driver, so is not expected to perform any flash memory access nor to fail if an image is not present

nordicjm (approved):

boot/boot_serial/src/boot_serial.c
nordicjm: have checked the out of tree code and it works with the changes

boot/boot_serial/src/boot_serial.c
```diff
@@ -295,11 +295,13 @@ bs_list(char *buf, int len)
...
-    image_index = 0;
-    IMAGES_ITER(image_index) {
+
+    IMAGES_ITER(BOOT_CURR_IMG(state)) {
 #if defined(MCUBOOT_SERIAL_IMG_GRP_IMAGE_STATE) || defined(MCUBOOT_SWAP_USING_OFFSET)
-        int swap_status = boot_swap_type_multi(image_index);
+        int swap_status = boot_swap_type_multi(BOOT_CURR_IMG(state));
 #endif
+        image_index = BOOT_CURR_IMG(state);
+        (void) image_index; /* Might be unused depending on the configuration */
```
de-nordic, line 304: These lines should probably be above line 300.

boot/boot_serial/src/boot_serial.c
```diff
@@ -544,40 +523,31 @@ bs_set(char *buf, int len)
...
 
     if (img_hash.len != 0) {
-        IMAGES_ITER(image_index) {
+        IMAGES_ITER(BOOT_CURR_IMG(state)) {
 #ifdef MCUBOOT_SWAP_USING_OFFSET
-            int swap_status = boot_swap_type_multi(image_index);
+            int swap_status = boot_swap_type_multi(BOOT_CURR_IMG(state));
 #endif
+            image_index = BOOT_CURR_IMG(state);
+            (void) image_index; /* Might be unused depending on the configuration */
```
de-nordic, line 531: If moved above `#ifdef MCUBOOT_SWAP_USING_OFFSET`, the `int swap_status = boot_swap_type_multi(image_index);` would not have to been modified.

PR #2362, opened 2025-07-01
bootutil: Use flash base address for direct hash

tomchy:
Fetch the flash base address if direct hash calculation is performed.

Fixes: https://github.com/mcu-tools/mcuboot/issues/2368

### DIFF
--- a/boot/bootutil/src/image_validate.c
+++ b/boot/bootutil/src/image_validate.c
@@ -84,6 +84,10 @@ bootutil_img_hash(struct boot_loader_state *state,
     uint32_t off;
     uint32_t blk_sz;
 #endif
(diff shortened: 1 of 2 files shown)


de-nordic:
> > @tomchy Is this a bug fix?
> 
> Indeed, this should be considered as a bugfix. I do not see why the hash function should accept the offset to the NVM memory mapped to `(void *)` instead of the absolute address 🙂

It used to work, at least for nrf platforms, because our flash starts at 0, so flash area offset is actually the offset in address space where that area actually starts; that is not true, though, for every device.

Can you report issue in MCUboot so that we can use it as a fix and add release notes snippet here https://github.com/mcu-tools/mcuboot/tree/main/docs/release-notes.d ?

de-nordic (approved):

PR #2370, opened 2025-07-02
boot: zephyr: RAM cleanup debug loop

michalek-no:
Option to put execution in infinite loop.
Meant to be used for debug.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -364,6 +364,13 @@ config MCUBOOT_CLEANUP_RAM
 	help
 	  Sets contents of memory to 0 before jumping to application.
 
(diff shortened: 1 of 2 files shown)


boot/zephyr/Kconfig
```diff
@@ -364,6 +364,13 @@ config MCUBOOT_CLEANUP_RAM
 	help
 	  Sets contents of memory to 0 before jumping to application.
 
+config MCUBOOT_LOOP_AFTER_RAM_CLEANUP
+	bool "Loop after RAM cleanup"
```
de-nordic:
```suggestion
	bool "Infinite loop after RAM cleanup"
```

nordicjm (approved):

PR #2373, opened 2025-07-03
boot: Provide default configuration for nRF54H20

tomchy:
Enable ed25519 signature as well as direct hashing while building for nRF54H20DK.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -146,6 +146,7 @@ config BOOT_IMG_HASH_ALG_SHA512_ALLOW
 
 config BOOT_IMG_HASH_DIRECTLY_ON_STORAGE
 	bool "Hash calculation functions access storage through address space"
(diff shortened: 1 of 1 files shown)


boot/zephyr/Kconfig
```diff
@@ -146,6 +146,7 @@ config BOOT_IMG_HASH_ALG_SHA512_ALLOW
 
 config BOOT_IMG_HASH_DIRECTLY_ON_STORAGE
 	bool "Hash calculation functions access storage through address space"
+	default y if NRF_SECURITY && SOC_NRF54H20
```
nordicjm, line 149: do not reference ncs variables in upstream

tomchy: In that case, it might be easier to keep it as noup.
This is just the `NRF_SECURITY` limitation that the current backend does not support RSA.

PR #2375, opened 2025-07-04
zephyr: boards: nrf - remove redundant multithreading configuration

nordic-mik7:
Remove configs that enable multithreading just because of SPI/QSPI use. Currently, nrf drivers do not depend on multithreading, so it is not needed and this change can save memory usage.


boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
```diff
@@ -1,3 +1,2 @@
 CONFIG_NORDIC_QSPI_NOR=n
 CONFIG_SPI=n
-CONFIG_MULTITHREADING=y
```
nordicjm: this is needed for usb

nordicjm (changes requested):

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
nordic-mik7: ack

de-nordic (approved):

PR #2377, opened 2025-07-04
Crypto key housekeeping

de-nordic:
Add generic definition for bootutil_crypto_key_houseekping function and call point in Zephyr mcuboot main.
The function is intended to be MCUboot, and provided by crypto backend, to be called once key access, which may be provided by the backend, is no longer needed.
Function should implement cleanup of any backend stored private key objects and apply key access policies that affect how code executed from the point of function invocation, including application code, will be able to use/access keys.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -108,6 +108,7 @@ zephyr_library_sources(
   ${BOOT_DIR}/bootutil/src/image_validate.c
   ${BOOT_DIR}/bootutil/src/tlv.c
   ${BOOT_DIR}/bootutil/src/encrypted.c
+  ${BOOT_DIR}/bootutil/src/crypto_common.c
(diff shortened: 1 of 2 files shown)


nordicjm:
typo in second commit

boot/bootutil/src/crypto_common.c
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Copyright (c) 2025 Nordic Semiconductor ASA
+ */
+
+#include <mcuboot_config/mcuboot_config.h>
+#include "bootutil/bootutil_log.h"
+
+BOOT_LOG_MODULE_DECLARE(mcuboot);
+
+__attribute__((__weak__)) void bootutil_crypto_key_housekeeping(void)
```
nordicjm, line 12: implement as same as https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/mcuboot_status.h#L24 does

PR #2382, opened 2025-07-10
bootutil: Fix comment formatting

maulik-arm:
This commit fixes the formatting of comments in the `image_validate.c` file to ensure they are properly aligned.


adeaarm:
> I'm a little curious about the Change-Id header in the commit. We don't use those on github, but I guess it doesn't really hurt anything, either.

We have internal mirrors of MCUboot that are hosted in Gerrit hence the Change-Id. It's harmless and makes our life easier

PR #2390, opened 2025-07-16
zephyr: Improve SHA support selectors

de-nordic:
Commit introduces BOOT_SOMETHING_USES_SHA<256,384,512> Kconfig options that can be used to control what algorithms should be compiled in with crypto backends.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -70,15 +70,20 @@ if BOOT_USE_PSA_CRYPTO
 
 config BOOT_PSA_IMG_HASH_ALG_SHA256_DEPENDENCIES
 	bool
-	default y if BOOT_IMG_HASH_ALG_SHA256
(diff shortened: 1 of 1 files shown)


boot/zephyr/Kconfig
```diff
@@ -627,14 +659,35 @@ config BOOT_ENCRYPT_X25519
...
-	  By default SHA256 is used for HKDF/HMAC in key exchange expansion
-	  and verification. This options switches to SHA512. The option is
-	  mainly useful to reduce numer of compiled in SHA algorithms.
+	  With ED25519, this option is worth selecting when SHA512 is used
+	  for hashing of an image to reduce number of compiled sha algorithms.
+
+endchoice # BOOT_HMAC_SHA
+
+endif # BOOT_ENCRYPT_X25519 && BOOT_USE_PSA_CRYPTO
+
```
nordicjm, line 690:
```suggestion
```
nit.

PR #2395, opened 2025-07-18
imgtool: add support for key password to enable non-interactive key loading with password

meremST:
This PR introduces password support for key files in imgtool to facilitate signing binaries for multiple boards requiring different key pairs. This can allow user to simplify automation by enabling  non-interactive key loading with password.


meremST:
Hello, 

I've improved the implementation and also added support of passphrase retries in case of error.

Cheers

PR #2398, opened 2025-07-22
espressif: fix esp32 console_write loop forever if cnt too large

moculll:
spilt buffer to make sure console_write won't occupy forever


almir-okato:
> > Looks good to me. Only 2 points, could you please sign your commit and also rebase it?
> 
> verified the commit and rebased it, thanks dude

Thanks!
I see you've authenticated your commit, however the CI is still failing when it checks for "Signed off by" in the commit. That would be done by using `git commit -s` instead of using the capital `-S` argument.

PR #2404, opened 2025-07-25
bootutil: Expose hardware security counter

tomchy:
Add a possibility to read the current value of the hardware security counter from the application.


boot/bootutil/src/boot_record.c
```diff
@@ -346,7 +353,28 @@ int boot_save_shared_data(const struct image_header *hdr, const struct flash_are
...
+            /* Some platforms support only a single security counter. */
+            continue;
+        }
+
+        rc = boot_add_data_to_shared_area(TLV_MAJOR_BLINFO,
+                                          BLINFO_SECURITY_COUNTER_IMAGE_0 + image,
+                                          sizeof(security_cnt),
+                                          (void *)&security_cnt);
+        if (!rc) {
+            break;
```
ahasztag, line 370: I see this is not present in the loop below. Perhaps similar code should be also added there?

de-nordic (approved):

boot/bootutil/src/boot_record.c
de-nordic: @ahasztag Can you re-review?

ahasztag (approved):

PR #2407, opened 2025-07-29
boot: zephyr: boards: set configs for nucleo_g071rb

djiatsaf-st:
Deactivate CONFIG_BOOT_MAX_IMG_SECTORS_AUTO by setting it to n, and set CONFIG_BOOT_MAX_IMG_SECTORS to 256.


boot/zephyr/boards/nucleo_g071rb.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_BOOT_MAX_IMG_SECTORS_AUTO=n
+CONFIG_BOOT_MAX_IMG_SECTORS=256
```
JarmouniA, line 2: This should not be needed, as it can be calculated automatically based on
https://github.com/zephyrproject-rtos/zephyr/blob/d53f324ca09dcafda0923a38f773b1538857f2f3/dts/arm/st/c0/stm32c0.dtsi#L150
and image size.
https://github.com/mcu-tools/mcuboot/blob/7a9f30f64729a91ee3f1efc1834b7cbeaf3dd7b4/boot/zephyr/CMakeLists.txt#L515
If it's not calculated properly when MAX_IMG_SECTORS_AUTO=y, the cause should be investigated, instead of patching the problem elsewhere which does not scale with the number of boards in Zephyr.

PR #2409, opened 2025-07-31
Add vendor and image class UUID support

tomchy:
According to the PSA Platform Security Boot Guide:
> R20_PSBG_MANIFEST: Each image manifest must contain the properties specified in Figure 2. It is permitted for
additional information to be included.

The manifest (which can be interpreted as image header and additional metadata stored inside TLVs) must include:
- Image type:  Specifically identifies the boot stage or peripheral. Also known as a purpose identifier. This field is not required if only one type of image is supported.
- Product class: This can refer to a vendor ID, product ID, or a more specific identifier, such as the
Instance ID.

This PR introduces two identifiers to meet those requirements:
- Image class ID, defined as RFC 9562 UUID Version 5, that may contain all: image type and product type values as input.
- Image vendor ID, defined as RFC 9562 UUID Version 5, that provides the namespace for the image class ID, so two devices with the same type/product name will have different image class UUIDs.


boot/bootutil/src/image_validate.c
```diff
@@ -749,6 +761,64 @@ bootutil_img_validate(struct boot_loader_state *state,
...
+            }
+
+            rc = LOAD_IMAGE_DATA(hdr, fap, off, img_uuid_vid.raw, len);
+            if (rc) {
+                goto out;
+            }
+
+            FIH_CALL(boot_uuid_vid_match, fih_rc, image_index, &img_uuid_vid);
+            if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+                FIH_SET(fih_rc, FIH_FAILURE);
```
de-nordic: This line should set uuid_vid_valid, not the assignment below.

boot/bootutil/src/image_validate.c
```diff
@@ -749,6 +761,64 @@ bootutil_img_validate(struct boot_loader_state *state,
...
+            }
+
+            rc = LOAD_IMAGE_DATA(hdr, fap, off, img_uuid_cid.raw, len);
+            if (rc) {
+                goto out;
+            }
+
+            FIH_CALL(boot_uuid_cid_match, fih_rc, image_index, &img_uuid_cid);
+            if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+                FIH_SET(fih_rc, FIH_FAILURE);
```
de-nordic: This line should set the uuid_cid_valid, not the assignment below.

boot/bootutil/src/image_validate.c
```diff
@@ -749,6 +761,64 @@ bootutil_img_validate(struct boot_loader_state *state,
...
+            if (len != sizeof(img_uuid_vid)) {
+                /* Vendor UUID is not valid. */
+                rc = -1;
+                goto out;
+            }
+
+            rc = LOAD_IMAGE_DATA(hdr, fap, off, img_uuid_vid.raw, len);
+            if (rc) {
+                goto out;
+            }
```
de-nordic, line 501: If it was decided to wrap uuid in fih, these should probably also report fih failure; i do not understand a difference between a failure in uuid check and inability to perform the check. And I know that we are inconsistent here, and maybe that is a good point to discuss it.

boot/bootutil/include/bootutil/image.h
```diff
@@ -142,6 +142,8 @@ extern "C" {
                                              * ...
                                              * 0xffa0 - 0xfffe
                                              */
+#define IMAGE_TLV_UUID_VID          0x80    /* Vendor unique identifier */
```
de-nordic: These go above comment explaining vendor specific ids. If these are vendor specific ids, then they have in-proper codes.

de-nordic: And should't really be here.

boot/zephyr/mcuboot_uuid.c
```diff
@@ -0,0 +1,73 @@
...
+        if (image_id == index) { \
+                *uuid_cid = &label; \
+                FIH_RET(FIH_SUCCESS); \
+        }
+#define CID_CHECK(index, prefix) \
+        IF_ENABLED(CID_CONFIG(index), (CID_CONDITION(index, prefix##index)))
+
+static fih_ret boot_uuid_compare(const struct image_uuid *uuid1, const struct image_uuid *uuid2)
+{
+        return fih_ret_encode_zero_equality(memcmp(uuid1->raw, uuid2->raw,
```
de-nordic: Are we sure we want to get fih heavy on uuid comparisons? What is the attack vector here? As far as i cen tell they are already protected tlvs and, as such, are signed with image, and why would you bypass uid check? You would have to bypass signature first.

tomchy: The attack vector would be to install the FW not designed for a particular core as a result of glitching the UUID checks (skipping them or altering comparison results).

Honestly, I do not know where to draw a line between checks that should use FIH and checks that should not. The code was based on the way the security counter is implemented, thus by default it uses FIH routines.

boot/bootutil/include/bootutil/image.h
tomchy: What do you mean by "comment explaining vendor specific ids"?

boot/bootutil/src/image_validate.c
tomchy: There is a translation logic, based on the rc value:
```
   if (rc) {
        FIH_SET(fih_rc, FIH_FAILURE);
```
and the default value of `uuid_vid_valid` and `uuid_cid_valid` will either way mask the other failure reasons.

The difference is just that - inability (if possible) can report "incompatible image format" and the uuid check can report "invalid UUID value".
There is though no way to distinguish between those error reasons as there is just a single `FIH_FAILURE` error code.

boot/bootutil/include/bootutil/image.h
nvlsianpu: @tomchy move this just after `#define IMAGE_TLV_COMP_DEC_SIZE     0x73    /* Compressed decrypted image size */` line.
@de-nordic Clarification: these are new MCUboot generic TLVs

boot/bootutil/include/bootutil/mcuboot_uuid.h
```diff
@@ -0,0 +1,63 @@
...
+ *  Copyright (c) 2025 Nordic Semiconductor ASA
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __MCUBOOT_UUID_H__
+#define __MCUBOOT_UUID_H__
+
+/**
+ * @file uuid.h
```
nvlsianpu: @file mcuboot_uuid.c

boot/bootutil/src/image_validate.c
nvlsianpu: out: seams can't mock success - it only could mock failure as a glith.

boot/zephyr/Kconfig
```diff
@@ -1071,6 +1071,44 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_LIMITED
...
+	default ""
+	depends on MCUBOOT_UUID_VID || MCUBOOT_UUID_CID
+	help
+	  The vendor unique identifier.
+	  The following formats are supported:
+	   - Domain name (i.e. amce.corp) used to generate RFC 9562 UUID5
+	     identifier.
+	   - Raw UUID (i.e. 12345678-1234-5678-1234-567812345678)
+
+if MCUBOOT_UUID_CID
```
nvlsianpu: can we have as many entries as images number allowed for given mcuboot assembly?

tomchy: Added dependency on `UPDATEABLE_IMAGE_NUMBER` Kconfig value.

nvlsianpu (approved):

boot/bootutil/include/bootutil/image.h
```diff
@@ -132,6 +132,8 @@ extern "C" {
                                              * signature
                                              */
 #define IMAGE_TLV_COMP_DEC_SIZE     0x73    /* Compressed decrypted image size */
+#define IMAGE_TLV_UUID_VID          0x80    /* Vendor unique identifier */
+#define IMAGE_TLV_UUID_CID          0x81    /* Device class unique identifier */
```
nordicjm: let's use IDs in a group above rather than a new set, we are unfortunately running out of ID groups, use the 0x7x

nordicjm (changes requested):
very quick comment, need to review

scripts/imgtool/image.py
```diff
@@ -254,6 +258,23 @@ def tlv_matches_key_type(tlv_type, key):
 
     return False
 
+def parse_uuid(namespace, value):
+    # Check if UUID is in the RAW format (12345678-1234-5678-1234-567812345678)
+    uuid_re = r'[0-9A-f]{8}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{12}'
```
nordicjm, line 263: dashes should be optional, formatting with them is just pointless overhead in e.g. a cli script

scripts/imgtool/image.py
```diff
@@ -612,6 +647,21 @@ def create(self, key, public_key_format, enckey, dependencies=None,
             if compression_tlvs is not None:
                 for tag, value in compression_tlvs.items():
                     prot_tlv.add(tag, value)
+
+            if self.vid is not None:
+                vid = parse_uuid(uuid.NAMESPACE_DNS, self.vid)
```
nordicjm, line 656: why is DNS used? What other options are there instead of DNS?

scripts/imgtool/image.py
```diff
@@ -612,6 +647,21 @@ def create(self, key, public_key_format, enckey, dependencies=None,
...
+                payload = struct.pack(e + '16s', vid)
+                prot_tlv.add('UUID_VID', payload)
+
+            if self.cid is not None:
+                if self.vid is not None:
+                    namespace = uuid.UUID(bytes=vid)
+                else:
+                    namespace = None
```
nordicjm, line 664: why does the above use DNS and this either use the above or nothing?

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
     )
 endif()
 
+# Include the sample implementation.
+if(CONFIG_MCUBOOT_UUID_VID OR CONFIG_MCUBOOT_UUID_CID)
+  zephyr_library_sources(
+    mcuboot_uuid.c
+    )
+endif()
```
nordicjm: sample? If this is for a sample it should be added specifically by the sample, it should not be in the cmake file and it should not be added for all builds when these Kconfigs are enabled

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+
+# Generate VID value and raw value definition
+if(CONFIG_MCUBOOT_UUID_VID OR CONFIG_MCUBOOT_UUID_CID)
+  string(REGEX MATCHALL "^([0-9A-F][0-9A-F]|\-)+$" match_parts ${CONFIG_MCUBOOT_UUID_VID_VALUE})
```
nordicjm: this runs if either is set, then assumes VID is the one that is set? Also make - optional

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+  if ("${match_parts}" STREQUAL "${CONFIG_MCUBOOT_UUID_VID_VALUE}")
```
nordicjm: no space between if and brackets in cmake

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+    set(UUID_VID ${match_parts})
+  else()
+    set(UUID_DNS_NAMESPACE 6ba7b810-9dad-11d1-80b4-00c04fd430c8)
```
nordicjm: what is this value?

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+  endif()
+
+  # Convert UUID into C array.
+  string(REGEX REPLACE "([0-9A-F][0-9A-F])\-?" "0x\\1, " UUID_VID_RAW ${UUID_VID})
+  add_compile_definitions(MCUBOOT_UUID_VID_VALUE=${UUID_VID_RAW})
+endif()
+
+# Generate VID value(s) and raw value definition(s)
+if(CONFIG_MCUBOOT_UUID_CID)
+  set(MCUBOOT_IMAGES_COUNT 4)
```
nordicjm: why is this using 4 images?

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+  foreach(image_id RANGE ${MCUBOOT_IMAGES_COUNT})
+    if(CONFIG_MCUBOOT_UUID_CID_IMAGE_${image_id})
+      # Check if RAW UUID format is used
+      string(REGEX MATCHALL "^([0-9A-F][0-9A-F]|\-)+$" match_parts ${CONFIG_MCUBOOT_UUID_CID_IMAGE_${image_id}_VALUE})
```
nordicjm: - to be optional

boot/zephyr/CMakeLists.txt
```diff
@@ -129,6 +129,59 @@ if(DEFINED CONFIG_MEASURED_BOOT OR DEFINED CONFIG_BOOT_SHARE_DATA)
...
+      if ("${match_parts}" STREQUAL "${CONFIG_MCUBOOT_UUID_CID_IMAGE_${image_id}_VALUE}")
```
nordicjm: as above

boot/zephyr/Kconfig
```diff
@@ -1071,6 +1071,44 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_LIMITED
...
+config MCUBOOT_UUID_VID_VALUE
+	string "Vendor name"
+	default ""
+	depends on MCUBOOT_UUID_VID || MCUBOOT_UUID_CID
+	help
+	  The vendor unique identifier.
+	  The following formats are supported:
+	   - Domain name (i.e. amce.corp) used to generate RFC 9562 UUID5
+	     identifier.
+	   - Raw UUID (i.e. 12345678-1234-5678-1234-567812345678)
```
nordicjm: also accept without dashes

boot/zephyr/mcuboot_uuid.c
```diff
@@ -0,0 +1,73 @@
+/*
+ *  Copyright (c) 2025 Nordic Semiconductor ASA
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+#include <bootutil/mcuboot_uuid.h>
+
+#define IMAGE_ID_COUNT 5
```
nordicjm: as above, if this is a sample it should be moved to a sample

scripts/imgtool/image.py
tomchy: Sure, I can add the RAW HEX option to the list.

scripts/imgtool/image.py
tomchy: Because it is the simplest thing with a registry to use to specify the vendor.
The python package includes the following namespaces:
 - DNS: When this namespace is specified, the name string is a fully qualified domain name.
 - URL: When this namespace is specified, the name string is a URL.
 - OID: When this namespace is specified, the name string is an ISO OID.
 - X500: When this namespace is specified, the name string is an X.500 DN in DER or a text output format.

It also aligns with the values used in SUIT:

> The RECOMMENDED method to create a vendor ID is:
> The "IANA UUID Namespace ID for DNS" is:
> 6ba7b810-9dad-11d1-80b4-00c04fd430c8
> Vendor ID = UUID5(<IANA UUID Namespace ID DNS>, vendor domain name)

scripts/imgtool/image.py
tomchy: Because it is easy to provide unique VIDs, based on DNS.
It is not easy to provide unique arbitrary CIDs, thus it is safer to use unique VID value as the namespace to calculate CID, thus providing unique, vendor-specific CIDs.

tomchy: The `None` option is valid if and only if the user provides RAW CID UUID value.

tomchy: It also aligns with the recommendations in SUIT specification:
> If the Vendor ID is a UUID, the RECOMMENDED method to create a Class ID is:
> Class ID = UUID5(Vendor ID, Class-Specific-Information)

boot/zephyr/CMakeLists.txt
tomchy: This is the IANA UUID Namespace ID for DNS.
Unfortunately I have not found a way to retrieve it from the CMake library, thus it is defined here.

boot/zephyr/CMakeLists.txt
tomchy: After second though - you're right: I cannot guarantee that this implementation will be sufficient for all use cases, thus it should not be placed inside MCUboot repository.

boot/zephyr/CMakeLists.txt
tomchy: Correct, the `MCUBOOT_UUID_VID_VALUE` symbol is available if either `CONFIG_MCUBOOT_UUID_VID` or `CONFIG_MCUBOOT_UUID_CID` is enabled.
The use case is pretty simple: you want a simplified check (only CID), but still want to generate UUID values, based on the VID namespace.

boot/zephyr/CMakeLists.txt
tomchy: I'll try to use `UPDATEABLE_IMAGE_NUMBER` instead.

boot/zephyr/mcuboot_uuid.c
tomchy: I'll move it outside of the MCUboot repo too.

de-nordic (approved):

PR #2414, opened 2025-08-06
bootutil: Fix PureEdDSA when flash base is not 0x0

nordic-mik7:
This commit introduces fix for PureEdDSA signature verification when the flash base address is not 0x0. The issue was that the flash base address was not taken into account when passing the image address to the signature verification function.

Additionally, FLASH_DEVICE_ID for nRF54H20 platform was corrected.


de-nordic:
> > No core and Zephyr specific change in one commit. Bootutil commits, docs and zephyr get different commits.
> 
> I understand that about Zephyr specific changes and will provide separate commit, but what about docs? Shouldn't release notes update be in the same commit as bugfix in bootutil ?

You can make them in one. Up to you.

nvlsianpu (approved):

PR #2426, opened 2025-08-12
Update FIH tests to use recent TFM and bump toolchain to 14.2 to ensure compatibility with the TFM version

maulik-arm:
- Update FIH tests to reference and work with the recent Trusted Firmware-M (TFM) version.
- Update Docker image to use toolchain version 14.2 for compatibility.
- Fix compatibility issues encountered with new TFM/toolchain.

### DIFF
--- a/ci/fih_test_docker/execute_test.sh
+++ b/ci/fih_test_docker/execute_test.sh
@@ -16,12 +16,62 @@
 
 set -e
 
+# Function to update/install native GCC inside the Docker container
+update_native_gcc() {
+    REQUIRED_MAJOR=12
+    INSTALLED_MAJOR=$(gcc -dumpversion | cut -d. -f1 || echo 0)
+
+    if [[ "$INSTALLED_MAJOR" -lt "$REQUIRED_MAJOR" ]]; then
+        echo "Installing native GCC $REQUIRED_MAJOR..."
+        apt-get update
+        apt-get install -y --no-install-recommends gcc-$REQUIRED_MAJOR g++-$REQUIRED_MAJOR \
+            cpp-$REQUIRED_MAJOR libgcc-$REQUIRED_MAJOR-dev libstdc++-$REQUIRED_MAJOR-dev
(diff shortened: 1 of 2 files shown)


ci/fih_test_docker/execute_test.sh
```diff
@@ -16,13 +16,63 @@
...
+        update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-$REQUIRED_MAJOR 60
+        rm -rf /var/lib/apt/lists/*
+    else
+        echo "Native GCC is already version $INSTALLED_MAJOR; skipping installation."
+    fi
+}
+
+# Function to update/install ARM Embedded GCC inside the Docker container
+update_cross_gcc() {
+    ARM_GCC_URL="https://developer.arm.com/-/media/Files/downloads/gnu/14.2.rel1/binrel/arm-gnu-toolchain-14.2.rel1-x86_64-arm-none-eabi.tar.xz"
```
nordicjm, line 39: why can this package not be got from the distro itself? How long will this file be available? How often can it be accessed per IP? Is it rate limited? I don't think we should be putting things like this in given that there's so many unknown variables and given that we've seen countless problems in the past trying to get even the tfm repo for CI checks

maulik-arm: You are right, the package has to be from the distro (which means I have to update the distro). But how do I do that? Did you missed the discord conversation where I raised the same question? https://discord.com/channels/1106321706588577904/1390625171785646151.  let me know how to update the distro and I will do the rest.

nordicjm: In that from @d3zd3z :

> I think the correct solution here is to figure out how to get the mcuboot docker images to be properly owned by trustedfirmware.org. I don't know if that moves us into a scenario where they are going to charge for this, though.

I see no comment suggesting that packages should be manually added like this, action is on @d3zd3z to deal with docket image ownership

PR #2427, opened 2025-08-12
zephyr: use REQUIRED in dt_nodelabel calls to fail early and with meaningful message

benedekkupper:
When the devicetree was missing a label, I'd get this kind of errors, which were not very helpful:
CMake Error at workspace/zephyr/cmake/modules/extensions.cmake:3979 (message):
  dt_prop(erase_size_slot1 ...) missing required argument: PATH
Call Stack (most recent call first):
  CMakeLists.txt:496 (dt_prop)

The new error message looks like this:
CMake Error at workspace/zephyr/cmake/modules/extensions.cmake:3748 (message):
  required nodelabel not found: slot1_partition


benedekkupper:
Sorry, I forgot the sign-off in the commit message.

de-nordic (approved):

PR #2436, opened 2025-08-27
Fixed str and int concat in key loading exceptions

fmasar:
There is string and int concatenation in handling unsupported key types which is leading to incorrect application crash.

### DIFF
--- a/scripts/imgtool/keys/__init__.py
+++ b/scripts/imgtool/keys/__init__.py
@@ -69,17 +69,17 @@ def load(path, passwd=None):
 
     if isinstance(pk, RSAPrivateKey):
         if pk.key_size not in RSA_KEY_SIZES:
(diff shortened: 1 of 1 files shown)


scripts/imgtool/keys/__init__.py
```diff
@@ -69,17 +69,17 @@ def load(path, passwd=None):
 
     if isinstance(pk, RSAPrivateKey):
         if pk.key_size not in RSA_KEY_SIZES:
-            raise Exception("Unsupported RSA key size: " + pk.key_size)
+            raise Exception("Unsupported RSA key size: " + str(pk.key_size))
```
fundakol, line 72: Should be done in more pythoning way with fstring formatting like:
```python
raise Exception(f"Unsupported RSA key size: {pk.key_size}")
```

nordicjm:
@fmasar please fix issues

PR #2439, opened 2025-09-03
imgtool: Add sha512 to allowed configurations for ECDSA256P1

fundakol:
Added SHA512 to the list of supported algorithms for ECDSA256P1.


nordicjm:
FYI @joerchan

de-nordic (changes requested):
change title from "bootutil: " to "imgtool: "

PR #2440, opened 2025-09-03
imgtool: max alignment to 128 bytes for Renesas

vjardin:
write-block-size is set to 128 bytes for the ra4, ra6, ra8's flash0 dtsi.

Fix error:
  Error: Invalid value for '--align': '128' is not one of '1', '2', '4', '8', '16', '32'.


vjardin:
Done: fixed author email to be aligned with the sign off.

vjardin:
Thanks @nordicjm for your analysis, your are right that I did miss it. I am still novice with it. Does it mean that there is no mean to use such method ?

If so, it means that none of the Renesas MCU can be supported since most are based on 128 bytes from their dt.

vjardin:
ok, closed.

PR #2446, opened 2025-09-08
zephyr: Add basic UUID checks implementation

tomchy:
Provide an implementation for MCUboot UUID checks that specify a single, common vendor identifier and a unique class identifier for each image.


boot/zephyr/uuid/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+  if(CONFIG_MCUBOOT_UUID_VID OR CONFIG_MCUBOOT_UUID_CID)
+    zephyr_library_sources(
+      uuid.c
+      )
+  endif()
+
+  # Generate VID value and raw value definition
+  if(CONFIG_MCUBOOT_UUID_VID OR CONFIG_MCUBOOT_UUID_CID)
+    if("${CONFIG_NCS_MCUBOOT_UUID_VID_VALUE}" STREQUAL "" AND CONFIG_MCUBOOT_UUID_VID)
+        message(FATAL_ERROR "VID value not set")
```
nordicjm: 2 space indent for cmake, also do not throw fatal error here (fine if the code itself does not compile) so that users can fix the issue without the whole cmake build being broken

boot/zephyr/uuid/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+
+    string(REGEX MATCHALL "^([0-9A-F][0-9A-F]|\-)+$" match_parts "${CONFIG_NCS_MCUBOOT_UUID_VID_VALUE}")
+    if("${match_parts}" STREQUAL "${CONFIG_NCS_MCUBOOT_UUID_VID_VALUE}")
+      set(UUID_VID ${match_parts})
+    else()
+      set(UUID_DNS_NAMESPACE 6ba7b810-9dad-11d1-80b4-00c04fd430c8)
+      string(
+        UUID UUID_VID
+        NAMESPACE ${UUID_DNS_NAMESPACE}
+        NAME ${CONFIG_NCS_MCUBOOT_UUID_VID_VALUE}
```
nordicjm: lots of NCS prefixes in here

boot/zephyr/uuid/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+    endif()
+  endif()
+
+  # Generate VID value(s) and raw value definition(s)
+  if(CONFIG_MCUBOOT_UUID_CID)
+    set(MCUBOOT_IMAGES_COUNT ${CONFIG_UPDATEABLE_IMAGE_NUMBER})
+    foreach(image_id RANGE ${MCUBOOT_IMAGES_COUNT})
+      if(CONFIG_NCS_MCUBOOT_UUID_CID_IMAGE_${image_id})
+        if("${CONFIG_NCS_MCUBOOT_UUID_CID_IMAGE_${image_id}_VALUE}" STREQUAL "")
+          message(FATAL_ERROR "CID value not set for image ${image_id}")
```
nordicjm: as above, `WARNING` then can fail in a .c file with an assert but not in cmake

boot/zephyr/uuid/Kconfig
```diff
@@ -0,0 +1,55 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+if MCUBOOT && (MCUBOOT_UUID_VID || MCUBOOT_UUID_CID)
```
nordicjm: MCUBOOT check is redundant

boot/zephyr/uuid/Kconfig
```diff
@@ -0,0 +1,55 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
```
nordicjm: apache license only

boot/zephyr/uuid/uuid.c
```diff
@@ -0,0 +1,74 @@
...
+
+#define CID_CONDITION(index, label) \
+	if (image_id == index) { \
+		*uuid_cid = &label; \
+		FIH_RET(FIH_SUCCESS); \
+	}
+#define CID_CHECK(index, prefix) \
+	IF_ENABLED(CID_CONFIG(index), (CID_CONDITION(index, prefix##index)))
+
+static fih_ret boot_uuid_compare(const struct image_uuid *uuid1, const struct image_uuid *uuid2)
```
nordicjm, line 26: not really inkeeping in the style of MCUboot - no other OS can use this, should be a base file in a common folder then the parts that zephyr needs in the zephyr specific file that gets included

tomchy: This implementation is based on the existence of Kconfigs, so I'd say that it is Zephyr-specific and does not present any value to other platforms.

PR #2449, opened 2025-09-10
ci: Use fih-test container from GHCR

davidvincze:
The `fih-test` Docker image (v0.0.3) was copied from Docker Hub to Github Container Registry under the `mcu-tools` to provide equal access to the images within the organization.

In case of acceptance: fih-test image may be removed from Docker Hub.


davidvincze:
Currently the [mcuboot-fih-test](https://github.com/orgs/mcu-tools/packages/container/package/mcuboot-fih-test) image is private as I didn't see any benefits in publishing it, although it wouldn't case any problems either.

ci/fih-tests_info.sh
```diff
@@ -0,0 +1,7 @@
+FIH_IMAGE_VERSION=0.0.3
```
de-nordic: New file missing license.

davidvincze: I missed it because I've just basically renamed the original fih-tests_version.sh file. I'm going to update it. Thanks!

PR #2451, opened 2025-09-12
boot: zephyr: Cleanup RAM by default

MarekPieta:
Enable the CONFIG_MCUBOOT_CLEANUP_RAM by deafult to improve security of the default configuration.


de-nordic (approved):

nordicjm (changes requested):
This causes https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/boot/mcuboot_data_sharing to continually bootloop, it also adds an unknown delay to booting applications where it's not needed (since the c start function in the application will do again when it starts)

PR #2453, opened 2025-09-15
boot: zephyr: config: Allow custom MCUBOOT_BOOT_MAX_ALIGN

pdgendt:
Allow setting CONFIG_MCUBOOT_BOOT_MAX_ALIGN to apply a custom programmable block in flash.

### DIFF
--- a/boot/zephyr/include/mcuboot_config/mcuboot_config.h
+++ b/boot/zephyr/include/mcuboot_config/mcuboot_config.h
@@ -430,7 +430,9 @@
 #endif
 
 /* Support 32-byte aligned flash sizes */
-#if DT_HAS_CHOSEN(zephyr_flash)
+#if defined(CONFIG_MCUBOOT_BOOT_MAX_ALIGN) && CONFIG_MCUBOOT_BOOT_MAX_ALIGN > 0
+#define MCUBOOT_BOOT_MAX_ALIGN CONFIG_MCUBOOT_BOOT_MAX_ALIGN
+#elif DT_HAS_CHOSEN(zephyr_flash)
     #if DT_PROP_OR(DT_CHOSEN(zephyr_flash), write_block_size, 0) > 8
         #define MCUBOOT_BOOT_MAX_ALIGN \
             DT_PROP(DT_CHOSEN(zephyr_flash), write_block_size)


boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -430,7 +430,9 @@
 #endif
 
 /* Support 32-byte aligned flash sizes */
-#if DT_HAS_CHOSEN(zephyr_flash)
+#if defined(CONFIG_MCUBOOT_BOOT_MAX_ALIGN) && CONFIG_MCUBOOT_BOOT_MAX_ALIGN > 0
+#define MCUBOOT_BOOT_MAX_ALIGN CONFIG_MCUBOOT_BOOT_MAX_ALIGN
+#elif DT_HAS_CHOSEN(zephyr_flash)
```
de-nordic: We have these dt_ macros to use with Kconfig, can we maybe move the entire logic to Kconfig?

pdgendt: Probably, yes but this fix should make that backwards compatible, no?

de-nordic: Yeah. So you set the Kconfig to defaults as the logic states here, but people can still override it. Wouldn't that make it backwards compatible?

de-nordic: This way we also have only one place where it is set: the Kconfig; there would no "if Kconfig is 0 then", because it could only be 0 if we allow somebody to set it as such in Kconfig.

de-nordic: Am I missing something here?

pdgendt: I've updated the PR per your suggestion. A single Kconfig symbol, which default to the chosen flash's write-block-size.

boot/zephyr/include/mcuboot_config/mcuboot_config.h
```diff
@@ -430,11 +430,8 @@
 #endif
 
 /* Support 32-byte aligned flash sizes */
-#if DT_HAS_CHOSEN(zephyr_flash)
-    #if DT_PROP_OR(DT_CHOSEN(zephyr_flash), write_block_size, 0) > 8
-        #define MCUBOOT_BOOT_MAX_ALIGN \
-            DT_PROP(DT_CHOSEN(zephyr_flash), write_block_size)
-    #endif
+#if defined(CONFIG_MCUBOOT_BOOT_MAX_ALIGN) && CONFIG_MCUBOOT_BOOT_MAX_ALIGN > 8
```
nordicjm:
```suggestion
#if CONFIG_MCUBOOT_BOOT_MAX_ALIGN > 8
```

pdgendt: Applied, thanks

nordicjm (approved):

boot/zephyr/include/mcuboot_config/mcuboot_config.h
pdgendt: As a side-note, we're backwards compatible like this indeed, but not forwards compatible (which was te case before my change).
So if this change is pulled in without https://github.com/zephyrproject-rtos/zephyr/pull/96001 the behavior changes.

nordicjm: That's fine, a specific commit of MCUboot is valid with a specific commit of zephyr, as the west.yml file points to

de-nordic:
@pdgendt We will need release notes entry for this change.

PR #2461, opened 2025-09-26
boot: bootutil: write_sz fix

michalek-no:
fixes missing array bug.

### DIFF
--- a/boot/bootutil/src/bootutil_priv.h
+++ b/boot/bootutil/src/bootutil_priv.h
@@ -253,7 +253,7 @@ struct boot_loader_state {
 #endif
 
     uint8_t swap_type[BOOT_IMAGE_NUMBER];
-    uint32_t write_sz;
+    uint32_t write_sz[BOOT_IMAGE_NUMBER];
 
 #if defined(MCUBOOT_SWAP_USING_OFFSET)
     uint32_t secondary_offset[BOOT_IMAGE_NUMBER];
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/bootutil_priv.h
```diff
@@ -253,7 +253,7 @@ struct boot_loader_state {
 #endif
 
     uint8_t swap_type[BOOT_IMAGE_NUMBER];
-    uint32_t write_sz;
+    uint32_t write_sz[BOOT_IMAGE_NUMBER];
```
de-nordic, line 256: I am not sure this is correct. The write size does not depend on the image, because you can have image slots on various images. The code that sets this still relies on the primary slot, so if you get image that has secondary slot with write block  size 16 and primary with 4, the value here will not be correct at all.

nvlsianpu: honestly - I don't know. But indeed it is line for assign that value; BOOT_WRITE_SZ(state) = boot_write_sz(state); in boot_read_sectors() function.

ArekBalysNordic: Maybe this should rely on something else than the image, but the issue was found in the following case:
- Primary slot for image 0 is on internal flash -> write_sz=4
- Primary slot for image 1 is on internal flash -> write_sz=4
- Primary slot for image 2 is on external flash through SPI -> write_sz=1

 boot_read_sectors()  sets write_sz according to the last known primary slot/image, so sets 1. As a result, NVM save fails during copying image 0, because the data chunk is aligned to 1, not to 4.

michalek-no:
rebase

nvlsianpu (approved):

boot/bootutil/src/bootutil_priv.h
de-nordic: We have several options here:
 - set it, for every image, as the highest value among slots - we would have to have slots of the same size, but there wasn't much to save anyway;
 - set that statically for all images, via Kconfig, to highest value - this would require verification run to confirm that the selected can be divided by write size of every slot; this would be bad for setups that have a single device with huge write size, like 128b. This one, except for verification run, would probably give us smaller code footprint, as you do not have to have the size discovery code.

de-nordic (approved):

PR #2464, opened 2025-09-27
imgtool: support producing images in test mode

danieldegrasse:
Add --test flag, which allows users to append a trailer that marks the image as ready for a test swap. This can be used for cases where the user wants to load an image to flash that MCUBoot will boot in test mode after system reset.


danieldegrasse:
> Rebase with `main` and CI should be fixed

Thanks, done

PR #2465, opened 2025-09-29
Feature: images in image support

ozersa:
This PR introduce an proposal method that can be used to handle multi image in a single image pack to decrease number of signature check during bootup. By this change multiple images can be loaded/updated on device with one signature check.
Depend on the product this feature might requires to decrease boot up time.

This PR includes:
- **combine_images.py** which is used to combine images and generate mcuboot image structure by using imgtool.
- **combine_images.yaml** is provide each images parameters. It is a reference file to demonstrate how can it be used.

The feature only support for if (BOOT_IMAGE_NUMBER == 1) and  MCUBOOT_RAM_LOAD defined case.
And assume on multicore heterogeneous system the host core which execute MCUBoot FW able to access sub core ITCM/DTCM by just memcpy operation.
The MCUBoot FW after copy them to the target address is going to execute first bootable image.

As an example 3 images can be concatenated as
image

Simply tested with [Zephyr](https://github.com/zephyrproject-rtos/zephyr/tree/main/boards/adi/max32690evkit) on [MAX32690EVKIT](https://www.analog.com/en/resources/evaluation-hardware-and-software/evaluation-boards-kits/max32690evkit.html#eb-overview)
image

image

Performance test output for 3 images:
As summary this method clearly decrease boot time. Please see:
image


nordicjm (changes requested):
This is really not at all how MCUboot is designed to operate, if you want this then you should use the hooks MCUboot has to run your functions for doing this in your own application or module

ozersa:
> This is really not at all how MCUboot is designed to operate, if you want this then you should use the hooks MCUboot has to run your functions for doing this in your own application or module

@nordicjm thanks for the feedback. While adding this feature we have not change behavior of mcuboot. It is just extended.
Isn't it be useful for anyone else? Especially for multi core heterogenous systems.
It helps to to decrease boot time & simplify fw update due to there is one main package that need to be managed.

nordicjm:
> > This is really not at all how MCUboot is designed to operate, if you want this then you should use the hooks MCUboot has to run your functions for doing this in your own application or module
> 
> @nordicjm thanks for the feedback. While adding this feature we have not change behavior of mcuboot. It is just extended. Isn't it be useful for anyone else? Especially for multi core heterogenous systems. It helps to to decrease boot time & simplify fw update due to there is one main package that need to be managed.

You have taken a mode and completely made up an entirely new container format without any discussion or thought about other MCUboot modes, security, previous version compatibility or design queries or how other supported OS's would implement such a feature notwithstanding a complete lack of documentation. Not to say that such a feature cannot be added, but this is really not at all the correct way to do it

boot/bootutil/src/loader.c
```diff
@@ -3083,6 +3208,13 @@ boot_go(struct boot_rsp *rsp)
     boot_state_clear(NULL);
 
     FIH_CALL(context_boot_go, fih_rc, &boot_data, rsp);
+
```
d3zd3z: This functionality should be integrated into `context_boot_go`. `boot_go` is a wrapper for on-target vs simulated usage, and any functionality here would be skipped in the simulator.

However, I don't think we should have the existing ram load load this into ram first, and then have process_sub_images split that out, but we should enhance the ram loading process to load the subimages if that is defined and used.

d3zd3z: This functionality should be integrated into `context_boot_go`. `boot_go` is a wrapper for on-target vs simulated usage, and any functionality here would be skipped in the simulator.

However, I don't think we should have the existing ram load load this into ram first, and then have process_sub_images split that out, but we should enhance the ram loading process to load the subimages if that is defined and used.

boot/bootutil/src/loader.c
```diff
@@ -3067,6 +3068,130 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 }
 #endif /* MCUBOOT_DIRECT_XIP || MCUBOOT_RAM_LOAD */
 
+
+#if (BOOT_IMAGE_NUMBER == 1) && defined(MCUBOOT_RAM_LOAD)
```
d3zd3z, line 2995: This should get it's own feature defined (maybe `MCUBOOT_SUBIMAGES`) instead of just assuming that if we have a single image that is ram loaded subimages would be wanted.

In addition, we should define a new flag in the header to indicate when subimages are present in the image itself. Functionality wise:

- If `MCUBOOT_SUBIMAGES` it not defined, but the flag is present, fail.
- if `MCUBOOT_SUBIMAGES` is defined, the flag will indicate whether a single image is loaded, or multiple images.

In other words, separate the implementation of the feature from the flags indicate whether it is present.

d3zd3z (changes requested):
In addition to the inline comments, before this can be merged, we will need to add simulator support to test this. The added test doesn't need to test upgrade robustness, as that should be handled the same as a regular single image case. But, it should test the subimages loading (providing buffers), and either be an extension of the existing ram loading tests, or an additional test.

In addition, please enhance the documentation to cover this feature. Likely add a mention in design.md that references a new documentation file describing how subimages work.

Beyond this. I do think this makes sense to include in mcuboot in a general sense, as long as the issues here are addressed. Going forward, I forsee ramloading becoming more common, and this addresses a few concerns with the current multiimage handling:
- The not-quite-working mutual dependency issues won't be a concern as this subimage always updates all images together.
- Future support for quantum cryptography, where signatures are often quite large, will benefit from only having a single signature.

boot/bootutil/src/loader.c
ozersa: >This functionality should be integrated into context_boot_go. boot_go is a wrapper for on-target vs simulated usage, and any functionality here would be skipped in the simulator.

There are two context_boot_go ( [1](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L2228) , [2](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L3020) ) function, the feature moved under the one that execute for (MCUBOOT_DIRECT_XIP || MCUBOOT_RAM_LOAD).  The executed rust tests as it is seems works, not applied any change yet.

image


> However, I don't think we should have the existing ram load load this into ram first, and then have process_sub_images split that out, but we should enhance the ram loading process to load the subimages if that is defined and used.

For RAM mode MCUboot first [load image into RAM](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L2808) than [validate it](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L2818C22-L2818C40) in RAM.
By the feature that propose in this PR main image (Image0) need to be validated first  before processing sub images then indirectly sub-image can be moved to target address. So that It seems not applicable it be moved under [ram loading function](https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/src/loader.c#L2808).

Please let me know if I misinterpreted the existing code.

boot/bootutil/src/loader.c
ozersa: Thanks for feedback, checked it, as per of my check it seems new flags may not be fit for this feature due to for this feature MCUBOOT_IMAGE_NUMBER should be 1, unless you do not want it supports "multi image X multi-subimages".
And for Zephyr port, MCUBOOT_IMAGE_NUMBER set by [zephyr Kconfig parameter](https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/include/mcuboot_config/mcuboot_config.h#L118), if we add MCUBOOT_SUBIMAGES that need to be set by each port of MCUboot.

Additionally if we define a new flag in mcuboot header, I am not clear how it should be set, please see existing options: https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/image.py#L804  The header flag set indirectly depend on some parameters specified.

So I think "#if (BOOT_IMAGE_NUMBER == 1) && defined(MCUBOOT_RAM_LOAD)" might be better,
Please let me know you would like to new flag be defined?

ozersa:
> My original NACK on this stands, fyi.

Thanks for comment, is there a way to address your concern, might be refactor...?

nordicjm:
Here are some links with things for you to study:

- https://docs.mcuboot.com/design.html#image-format (the whole document, and other pages on the doc site)
- https://docs.mcuboot.com/encrypted_images.html
- https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L167
- https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/dfu/img_util/flash_img.c#L26
- https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/dfu/boot/mcuboot.c
- https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt.c
- https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt_state.c
- https://github.com/apache/mynewt-mcumgr/tree/master/cmd/img_mgmt (the original MCUmgr code)

From there I suggest you look into things like the image structure, the fields in the image header, how image identification works, how image updates and uploads work. Then bearing in mind the following requirements:

- System should not break how existing bootloaders work or cause them to malfunction
- MCUboot header size shall not be changed
- MCUmgr support must be maintained including support for existing builds of firmware and new firmware should not break how it functions on those devices
- Information available for things like where images should be loaded must not be buried part the way through firmware
- Other features should remain fully functional and remain secure e.g. monotonic counter support, encrypted image support
- Feature should be usable (not that support needs adding for it in the addition PR, but that support _could_ be made to other RTOS's if other users wanted to) on any RTOS/OS
- Be mindful that MCUboot should be optimised for constrained devices and not use excessive flash/RAM, not that usage cannot increase but it should not increase by a large amount that prevents usage on commonly supported (and used) devices

With the above in mind, please have a think and update your RFC with your proposed solution for this which meets the above requirements and I shall review it, thanks!

d3zd3z:
We discussed this change today during the tech call. The basic conclusion we came to can be summarized as:

- There is value in multiple image support, however this implementation does not fit well with the rest of MCUboot.
- There are two approaches we can take to multiple image support:
  - Add a new signature type that can cover multiple images. This is fairly invasive, and only solves the minor problem of signature verification time.
  - Allow a richer image type that supports loading a single image into possibly multiple areas of RAM.
- The best approach, then, is likely to be adding support for richer images. However, it needs to be implemented in a way that works with MCUboot.
- There is an existing TOCTOU vulnerability in ram loading, where the signature is validated, and then in a separate pass the image is loaded into RAM. To fix this, the image validation needs to be combined with the image loading as a single pass. This will need to be fixed before this multiple image support should be added.
- Then with this new loader, it can be enhanced to support these multiple images. To do this, the file format of the extended image should be clearly defined and documented, and discussed, perhaps in some kind of RFC change that just adds the documentation. We will need some time to review this and consider issues.
- When using multiple images, there is no reason to have headers on the individual images, they can just be the start of the OS image. This actually applies even to single images, so a "multiple-image" that has a single image in it makes sense, and allows MCUboot, in configurations that ram load, to not need _any_ modification in the target OS to support or work around the header.
- The implementation should then clearly be defined in two ways:
  - There should be a change to the regular image header, perhaps a flag bit, to indicate that the image is a "rich image" or something similar. This has a defined format, either with a single header at the start, or more likely, individual headers before each segment.
  - This needs to be done in such a way that the loader can process both the signature as it loads, as well as placing them in the correct destinations within RAM. There can be a feature in MCUboot to enable this rich image support, but it would only be for code-size, as the load would need to continue to support the current plain images as well.
- As a simple workaround, the authors of this change could consider disabling image verification on each boot. The image signatures would still be verified when doing upgrades, but not on every boot. This still has some security, as presumably the internal flash is reasonably protected (and if it isn't mcuboot itself could be modified to bypass security).
- In addition, each change/pr that implements this needs to have support added to the simulator to both verify that the existing use cases are not broken, and to verify that the rich images are being handled properly.

PR #2468, opened 2025-09-29
boot: zephyr: Fix IO-based entrance method

tomchy:
If a device uses RESETINFO, than there are some bits set in the resetinfo, even for reboots that should allow to interpret and enter the device recovery.
That way it is possible to recover a device through serial recovery if the main application resets due to i.e. watchdog.


boot/zephyr/include/io/io.h
```diff
@@ -67,6 +67,45 @@ static inline bool io_boot_skip_serial_recovery()
...
+        | NRFX_RESET_REASON_DOG_MASK
+        /* Reset from CTRL-AP. */
+        | NRFX_RESET_REASON_CTRLAP_MASK
+        /* Reset due to secure domain system reset request. */
+        | NRFX_RESET_REASON_SREQ_MASK
+        /* Reset due to secure domain watchdog 0 timer. */
+        | NRFX_RESET_REASON_SECWDT0_MASK
+        /* Reset due to secure domain watchdog 1 timer. */
+        | NRFX_RESET_REASON_SECWDT1_MASK
+        /* Reset due to secure domain lockup. */
```
nordicjm, line 92: not sure what all these are defined on but will this file compile e.g. for nrf51 and nrf52 devices which I assume don't have these?

tomchy: To my understanding - the NRF_RESETINFO is secific to nRF54H and nRF92 families.
I tried to follow the definitions from here: https://github.com/zephyrproject-rtos/hal_nordic/blob/979a58d0829108b57f10e90b6c3fc35ab6206e4b/nrfx/helpers/nrfx_reset_reason.h#L225

The main issue it that on those SoCs either `NRFX_RESET_REASON_POR_MASK` or `RESETINFO_RESETREAS_GLOBAL_RESETPOR_Msk` is set, making it impossible to pass the `0` test.

nordicjm (approved):

de-nordic (approved):

PR #2479, opened 2025-10-06
Boot: Zephyr: Boards: Fix STM32H750B-DK App in Ext Flash support

JarmouniA:
Following the introduction of stm32h750b-dk 'ext_flash_app' board variant (https://github.com/zephyrproject-rtos/zephyr/pull/97037), for storing apps in external Flash and chainloading them with MCUboot that's placed in internal Flash, we need a DT overlay to designate internal Flash & controller as the chosen 'zephyr,flash' & 'zephyr,flash-controller' for the MCUboot Zephyr Application. This way, no DT overlay will be needed at the Zephyr user app level.

Turn the board's `.conf` & `.overlay` files into a SoC ones, so they apply to all boards using STM32H750XX SoC.

`BOOT_MAX_IMG_SECTORS_AUTO=n` no longer needed after adding an external flash node to board's common `.dtsi` with 'soc-nv-flash' compatible, & declaring the 'erase-block-size' property.


nordicjm (approved):

boot/zephyr/socs/stm32h750xx_ext_flash_app.overlay
erwango, line 1: I wonder why this has to be applied to a `stm32h750b_dk/stm32h750xx/ext_flash_app` target.
My understanding of https://github.com/zephyrproject-rtos/zephyr/pull/97037, is that `stm32h750b_dk/stm32h750xx` is the target running on internal flash.

This way, we re-use `zephyr,flash` and `zephyr,flash-controller`, which are already defined for `stm32h750b_dk.dts`. And since `zephyr,code-partition = &boot_partition;` is already defined in app.overlay we don't have any dependency on MCUBoot anymore

JarmouniA: The memory layout for running a user app from external Flash is in ext_flash_app variant dts, and we use stm32h750b_dk/stm32h750xx/ext_flash_app target when building with sysbuild.

PR #2487, opened 2025-10-09
loader: Brute force solution for dependencies

tomchy:
Move to a brute force solution for image dependency resolution in case of builds, that do not always prefer image with the highest version number.


tomchy:
Decided to fix the issue by removing the slotted dependency feature.

PR #2490, opened 2025-10-09
boot: zephyr: fix getting base address for certain boards

AndreHeinemans-NXP:
This fixes the build for boards who have have a flexspi controller but having different controller assigned to zephyr,flash-controller in the DTS.

This solves this issue: zephyrproject-rtos/zephyr#97021


ZhaoxiangJin:
This PR fixed this [issue](https://github.com/zephyrproject-rtos/zephyr/issues/97021). But do we really need to obtain the base address here via DT? This base address has already been obtained [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/nxp/common/Kconfig.flexspi_xip#L14) and assigned to CONFIG_FLASH_BASE_ADDRESS.

PR #2491, opened 2025-10-09
bootutil: Move boot_enc_init in boot_swap_image

de-nordic:
Encryption context intialization can happen before we read the key.


boot/bootutil/src/loader.c
```diff
@@ -1698,6 +1698,9 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
 
 #ifdef MCUBOOT_ENC_IMAGES
         for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
+
```
nordicjm, line 1701:
```suggestion
```
nit.

PR #2500, opened 2025-10-16
boot: zephyr: board: stm32h573i_dk: Add support for ext_flash_app variant

tpambor:
Following the introduction of stm32h573i-dk 'ext_flash_app' board variant, for storing apps in external flash and chainloading them with MCUboot that's placed in internal flash, we need a devicetree overlay to designate the internal flash and flash controller as the chosen 'zephyr,flash' and 'zephyr,flash-controller' for the MCUboot zephyr application.

STM32_MEMMAP is no longer needed since it is now set on the Zephyr side in the board's 'Kconfig.defconfig' when building MCUboot app. Also, BOOT_MAX_IMG_SECTORS_AUTO is no longer needed to be disabled after including the 'soc-nv-flash' binding in the 'st,stm32-xspi-nor' binding, then adding the 'erase-block-size' and 'write-block-size' properties
in external Flash DT node. Therefore, remove the board conf file.

This PR follow along the changes for STM32H750B-DK in #2479. Aim is to get rid of the overlay in the future together with the overlay for STM32H750B-DK.


erwango:
See comment here https://github.com/mcu-tools/mcuboot/pull/2479#pullrequestreview-3345576429

boot/zephyr/socs/stm32h573xx_ext_flash_app.overlay
JarmouniA, line 1: should go in `boot/zephyr/socs` folder.

boot/zephyr/socs/stm32h573xx_ext_flash_app.conf
JarmouniA, line 1: ditto

boot/zephyr/socs/stm32h573xx_ext_flash_app.overlay
tpambor: Fixed, missed that, thanks

nordicjm (approved):

PR #2501, opened 2025-10-17
bootutil: Drop slot number and boot_state from most boot_enc functions

de-nordic:
Step forward in reducing entanglement between internal APIs.


boot/bootutil/src/encrypted.c
```diff
@@ -587,13 +587,13 @@ boot_enc_load(struct boot_loader_state *state, int slot,
     BOOT_LOG_DBG("boot_enc_load: slot %d", slot);
 
     /* Already loaded... */
-    if (enc_state[slot].valid) {
+    if (boot_enc_valid(enc_state)) {
```
nordicjm, line 590: not for this PR but in future we need to do rework the code and do expression checks properly, e.g. `if (boot_enc_valid(enc_state) == true) {`

PR #2503, opened 2025-10-20
bootutil: Extract common loader functions

tomchy:
Reduce the size of loader.c as well as bootutil_misc.c files by extracting common functions into a separate files.


boot/bootutil/src/loader.c
de-nordic, line 1: Commit title: `loader: Unifa image check API`, `Unify`?

de-nordic (approved):

almir-okato (approved):

PR #2504, opened 2025-10-20
boot: zephyr: sysbuild: fix CONFIG_MCUBOOT_UPDATE_FOOTER_SIZE leak

JarmouniA:
`CONFIG_MCUBOOT_UPDATE_FOOTER_SIZE` depends on `CONFIG_MCUBOOT_IMG_MANAGER` on Zephyr side, but we try to set it here always which results in the build warning:

"warning: MCUBOOT_UPDATE_FOOTER_SIZE (defined at subsys/dfu/Kconfig:55) was assigned the value '0x30' but got the value ''. Check these unsatisfied dependencies: MCUBOOT_IMG_MANAGER (=n), IMG_MANAGER (=n)"

### DIFF
--- a/boot/zephyr/sysbuild/CMakeLists.txt
+++ b/boot/zephyr/sysbuild/CMakeLists.txt
@@ -33,10 +33,13 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_image_cmake)
       sysbuild_get(mcuboot_image_footer_size IMAGE mcuboot CACHE)
       sysbuild_get(mcuboot_image_upgrade_footer_size IMAGE mcuboot CACHE)
       math(EXPR mcuboot_image_footer_size "${mcuboot_image_footer_size}" OUTPUT_FORMAT HEXADECIMAL)
-      math(EXPR mcuboot_image_upgrade_footer_size "${mcuboot_image_upgrade_footer_size}" OUTPUT_FORMAT HEXADECIMAL)
-
+      
       set_property(TARGET ${image} APPEND_STRING PROPERTY CONFIG "CONFIG_ROM_END_OFFSET=${mcuboot_image_footer_size}\n")
-      set_property(TARGET ${image} APPEND_STRING PROPERTY CONFIG "CONFIG_MCUBOOT_UPDATE_FOOTER_SIZE=${mcuboot_image_upgrade_footer_size}\n")
+      
+      if(CONFIG_MCUBOOT_IMG_MANAGER)
+        math(EXPR mcuboot_image_upgrade_footer_size "${mcuboot_image_upgrade_footer_size}" OUTPUT_FORMAT HEXADECIMAL)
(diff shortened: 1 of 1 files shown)


boot/zephyr/sysbuild/CMakeLists.txt
```diff
@@ -33,10 +33,13 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_image_cmake)
       sysbuild_get(mcuboot_image_footer_size IMAGE mcuboot CACHE)
       sysbuild_get(mcuboot_image_upgrade_footer_size IMAGE mcuboot CACHE)
       math(EXPR mcuboot_image_footer_size "${mcuboot_image_footer_size}" OUTPUT_FORMAT HEXADECIMAL)
-      math(EXPR mcuboot_image_upgrade_footer_size "${mcuboot_image_upgrade_footer_size}" OUTPUT_FORMAT HEXADECIMAL)
-
+      
       set_property(TARGET ${image} APPEND_STRING PROPERTY CONFIG "CONFIG_ROM_END_OFFSET=${mcuboot_image_footer_size}\n")
-      set_property(TARGET ${image} APPEND_STRING PROPERTY CONFIG "CONFIG_MCUBOOT_UPDATE_FOOTER_SIZE=${mcuboot_image_upgrade_footer_size}\n")
+      
+      if(CONFIG_MCUBOOT_IMG_MANAGER)
```
nordicjm, line 39: this is not a sysbuild variable and does not exist, thus the changes here mean this if block never runs

JarmouniA: Ah right! Should we create a sysbuild one corresponding to IMG_MANAGER, or is there another sysbuild symbol that can be used to guard this?

nordicjm: there is no symbol and you cannot make one, this sets the config of the application before the application is configured (it works because system sets MCUboot to be configured first, otherwise it would fail), you cannot read any application configuration and there is nothing here that can be changed. Also note that a CMake warning is just that, it indicates something that should be read and _might_ need user action, but does not indicate an error, having them in production application builds is not an issue

JarmouniA: Would it be acceptable to remove the dependency of `MCUBOOT_UPDATE_FOOTER_SIZE` on `MCUBOOT_IMG_MANAGER`? This way we get rid of the warning with the price of a Kconfig leak.

nordicjm: not really no, zephyr only supports MCUboot but it's possible for out of tree users now or even maybe in future zephyr supports more than just MCUboot and MCUboot configurations should not leak out if it's not enabled

JarmouniA: I actually meant to say: replace the dependency on `MCUBOOT_IMG_MANAGER` with `BOOTLOADER_MCUBOOT`.

PR #2510, opened 2025-10-21
boot: zephyr: socs: stm32h7s7xx: Add support for ext_flash_app variant

erwango:
On the same example as stm23h750xx and stm32h573xx, introduce ext_flash_app variant for stm32h7s7xx in mcuboot.

This allows chainloading applications in ext flash from a mcuboot application running in internal flash.

Similarly, BOOT_MAX_IMG_SECTORS_AUTO is no longer needed to be disabled after including the 'soc-nv-flash' binding in the 'st,stm32-xspi-nor' binding, whihc provides 'erase-block-size' and 'write-block-size' to external Flash DT node.

Remove the no longer required Kconfig fragment for stm32h7s78_dk.


erwango:
@nordicjm Do you think it is possible to get that along with the board support in v4.3.0 ?

PR #2511, opened 2025-10-21
bootutil: Add manifest-based loader for Direct XIP

tomchy:
Add a loader variant that is capable of booting images in Direct XIP mode, based on a simple manifest.

* The manifest image, configured through `MCUBOOT_MANIFEST_IMAGE_INDEX` (slot pair) is the main/lead image - it describes firmware bundle for the device.
* The firmware bundle description is provided in the protected TLV of the manifest image. Images relations are expressed as their hashes. Hash of the manifest image is not in the TLV as it covers it as well. 

For more details, please take a look at updates inside the `design.md` file.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -4,6 +4,7 @@
   - Copyright (c) 2017-2020 Linaro LTD
   - Copyright (c) 2017-2019 JUUL Labs
   - Copyright (c) 2019-2024 Arm Limited
+  - Copyright (c) 2025 Nordic Semiconductor ASA
 
   - Original license:
 
@@ -986,6 +987,147 @@ strategy but there is no need for Scratch area.
     + Do the measured boot and the data sharing if needed.
 
 + Boot the loaded slot of image 0.
+
+### [Multiple image boot using manifest](#multiple-image-boot-using-manifest)
+
+
+Deployments that use multiple images typically require strict control over
+versions of firmware components.
+There is a dependency TLV that can be used to specify dependencies between
+semantic versions of multiple components; however, since these only describe
+a minimally compatible version of a counterpart component, there is no mechanism
+to enforce a specific revision of the other image;
+therefore, the publisher must ensure that all combinations that satisfy the
+dependencies are compatible with each other and are tested before deploying
+a new version of the firmware bundle.
+One way to simplify the process is to add dependencies to all parts that
+enforce the latest revision of all other parts of the bundle.
+This effectively enforces equality between the revisions of all parts.
+This solution becomes even more problematic when using Direct-XIP mode, since
+the bootloader typically only chain-loads the next stage, the next stage must
+select and boot the correct slot of the next part of the firmware.
+Although the dependencies ensure that the next part with a specific
+version is present on the device, there is no guarantee in which slot it
+was validated.
+The publisher may use different version numbers for the same firmware
+created for different slots, but this raises the question of whether there is
+a better way to manage dependencies between images.
+
+The bootloader manifest is a protected TLV that serves as the sole
+source of information about the compatibility and bootability of the multipart
+firmware. The basic rules for all types of manifests are:
+
+  * A manifest must be protected (directly or indirectly) by a cryptographic
+    signature.
+  * There must be exactly one selected image that can provide the manifest TLV.
+  * There must be only one manifest per slot.
+  * Processing of the manifest must validate the full functional set of firmware
+    components.
+  * If a manifest does not describe a part of the firmware, it must be
+    considered invalid.
+  * Each update candidate must provide a new manifest.
+
+
+```
+                                           +-------------------+
+                        +----------------->|     Mainfest      |
+                        |                  +===================+
+                        |                  |+-----------------+|
+                        |                  || format          ||
+                        |                  |+-----------------+|
+                        |                  || image count     ||
+                        |                  |+-----------------+|
+                        |     +------------|| digest(Image 1) ||
+                        |     |            |+-----------------+|
+                        |     |            || digest(Image 2) ||-------------+
+                        |     |            |+-----------------+|             |
+                        |     |            +-------------------+             |
+                        |     |                                              |
+    +-----------------+ |     |  +-----------------+    +-----------------+  |
+    | Manifest image  | |     |  |     Image 1     |    |     Image 2     |  |
+    +=================+ |     |  +=================+    +=================+  |
+    |+---------------+| |     |  |+---------------+|    |+---------------+|  |
+    || header        || |     |  || header        ||    || header        ||  |
+    |+---------------+| |     |  |+---------------+|    |+---------------+|  |
+    || manifest TLV  ||-+     |  || firmware      ||    || firmware      ||  |
+    |+---------------+|       |  |+---------------+|    |+---------------+|  |
+    || digest TLV    ||       +->|| digest TLV    ||    || digest TLV    ||<-+
+    |+---------------+|          |+---------------+|    |+---------------+|
+    || signature TLV ||          || signature TLV ||    || signature TLV ||
+    |+---------------+|          |+---------------+|    |+---------------+|
+    +-----------------+          +-----------------+    +-----------------+
+```
+
+The manifest TLV has a format field that allows for the development of complex
+boot logic in the future. The default manifest is structured as a list of
+digests of firmware parts.
+
+A manifest is transferred as a protected TLV in a dedicated "Manifest image."
+This image is updated using the same mechanisms as regular images.
+
+A manifest image may contain firmware - if so, this implies that this part of
+the firmware must be updated with any firmware update.
+
+The bootloader's behavior changes once manifest-based updates
+and booting are enabled.
+
+Boot process for Direct-XIP modes:
+
++  Loop 1. Until all images are loaded and validated against the active manifest
+    1. Subloop 1. Iterate over the manifest image slots
+        + Does any of the slots contain a manifest?
+            + Yes:
+                + Select the newer manifest.
+                + Copy it to the bootloader state.
+                + Validate the manifest image (integrity and security check).
+                + If validation fails mark the active manifest slot as
+                  unavailable and try the other slot.
+            + No: Return with an error.
+
+    2. Subloop 2. Iterate over all images except manifest image
+        + Does the current image contain a valid header in the same slot
+          as the selected (active) manifest?
+            + Yes: Is the image valid (integrity and security check) and its
+              digest matches the manifest?
+                + Yes: Skip to the next image.
+                + No:
+                    + Mark the active manifest slot as unavailable.
+                    + Restart main loop.
+            + No:
+                + Mark the active manifest slot as unavailable.
+                + Restart main loop.
+
++  Loop 2. Iterate over all images
+    + Increase the security counter if needed.
+    + Do the measured boot and the data sharing if needed.
+
++ Boot the loaded slot of image 0.
+
+```
+  Slot 0:
+                         +----------------------- +
+                         |                        |
+    +-----------------+  |   +-----------------+  |   +-----------------+
+    | Manifest image  |  |   |     Image 1     |  |   |     Image 2     |
+    +=================+  |   +=================+  |   +=================+
+    |+---------------+|  |   |+---------------+|  |   |+---------------+|
+    || manifest TLV  ||--+-->|| digest TLV    ||  +-->|| digest TLV    ||
+    |+---------------+|      |+---------------+|      |+---------------+|
+    +-----------------+      +-----------------+      +-----------------+
+
+  Slot 1:
+                         +----------------------- +
+                         |                        |
+    +-----------------+  |   +-----------------+  |   +-----------------+
+    | Manifest image  |  |   |     Image 1     |  |   |     Image 2     |
+    +=================+  |   +=================+  |   +=================+
+    |+---------------+|  |   |+---------------+|  |   |+---------------+|
+    || manifest TLV  ||--+-->|| digest TLV    ||  +-->|| digest TLV    ||
+    |+---------------+|      |+---------------+|      |+---------------+|
+    +-----------------+      +-----------------+      +-----------------+
+```
+
+Manifest-based updates and booting for other modes are not yet implemented.
 
 ## [Image swapping](#image-swapping)
 


boot/zephyr/Kconfig
```diff
@@ -1062,6 +1062,33 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_LIMITED
 
 endchoice
 
+config MCUBOOT_MANIFEST_UPDATES
+	bool "Enable transactional updates"
```
tomchy, line 1087: Add experimental

boot/bootutil/src/loader_manifest_xip_ram.c
```diff
@@ -0,0 +1,651 @@
...
+            /* All images have been loaded and validated successfully. */
+            break;
+        }
+
+        BOOT_LOG_DBG("Manifest in slot %d is invalid", active_slot);
+
+        /* Invalidate manifest */
+        state->slot_usage[MCUBOOT_MANIFEST_IMAGE_NUMBER].slot_available[active_slot] = false;
+        state->slot_usage[MCUBOOT_MANIFEST_IMAGE_NUMBER].active_slot = BOOT_SLOT_NONE;
+#ifdef MCUBOOT_RAM_LOAD
```
tomchy: Remove RAM_LOAD

michalek-no (approved):

boot/bootutil/include/bootutil/mcuboot_manifest.h
```diff
@@ -0,0 +1,96 @@
...
+    if (manifest == NULL) {
+        return false;
+    }
+
+    /* Currently only the simplest manifest format is supported */
+    if (manifest->format != 0x1) {
+        return false;
+    }
+
+    if (manifest->image_count != BOOT_IMAGE_NUMBER - 1) {
```
de-nordic: () around arithmetic.

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,623 @@
...
+ * Checks whether the active slot of the current image was previously selected
+ * to run. Erases the image if it was selected but its execution failed,
+ * otherwise marks it as selected if it has not been before.
+ *
+ * @param  state        Boot loader status information.
+ *
+ * @return              0 on success; nonzero on failure.
+ */
+static int
+boot_select_or_erase(struct boot_loader_state *state)
```
tomchy, line 316: TODO: This should be executed only on the manifest image. If a manifest image is erased/not present, all other images will be rejected on the digest check.

boot/bootutil/include/bootutil/mcuboot_manifest.h
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    if (image_index < MCUBOOT_MANIFEST_IMAGE_NUMBER) {
+        return manifest->image_hash[image_index];
+    } else if (image_index > MCUBOOT_MANIFEST_IMAGE_NUMBER) {
+        return manifest->image_hash[image_index - 1];
+    }
+
+    return NULL;
+}
```
de-nordic: I am not sure this should be able to return NULL in any case.
This should panic on any parameter that is not expected.
Since the function is always called before `boot_fih_memequal` I think that both functions should be merged and the `bootutil_get_image_hash` should be changed to `FIH` return function that will be given enough parameter to compare a hash without pulling it out of the manifest object.
This should also reduce code size, since the `boot_fih_memequal` and will be combined here and no need for external null checks.

boot/bootutil/src/bootutil_priv.h
```diff
@@ -271,6 +275,14 @@ struct boot_loader_state {
 #endif
     } slot_usage[BOOT_IMAGE_NUMBER];
 #endif /* MCUBOOT_DIRECT_XIP || MCUBOOT_RAM_LOAD */
+
+#if defined(MCUBOOT_MANIFEST_UPDATES)
+    struct mcuboot_manifest manifest[BOOT_NUM_SLOTS];
+    bool manifest_valid[BOOT_NUM_SLOTS];
```
de-nordic, line 281: Can we move that into the `mcuboot_manifest`? We can have different structs for TLV and internal usage.

boot/bootutil/src/image_validate.c
```diff
@@ -356,6 +366,82 @@ bootutil_img_validate(struct boot_loader_state *state,
                 goto out;
             }
 
+#ifdef MCUBOOT_MANIFEST_UPDATES
+            if (image_index == MCUBOOT_MANIFEST_IMAGE_NUMBER) {
+                if (!state->manifest_valid[slot]) {
+                    /* Manifest TLV must be processed before any of the image's hash TLV. */
+                    BOOT_LOG_INF("bootutil_img_validate: image rejected, manifest not found before image %d hash",
```
de-nordic: Isn't that at least warning? It ends up in rc = -1, so basically fails the function.

boot/bootutil/src/image_validate.c
```diff
@@ -356,6 +366,82 @@ bootutil_img_validate(struct boot_loader_state *state,
...
+                    FIH_CALL(boot_fih_memequal, fih_rc, hash, image_hash, sizeof(hash));
+                    if (FIH_EQ(fih_rc, FIH_SUCCESS)) {
+                        state->matching_manifest[image_index][slot] = BOOT_SLOT_SECONDARY;
+                    }
+                }
+            }
+
+            /* No matching manifest found. */
+            if (state->matching_manifest[image_index][slot] == BOOT_SLOT_NONE) {
+                BOOT_LOG_INF("bootutil_img_validate: image rejected, no valid manifest for image %d slot %d",
```
de-nordic: This is either warning or error, we are failing function here.

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,623 @@
...
+ * Prints the state of the loaded images.
+ *
+ * @param  state        Boot loader status information.
+ */
+static void
+print_loaded_images(struct boot_loader_state *state)
+{
+    uint32_t active_slot;
+
+    (void)state;
```
de-nordic: Seems to be used.

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,623 @@
...
+#if (BOOT_IMAGE_NUMBER > 1)
+        BOOT_CURR_IMG(state) = MCUBOOT_MANIFEST_IMAGE_NUMBER;
+#endif
+        rc = BOOT_HOOK_FIND_SLOT_CALL(boot_find_next_slot_hook, BOOT_HOOK_REGULAR,
+                                      state, BOOT_CURR_IMG(state), &active_slot);
+        if (rc == BOOT_HOOK_REGULAR) {
+            active_slot = find_slot_with_highest_version(state);
+        }
+        if (active_slot == BOOT_SLOT_NONE) {
+            BOOT_LOG_INF("No more manifest slots available");
```
de-nordic: Error?

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,623 @@
...
+        /* Attempt to read an image header from each slot. */
+        rc = boot_read_image_headers(state, false, NULL);
+        if (rc != 0) {
+            BOOT_LOG_WRN("Failed reading image headers.");
+            return rc;
+        }
+
+        /* Check headers in all slots */
+        for (slot = 0; slot < BOOT_NUM_SLOTS; slot++) {
+            hdr = boot_img_hdr(state, slot);
```
de-nordic: Can be moved into if under 231, there is only usage of this.

boot/bootutil/src/image_validate.c
```diff
@@ -564,6 +683,13 @@ bootutil_img_validate(struct boot_loader_state *state,
     }
 #endif
 
+#ifdef MCUBOOT_MANIFEST_UPDATES
+    if (image_index == MCUBOOT_MANIFEST_IMAGE_NUMBER && (!manifest_found || !manifest_valid)) {
+        BOOT_LOG_INF("bootutil_img_validate: slot %d rejected, manifest missing or invalid", slot);
```
de-nordic: Isn't this error?

boot/bootutil/include/bootutil/mcuboot_manifest.h
tomchy: I agree that there is no need to expose the manifest hash, however since this function is likely to be shared between the application and bootloader, I will remove the FIH part from the logic, so it does not rely on additional external modules.

boot/bootutil/src/bootutil_priv.h
tomchy: That'd need a struct within a struct, as the `struct mcuboot_manifest` is directly loaded from the TLV and the `manifest_valid` is not part of the incoming data, but an internal flag, indicating that the structure was loaded.

If you find the "valid" sate flag to be non-universal, I may move everything inside (including `manifest_valid` as well as `matching_manifest`), but I am hesitant to do so as those are not variables that holds information from the manifest, but a state variables for a loader logic and are not needed outside of the bootloader scope.

boot/bootutil/src/image_validate.c
tomchy: I'm OK with increasing the log level. Generally speaking - this function is not too verbose (most errors logged as debug information), but I do not fully understand the reason behind it (code size?).

boot/bootutil/src/image_validate.c
tomchy: Changed to error.

boot/bootutil/src/loader_manifest_xip.c
tomchy: Interestingly, this is copied from the regular `loader.c` file.

boot/bootutil/src/loader_manifest_xip.c
tomchy: Definitely!

boot/zephyr/Kconfig
```diff
@@ -1083,6 +1083,31 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION_LOCK
...
+	  that is signed to prevent tampering.
+	  The manifest must be transferred as part of the image with index 0.
+	  It can be a dedicated image, or part of an existing image.
+	  If the second option is selected, all updates must contain an update
+	  for image 0.
+
+if MCUBOOT_MANIFEST_UPDATES
+
+config MCUBOOT_MANIFEST_IMAGE_NUMBER
+	int "Number of image that must include manifest"
```
nvlsianpu: `the image`

boot/zephyr/Kconfig
```diff
@@ -1083,6 +1083,31 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION_LOCK
...
+	  their expected hash values. The manifest TLV is a part of an image
+	  that is signed to prevent tampering.
+	  The manifest must be transferred as part of the image with index 0.
+	  It can be a dedicated image, or part of an existing image.
+	  If the second option is selected, all updates must contain an update
+	  for image 0.
+
+if MCUBOOT_MANIFEST_UPDATES
+
+config MCUBOOT_MANIFEST_IMAGE_NUMBER
```
nvlsianpu: consider NUMBER -> INDEX
rationale: can't mistake a index with an item count.

boot/zephyr/CMakeLists.txt
```diff
@@ -157,6 +157,26 @@ elseif(CONFIG_BOOT_FIRMWARE_LOADER)
     ${BOOT_DIR}/zephyr/firmware_loader.c
     )
   zephyr_library_include_directories(${BOOT_DIR}/bootutil/src)
+elseif(CONFIG_MCUBOOT_MANIFEST_UPDATES)
```
nvlsianpu, line 161: that's good :)

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,639 @@
...
+ * Check if image in slot has been set with specific ROM address to run from
+ * and whether the slot starts at that address.
+ *
+ * @returns 0 if IMAGE_F_ROM_FIXED flag is not set;
+ *          0 if IMAGE_F_ROM_FIXED flag is set and ROM address specified in
+ *            header matches the slot address;
+ *          1 if IMF_F_ROM_FIXED flag is set but ROM address specified in header
+ *          does not match the slot address.
+ */
+static bool
```
nvlsianpu, line 120: f. name and bool return value type are suggesting positive logic, while the logic is actually reversed

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,639 @@
...
+ */
+fih_ret
+boot_load_and_validate_images(struct boot_loader_state *state)
+{
+    uint32_t active_slot;
+    int rc;
+    fih_ret fih_rc;
+
+    while (true) {
+#if (BOOT_IMAGE_NUMBER > 1)
```
nvlsianpu: Isn't that always the image number is at last 2? Any sense to have this for one image?

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,639 @@
...
+        BOOT_CURR_IMG(state) = MCUBOOT_MANIFEST_IMAGE_NUMBER;
+#endif
+        rc = BOOT_HOOK_FIND_SLOT_CALL(boot_find_next_slot_hook, BOOT_HOOK_REGULAR,
+                                      state, BOOT_CURR_IMG(state), &active_slot);
```
nvlsianpu: transfer MCUBOOT_MANIFEST_IMAGE_NUMBER instead of BOOT_CURR_IMG(state)
(gives +10 to legibility)

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,639 @@
...
+        if (rc == BOOT_HOOK_REGULAR) {
+            active_slot = find_slot_with_highest_version(state);
+        }
+        if (active_slot == BOOT_SLOT_NONE) {
+            BOOT_LOG_ERR("No more manifest slots available");
+            FIH_RET(FIH_FAILURE);
+        }
+
+        /* Save the number of the active manifest slot. */
+        state->slot_usage[BOOT_CURR_IMG(state)].active_slot = active_slot;
```
nvlsianpu: transfer MCUBOOT_MANIFEST_IMAGE_NUMBER instead of BOOT_CURR_IMG(state)
(gives +10 to legibility)

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,639 @@
...
+        }
+
+        BOOT_LOG_DBG("Manifest in slot %d is invalid", active_slot);
+
+        /* Invalidate manifest */
+        state->slot_usage[MCUBOOT_MANIFEST_IMAGE_NUMBER].slot_available[active_slot] = false;
+        state->slot_usage[MCUBOOT_MANIFEST_IMAGE_NUMBER].active_slot = BOOT_SLOT_NONE;
+    }
+
+    FIH_RET(FIH_SUCCESS);
```
nvlsianpu: always success? whenever we get into the IMAGES_ITER() loop above.

docs/design.md
```diff
@@ -987,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+```
+
+The manifest TLV has a format field that allows for the development of complex
+boot logic in the future. The default manifest is structured as a list of
+digests of firmware parts.
+
+The manifest is transferred as a protected TLV in a dedicated "Manifest image."
+This image is updated using the same mechanisms as regular images.
+
+The manifest image may contain firmware - if so, this part of the firmware must
```
nvlsianpu: if so, this implies ...

boot/bootutil/src/loader_manifest_xip.c
tomchy: fixed. Also - changed the call to receive `bool` instead of `int`.

boot/bootutil/src/loader_manifest_xip.c
tomchy: Correct. I also need to express it within Kconfigs then.

boot/bootutil/src/loader_manifest_xip.c
tomchy: I do not see it. The `break` inside the `IMAGES_ITER()` should not bring us out of the `while(true)` which has two exit points:
```
        if (active_slot == BOOT_SLOT_NONE) {
            BOOT_LOG_ERR("No more manifest slots available");
            FIH_RET(FIH_FAILURE);
        }
```
and
```
        if (FIH_EQ(fih_rc, FIH_SUCCESS)) {
            /* All images have been loaded and validated successfully. */
            break;
        }
```
either way - changed to the `FIH_RET(FIH_FAILURE);` and replaced the `break` with `FIH_RET(FIH_SUCCESS);`.

boot/bootutil/src/loader_manifest_xip.c
nvlsianpu: shouldn't be actually int as return value type?

boot/bootutil/src/loader_manifest_xip.c
nvlsianpu: My bad - missed while(true) loop. No issue there

boot/bootutil/src/loader_manifest_xip.c
tomchy: This function returns only one out of two values, so I think `bool` is more appropriate.
This is also a `static` function, so any modification of its signature should not affect other modules.

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
 
 + Boot the loaded slot of image 0.
 
+### [Multiple image boot using manifest](#multiple-image-boot-using-manifest)
+
+
+Deployments that use multiple images typically require strict control over
+versions of firmware components.
+There is a dependency TLV that can be used to specify dependencies between
+semantic versions of multiple components. However, since these only describe
```
de-nordic:
```suggestion
semantic versions of multiple components; however, since these only describe
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+a minimally compatible version of a counterpart component, there is no mechanism
+to enforce a specific revision of the other image.
+Therefore, the publisher must ensure that all combinations that satisfy the
```
de-nordic:
```suggestion
to enforce a specific revision of the other image;
therefore, the publisher must ensure that all combinations that satisfy the
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+dependencies are compatible with each other and are tested before deploying
+a new version of the firmware bundle.
+One way to simplify the process is to add dependencies to all parts that
+enforce the latest revision of all other parts of the bundle.
+This effectively enforces equality between the revisions of all parts.
+This solution becomes even more problematic when using Direct-XIP mode - since
```
de-nordic:
```suggestion
This solution becomes even more problematic when using Direct-XIP mode, since
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+the bootloader typically only chain-loads the next stage, the next stage must
+select and boot the correct slot of the next part of the firmware.
+Although the dependencies ensure that the next part with a specific
+version is present on the device, there is no guarantee in which slot it
+was validated.
+Again, the publisher may use different version numbers for the same firmware
```
de-nordic:
```suggestion
The publisher may use different version numbers for the same firmware
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+created for different slots, but this raises the question of whether there is
+a better way to manage dependencies between images.
+
+The bootloader manifest is an additional, protected TLV that serves as the sole
```
de-nordic:
```suggestion
The bootloader manifest is a protected TLV that serves as the sole
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+source of information about the compatibility and bootability of the multipart
+firmware. The basic rules for all types of manifests are:
+
+  * The manifest must be protected (directly or indirectly) by a cryptographic
```
de-nordic:
```suggestion
  * A manifest must be protected (directly or indirectly) by a cryptographic
```
We describe any manifest?

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+    signature.
+  * There must be exactly one selected image that can provide the manifest TLV.
+  * There must be only one manifest per slot.
+  * Processing of the manifest must validate the full functional set of firmware
+    components.
+  * If the manifest does not describe a part of the firmware, it must be
```
de-nordic:
```suggestion
  * If a manifest does not describe a part of the firmware, it must be
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+    || signature TLV ||          || signature TLV ||    || signature TLV ||
+    |+---------------+|          |+---------------+|    |+---------------+|
+    +-----------------+          +-----------------+    +-----------------+
+```
+
+The manifest TLV has a format field that allows for the development of complex
+boot logic in the future. The default manifest is structured as a list of
+digests of firmware parts.
+
+The manifest is transferred as a protected TLV in a dedicated "Manifest image."
```
de-nordic:
```suggestion
A manifest is transferred as a protected TLV in a dedicated "Manifest image."
```

docs/design.md
```diff
@@ -986,6 +988,147 @@ strategy but there is no need for Scratch area.
...
+This image is updated using the same mechanisms as regular images.
+
+The manifest image may contain firmware - if so, this implies that this part of
```
de-nordic:
```suggestion
A manifest image may contain firmware - if so, this implies that this part of
```

docs/design.md
```diff
@@ -150,8 +150,9 @@ struct image_tlv {
                                              * ...
                                              * 0xffa0 - 0xfffe
                                              */
-#define IMAGE_TLV_UUID_VID          0x80    /* Vendor unique identifier */
-#define IMAGE_TLV_UUID_CID          0x81    /* Device class unique identifier */
+#define IMAGE_TLV_UUID_VID          0x74    /* Vendor unique identifier */
+#define IMAGE_TLV_UUID_CID          0x75    /* Device class unique identifier */
```
de-nordic, line 155: This correction should go to separate commit.

boot/bootutil/include/bootutil/mcuboot_manifest.h
```diff
@@ -0,0 +1,99 @@
...
+ * @brief Get the image hash from the manifest.
+ *
+ * @param[in] manifest     The reference to the manifest structure.
+ * @param[in] image_index  The index of the image to get the hash for.
+ *                         Must be in range <0, MCUBOOT_IMAGE_NUMBER - 1>, but
+ *                         must not be equal to MCUBOOT_MANIFEST_IMAGE_INDEX.
+ *
+ * @return true if hash matches with the manifest, false otherwise.
+ */
+static inline bool bootutil_verify_manifest_image_hash(const struct mcuboot_manifest *manifest,
```
de-nordic: I am not sure there is sense in inlining this, there are at least two calls here that will require stack operations, so the entire inline will probably be larger than just calling the function.

boot/bootutil/src/loader_manifest_xip.c
```diff
@@ -0,0 +1,648 @@
...
+    fih_ret fih_rc;
+
+    while (true) {
+        BOOT_CURR_IMG(state) = MCUBOOT_MANIFEST_IMAGE_INDEX;
+
+        rc = BOOT_HOOK_FIND_SLOT_CALL(boot_find_next_slot_hook, BOOT_HOOK_REGULAR,
+                                      state, MCUBOOT_MANIFEST_IMAGE_INDEX, &active_slot);
+        if (rc == BOOT_HOOK_REGULAR) {
+            active_slot = find_slot_with_highest_version(state);
+        }
```
de-nordic, line 452: Took me a moment to figure out that this searches for newest image with manifest to use. Some comment could be added here.

docs/design.md
tomchy: Those points were meant to describe assumptions, that affected the design.
I wanted to leave them, so in case there will be a second version of the manifest format, there will be some basis to start with.

boot/bootutil/src/loader_manifest_xip.c
tomchy: Comments added. Please check if it is easier to grasp the idea now.

docs/design.md
de-nordic: What I mean "the" specifies some single already mentioned item, "a" is general item.

boot/bootutil/include/bootutil/mcuboot_manifest.h
tomchy: I'll try to separate it into a new C file then.

PR #2515, opened 2025-10-22
loader: Fix compile-time issues in loader.c

tomchy:
Fix a few minor issues in loader.c file when MCUBOOT_VERIFY_IMG_ADDRESS is enabled.


boot/bootutil/src/loader.c
de-nordic, line 1: Typo in commin message IG->IMG

tomchy: Fixed!

PR #2516, opened 2025-10-22
Support for Embedded Plain AES keys

de-nordic:
Support in code for querying user provided function for an AES encryption key.

Extra Zephyr code and template that allows to use the new functionality.

Although I think this is ready for review I am still working, in different set of commits, on redefining encryption key type so that we could abstract that type across the MCUboot code; this means that this PR may change in how boot_take_enc_key function is defined.

~~Update: I will move imgtool script to separate PR, at this point. Fighting with FIH tests that from, what seems to be a timeout, without any feedback is impossible. Running personal docker is impossible either, because after getting authorized with docker registry I am denied access to required data anyway.~~
Finally fixed the scripts to make FIH and TrustedFirmware-M happy.


nvlsianpu (commented):
changes in scripts/imgtool/image.py looks good.

boot/boot_serial/src/boot_serial_encryption.c
```diff
@@ -235,7 +235,11 @@ decrypt_image_inplace(const struct flash_area *fa_p,
 #endif
         memset(&boot_data, 0, sizeof(struct boot_loader_state));
         /* Load the encryption keys into cache */
+#if !defined(MCUBOOT_BUILTIN_ENC_KEY)
```
nvlsianpu: just minor thing - I would prefer positive case testing over negative, especially when this is just about swapping mutually exclusive code sequence.

boot/bootutil/src/loader.c
```diff
@@ -1007,8 +1008,12 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
 #ifdef MCUBOOT_ENC_IMAGES
     if (IS_ENCRYPTED(boot_img_hdr(state, BOOT_SLOT_SECONDARY))) {
         rc = boot_enc_load(state, BOOT_SLOT_SECONDARY,
+#if !defined(MCUBOOT_BUILTIN_ENC_KEY)
```
nvlsianpu: move #if one line up.

boot/bootutil/src/loader.c
```diff
@@ -1201,9 +1217,15 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
                 }
             }
 
-            if (i != BOOT_ENC_KEY_SIZE) {
-                boot_enc_set_key(BOOT_CURR_ENC_SLOT(state, slot), bs->enckey[slot]);
+            if (i == BOOT_ENC_KEY_SIZE) {
```
nvlsianpu: `i` is not initialized.

boot/zephyr/Kconfig
```diff
@@ -653,12 +658,66 @@ config BOOT_ENCRYPTION_SUPPORT
...
+	string "Hexadecimal string representing AES key"
+	depends on BOOT_ENCRYPT_IMAGE_GENERATE_BASIC_KEY_PROVIDER
+	help
+	  AES key in form of hexadecimal string that will be used to
+          generate boot_enc_take_key function, returning the key for
+          decryption and encryption of image.
+	  The key character length should be the double of expected
+	  AES key length in bytes.
+
+config BOOT_ENCRYPT_IMAGE_EMBEDDED_CUSTOM_KEY_PROVIDER_DIR
```
nvlsianpu: User need to modify project on there own anywaey, In a fork or maybe in the feature in this CMakeList.stxt. No better to instruct user to include relevant sources on their own?

de-nordic: General idea is that user needs to copy template and do what they need. They can add this as part of MCUboot or have separate repo for their key routines. Advantage of this is that you do not have to modify MCUboot CMakeLists.txt.

scripts/imgtool/image.py
```diff
@@ -742,12 +741,18 @@ def create(self, key, public_key_format, enckey, dependencies=None,
         if protected_tlv_off is not None:
             self.payload = self.payload[:protected_tlv_off]
 
-        if enckey is not None and dont_encrypt is False:
-            if encrypt_keylen == 256:
-                plainkey = os.urandom(32)
-            else:
-                plainkey = os.urandom(16)
+        # When passed AES key and clear flag, then do not encrypt, because the key
+        # is only passed to be stored in encryption key TLV, the paylad stays clear text.
```
nordicjm: *payload

boot/zephyr/Kconfig
```diff
@@ -653,12 +658,66 @@ config BOOT_ENCRYPTION_SUPPORT
...
+	string "Path to AES key provider, boot_enc_take_key, implementation"
+	depends on BOOT_ENCRYPT_IMAGE_USE_CUSTOM_KEY_PROVIDER
+	default "templates/custom_boot_enc_take_key"
+	help
+	  Use templates/custom_boot_enc_take_key as a template.
```
nordicjm: yeah let's not have this as Kconfig as this is not zephyr-like, have an option to use this (without a path for it in Kconfig) and an option which includes nothing (part of same choice), and in the other case users can expand it with zephyr modules

boot/zephyr/templates/custom_boot_enc_take_key/README.rst
```diff
@@ -0,0 +1,11 @@
+This is template for custom AES key provider.
+
+The template can be used as a starter for your own implementation.
+To use the template, copy it somewhere where you want to continue
```
nordicjm: would leave instructions off, modules are documented in zephyr

nordicjm (changes requested):

boot/zephyr/Kconfig
de-nordic: OK, so generally just build nothing more in MCUboot and leave it for user to link something here?

boot/zephyr/templates/custom_boot_enc_take_key/README.rst
de-nordic: Maybe just remove the template then? If we are not going to point to it anyway.

boot/zephyr/Kconfig
nordicjm: if there is a sample it can be used (probably best in https://github.com/mcu-tools/mcuboot/tree/main/samples ) but for actual user implementation they would just use a module (so it's reusable)

boot/zephyr/Kconfig
de-nordic: Removed

boot/bootutil/src/loader.c
de-nordic: No longer applies

boot/bootutil/src/loader.c
de-nordic: Fixed

boot/boot_serial/src/boot_serial_encryption.c
de-nordic: Fixed

scripts/imgtool/image.py
de-nordic: Fixed

boot/zephyr/Kconfig
de-nordic: Removed template

boot/zephyr/templates/custom_boot_enc_take_key/README.rst
de-nordic: Removed template

boot/boot_serial/src/boot_serial_encryption.c
```diff
@@ -31,7 +31,11 @@ boot_image_validate_encrypted(struct boot_loader_state *state,
     int rc;
 
     if (MUST_DECRYPT(fa_p, BOOT_CURR_IMG(state), hdr)) {
+#ifdef MCUBOOT_EMBEDDED_ENC_KEY
+        rc = boot_en_take_key(bs->enckey[BOOT_SLOT_SECONDARY], BOOT_CUR_IMG(state), BOOT_SLOT_SECONDARY);
```
nvlsianpu, line 35: I guess wrapping this API function under boot_enc_load() doesn't make sense.

nvlsianpu (approved):

boot/boot_serial/src/boot_serial_encryption.c
de-nordic: Crap, there is bug in this line actually.
Anyway, I was considering doing the opposite so that boot_enc_load would become one of boot_take_enc_key implementations, but could not get rid of `boot_loader_state` for various reasons.

PR #2523, opened 2025-10-24
Remove tinycrypt

d3zd3z:
Remove tinycrypt entirely as a dependency from mcuboot.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/ecdh_p256.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdh_p256.h
@@ -23,50 +23,9 @@
     #define EC256_PUBK_LEN (65)
 #endif /* MCUBOOT_USE_MBED_TLS */
 
-#if defined(MCUBOOT_USE_TINYCRYPT)
-    #include <tinycrypt/ecc_dh.h>
-    #include <tinycrypt/constants.h>
-    #define BOOTUTIL_CRYPTO_ECDH_P256_HASH_SIZE (4 * 8)
-#endif /* MCUBOOT_USE_TINYCRYPT */
-
 #ifdef __cplusplus
 extern "C" {
 #endif
-
-#if defined(MCUBOOT_USE_TINYCRYPT)
-typedef uintptr_t bootutil_ecdh_p256_context;
-typedef bootutil_ecdh_p256_context bootutil_key_exchange_ctx;
-static inline void bootutil_ecdh_p256_init(bootutil_ecdh_p256_context *ctx)
-{
-    (void)ctx;
-}
-
-static inline void bootutil_ecdh_p256_drop(bootutil_ecdh_p256_context *ctx)
-{
-    (void)ctx;
-}
-
-static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *ctx, const uint8_t *pk, const uint8_t *sk, uint8_t *z)
-{
-    int rc;
-    (void)ctx;
-
-    if (pk[0] != 0x04) {
-        return -1;
-    }
-
-    rc = uECC_valid_public_key(&pk[1], uECC_secp256r1());
-    if (rc != 0) {
-        return -1;
-    }
-
-    rc = uECC_shared_secret(&pk[1], sk, z, uECC_secp256r1());
-    if (rc != TC_CRYPTO_SUCCESS) {
-        return -1;
-    }
-    return 0;
-}
-#endif /* MCUBOOT_USE_TINYCRYPT */
 
 #if defined(MCUBOOT_USE_MBED_TLS)
 #define NUM_ECC_BYTES 32
--- a/boot/bootutil/include/bootutil/crypto/ecdsa.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdsa.h
@@ -8,7 +8,7 @@
  * This module provides a thin abstraction over some of the crypto
  * primitives to make it easier to swap out the used crypto library.
  *
- * At this point, the choices are: MCUBOOT_USE_TINYCRYPT, MCUBOOT_USE_CC310,
+ * At this point, the choices are: MCUBOOT_USE_CC310,
  * MCUBOOT_USE_MBED_TLS, MCUBOOT_USE_PSA_CRYPTO. Note that support for
  * MCUBOOT_USE_PSA_CRYPTO is still experimental and it might not support all
  * the crypto abstractions that MCUBOOT_USE_MBED_TLS supports. For this
@@ -32,16 +32,10 @@
     #error "P384 requires PSA_CRYPTO to be defined"
 #endif
 
-#if (defined(MCUBOOT_USE_TINYCRYPT) + \
-     defined(MCUBOOT_USE_CC310) + \
+#if defined(MCUBOOT_USE_CC310) + \
      defined(MCUBOOT_USE_PSA_OR_MBED_TLS)) != 1
-    #error "One crypto backend must be defined: either CC310/TINYCRYPT/MBED_TLS/PSA_CRYPTO"
+    #error "One crypto backend must be defined: either CC310/MBED_TLS/PSA_CRYPTO"
 #endif
-
-#if defined(MCUBOOT_USE_TINYCRYPT)
-    #include <tinycrypt/ecc_dsa.h>
-    #include <tinycrypt/constants.h>
-#endif /* MCUBOOT_USE_TINYCRYPT */
 
 #if defined(MCUBOOT_USE_CC310)
     #include <cc310_glue.h>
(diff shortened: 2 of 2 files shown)


JayToltTech:
So I have some concerns here based on my personal experience.

I went to modify my MCUBoot to use EC-384 as part of working on my MDR Cybersecurity (IEC 81001-5-1:2022) work.  This led me to replace Tinycrypt (which includes EC-256 but not EC-384) with PSA/mbedTLS.  This in turn grew my MCUBoot flash image beyond its 84 kb size to over 100 kb, with a +24 kb flash image size.  I asked Claude to do a thorough analysis of the impact of replacing Tinycrypt with PSA/mbedTLS.  The short of the matter was I am going to port microECC over to replace Tinycrypt for my specific use case due to the flash growth that bringing PSA/mbedTLS brought.

# Why pruned mbedTLS PSA is 24 KB for one ECDSA-P384 verify

  ## Summary

  Measured on the ADT-25 bootloader (Zephyr 4.4, nRF5340 cpuapp, MCUboot
  26.4.15 + Tolt fork patches) with `mcuboot/overlay-mbedtls-p384-measure.conf`
  applied (aggressive prune of SSL/TLS, X.509, RSA, DH, ECDH, ciphers, MACs,
  KDFs, all other curves, all other hashes, entropy, RNG):

  | Measurement | Value |
  |---|---|
  | Baseline (P-256 + TinyCrypt + SHA-256) | 80,432 B |
  | Pruned mbedTLS PSA (P-384 + SHA-384) | 105,032 B |
  | **Δ (mbedTLS PSA cost over baseline)** | **+24,600 B** |
  | Budget (84 KB `boot_partition`) | 86,016 B |
  | **Overflow** | **+18,880 B** |

  Direct symbol-level attribution of the mbedTLS contribution: **24,458 B**
  (matches the delta within 0.6% — there is no hidden bloat elsewhere).

  ## Where the 24 KB goes

  Per-component breakdown from `arm-none-eabi-nm --size-sort --print-size`
  on `build-mcuboot/zephyr/zephyr.elf`:

  | Component | Size | Purpose |
  |---|---:|---|
  | MPI (big-number arithmetic) | 6,988 B | Generic arbitrary-precision integer ops: add, sub, mul, div, mod, modinv,
  gcd, `exp_mod_optionally_safe` |
  | SHA-384/512 | 4,098 B | Generic `mbedtls_md_context_t` + SHA-512 core (SHA-384 reuses this);
  `mbedtls_internal_sha512_process` alone is 2,452 B |
  | ECP static helpers | 3,066 B | `ecp_mul_comb`, `ecp_double_jac`, `ecp_add_mixed`, `ecp_mul_restartable_internal` |
  | PSA dispatcher / key slots | 2,812 B | `psa_crypto_init`, `psa_verify_hash`, slot mgmt, policy enforcement, error
  translation |
  | Other mbedTLS (platform, CT, helpers) | 2,426 B | Constant-time utilities, platform abstraction, memory-buffer
  allocator |
  | ECP curve math (`mbedtls_ecp_*`) | 1,778 B | Public ECC API: point read/write, mul, muladd, binary I/O |
  | mbedTLS ↔ PSA bridge | 1,270 B | Translation between PSA API surface and internal mbedTLS calls |
  | SHA-512 K constants (`.rodata`, 2 copies) | 896 B | 640 B + 256 B of round constants |
  | `secp384r1_T` precomputed lookup (`.rodata`) | 768 B | Precomputed multiples of generator for scalar-mult speedup |
  | ECDSA wrapper | 350 B | Thin layer over ECP |
  | **Total** | **~24,500 B** | |

  ## Five reasons the pruning floor is so high

  ### 1. Generic vs fixed-width arithmetic (~7 KB)

  `mbedtls_mpi` supports arbitrary-precision integers with dynamic
  allocation. Even for P-384 (known-size 384-bit operands) the routines
  are written generically:

  - `mbedtls_mpi_exp_mod_optionally_safe` — 438 B
  - `mbedtls_mpi_core_gcd_modinv_odd` — 472 B
  - `mbedtls_mpi_div_mpi` — 1,008 B

  These are tightly interconnected. Pulling in ECDSA verify drags in
  ~70 MPI functions via the linker, even with `-ffunction-sections`
  `--gc-sections`, because `ecp_mul` needs `mpi_exp_mod` needs `mpi_mul`
  needs `mpi_div` needs the generic primitives.

  **micro-ecc** implements P-384 as fixed 12×32-bit word operations inline
  with no length parameters, no allocation, no generic shape.

  ### 2. Generic ECC infrastructure (~3–4 KB)

  mbedTLS's ECP is parameterized over a curve descriptor
  `(p, a, b, n, G, cofactor)` so one code path handles secp256r1,
  secp384r1, secp521r1, curve25519, brainpool, and so on.

  The `mbedtls_ecp_mul_restartable` (524 B) + `mbedtls_ecp_muladd_restartable`
  (480 B) machinery carries restartability support that MCUboot's verify
  will never use.

  **micro-ecc** has P-384 constants and code baked in as a single fixed
  implementation.

  ### 3. PSA service layer (~4 KB)

  PSA is a real service API:

  - 32-bit algorithm encoding: `PSA_ALG_ECDSA(PSA_ALG_SHA_384)`
  - Key slot table with full lifecycle
  - Key attributes (type, bits, usage, algorithm, lifetime)
  - Driver-wrapper dispatch (transparent vs opaque)
  - Policy enforcement
  - Error translation

  Largest single symbol: `psa_key_algorithm_permits` at **480 B** just to
  check whether a given key is allowed to be used with a given algorithm.
  `psa_import_key` (218 B) + `psa_get_and_lock_key_slot_with_policy` (212 B)
  + `mbedtls_to_psa_error` (208 B) add up fast.

  **micro-ecc** has `uECC_verify(pk, hash, sig, curve)` — four raw
  pointers, zero state.

  ### 4. Generic hash abstraction (~3 KB)

  `mbedtls_internal_sha512_process` at **2,452 B** is the largest function
  in the entire binary. Plus **896 B** of rodata for the SHA-512 K constants
  (two copies at 640 B and 256 B).

  This is a generic-context SHA-384/512 that produces either output length
  from the same core with full update/finish ceremony.

  A standalone SHA-384 state machine in ~200 LoC would be ~1–2 KB combined.

  ### 5. Precomputed tables trade rodata for speed (~3 KB rodata)

  `secp384r1_T` is 768 B of precomputed multiples of the P-384 generator —
  an 8× scalar-mult speedup that mbedTLS opts into by default.

  For a bootloader that verifies once at boot, the ~50 ms wall-clock saved
  is not meaningful.

  **micro-ecc** uses inline doubling/addition without a precomputed table.

  ## The structural argument

  **Pruning cannot beat the generality floor.** Everything we disabled in
  the overlay (SSL, X.509, RSA, DH, ECDH, ciphers, MACs, KDFs, all other
  curves, all other hashes, entropy, RNG) either isn't compiled or gets
  `--gc-section`-stripped. What remains is the *irreducible* cost of
  mbedTLS PSA for one ECDSA-P384-verify + SHA-384-compute operation:

  - **~14 KB** is generic arithmetic (MPI 7 KB + generic ECC 5 KB + ECDSA
    wrapper 2 KB) — written to handle any curve, any width, any restart point
  - **~4 KB** is PSA service plumbing
  - **~5 KB** is generic hash (SHA-384/512 core + K tables + context machinery)
  - **~1 KB** is platform/utility

  This is *by design* a general-purpose crypto stack. It is excellent for
  an application tier that needs PSA's API guarantees, driver wrappers,
  opaque keys, certificate parsing, and policy enforcement. **It is
  architecturally wrong for a bootloader that does one operation.**

  ## What micro-ecc would look like instead

  | Component | Estimated size |
  |---|---:|
  | `uECC.c` with only `uECC_secp384r1` + verify: curve constants, point math, scalar mult, verify | ~5–6 KB |
  | Minimal SHA-384 (self-contained state machine) | ~1–2 KB |
  | MCUboot backend shim (DER→raw signature, SPKI→raw pubkey) | ~300–500 B |
  | **Total** | **~7–8 KB** |

  That is **~16–17 KB smaller than mbedTLS PSA for equivalent
  cryptographic function.** The delta is entirely spent on generality
  (arbitrary curves, arbitrary widths, service API, driver wrappers) that
  the bootloader will never use.

  ## The "strategic" re-framing

  The earlier argument was that mbedTLS is strategic because it is the
  ecosystem crypto stack. The measurement flips the framing:

  - **App side**: mbedTLS PSA is the right call. The app does ECDSA, ECDH,
    AES, and hashing across multiple key lifetimes, with certificate
    parsing. Generality earns its keep.
  - **Bootloader side**: the bootloader does one verify operation per boot.
    It has no cert chains, no key lifecycle, no algorithm negotiation.
    Generality is all cost and no benefit.

  Using micro-ecc in the bootloader **does not** mean diverging from
  mbedTLS on the app side — they can coexist cleanly. The app layer stays
  on PSA as a first-class citizen. The bootloader picks the right tool for
  a narrow, fixed job. This is good factoring, not a compromise.

  **The concrete payoff**: staying under 84 KB means **no partition
  resize**, which means **no mandatory J-Link reflash of fielded devices**
  — which is the blocker we have been working around.

JayToltTech:
Note that Claude called this MCUBoot 26.4.15 (which is my bootloader ver override) but this research is based on a lightly modified fork of 2.4.0RC1.

PR #2526, opened 2025-10-28
Add boot_state_init and fix boot_state_clear (and usage)

de-nordic:
PR adds `boot_state_init` that is intended to be used for boot_loader_state object initialization.
The `boot_state_init` is then used in places where  `boot_state_clear` has been previously used for state initialization and paired with `boot_state_clear` to cleanup state when it is no longer used.

### DIFF
--- a/boot/bootutil/src/bootutil_misc.c
+++ b/boot/bootutil/src/bootutil_misc.c
@@ -647,5 +647,18 @@ const struct image_max_size *boot_get_max_app_size(void)
  */
 void boot_state_clear(struct boot_loader_state *state)
 {
+#if defined(MCUBOOT_ENC_IMAGES)
+    int image;
+    int slot;
+
+    for (image = 0; image < BOOT_IMAGE_NUMBER; ++image) {
+        for (slot = 0; slot < BOOT_NUM_SLOTS; ++slot) {
+            /* Not using boot_enc_zeorize here, as it is redundant
+             * to the memset below that clears entire boot_loader_state.
+             */
+            boot_enc_drop(&state->enc[image][slot]);
+        }
+    }
+#endif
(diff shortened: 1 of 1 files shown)


boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -96,6 +96,14 @@ fih_ret context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp);
...
 struct boot_loader_state *boot_get_loader_state(void);
 
+/**
+ * Initialize boot_loader_state object
+ *
+ * @param state Bootloader state.
+ */
+void boot_state_init(struct boot_loader_state *state);
+
+
```
nordicjm:
```suggestion
```

nordicjm (approved):

boot/bootutil/src/bootutil_loader.c
```diff
@@ -381,3 +381,21 @@ boot_close_all_flash_areas(struct boot_loader_state *state)
...
+    int image;
+    int slot;
+#endif
+
+    memset(state, 0, sizeof(*state));
+
+#if defined(MCUBOOT_ENC_IMAGES)
+    for (image = 0; image < BOOT_IMAGE_NUMBER; ++image) {
+        for (slot = 0; slot < BOOT_NUM_SLOTS; ++slot) {
+            boot_enc_init(&state->enc[image][slot]);
```
nvlsianpu, line 397: whatever was pointed by state get zeroized already.

de-nordic: Yes, but the boot_enc_init may be doing extra stuff on the context, and may actually set here something different than 0.

nvlsianpu: good point

PR #2527, opened 2025-10-28
bootutil: Split header with encryption functions

de-nordic:
Split definitions to crypto backend specific headers.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
@@ -1,0 +1,77 @@
+/*
+ * This module provides a thin abstraction over some of the crypto
+ * primitives to make it easier to swap out the used crypto library.
+ *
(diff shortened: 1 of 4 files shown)


boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
```diff
@@ -0,0 +1,77 @@
...
+{
+    int rc;
+    rc = tc_aes128_set_encrypt_key(ctx, k);
+    if (rc != TC_CRYPTO_SUCCESS) {
+        return -1;
+    }
+    return 0;
+}
+
+static int _bootutil_aes_ctr_crypt(bootutil_aes_ctr_context *ctx, uint8_t *counter, const uint8_t *in, uint32_t inlen, uint32_t blk_off, uint8_t *out)
```
nordicjm: we should rename these, no functions should start with underscores, that is a compiler reserved token

PR #2532, opened 2025-11-03
espressif: fix UART init for Serial Recovery and unify adapter code

almir-okato:
Fix UART initialization for Espressif chips and unify their serial adapter code.


boot/espressif/port/serial_adapter.c
```diff
@@ -57,30 +76,70 @@
...
 #ifdef CONFIG_ESP_MCUBOOT_SERIAL_USB_SERIAL_JTAG
-    usb_serial_jtag_ll_txfifo_flush();
-    while (!usb_serial_jtag_ll_txfifo_writable()) {
-        MCUBOOT_WATCHDOG_FEED();
+    int write_len = 0;
+    while (cnt > 0) {
+        while (!usb_serial_jtag_ll_txfifo_writable()) {
+            MCUBOOT_WATCHDOG_FEED();
+        }
+        usb_serial_jtag_ll_write_txfifo((const uint8_t *)&str[write_len], 1);
```
sylvioalves: This sounds inefficient. Why sending 1 byte at a time?

```c
void console_write(const char *str, int cnt)
{
#ifdef CONFIG_ESP_MCUBOOT_SERIAL_USB_SERIAL_JTAG
    int remaining = cnt;
    int offset = 0;
    
    while (remaining > 0) {
        while (!usb_serial_jtag_ll_txfifo_writable()) {
            MCUBOOT_WATCHDOG_FEED();
        }
        
        int chunk_size = remaining;  // Perhaps using MIN/MAX here based on SoC FIFO.
        int written = usb_serial_jtag_ll_write_txfifo(
            (const uint8_t *)&str[offset], 
            chunk_size
        );
        
        offset += written;
        remaining -= written;
        
        usb_serial_jtag_ll_txfifo_flush();
        MCUBOOT_WATCHDOG_FEED();
    }
#else
    const uint8_t *data = (const uint8_t *)str;
    int remaining = cnt;
    
    while (remaining > 0) {
        uint32_t tx_len = uart_ll_get_txfifo_len(serial_boot_uart_dev);
        if (tx_len > 0) {
            uint32_t write_len = (tx_len < remaining) ? tx_len : remaining;
            uart_ll_write_txfifo(serial_boot_uart_dev, data, write_len);
            data += write_len;
            remaining -= write_len;
        }
        MCUBOOT_WATCHDOG_FEED();
    }
#endif
}
```

almir-okato: It was roughly based on serial_esp32_usb_poll_out from serial_usb_esp2_usb.c from Zephyr. I modified it with what you've suggested.

sylvioalves (approved):

PR #2549, opened 2025-11-17
boot: avoid increasing rollback counter during RAM load FW update

meremST:
When MCUBOOT_RAM_LOAD_REVERT is enabled, the rollback counter must not be incremented until image_ok is set. Otherwise, in the event of a revert, the boot process may become bricked.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2338,10 +2338,12 @@ boot_update_hw_rollback_protection(struct boot_loader_state *state)
     /* Update the stored security counter with the newer (active) image's
      * security counter value.
      */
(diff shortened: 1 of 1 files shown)


tomchy (approved):

d3zd3z (changes requested):
Unfortunately, I think this makes an existing vulnerability window happen in even more cases. The problem is that we can't wait until the next reboot to update the counter, or this can be tricked by not allowing the device to boot a second time after each update.

I think the right fix is to, for test boots, to have the security counter updated as part of "confirmation", not part of the normal boot. However, there are challenges to doing this, so I'm not quite sure what the right solution is.

PR #2553, opened 2025-11-19
fix(bootutil): correct copy size when bootstrapping and swapping with…

TomLier:
… MOVE

Previously, the copy size was calculated using the primary region size, which could be larger than the secondary region. This fix ensures that the size of the secondary region (excluding the swap sector) is used, preventing over-copying and related issues during image upgrade or bootstrap operations.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -990,6 +990,27 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
         size += this_size;
     }
 
+#if defined(MCUBOOT_SWAP_USING_MOVE)
+    /* When using MCUBOOT_SWAP_USING_MOVE, primary region is larger then the secondary region
+     * Optimal region configuration: # useful regions in primary region = # regions in secondary region + 1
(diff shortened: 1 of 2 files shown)


boot/bootutil/src/loader.c
```diff
@@ -990,6 +990,19 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
         size += this_size;
     }
 
+#if defined(MCUBOOT_SWAP_USING_MOVE)
+    /* When using MCUBOOT_SWAP_USING_MOVE, primary region is larger then the secondary region
```
nordicjm: adjust commit title to: `boot: bootutil: loader: Fix bootstrap copying in swap move mode`

nordicjm:
@TomLier we rebased to fix a CI change and fixed an incorrect define being used in an attempt to get this into the 2.3 release but it seems your changes break CI due to writing to an unerased page on https://github.com/mcu-tools/mcuboot/actions/runs/20139212331/job/57801476343?pr=2553 can you take a look at fix?

PR #2564, opened 2025-11-28
sim: Add Infineon PSOC Edge E8x device with the new "max-align-16" feature

Ta-ras:
- Add simulator support for Infineon PSOC Edge E81–E84 silicon, enforce a minimum write size of 16 bytes
- Add the "max-align-16" feature to the simulator and CI/CD test workflows


d3zd3z (commented):
I'm ok with this change, but I do think we should hold off merging until 2.3.0 is released.

PR #2577, opened 2025-12-05
bootutil: Avoid re-definition of MCUBOOT_SWAP_USING_SCRATCH

kkrentz:
Two identical definitions of `MCUBOOT_SWAP_USING_SCRATCH` occur in `bootutil_priv.h` and `bootutil_area.c`. This commit removes the duplicate in `bootutil_area.c`.

### DIFF
--- a/boot/bootutil/src/bootutil_area.c
+++ b/boot/bootutil/src/bootutil_area.c
@@ -27,6 +27,7 @@
  */
 
 #include "bootutil_area.h"
+#include "bootutil_priv.h"
 #include "bootutil/image.h"
 #include "bootutil/bootutil_public.h"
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/bootutil_area.c
```diff
@@ -5,6 +5,7 @@
  * Copyright (c) 2016-2019 JUUL Labs
  * Copyright (c) 2019-2020 Arm Limited
  * Copyright (c) 2025 Nordic Semiconductor ASA
+ * Copyright (c) 2025 Siemens AG
```
nordicjm: adding an include and removing some lines does not entitle you to add a copyright to this file

nordicjm (changes requested):

PR #2579, opened 2025-12-09
bootutil/crypto: Handle hashing errors and refactorings

kkrentz:
This PR complements invocations of hashing-related function with error handling. This is important in various occasions, such as when verifying the signature of an image in "pure" mode. In this case, an error during hashing may cause the computed hash digest to become the all-zero string. However, an arbitrary ECDSA signature over the all-zero hash digest will turn out authentic.

Also, this PR adds a new file called `sha.c`, which encapsulates shorthands for common invocation patterns of SHA and HMAC, as well as an HKDF implementation. This simplifies matters in other files and having HKDF at a reusable place helped us integrate TCG's Device Identifier Composition Engine with MCUboot, too. This is because DICE uses the "expand" subroutine of HKDF.

Besides, this PR makes HMAC available regardless of the crypto backend. Currently, only mbedTLS and TinyCrypt provide this function, thereby limiting firmware encryption to configurations with one of these crypto backends.


boot/bootutil/src/encrypted.c
```diff
@@ -380,9 +269,8 @@ int
 boot_decrypt_key(const uint8_t *buf, uint8_t *enckey)
 {
 #if defined(MCUBOOT_ENCRYPT_EC256) || defined(MCUBOOT_ENCRYPT_X25519)
-    bootutil_hmac_sha256_context hmac;
     bootutil_aes_ctr_context aes_ctr;
-    uint8_t tag[BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE];
+    uint8_t tag[IMAGE_HASH_SIZE];
     uint8_t shared[EC_SHARED_LEN];
     uint8_t derived_key[BOOT_ENC_KEY_SIZE + BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE];
```
d3zd3z, line 284: Is there a concern about still having the SHA256 digest size here?

d3zd3z (commented):
So, this will need to be rebased on a recent version to be workable. I apologize for missing it back when it was fresher.

Some concerns here:
- The included target where error handling might make sense (cc310) doesn't return status on it's hash functions. Presumably, this Pr is to support out of tree HW devices?
- The new `bootutil_sha_finish` doesn't have any testing. Admittedly, this is probably best checked by inspection, though.

boot/bootutil/src/encrypted.c
kkrentz: Thank you for your review. The new `bootutil_sha_hmac` creates tags of length `IMAGE_HASH_SIZE`. Nevertheless, we could leave `BOOTUTIL_CRYPTO_SHA256_DIGEST_SIZE` because the encryption and signature mechanism must share the same hash function. On the other hand, I now chose to declare `tag` more appropriately as `uint8_t tag[EC_TAG_LEN]`. Besides, I added an assertion to ensure that the encryption and signature mechanisms are sharing the same hash function.

kkrentz:
> * The included target where error handling might make sense (cc310) doesn't return status on it's hash functions. Presumably, this Pr is to support out of tree HW devices?

If Mbed TLS is configured to leverage HW acceleration, this might cause issues in MCUboot. However, I did not track this down. Anyway, I would like to have error handling to make proper use of the HW acceleration of my TI chips.

boot/bootutil/src/sha.c
```diff
@@ -0,0 +1,238 @@
...
+    rc = bootutil_sha_finish(&ctx, digest);
+    bootutil_sha_drop(&ctx);
+    return rc;
+}
+
+#if defined(MCUBOOT_USE_MBED_TLS)
+
+static const mbedtls_md_type_t md_type =
+#if defined(MCUBOOT_SHA512)
+        MBEDTLS_MD_SHA384;
```
d3zd3z: Shouldn't this be selecting MBEDTLS_MD_SHA512?

boot/zephyr/Kconfig
```diff
@@ -804,11 +812,13 @@ config BOOT_ENCRYPT_RSA
 
 config BOOT_ENCRYPT_EC256
 	bool
+	select SOMETHING_USES_HKDF
```
d3zd3z: These selects will need `BOOT_SOMETHING_USES_HKDF` to match the config above.

d3zd3z (changes requested):
There is no test coverage for the configurations describe, which is why the two mistakes didn't get caught.

boot/bootutil/src/sha.c
kkrentz: Thank you for the hint. I fixed that.

boot/zephyr/Kconfig
kkrentz: Thank you for the hint. I fixed that.

boot/bootutil/src/sha.c
```diff
@@ -0,0 +1,238 @@
...
+    struct hmac_context hmac_ctx;
+    uint8_t t_i[IMAGE_HASH_SIZE];
+    int rc;
+
+    if (okm_len > 255 * IMAGE_HASH_SIZE) {
+        return 1;
+    }
+    n = okm_len / IMAGE_HASH_SIZE + (okm_len % IMAGE_HASH_SIZE ? 1 : 0);
+
+    for (i = 1; i <= n; i++) {
```
d3zd3z: `i` here is `uint8_t`, but it the image is `255 * IMAGE_HASH_SIZE`, this loop will never terminate.

d3zd3z (changes requested):
Do the sim changes just support this, or is there any test coverage that say injects faults into the sha256 hashing.

kkrentz:
Testing the error handling paths would be a useful addition, but the test automation business looks intimidating to me.

d3zd3z (approved):
Even without direct tests, I think this is probably still useful to try to get in.

PR #2586, opened 2025-12-18
bootutil: mynewt: Imporve BOOTUTIL configuration

m-gorecki:
This:
1. Removes a variable with external linkage which is only used in imgmgr but it's never set anywhere so it's always 0.
2. Makes bootutil_public.c the only source file built for non-bootloader builds, so we don't have to set bootutil syscfgs that are relevant only to the bootloader when building an application.


m-gorecki:
@de-nordic Could you please explain how this affects mbed platform? This only modifies files from bootutil package

PR #2590, opened 2025-12-22
samples: zephyr: Remove hello-world

nordicjm:
This sample is very outdated and there has been a sysbuild MCUboot sample in zephyr for a number of years now


nordicjm:
> So, this used to be how I tested mcuboot upgrades. I guess the real question is, does any of this directory belong? I'm not sure I see the point in removing just the hello-world part. Or should the `samples/zephyr` scripts and such just be configured to build something out of the Zephyr tree?

There is https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/sysbuild/with_mcuboot which includes an application and MCUboot and is tested in zephyr's CI

d3zd3z (approved):
This is indeed dated, and redundant with tests already in Zephyr.

nandojve (approved):

PR #2591, opened 2025-12-25
Support resolving relative key file path via WEST_TOPDIR, as supported by Zephyr

silas-xing:
Zephyr supports  resolving relative key file path via WEST_TOPDIR, refer to https://github.com/zephyrproject-rtos/zephyr/blob/23354f3cad32cf77ac7ca9fc82450b183601277e/cmake/mcuboot.cmake#L62

```c
      if(NOT IS_ABSOLUTE "${${file}}")
        if(EXISTS "${APPLICATION_CONFIG_DIR}/${${file}}")
          set(${file} "${APPLICATION_CONFIG_DIR}/${${file}}")
        else()
          # Relative paths are relative to 'west topdir'.
          #
          # This is the only file that has a relative check to topdir likely
          # from the historical callouts to "west" itself before using
          # imgtool. So, this is maintained here for backward compatibility
          #
          if(NOT WEST OR NOT WEST_TOPDIR)
            message(FATAL_ERROR "Can't sign images for MCUboot: west workspace undefined. "
                                "To fix, ensure `west topdir` is a valid workspace directory.")
          endif()
          set(${file} "${WEST_TOPDIR}/${${file}}")
        endif()
      endif()
```

But resolving with `APPLICATION_CONFIG_DIR` seems not correct, because in my test:
- In zephyr's code above, `APPLICATION_CONFIG_DIR` value is `source_dir` specified in west arguments
- In mcuboot's code in the PR, `APPLICATION_CONFIG_DIR` value is `bootloader/mcuboot/boot/zephyr`
They are not match(If my test is correct)


boot/zephyr/CMakeLists.txt
```diff
@@ -347,6 +347,8 @@ if(NOT CONFIG_BOOT_SIGNATURE_TYPE_NONE AND NOT CONFIG_BOOT_SIGNATURE_KEY_FILE ST
     set(signing_key_file ${key_file})
   elseif(EXISTS ${APPLICATION_CONFIG_DIR}/${key_file})
     set(signing_key_file ${APPLICATION_CONFIG_DIR}/${key_file})
+  elseif((DEFINED WEST_TOPDIR) AND (EXISTS ${WEST_TOPDIR}/${key_file}))
```
nordicjm, line 350: Not acceptable, what you mention is the signing for the application which is relative to the application. Instead of doing this, use sysbuild and use an escaped path e.g. `\${APP_DIR}/my.pem`

nordicjm (changes requested):

boot/zephyr/Kconfig
```diff
@@ -393,11 +393,11 @@ config BOOT_SIGNATURE_KEY_FILE
...
-	  The key file will be parsed by imgtool's getpub command and a .c source
-	  with the public key information will be written in a format expected by
-	  MCUboot.
+	  from the APPLICATION_CONFIG_DIR or WEST_TOPDIR directory. If the key file
+	  is not there, the build system uses relative path that starts from the
+	  MCUBoot repository root directory. The key file will be parsed by imgtool's
+	  getpub command and a .c source with the public key information will be written in
+	  a format expected by MCUboot.
 
 	  Note: In configuration files, escaped CMake variables can be used to refer to paths
```
nordicjm, line 402: it's also described here

boot/zephyr/CMakeLists.txt
silas-xing: Thanks! \${APP_DIR}/my.pem is better solution for me, I will close the PR

PR #2592, opened 2026-01-08
Feat/image binding

maulik-arm:
Binding is a post-verification step that shifts authentication from asymmetric signature checks to symmetric MAC-based checks by embedding authorisation into the image metadata. Once bound, future boots can validate images with a MAC instead of re-running the more expensive asymmetric operation.

### DIFF
--- a/boot/bootutil/src/encrypted.c
+++ b/boot/bootutil/src/encrypted.c
@@ -662,7 +662,7 @@ boot_enc_load(struct boot_loader_state *state, int slot,
         return -1;
     }
 
-    return boot_decrypt_key(buf, bs->enckey[slot], state->curr_img_idx);
+    return boot_decrypt_key(buf, bs->enckey[slot], BOOT_CURR_IMG(state));
 }
 
 int


boot/bootutil/include/bootutil/image.h
```diff
@@ -145,6 +145,10 @@ extern "C" {
                                              * ...
                                              * 0xffa0 - 0xfffe
                                              */
+
+#define IMAGE_TLV_BIND_METADATA  0xA4   /* Image binding metadata placeholder TLV */
+#define IMAGE_TLV_BIND_TAG       0xA5   /* Image binding TAG placeholder TLV */
+
```
de-nordic, line 151: I do not think that vendor TLVs should be registered here.
Ranges are, but they are free to use by vendors which means you might have just collided with somebody using these in their code.
I do not think vendor TLVs should be processed nor exist in common code.
Also, if they are vendor specific, nothing indicates they are such.
It seems like this is regular feature that should have been in common code, but for some reason got here.

PR #2614, opened 2026-01-29
boot:zephyr:boards: Add configuration for nucleo_n657x0_q

zacck:
Add mcuboot configuration for nucleo_n657x0_q.


erwango (commented):
This shouldn't be required. You should be able to rely on: https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/socs/stm32n657xx_fsbl.conf

PR #2616, opened 2026-01-29
sim: init mbedtls allocator for KW

ledvinap:
- The KW unwrap path (bootloader / bootutil_aes_kw_unwrap) uses mbedtls and assumes allocators are already initialized.
- Internal sim images call the KW wrap helper (kw_encrypt_) before the bootloader runs.
- That helper calls mbedtls_platform_set_calloc_free(calloc, free), so the allocator is initialized.

Skipping the wrap step makes bootutil_aes_kw_unwrap fail.


ledvinap:
Here is code that triggered this bug:
https://github.com/mcu-tools/mcuboot/compare/main...ledvinap:mcuboot:sim-imgtool-kw

d3zd3z (commented):
The change itself looks good, but rebasing will be necessary. I don't think the failure was caused by this change, so hopefully that will be better after the rebase.

PR #2622, opened 2026-02-04
imgtool: Output format option to dumpinfo

5frank:
Add -f, --format option to support multiple output formats (human, yaml,
json). Default yaml for file, default human for stdout - backward
compatible.

Details:
- Human output is unchanged
- Refactored dump_imginfo() to separate data reading from formatting
- JSON output uses hex strings for binary data (e.g., "e3a333ca...")
- YAML and JSON have both "type" (int) and "type_name" (str) for TLVs
- Image filename (basename) is included in all output formats
- Argument `--silent` is silently ignored. Doesn't make sense for a
  command that outputs data. i.e. No output on success. This is the same
  as for other sub commands

### DIFF
--- a/scripts/imgtool/dumpinfo.py
+++ b/scripts/imgtool/dumpinfo.py
@@ -17,8 +17,10 @@
 """
 Parse and print header, TLV area and trailer information of a signed image.
 """
+import json
 import os.path
 import struct
+import sys
 
 import click
 import yaml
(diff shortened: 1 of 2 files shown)


nordicjm:
@fundakol please review

nordicjm:
@5frank see failing CI

PR #2633, opened 2026-02-18
imgtool: add --custom-tlv-file option

jonathannilsen:
Add a new option --custom-tlv-file that works exactly like --custom-tlv, but takes a binary file to read the TLV data from instead of taking the data directly on the command line. This makes it easier to provide custom TLV data in certain scenarios, e.g. if the TLV data is stored on disk or is generated from a different build artifact.


fundakol (approved):
LGTM

Issue #2596, opened 2026-01-13
Migrate MCUboot Zephyr USB support to USB_DEVICE_STACK_NEXT before Zephyr v4.5.0


JayToltTech:
Problem

  MCUboot's Zephyr USB serial recovery and USB DFU features currently use the legacy USB_DEVICE_STACK API, which has been deprecated in Zephyr and is scheduled for removal in Zephyr v4.5.0.

  Current Impact (Zephyr v4.3.0)

  Building MCUboot with USB features enabled generates numerous deprecation warnings:

  warning: 'usb_enable' is deprecated [-Wdeprecated-declarations]
  warning: 'usb_disable' is deprecated [-Wdeprecated-declarations]
  warning: 'usb_dc_ep_write' is deprecated [-Wdeprecated-declarations]
  warning: 'usb_dc_ep_read' is deprecated [-Wdeprecated-declarations]
  ... (50+ similar warnings from zephyr/subsys/usb/device/)

  These warnings occur throughout Zephyr's USB subsystem code when compiling MCUboot with:
  - CONFIG_BOOT_SERIAL_CDC_ACM=y (USB serial recovery)
  - CONFIG_BOOT_USB_DFU_WAIT=y (USB DFU mode)

  Timeline

  - Zephyr v4.3.0 (current): Legacy USB stack deprecated, warnings generated
  - Zephyr v4.5.0 (upcoming): Legacy USB stack will be removed entirely
  - Result: MCUboot's USB features will break unless migrated

  Workaround Currently in Place

  Zephyr has a temporary workaround that prevents the new stack from being used with MCUboot:

  # boards/common/usb/Kconfig.cdc_acm_serial.defconfig
  config USB_DEVICE_STACK_NEXT
      default y if !MCUBOOT  # Explicitly disabled when MCUboot is present

  This confirms the Zephyr team is aware of the incompatibility but cannot fix it directly since MCUboot is a separate project.

  Proposed Solution

  Migrate MCUboot's USB implementation to use the new USB_DEVICE_STACK_NEXT API.

  API Changes Required

  Files to modify:
  - boot/zephyr/main.c - 3 locations using usb_enable() / usb_disable()
  - boot/zephyr/serial_adapter.c - 1 location using usb_enable()

  Legacy API:
  usb_enable(NULL);
  usb_disable();

  New API:
  struct usbd_context *uds_ctx = /* context */;
  usbd_init(uds_ctx);
  usbd_enable(uds_ctx);
[...]

PR #2634, opened 2026-02-19
boot: zephyr: Migrate BOOT_SERIAL_CDC_ACM to USB_DEVICE_STACK_NEXT

JayToltTech:
This change includes a series of updates I found necessary to accomplish the goal.  It is part 1 of 2 for moving to USB_DEVICE_STACK_NEXT (see #2596), the next PR will include updates to DFU.  Changes include:

- Moving away from CDC_ACM_SERIAL_INITIALIZE_AT_BOOT to initializing USBD in boot_uart_fifo_init so that USBD is not initialized unless needed by serial recovery

- A few fixes to the logging infrastructure so that logging messages were preserved and flushed, while still ensuring if logging to uart is used it does not interfere with serial recovery itself

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -470,12 +470,15 @@ void boot_log_thread_func(void *dummy1, void *dummy2, void *dummy3)
 
 void zephyr_boot_log_start(void)
 {
-    /* start logging thread */
+    /* start logging thread immediately — the 30ms polling interval is only
+     * used between processing rounds, not as a startup delay. On fast boots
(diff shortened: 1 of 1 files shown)


JayToltTech:
I have run a set of tests against my own board which is an nRF53 based system which uses SWO for logging.  I have a set of follow-on PRs which I will submit independently around logging and watchdog and serial recovery performance changes that I found as part of testing this functionality, but they are not necessary for the base CDC-ACM port.

PR #2635, opened 2026-02-19
boot: zephyr: migrate CDC ACM serial recovery to USB_DEVICE_STACK_NEXT

JayToltTech:
## Summary

- Replace the legacy USB device stack with `USB_DEVICE_STACK_NEXT` for CDC-ACM serial recovery
- Initialize USB lazily in `boot_uart_fifo_init()` only when serial recovery is triggered, avoiding interference with application USB on normal boots
- Add self-contained USBD CDC ACM module (`usbd_cdc_serial.c/.h`) with configurable USB descriptors (VID, PID, manufacturer, product, self-powered, max power) via new Kconfig options
- Feed the watchdog while waiting for USB host connection to prevent timeout before the serial recovery loop is reached
- Enlarge nrf52840 boot partitions to accommodate the larger USB stack
- Remove stale legacy USB Kconfig options from board configs
- Add flash-saving options (disable BOS, vendor requests, deferred messages) to recovery conf files

Supersedes #2634 (narrowed scope to CDC-ACM only, removed unrelated logging/SWO changes).

Builds on the approach from #2346 with key improvements:
- Lazy USB init instead of `CDC_ACM_SERIAL_INITIALIZE_AT_BOOT`
- Watchdog feeding during USB host wait
- Dongle partition overlay for the larger USB stack
- Direct context accessor instead of fragile `STRUCT_SECTION_GET` index

## Test plan

- [x] Build nrf52840dk with `usb_cdc_acm_log_recovery.conf` (92.49% flash)
- [x] Build nrf52840dongle with `usb_cdc_acm_recovery.conf` (99.71% flash)
- [x] Functional test: enter serial recovery, upload image via mcumgr over USB CDC-ACM
- [x] Verify normal boot path does not initialize USB


boot/zephyr/serial_adapter.c
```diff
@@ -246,5 +237,41 @@ boot_uart_fifo_init(void)
...
+	}
+
+	/* Feed WDT while waiting for USB host to open the serial port.
+	 * Without this, the hardware watchdog fires before the serial
+	 * recovery loop (which feeds the WDT) is reached. */
+	while (k_sem_take(&boot_cdc_acm_ready, K_MSEC(5000)) != 0) {
+#if CONFIG_BOOT_WATCHDOG_FEED
+		MCUBOOT_WATCHDOG_FEED();
+#endif
+	}
```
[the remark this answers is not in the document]
JayToltTech: Replaces an infinite wait.  Indefinite hang is preexisting behavior.

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +47,51 @@ config BOOT_SERIAL_CDC_ACM
...
+	depends on BOOT_SERIAL_CDC_ACM
+	help
+	  USB Vendor ID presented during serial recovery mode.
+
+config BOOT_SERIAL_CDC_ACM_PID
+	hex "USB Product ID for CDC ACM serial recovery"
+	default 0x0004
+	depends on BOOT_SERIAL_CDC_ACM
+	help
+	  USB Product ID presented during serial recovery mode.
```
nordicjm, line 63: should probably have notes that these are intended for sample demonstration usage only and MUST be set properly for applications

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +47,51 @@ config BOOT_SERIAL_CDC_ACM
 
 endchoice
 
+config BOOT_SERIAL_CDC_ACM_VID
```
nordicjm, line 52: put an if here instead of having depends on on all symbols

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +47,51 @@ config BOOT_SERIAL_CDC_ACM
...
+	help
+	  USB product string descriptor for serial recovery mode.
+
+config BOOT_SERIAL_CDC_ACM_SELF_POWERED
+	bool "USB self-powered attribute for CDC ACM serial recovery"
+	depends on BOOT_SERIAL_CDC_ACM
+	help
+	  Set the Self-powered attribute in the USB configuration descriptor.
+	  When enabled, the device reports that it does not draw power from
+	  the USB bus. Only enable this if the device is powered by an
```
nordicjm, line 83: actually bit confused on all these options, they already exist in zephyr? `CDC_ACM_SERIAL_SELF_POWERED`, `CDC_ACM_SERIAL_MANUFACTURER_STRING`, `CDC_ACM_SERIAL_PRODUCT_STRING`, `CDC_ACM_SERIAL_VID`, `CDC_ACM_SERIAL_PID`?

nordicjm (commented):
think there is some confusion here, the CDC ACM porting work was already done in https://github.com/mcu-tools/mcuboot/pull/2346/files, that commit should be used **as-is**, the additional part that is needed is USB DFU support which needs to be ported (comments below were left before posting this message, so can be ignored, just pull in the existing commit from that PR, which I assume works?)

JayToltTech:
Thank you for the review @nordicjm .

> think there is some confusion here, the CDC ACM porting work was already done in https://github.com/mcu-tools/mcuboot/pull/2346/files, that commit should be used **as-is**, the additional part that is needed is USB DFU support which needs to be ported (comments below were left before posting this message, so can be ignored, just pull in the existing commit from that PR, which I assume works?)

I would recommend against that.  There are a number of issues with that PR which I fixed when cherry picking those files into this branch, including:

- Enlarge nrf52840 dongle mcuboot partition to accommodate the larger USB stack
- Call usbd_disable on exiting main (this surfaced a silent failure returning success in nrfx_power_usbevt_enable, see https://github.com/zephyrproject-rtos/zephyr/issues/104190)
- Add lazy initialization of USBD so the USB stack is not enabled and disabled every time MCUBoot runs.  The 'happy path' that occurs most of the time is that MCUBoot runs and finds that image slot 0 is valid and ready to execute, and we only need to initialize USBD if we need to enter recovery/dfu mode.

And probably one of two other small things I fixed along the way.

I will address the rest of your feedback once we have consensus that we are going to move this PR forward instead of go back to #2346.

boot/zephyr/Kconfig.serial_recovery
henrikbrixandersen: It is correct that these are defined in https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/device_next/app/Kconfig.cdc_acm_serial, but I am not sure if these are to be used outside the Zephyr tree? The comment in https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/device_next/app/cdc_acm_serial.c seems to indicate that their usage is limited:

> ```
> /*
>  * This is intended for use with cdc-acm-snippet or as a default serial backend
>  * only in applications where no other USB features are required, configured,
>  * and enabled. This code only registers the first CDC-ACM instance.
>  */
> ```

henrikbrixandersen: When it comes to USB DFU, MCUboot will have to define similar Kconfigs on its own anyway (and reviewing this without seeing how these two "sets" of Kconfig options will co-exist is difficult at best).

JayToltTech: Right, if you look in that source, you see that `CDC_ACM_SERIAL_SELF_POWERED` is conditionally included `if CDC_ACM_SERIAL_INITIALIZE_AT_BOOT`.

This design does not use `CDC_ACM_SERIAL_INITIALIZE_AT_BOOT` because that causes the USBD enable and disable code to be run, even in the happy case when mcuboot runs and sees that slot0 is valid and recovery/dfu are not needed.

JayToltTech: We could start with a refactoring of Zephyr's USBD CDC_ACM_SERIAL code, in order to cleanly support lazy initialization better, but...

henrikbrixandersen: I honestly think MCUboot should implement this on the application layer itself.

nordicjm: The description seems apt for MCUboot, it would only use one instance of it, but self powered being gated on initialising at boot seems a bit of a random condition

JayToltTech:
@nordicjm I went ahead and addressed your prior round of review feedback.  I will pause here and will join the mcuboot tech forum call next Thursday to be available for discussion.

nvlsianpu:
> It looks that when `CONFIG_BOOT_SERIAL_WAIT_FOR_DFU` is enabled, `boot_console_init` is blocked (and MCUboot) until the USB CDC port is opened... I was expecting to timeout after `CONFIG_BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT` and the boot process to continue...

Unfortunately initialization of usbd stacks while USB is unattached (by zephyr-rtos implementation).

nvlsianpu (commented):
LGTM

CONFIG_BOOT_SERIAL_WAIT_FOR_DFU should be unelectable with this implementation.

docs/readme-zephyr.md
```diff
@@ -220,6 +220,20 @@ Which interface belongs to the protocol shall be set by the devicetree-chosen no
...
+The USB device descriptors can be configured with the following Kconfig options:
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_VID`` | ``0x2fe3`` | USB Vendor ID |
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_PID`` | ``0x0004`` | USB Product ID |
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_MANUFACTURER_STRING`` | ``"Zephyr Project"`` | Manufacturer string descriptor |
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_PRODUCT_STRING`` | ``"CDC ACM serial recovery"`` | Product string descriptor |
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_SELF_POWERED`` | ``n`` | Set the self-powered attribute in the USB configuration descriptor |
+| ``CONFIG_BOOT_SERIAL_CDC_ACM_MAX_POWER`` | ``125`` | ``bMaxPower`` value in the USB configuration descriptor, in 2 mA units (default: 250 mA) |
```
josuah, line 235: Kconfigs are often documented in the Kconfig files, so might not be needed to repeat them here.

boot/zephyr/serial_adapter.c
```diff
@@ -246,5 +237,41 @@ boot_uart_fifo_init(void)
...
+
+	rc = usbd_enable(boot_usb_cdc_serial_get_context());
+	if (rc) {
+		return (-1);
+	}
+
+	/* Feed WDT while waiting for USB host to open the serial port.
+	 * Without this, the hardware watchdog fires before the serial
+	 * recovery loop (which feeds the WDT) is reached. */
+	while (k_sem_take(&boot_cdc_acm_ready, K_MSEC(5000)) != 0) {
```
josuah, line 257: alternative, replicate this:
https://github.com/zephyrproject-rtos/zephyr/blob/12759875f6181cf3a3b520bae77f1065c0b859fd/samples/subsys/usb/common/sample_usbd_init.c#L193-L210

This would assume a global uds_ctx rather than `boot_usb_cdc_serial_get_context()` maybe.

Just a matter of having similar boilerplate across projects to reduce maintainance.

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +47,51 @@ config BOOT_SERIAL_CDC_ACM
...
+config BOOT_SERIAL_CDC_ACM_VID
+	hex "USB Vendor ID for CDC ACM serial recovery"
+	default 0x2fe3
+	help
+	  USB Vendor ID presented during serial recovery mode.
+		You must use your own VID for samples and applications outside of Zephyr Project.
+
+config BOOT_SERIAL_CDC_ACM_PID
+	hex "USB Product ID for CDC ACM serial recovery"
+	default 0x0004
```
josuah, line 61: I think this for the samples only
https://docs.zephyrproject.org/latest/services/connectivity/usb/device_next/vid_pid.html

This would need its own VID:PID, which currently piggy-back on Nordic vendor ID.

maybe @jfischer-no and @tmon-nordic know about it?

boot/zephyr/main.c
```diff
@@ -189,6 +198,21 @@ static void do_boot(struct boot_rsp *rsp)
...
+        usbd_rc = usbd_disable(boot_usb_cdc_serial_get_context());
+
+        /* -EALREADY is expected on normal boot: USB was never enabled
+         * (lazy init -- only initialized when recovery is triggered).
+         * Any other error indicates a real problem.
+         */
+        if (usbd_rc != 0 && usbd_rc != -EALREADY) {
+            BOOT_LOG_WRN("USB disable failed: %d", usbd_rc);
+        }
+    }
```
josuah, line 214: How about using `rc` instead of creating a new scope `{}` and introducing `usbd_rc;`?

docs/readme-zephyr.md
```diff
@@ -220,6 +220,20 @@ Which interface belongs to the protocol shall be set by the devicetree-chosen no
 - `zephyr,console` - If a hardware serial port is used.
 - `zephyr,cdc-acm-uart` - If a virtual serial port is used.
 
+When ``CONFIG_BOOT_SERIAL_CDC_ACM=y`` is selected, MCUboot manages the USB device stack
+internally and only initializes it when serial recovery is actually triggered — not on every boot.
+This avoids consuming hardware USB events that the application would otherwise receive.
+The USB device descriptors can be configured with the following Kconfig options:
```
josuah, line 226: This sounds more like a design decision than an user guide/readme.

It is often easier to inspect this directly in the source which is never out of date as opposed to comments, rarely updated.

How about removing it?

boot/zephyr/serial_adapter.c
```diff
@@ -246,5 +237,41 @@ boot_uart_fifo_init(void)
...
+	}
+
+	rc = usbd_enable(boot_usb_cdc_serial_get_context());
+	if (rc) {
+		return (-1);
+	}
+
+	/* Feed WDT while waiting for USB host to open the serial port.
+	 * Without this, the hardware watchdog fires before the serial
+	 * recovery loop (which feeds the WDT) is reached. */
```
josuah, line 256: Somewhat self-explanatory given what's below, but not inaccurate either.

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +47,51 @@ config BOOT_SERIAL_CDC_ACM
...
+	  When enabled, the device reports that it does not draw power from
+	  the USB bus. Only enable this if the device is powered by an
+	  independent supply during serial recovery mode.
+
+config BOOT_SERIAL_CDC_ACM_MAX_POWER
+	int "USB bMaxPower value for CDC ACM serial recovery"
+	default 125
+	range 0 250
+	help
+	  bMaxPower value in the configuration in 2 mA units.
```
josuah, line 91: How about keeping the unit in mA directly and make the division in the `.c`?

Assert might/might not be needed as 21 mA turning into 20 mA might be ok.

boot/zephyr/usbd_cdc_serial.c
```diff
@@ -0,0 +1,180 @@
...
+	int err;
+
+	err = usbd_add_configuration(&boot_usbd, USBD_SPEED_FS,
+				     &boot_usbd_fs_config);
+	if (err) {
+		BOOT_LOG_ERR("Failed to add FS configuration: %d", err);
+		return err;
+	}
+
+	err = usbd_register_class(&boot_usbd, "cdc_acm_0", USBD_SPEED_FS, 1);
```
josuah, line 71: This could be `usbd_register_all_classes()` to simplify.

This is probably a good starting point as helps co-maintaining the boilerplate across Zephyr and this.

It does not need to be different from the "sample-only" example I believe.

josuah (commented):
In general, feedback is to try to make it a bit closer to upstream's sample just to reduce maintenance. For instance, when Zephyr is updated, update this as well.

Otherwise, seems to me that it works.

Thank you for the initiative!

kartben:
Revived the work from this PR in #2849 since it's been a bit quiet on this one for a bit. @JayToltTech your work is very much attributed there, so no worries about that!

PR #2638, opened 2026-02-20
boot: bootutil: feed watchdog during swap operations

JayToltTech:
## Problem

Flash swap was exceeding 30 seconds on my device.  I had set mcuboot to have a watchdog set to 30 second reset (consistent with my app behavior).  Lack of watchdog feed during flash swap caused a watchdog reset to occur.  56d28f0c started this work with adding watchdog feed in serial recovery, adjusting flash swap to match. 

## Summary

- Add `MCUBOOT_WATCHDOG_FEED()` to swap loops in `swap_offset.c`, `swap_move.c`, and `swap_scratch.c`
- On devices with many flash sectors or slow flash, swap operations can exceed the watchdog timeout and reset the device mid-swap
- The existing `loader.c` copy loop already feeds the watchdog; this closes the gap in the swap implementations

## Test plan

- [x] Build with `CONFIG_BOOT_WATCHDOG_FEED=y` and verify swap completes without watchdog reset

### DIFF
--- a/boot/bootutil/src/swap_offset.c
+++ b/boot/bootutil/src/swap_offset.c
@@ -682,6 +682,7 @@ void swap_run(struct boot_loader_state *state, struct boot_status *bs,
                                          (mirror_idx > used_sectors_sec ? true : false));
             }
 
+            MCUBOOT_WATCHDOG_FEED();
             idx++;
         }
 
@@ -702,6 +703,7 @@ void swap_run(struct boot_loader_state *state, struct boot_status *bs,
                                   (idx > used_sectors_sec ? true : false));
             }
 
+            MCUBOOT_WATCHDOG_FEED();
             idx++;
         }
     }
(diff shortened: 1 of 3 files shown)


boot/bootutil/src/swap_offset.c
```diff
@@ -682,6 +682,7 @@ void swap_run(struct boot_loader_state *state, struct boot_status *bs,
                                          (mirror_idx > used_sectors_sec ? true : false));
             }
 
+            MCUBOOT_WATCHDOG_FEED();
```
nordicjm, line 685: not following how this helps, the watchdog feed function is called twice in the above function when the `boot_copy_region` function is called, this isn't actually doing anything

boot/bootutil/src/swap_scratch.c
```diff
@@ -935,6 +935,7 @@ swap_run(struct boot_loader_state *state, struct boot_status *bs,
             boot_swap_sectors(first_sector_idx, sz, state, bs);
         }
 
+        MCUBOOT_WATCHDOG_FEED();
```
nordicjm, line 938: as above

boot/bootutil/src/swap_offset.c
JayToltTech: Excellent point.  I missed the call to MCUBOOT_WATCHDOG_FEED which is embedded in boot_copy_region in all three cases.

Thank you!

PR #2640, opened 2026-02-21
docs: add serial recovery throughput tuning guide

JayToltTech:
## Summary
- Document the Kconfig options that control serial recovery upload throughput (`CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE`, `CONFIG_BOOT_MAX_LINE_INPUT_LEN`, `CONFIG_BOOT_LINE_BUFS`)
- Show how to pass a matching MTU in the `mcumgr` connstring
- Include measured upload times on nRF52840 over USB CDC-ACM (1 KB/s default vs 4 KB/s with mtu=4096)

### DIFF
--- a/docs/serial_recovery.md
+++ b/docs/serial_recovery.md
@@ -54,6 +54,48 @@ As a result, a device can receive images smoothly, and can erase required part o
 How to enable and configure the serial recovery feature depends on the given mcuboot-port implementation.
 Refer to the respective documentation and source code for more details.
 
+## Throughput tuning
+
+Upload throughput is determined by how much image data fits in each SMP
+round-trip. Three Kconfig options control the receive-side buffer on the device
+(Zephyr port):
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE`` | 1024 | Maximum reassembled SMP frame size (bytes) |
+| ``CONFIG_BOOT_MAX_LINE_INPUT_LEN`` | 128 | SMP serial line fragment size (bytes) |
+| ``CONFIG_BOOT_LINE_BUFS`` | 8 | Number of fragment reassembly buffers |
+
+The product ``CONFIG_BOOT_MAX_LINE_INPUT_LEN × CONFIG_BOOT_LINE_BUFS`` should be at least
+``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE``. With the defaults, each upload request carries
+roughly 228 bytes of image data after base64 and CBOR overhead.
+
+Increasing the receive buffer allows the host to send larger chunks per
+round-trip. For example, setting ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE=4096``
+and ``CONFIG_BOOT_MAX_LINE_INPUT_LEN=512`` allows approximately 3 KB of image
+data per round-trip at the cost of roughly 6 KB additional RAM (the ``in_buf``
+and ``dec_buf`` static arrays in ``boot_serial.c``).
+
+On the host side, pass a matching MTU in the ``mcumgr`` connstring:
+
+``` console
+mcumgr --conntype serial --connstring "dev=/dev/ttyACM0,baud=115200,mtu=4096" image upload <image>
+```
+
+The following table shows approximate upload times for a 374 KB image
+over USB CDC-ACM (tested on nRF52840 with
+[apache/mynewt-mcumgr-cli@5c56bd2](https://github.com/apache/mynewt-mcumgr-cli/commit/5c56bd24066c)):
+
+| Configuration | Throughput | Approximate time |
+|---------------|-----------|------------------|
+| Default (no MTU) | ~1 KB/s | ~6 minutes |
+| mtu=4096 | ~4 KB/s | ~1.5 minutes |
+
+**Note:** When using USB CDC-ACM as the serial transport, the baud rate
+parameter is ignored because USB bulk transfers are not rate-limited by a
+physical UART clock. The baud rate setting only affects physical UART
+transports.
+
 ## Entering serial recovery mode
 
 Entering the serial recovery mode is usually possible right after a device reset, for instance as a reaction on a GPIO pin state.


docs/serial_recovery.md
```diff
@@ -54,6 +54,48 @@ As a result, a device can receive images smoothly, and can erase required part o
 How to enable and configure the serial recovery feature depends on the given mcuboot-port implementation.
 Refer to the respective documentation and source code for more details.
 
+## Throughput tuning
```
nordicjm, line 57: This is a generic MCUboot doc page, the parts added here are zephyr specific, they cannot go on this page and would need to go on a separate page or under a separate heading - thoughts on which @de-nordic @d3zd3z ?

JayToltTech: Yes and no.  Yes, these are Zephyr KConfig values.  No, in that the connection string sample is specific to MCUMgr so that MCUMgr image upload performs better.

The performance measurements are specific to mcumgr/cbor encoded serial streams sending image upload frames, and while cbor isn't a huge overhead here, I wouldn't represent the measurements as 'generic CDC-ACM performance'.  I also noticed that the throughput started high and then stagnated as the size of the transmitted data increased, so I suspect that the default buffer settings may be tuned low ram / good performance of bursty small transmissions.

If you are developing a custom CDC-ACM transmission app, I imagine you are going to pay attention and analyze the buffer settings.  If you are coming to MCUBoot and flipping on a serial recovery kconfig, you may assume the default settings are tuned for reasonable performance.  In my case, I found my image upload went from 6 minutes to 90 seconds with these changes, so there's a pretty big performance delta.  On the other hand, these settings increase the RAM usage of mcuboot by 6k and with all the talk around optimization and fitting on smaller MCUs, I didn't think it was wise to add this increase.

Documenting this, in the mcumboot serial recovery doc, under the 'Configuration of serial recovery' section, seemed like the most prudent approach.  Especially because if it's in a generic 'Zephyr tuning CDC-ACM' document, it's not going to be terribly discoverable or identifiable as applicable to your standard 'welcome to mcuboot' customer.

JayToltTech:
@nordicjm Per conversation on the last MCUBoot call, I have moved the doc edits into docs/readme-zephyr.md.

docs/readme-zephyr.md
```diff
@@ -250,6 +250,47 @@ The mapping of image number to partition is as follows:
...
 
+### Throughput tuning
+
+Upload throughput is determined by how much image data fits in each SMP
+round-trip. Three Kconfig options control the receive-side buffer on the device:
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE`` | 1024 | Maximum reassembled SMP frame size (bytes) |
+| ``CONFIG_BOOT_MAX_LINE_INPUT_LEN`` | 128 | SMP serial line fragment size (bytes) |
```
nordicjm:
```suggestion
```
this value should never be changed, in fact it needs to be removed, 128 is per the specification

nordicjm (changes requested):

docs/serial_recovery.md
d3zd3z: I think putting them in readme-zephyr.md makes sense here.

d3zd3z (approved):

de-nordic (approved):

docs/readme-zephyr.md
```diff
@@ -250,6 +250,49 @@ The mapping of image number to partition is as follows:
...
+|--------|---------|-------------|
+| ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE`` | 1024 | Maximum reassembled SMP frame size (bytes) |
+| ``CONFIG_BOOT_LINE_BUFS`` | 8 | Number of fragment reassembly buffers |
+
+The SMP serial line fragment size is fixed at 128 bytes per specification
+(``CONFIG_BOOT_MAX_LINE_INPUT_LEN``). The product
+``128 × CONFIG_BOOT_LINE_BUFS`` must be at least
+``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE``. With the defaults
+(8 buffers × 128 = 1024), each upload request carries roughly 228 bytes of
+image data after base64 and CBOR overhead.
```
nordicjm: this does not seem right, if you send 1024 bytes of data (total size after SMP over console) then you get the following in the first request (for a dummy file I have):
 - 4-byte Length
 - 32-byte SHA256 hash
 - 0 byte offset
 - 633 data bytes

so something is amiss if you are only get 228 bytes here

docs/readme-zephyr.md
```diff
@@ -250,6 +250,49 @@ The mapping of image number to partition is as follows:
...
+Increasing the receive buffer allows the host to send larger chunks per
+round-trip. For example, setting ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE=4096``
+and ``CONFIG_BOOT_LINE_BUFS=32`` allows approximately 3 KB of image data per
+round-trip at the cost of roughly 4 KB additional RAM (the ``in_buf`` static
+array in ``boot_serial.c``).
+
+On the host side, pass a matching MTU in the ``mcumgr`` connstring:
+
+``` console
+mcumgr --conntype serial --connstring "dev=/dev/ttyACM0,baud=115200,mtu=4096" image upload <image>
```
nordicjm: never use this garbage tool

nordicjm (changes requested):

docs/serial_recovery.md
```diff
@@ -54,6 +54,48 @@ As a result, a device can receive images smoothly, and can erase required part o
...
+
+Upload throughput is determined by how much image data fits in each SMP
+round-trip. Three Kconfig options control the receive-side buffer on the device
+(Zephyr port):
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| ``CONFIG_BOOT_SERIAL_MAX_RECEIVE_SIZE`` | 1024 | Maximum reassembled SMP frame size (bytes) |
+| ``CONFIG_BOOT_MAX_LINE_INPUT_LEN`` | 128 | SMP serial line fragment size (bytes) |
```
JPHutchins, line 66: We have discussed in Zephyr that the spec actually requires 128. Ideally the configuration options would be removed from MCUboot and Zephyr because it's misleading.

JPHutchins (commented):
I'm going to open a PR for adding mcumgr params so that clients can actually use this without having to be given the buffer settings. We do have a branch for it (on a zephyr smp server) but I really want to get it under emulated testing before releasing: https://github.com/intercreate/smpclient/pull/81

PR #2648, opened 2026-03-04
Interrupted-OTA swap-move coverage: sim torn-write test and CI canary

neilberkman:
MCUboot has had real interrupted-OTA regressions in swap-using-move revert handling where a power cut leaves the device stuck or booting the wrong image (#1966 / #2100, #2199). This PR adds coverage for that failure class at two layers.

## Simulator: torn-write power-loss testing

The simulator's power-loss testing interrupts `boot_go` between whole flash operations: when the flash counter expires, `flash_area_write` in `sim/mcuboot-sys/csupport/run.c` returns via longjmp before calling the simulated flash write, so every simulated power cut is write-atomic. That is why the revert interruption sweep did not catch #1966: that failure needs power loss inside `boot_write_magic` during swap-move revert fixup, leaving a partially written magic that reads back as `BOOT_MAGIC_BAD` and matches no swap table row.

The sim commit adds a torn-write mode:

- When the interruption counter expires in torn mode, the write programs all but the last byte of its payload before the simulated power loss. Write alignment is a driver interface requirement, not a power-fail atomicity guarantee, so the tear point falls inside the final write unit regardless of `max-align`.
- The simulated flash marks torn bytes as retryable: one subsequent write of the same values is allowed, since recovery replays the interrupted operation. The write-once discipline stays enforced everywhere else, with unit tests for the tear, the single replay, a mismatched replay, and erase restoring normal behavior.
- `run_revert_with_torn_writes` sweeps the revert path: perform a test upgrade, count the flash operations of an uninterrupted revert, then for each operation clone the post-upgrade state, tear that write during revert, and verify recovery to the original image with clean trailers. Gated to swap-using-move.

Validation: on current main, both the existing `revert_with_fails` sweep and the new torn-write sweep pass (`--features swap-move`, and `swap-move,max-align-32`). With the one-word fix from 4f393563 reverted, the new sweep fails at the fixup_revert magic write, including under `max-align-32`, while the existing whole-operation sweep still passes. That reproduces #1966 offline in the simulator.

## CI: post-merge canary on real firmware

The workflow builds MCUboot swap-using-move for nrf52840dk/nrf52840 against Zephyr main, using the same container image, setup action, and SDK as `zephyr_build.yaml`, signs test images with the RSA-2048 test key, and runs two power-loss fault-injection sweeps under Renode via [tardigrade](https://github.com/neilberkman/tardigrade), pinned by full commit SHA: one across the upgrade, one across the revert of a completed-but-unconfirmed upgrade. It runs post-merge on `main` when OTA-relevant paths change, weekly, and manually.

- A run of this exact workflow is green on current main: 941 and 948 fault points, all passing ([run](https://github.com/neilberkman/mcuboot/actions/runs/29588412356)).
- Detection is A/B validated at the pinned harness revision: an exhaustive sweep against the commit before the #2100 fix flags 3 power-loss points in the revert fixup where the device ends up stuck on the unconfirmed image ([run](https://github.com/neilberkman/tardigrade/actions/runs/29561593489)); the fixed commit runs the same scenario clean ([run](https://github.com/neilberkman/tardigrade/actions/runs/29559765014)).
- The tardigrade action is a stateless composite action: it downloads Renode and Python dependencies, runs the audit CLI, and uploads a JSON report. No external service or telemetry.

The sim test is the primary regression coverage for this class; the canary is the narrower integration layer on top, exercising the real Zephyr-linked binary, image signing, and reset/recovery behavior for one representative configuration.

### DIFF
--- a/.github/workflows/ota-resilience-canary.yml
+++ b/.github/workflows/ota-resilience-canary.yml
@@ -1,0 +1,253 @@
+name: OTA resilience canary
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/ota-resilience-canary.yml
+      - boot/**
+      - scripts/**
+      - zephyr/**
+      - root-rsa-2048.pem
+  schedule:
+    - cron: '0 8 * * 1'
+  workflow_dispatch:
+    inputs:
+      zephyr_ref:
+        description: 'Which Zephyr ref should be used?'
+        required: true
+        default: 'main'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  ZEPHYR_VERSION: main
+  # Same Renode source the pinned tardigrade revision's own CI runs with.
+  # Versioned 1.16.x mono-portable archives fail to compile the flash
+  # controller peripheral under their bundled mono; the current portable
+  # build is the combination validated by that revision's profile sweeps.
+  RENODE_URL: https://builds.renode.io/renode-latest.linux-portable.tar.gz
+
+jobs:
+  swap-move-canary:
+    runs-on: ubuntu-24.04
+    container:
+      image: zephyrprojectrtos/ci-base:v0.29.0
+      options: '--entrypoint /bin/bash'
+    timeout-minutes: 90
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Set Zephyr version for manual runs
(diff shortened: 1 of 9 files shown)


d3zd3z (commented):
This seems like a good thing to test, although two questions:
- Does it demonstrate a bug that needs to be fixed?
- Is there something that could be added to the sim to demonstrate the failure in the offline simulator, which should be doing adversarial testing.
The build failures likely can be resolved by rebasing onto the latest 'main' which fixes these.

PR #2665, opened 2026-03-18
Add support of shared upgrade slot feature in simulator

DOAR-Infineon:
Summary
Adds simulator device configurations and tests that validate the shared upgrade slot feature — a multi-image layout where multiple images share a single physical flash region for their secondary (upgrade) slots. This significantly reduces flash usage in systems where only one image is upgraded at a time.

Motivation
In the standard MCUboot multi-image configuration, each image requires its own dedicated secondary slot. For a 2-image system with 64 KB slots, this means 128 KB of flash reserved for upgrades. With the shared upgrade slot layout, both secondary slots overlap on the same physical SPI flash region, reducing upgrade flash to ~68 KB — a savings of ~47%.

### DIFF
--- a/docs/shared_upgrade_slot.md
+++ b/docs/shared_upgrade_slot.md
@@ -1,0 +1,532 @@
+# Shared Upgrade Slot — User Guide
+
+## Overview
+
+In a standard MCUboot multi-image configuration every image has its own
+dedicated **primary slot** (boot) and **secondary slot** (upgrade).  The
+*shared upgrade slot* feature allows multiple images to share **one physical
+flash region** for their secondary slots, reducing flash usage significantly
+in systems where only one image is upgraded at a time.
+
+```
+Standard layout                      Shared-slot layout
+┌──────────────┐ ┌──────────────┐    ┌──────────────┐ ┌──────────────┐
+│  Primary 0   │ │  Primary 1   │    │  Primary 0   │ │  Primary 1   │
+│  (internal)  │ │  (internal)  │    │  (internal)  │ │  (internal)  │
+├──────────────┤ ├──────────────┤    ├──────────────┴─┴──────────────┤
+│ Secondary 0  │ │ Secondary 1  │    │    Shared Secondary Region    │
+│  (upgrade)   │ │  (upgrade)   │    │  (one physical flash area)    │
+└──────────────┘ └──────────────┘    └───────────────────────────────┘
+      2× secondary flash                  ~1× secondary flash
+```
+
+### Key constraint
+
+> **Only one image may be upgraded at a time.** The secondary slots
+> physically overlap; writing upgrade data for two images simultaneously
+> would corrupt both.
+
+---
+
+## Supported Upgrade Modes
+
+| Mode | Supported | Scratch needed | Device | Notes |
+|------|-----------|----------------|--------|-------|
+| Overwrite-only | **Yes** | No | `PSOCEdgeE8xSpiFlash` | Simplest layout |
+| Swap-using-scratch | **Yes** | Yes (1 sector) | `PSOCEdgeE8xSpiFlash` | **Deprecated** — prefer swap-using-offset |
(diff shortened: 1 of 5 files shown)


sim/src/lib.rs
```diff
@@ -64,7 +64,8 @@ struct Args {
 #[derive(Copy, Clone, Debug, Deserialize)]
 pub enum DeviceName {
     Stm32f4, Stm32f4SpiFlash, K64f, K64fBig, K64fMulti, Nrf52840, Nrf52840SpiFlash,
-    Nrf52840UnequalSlots, Nrf52840UnequalSlotsLargerSlot1,PSOCEdgeE8x,
+    Nrf52840UnequalSlots, Nrf52840UnequalSlotsLargerSlot1, PSOCEdgeE8x,
+    PSOCEdgeE8xSpiFlash, PSOCEdgeE8xSpiFlashSwapOffset,
 }
 
 pub static ALL_DEVICES: &[DeviceName] = &[
```
d3zd3z, line 71: Given that you've added two devices, and have not added them to `ALL_DEVICES`, I think it warrants an explanation here. For example, maybe a line for each below where they would go, but with a comment explaining that they are special cases and don't with the ordinary multi-image code.

docs/shared_upgrade_slot.md
```diff
@@ -0,0 +1,532 @@
...
+> would corrupt both.
+
+---
+
+## Supported Upgrade Modes
+
+| Mode | Supported | Scratch needed | Device | Notes |
+|------|-----------|----------------|--------|-------|
+| Overwrite-only | **Yes** | No | `PSOCEdgeE8xSpiFlash` | Simplest layout |
+| Swap-using-scratch | **Yes** | Yes (1 sector) | `PSOCEdgeE8xSpiFlash` | Equal primary/secondary sizes |
```
d3zd3z: Given that we're trying to deprecate swap-using-scratch, I wonder if we should have this line at all (or the scratch column). I suppose it is here for completeness.

d3zd3z (approved):
Overall, this looks really good. It isn't clear to me where this will get covered in CI tests. We might want to consider, in the future, whether we want a bit more coverage of some of the power-loss test cases in this configuration.

But this is really good for testing what we do have.

d3zd3z (dismissed):

docs/shared_upgrade_slot.md
```diff
@@ -0,0 +1,532 @@
...
+trailer.
+
+### Additional swap-mode constraint
+
+> **Primary slot size must equal secondary slot size** for each image.
+> This is enforced by `boot_slots_compatible()` in `swap_scratch.c`.
+> If they differ, boot will print *"Cannot upgrade: slots are not
+> compatible"* and refuse to swap.
+
+### Staggering the trailers
```
nordicjm, line 167: can you explain why this is needed? The MCUboot header https://github.com/mcu-tools/mcuboot/blob/main/boot/bootutil/include/bootutil/image.h#L166 includes `uint32_t ih_load_addr;` which gives the exact address to load to which maps to one image only

DOAR-Infineon: To prevent the image header and trailers to colide with each other - to overcome the case when both headers and trailers are valid, but only 2nd application slot is programmed - in such case, during the MCUBoot flow (for multiimage) 1st application integrity will be broken and mcuboot gonna erase 1st application upgrade image slot, causing the second application image to be corrupted/erased.

nordicjm: but it's not valid to have an update to image 2 without a valid image 1, MCUboot will not boot, given that it cannot boot, so the workflow here makes no sense

DOAR-Infineon: > but it's not valid to have an update to image 2 without a valid image 1, MCUboot will not boot, given that it cannot boot, so the workflow here makes no sense

@nordicjm - so that original MCUBoot logic.


Our proposed variant does not change anything in MCUBoot code - the solution is implemented by specific memory layout.

Shall we proceed or you prefer to reject this PR?

nordicjm: it should be done using the proper method of the same slot, having different overlapped parts is a hack

PR #2679, opened 2026-03-27
imgtool: Add TLV IDs for SHA3 in image.py

zhovtobruhd:
Infineon Technologies is working on a feature that involves SHA3 hashes. We are known to have our own fork of imgtool, but we pull sometimes features from the upstream. For that reason, we intend to reserve TLV IDs for SHA3 hashes for future PRs when we need to add features to upstream.


scripts/imgtool/image.py
```diff
@@ -74,6 +74,9 @@
         'SHA256': 0x10,
         'SHA384': 0x11,
         'SHA512': 0x12,
+        'SHA3_256': 0x13,
```
d3zd3z, line 77: To truly reserve these, they should added to `boot/bootutil/include/bootutil/image.h` as that is the authoritative list. They should also be added to `docs/design.md`.

d3zd3z (changes requested):
The commit text itself needs two things:

- a more full description. The description you have as the first comment here is a good starting point. The commit text should describe *why* the change is needed, not just what it is.
- A Signed-off-by line at the end.

PR #2680, opened 2026-03-30
boot:crypto Add custom crypto support

OleksandrShkurchenko:
The `MCUBOOT_USE_CUSTOM_CRYPTO` option allows to implement a custom
backend that lets users plug in any crypto library, hardware
accelerator, proprietary SDK, or another software implementation
without modifying MCUboot's own source.

### DIFF
--- a/.github/workflows/sim.yaml
+++ b/.github/workflows/sim.yaml
@@ -48,6 +48,8 @@ jobs:
         - "sig-ecdsa-psa enc-ec256 max-align-16, sig-ecdsa-psa enc-ec256 swap-offset validate-primary-slot max-align-16"
         - "ram-load enc-aes256-kw multiimage"
         - "ram-load enc-aes256-kw sig-ecdsa-mbedtls multiimage"
+        - "custom-crypto,custom-crypto overwrite-only,custom-crypto validate-primary-slot,custom-crypto swap-offset"
+        - "custom-enc-crypto,custom-enc-crypto validate-primary-slot,custom-enc-crypto swap-offset validate-primary-slot max-align-32"
     runs-on: ubuntu-latest
     env:
       MULTI_FEATURES: ${{ matrix.features }}
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr.h
@@ -2,9 +2,10 @@
  * This module provides a thin abstraction over some of the crypto
  * primitives to make it easier to swap out the used crypto library.
  *
- * At this point, there are two choices: MCUBOOT_USE_MBED_TLS, or
- * MCUBOOT_USE_TINYCRYPT.  It is a compile error there is not exactly
- * one of these defined.
+ * At this point, there are four choices: MCUBOOT_USE_MBED_TLS,
+ * MCUBOOT_USE_TINYCRYPT, MCUBOOT_USE_PSA_CRYPTO, or
+ * MCUBOOT_USE_CUSTOM_CRYPTO.  It is a compile error if there is not
+ * exactly one of these defined.
  */
 
 #ifndef __BOOTUTIL_CRYPTO_AES_CTR_H_
@@ -13,8 +14,10 @@
 #include "mcuboot_config/mcuboot_config.h"
 
 #if (defined(MCUBOOT_USE_MBED_TLS) + \
-     defined(MCUBOOT_USE_TINYCRYPT) + defined(MCUBOOT_USE_PSA_CRYPTO)) != 1
-    #error "One crypto backend must be defined: either MBED_TLS or TINYCRYPT or PSA"
+     defined(MCUBOOT_USE_TINYCRYPT) + \
+     defined(MCUBOOT_USE_PSA_CRYPTO) + \
+     defined(MCUBOOT_USE_CUSTOM_CRYPTO)) != 1
+    #error "One crypto backend must be defined: either MBED_TLS or TINYCRYPT or PSA or CUSTOM_CRYPTO"
 #endif
 
 #if defined(MCUBOOT_USE_MBED_TLS)
@@ -27,6 +30,6 @@
 
 #if defined(MCUBOOT_USE_PSA_CRYPTO)
     #include "bootutil/crypto/aes_ctr_psa.h"
-#endif
+#endif /* MCUBOOT_USE_PSA_CRYPTO */
 
 #endif /* __BOOTUTIL_CRYPTO_AES_CTR_H_ */
--- a/boot/bootutil/include/bootutil/crypto/ecdsa.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdsa.h
@@ -9,12 +9,13 @@
  * primitives to make it easier to swap out the used crypto library.
  *
  * At this point, the choices are: MCUBOOT_USE_TINYCRYPT, MCUBOOT_USE_CC310,
- * MCUBOOT_USE_MBED_TLS, MCUBOOT_USE_PSA_CRYPTO. Note that support for
- * MCUBOOT_USE_PSA_CRYPTO is still experimental and it might not support all
- * the crypto abstractions that MCUBOOT_USE_MBED_TLS supports. For this
- * reason, it's allowed to have both of them defined, and for crypto modules
- * that support both abstractions, the MCUBOOT_USE_PSA_CRYPTO will take
- * precedence.
+ * MCUBOOT_USE_MBED_TLS, MCUBOOT_USE_PSA_CRYPTO, MCUBOOT_USE_CUSTOM_CRYPTO.
+ * Note that support for MCUBOOT_USE_PSA_CRYPTO is still experimental and it
+ * might not support all the crypto abstractions that MCUBOOT_USE_MBED_TLS
+ * supports. For this reason, it's allowed to have both of them defined, and
+ * for crypto modules that support both abstractions, the MCUBOOT_USE_PSA_CRYPTO
+ * will take precedence. MCUBOOT_USE_CUSTOM_CRYPTO delegates all operations to
+ * a platform-supplied <mcuboot_custom_crypto.h> resolved via the include path.
  */
 
 #ifndef __BOOTUTIL_CRYPTO_ECDSA_H_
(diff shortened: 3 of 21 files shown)


docs/custom_crypto.md
```diff
@@ -0,0 +1,192 @@
+<!--
+  - SPDX-License-Identifier: Apache-2.0
+-->
+
+# Custom Crypto Backend
+
+MCUboot's crypto abstraction layer supports several open source backends like
+`MCUBOOT_USE_MBED_TLS`, `MCUBOOT_USE_TINYCRYPT`, etc. The `MCUBOOT_USE_CUSTOM_CRYPTO` option allows to implement a custom backend that lets users plug in any crypto library —
```
d3zd3z: This line can be wrapped at < 78 or so solumns, which will help when reading these files directly.

sim/Cargo.toml
```diff
@@ -36,6 +36,7 @@ max-align-16 = ["mcuboot-sys/max-align-16"]
 max-align-32 = ["mcuboot-sys/max-align-32"]
 hw-rollback-protection = ["mcuboot-sys/hw-rollback-protection"]
 check-load-addr = ["mcuboot-sys/check-load-addr"]
+custom-crypto = ["mcuboot-sys/custom-crypto"]
```
d3zd3z, line 39: If we add a custom-crypto configuration, we should also add to `.github/workflows/sim.yaml` to the build matrix to include testing of this configuration.

d3zd3z (changes requested):
Looks good. A minor textual change, and adding CI support.

In addition, can you please format the commit text to be less than 72 columns.

docs/custom_crypto.md
OleksandrShkurchenko: Fixed

sim/Cargo.toml
OleksandrShkurchenko: fixed

boot/bootutil/src/encrypted_psa.c
```diff
@@ -1,11 +1,15 @@
 /*
  * Copyright (c) 2025 Nordic Semiconductor ASA
+ * Copyright (c) 2026 Infineon Technologies AG, or an affiliate of Infineon
+ * Technologies AG
```
nordicjm:
```suggestion
```
let's not add copyrights for 2 line #ifdef changes

boot/bootutil/include/bootutil/crypto/sha.h
```diff
@@ -4,18 +4,22 @@
  * Copyright (c) 2017-2019 Linaro LTD
  * Copyright (c) 2017-2019 JUUL Labs
  * Copyright (c) 2021-2023 Arm Limited
+ * Copyright (c) 2026 Infineon Technologies AG, or an affiliate of Infineon
+ * Technologies AG
```
nordicjm:
```suggestion
```

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -2,19 +2,22 @@
  * SPDX-License-Identifier: Apache-2.0
  *
  * Copyright (c) 2023-2025 Arm Limited
+ * Copyright (c) 2026 Infineon Technologies AG, or an affiliate of Infineon
+ * Technologies AG
```
nordicjm:
```suggestion
```

nordicjm (changes requested):

docs/custom_crypto.md
```diff
@@ -0,0 +1,199 @@
...
+int  boot_enc_init(struct enc_key_data *enc_state, uint8_t slot);
+int  boot_enc_drop(struct enc_key_data *enc_state, uint8_t slot);
+int  boot_enc_set_key(struct enc_key_data *enc_state, uint8_t slot,
+                      const struct boot_status *bs);
+int  boot_enc_load(struct boot_loader_state *state, int image_index,
+                   const struct image_header *hdr,
+                   const struct flash_area *fap,
+                   struct boot_status *bs);
+bool boot_enc_valid(struct enc_key_data *enc_state, uint8_t slot);
+void boot_enc_encrypt(struct enc_key_data *enc_state, uint8_t slot,
```
d3zd3z: The code for this function doesn't have a slot argument.

.github/workflows/sim.yaml
```diff
@@ -48,6 +48,8 @@ jobs:
         - "sig-ecdsa-psa enc-ec256 max-align-16, sig-ecdsa-psa enc-ec256 swap-offset validate-primary-slot max-align-16"
         - "ram-load enc-aes256-kw multiimage"
         - "ram-load enc-aes256-kw sig-ecdsa-mbedtls multiimage"
+        - "custom-crypto,custom-crypto overwrite-only,custom-crypto validate-primary-slot"
+        - "custom-enc-crypto,custom-enc-crypto overwrite-only,custom-enc-crypto validate-primary-slot,custom-enc-crypto max-align-32"
```
d3zd3z: Should we be testing custom crypto with at least swap-offset, as that is the mode we are generally encouraging people to use.

d3zd3z (changes requested):

boot/bootutil/src/encrypted_psa.c
OleksandrShkurchenko: Fixed

boot/bootutil/include/bootutil/crypto/sha.h
OleksandrShkurchenko: Fixed

boot/bootutil/include/bootutil/crypto/ecdsa.h
OleksandrShkurchenko: Fixed

.github/workflows/sim.yaml
OleksandrShkurchenko: Add swap offset

docs/custom_crypto.md
OleksandrShkurchenko: Fixed. Also reviewed whole .md for such issues in other sections

sim/mcuboot-sys/csupport/custom_crypto/sim_custom_hmac_sha256.h
```diff
@@ -0,0 +1,66 @@
...
+}
+
+static inline int bootutil_hmac_sha256_update(bootutil_hmac_sha256_context *ctx,
+                                              const void *data,
+                                              unsigned int data_length)
+{
+    return mbedtls_md_hmac_update(ctx, data, data_length);
+}
+
+static inline int bootutil_hmac_sha256_finish(bootutil_hmac_sha256_context *ctx,
```
d3zd3z, line 54: I wonder if we should have a comment explaining why we ignore taglen, and specifically, what taglen mbedtls is expecting?

d3zd3z (approved):
Looks good. Thanks.

boot/bootutil/include/bootutil/crypto/aes_ctr.h
```diff
@@ -27,6 +30,10 @@
 
 #if defined(MCUBOOT_USE_PSA_CRYPTO)
     #include "bootutil/crypto/aes_ctr_psa.h"
-#endif
+#endif /* MCUBOOT_USE_PSA_CRYPTO */
+
+#if defined(MCUBOOT_USE_CUSTOM_CRYPTO)
+    #include "mcuboot_custom_crypto.h"
```
nordicjm: I don't see why this should be including a file, if you have custom crypto, you can add your custom source/headers to the build, why is this needed?

OleksandrShkurchenko: Yes we can but
 if we consider MCUBOOT_USE_CUSTOM_CRYPTO as one of crypto backend(along with MCUBOOT_USE_PSA_CRYPTO, MCUBOOT_USE_MBED_TLS, etc) it would be good to follow the same approach  and include custom crypto header as it done for PSA or MbedTLS backends. Since each custom crypto implementation may have different header names, I included  implementation agnostic header in the bootutil files.

OleksandrShkurchenko: and  for some ecosystem is more convenient to  configure mcuboot features via mcuboot_config.h because  they include to build all source files and features should be selected by defines.

nordicjm: those files are included because they exist in MCUboot, you are including a file that does not exist, there isn't really any reason do to that. Add the .c file to your build with the same header definitions and that's all you need. I really don't see the point in adding an include for a file that does not exist when it isn't needed

de-nordic (approved):

sim/mcuboot-sys/csupport/custom_crypto/sim_custom_hmac_sha256.h
OleksandrShkurchenko: Added some explanation to this file and  hmac_sha256.h in bootutils

boot/bootutil/include/bootutil/crypto/aes_ctr.h
OleksandrShkurchenko: #include "mcuboot_custom_crypto.h" has been removed from bootutil include files. Now there are 2 options to include "mcuboot_custom_crypto.h" into build process:

Option 1: Force-Include via Compiler Flag
Platform header included before all source files via compiler's force-include flag.
```
GCC/Clang:  -include mcuboot_custom_crypto.h
ARM CC6:    -include mcuboot_custom_crypto.h
ARM CC5:    --preinclude=mcuboot_custom_crypto.h
IAR:        --preinclude mcuboot_custom_crypto.h
```
Option 2: Include via mcuboot_config.h
Users adds include to their mcuboot_config.h:
```
// mcuboot_config.h
#define MCUBOOT_USE_CUSTOM_CRYPTO

#ifdef MCUBOOT_USE_CUSTOM_CRYPTO
#include "mcuboot_custom_crypto.h"
#endif
```

@d3zd3z @nordicjm @de-nordic  Could you please review latest changes

d3zd3z (approved):

davidvincze (approved):
Thanks for the contribution.

d3zd3z:
@nordicjm does this address your concern? The include is still there, but only used in the sim.

nordicjm:
> @nordicjm does this address your concern? The include is still there, but only used in the sim.

Yep is fine, though the doc is not exactly correct, you do **not** need to adjust any sort of include order if using e.g. cmake, you just add files to the build, which for zephyr you can do with a zephyr module. But documenting how to do that is not necessary so is OK

PR #2681, opened 2026-03-31
boot: zephyr: socs: Add nrf54h20 Kconfig fragment

nordicjm:
Adds a Kconfig fragment needed for MRAM access


rghaddab (approved):

d3zd3z (changes requested):
The change itself looks straightforward, but the commit text really needs to explain a bit more of "why" this is needed, rather than just re-iterating the summary line. What's going on with MRAM that needs multithreading support in Zephyr?

rghaddab:
> The change itself looks straightforward, but the commit text really needs to explain a bit more of "why" this is needed, rather than just re-iterating the summary line. What's going on with MRAM that needs multithreading support in Zephyr?

@d3zd3z For the nrf54h20 board family, MRAM writes could stall if the Ready register is not checked before any write, as multiple cores can trigger a write at the same time.
But in order to check that register we need to make sure that the MRAM controller is powered up otherwise we won't get a Ready register that is set to '1'. This is due to the Autopowerdown mode in the controller.
In order to disable that mode in the controller before any write, we need to use the mram_latency service.
This is the only way to disable Autopowerdown mode.
mram_latency service, is of course a service that needs EVENTS, and EVENTS depends on MULTITHREADING.
That's the whole story behind this little change.

de-nordic (approved):

carlescufi:
> The change itself looks straightforward, but the commit text really needs to explain a bit more of "why" this is needed, rather than just re-iterating the summary line. What's going on with MRAM that needs multithreading support in Zephyr?

@nordicjm can you add the info to the commit message please?

d3zd3z:
> @d3zd3z For the nrf54h20 board family, MRAM writes could stall if the Ready register is not checked before any write, as multiple cores can trigger a write at the same time.
> But in order to check that register we need to make sure that the MRAM controller is powered up otherwise we won't get a Ready register that is set to '1'. This is due to the Autopowerdown mode in the controller.
> In order to disable that mode in the controller before any write, we need to use the mram_latency service.
> This is the only way to disable Autopowerdown mode.
> mram_latency service, is of course a service that needs EVENTS, and EVENTS depends on MULTITHREADING.
> That's the whole story behind this little change.

This would be good to put in the body of the commit. It is significantly more work (especially because we rebase) to trace individual commits back to comments in a PR thread, so this kind of explanation really should be in the commit text itself.

nordicjm:
> > @d3zd3z For the nrf54h20 board family, MRAM writes could stall if the Ready register is not checked before any write, as multiple cores can trigger a write at the same time.
> > But in order to check that register we need to make sure that the MRAM controller is powered up otherwise we won't get a Ready register that is set to '1'. This is due to the Autopowerdown mode in the controller.
> > In order to disable that mode in the controller before any write, we need to use the mram_latency service.
> > This is the only way to disable Autopowerdown mode.
> > mram_latency service, is of course a service that needs EVENTS, and EVENTS depends on MULTITHREADING.
> > That's the whole story behind this little change.
> 
> This would be good to put in the body of the commit. It is significantly more work (especially because we rebase) to trace individual commits back to comments in a PR thread, so this kind of explanation really should be in the commit text itself.

Has been added

d3zd3z (approved):

PR #2683, opened 2026-04-01
boot: zephyr: Add HW Image Swapping into boot flow

PetrBuchtaNXP:
This adds support for HW Image Swapping used with DIRECT_XIP modes. This enables zero-overhead image swapping that is done instantly and with no flash wear. This PR is for changes on MCUboot side.

After the image to be booted is selected it passes the image offset to the HW Image Swapping logic that controls the underlying HW swapping mechanism - this is implemented on the Zephyr side.

First NXP boards to use this are frdm_rw612 and rd_rw612_bga that use FlexSPI memory controller with address remapping feature. There is a number of other boards that can use this.

Zephyr side PR - https://github.com/zephyrproject-rtos/zephyr/pull/106683

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -1269,6 +1269,16 @@ config MCUBOOT_USE_TLV_ALLOW_LIST
 	  Disabling this option will cut down MCUboot size.
 	  The Kconfig controlls MCUboot configuration option MCUBOOT_USE_TLV_ALLOW_LIST.
 
+config BOOT_HW_IMAGE_SWAP
+	bool "Image swapping done by hardware"
+	depends on BOOT_DIRECT_XIP || BOOT_DIRECT_XIP_REVERT
+	help
+	  Image swapping will be done using hardware means which will result in zero flash wear compared
+	  to techniques like SWAP_MOVE or SWAP_SCRATCH that physically move data in flash.
+	  This requires support in hardware (e.g. address remapping, flash swapping...).
+	  This feature can only be used with MCUboot modes based on DIRECT_XIP.
+
+
 config BOOT_DECOMPRESSION_SUPPORT
 	bool
 	help
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -30,6 +30,10 @@
 #include <soc.h>
 #include <zephyr/linker/linker-defs.h>
 
+#if defined(CONFIG_BOOT_HW_IMAGE_SWAP)
+#include <zephyr/drivers/memc/hwimageswap.h>
+#endif
+
 #if defined(CONFIG_BOOT_DISABLE_CACHES)
 #include <zephyr/cache.h>
 #endif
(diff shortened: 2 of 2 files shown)


nordicjm (changes requested):
This isn't really acceptable, and wouldn't even work since the whole update system in e.g. MCUmgr is done based on addresses

PetrBuchtaNXP:
Hi @nordicjm , please have a look here on the Zephyr side PR - https://github.com/zephyrproject-rtos/zephyr/pull/106683

I did a little refactoring in DFU subsystems to make this work and those are only small changes that don't brake any current logic.

I don't understand your remark about addresses - all stays compatible with current approach. The remapping mechanism doesn't brake anything. It just allows to run image stored in slot1 that was linked for slot0 as if it was stored in slot0.

Could you also further elaborate why this isn't acceptable?

PetrBuchtaNXP:
Hi @nordicjm, could you please comment on my previous message? Also I was thinking about other ways this could be done - maybe a generic pre-boot hook in the same place where I call the HW Image Swapping control function would be more acceptable and could be used for similar functions? There are already several projects that use the commits I'm trying to upstream so coming up with some acceptable form would be great.

nordicjm:
What you said is irrelevant to my comment

> This isn't really acceptable, and wouldn't even work since the whole update system in e.g. MCUmgr is done based on addresses

As said, this does not work with MCUmgr since it uses the address to determine what it is, you cannot upload if your image always has the same address for both slots, it is broken by design

PetrBuchtaNXP:
I used MCUmgr for development, did all my testing with it, it works. Nothings changes about addresses - there is one address for slot0 and another different one for slot1.

nordicjm:
> I used MCUmgr for development, did all my testing with it, it works. Nothings changes about addresses - there is one address for slot0 and another different one for slot1.

Then I don't understand the point in this at all, if it does remapping then it means when the image in either slot is running, with remap, it will be running from what it _thinks_ is the exact same address, which is where the problem would be. If it doesn't and you still have 2 different images and you can only load the correct image to each slot (you can't load either) then that is what is already supported in MCUboot as direct-xip and I don't see why you need any changes to support what is already supported (and used) by many boards

PetrBuchtaNXP:
I'm using DIRECT-XIP as my base mode. The remapping allows me to have image in slot1 linked for slot0, as would be the case with the SWAP modes. Otherwise without the remapping I would need to have two separate images for image update - one linked for slot0 and other one for slot1 - this is not acceptable for many. With the remapping I'm able to have an image linked for slot0 stored in slot1 and once the remapping is active MCUboot can run the image stored in slot1. Then download a new image to whatever slot is the old one. So I reuse everything from the DIRECT-XIP mode and I just need a point right before boot where I need to activate the remapping in case I need to boot from slot1 (because the image is linked for slot0). **It's all about using DIRECT-XIP mode and being able to have only one image update that can run from both slots**.

nordicjm:
> The remapping allows me to have image in slot1 linked for slot0

Then as per what I said above, it does not work. The image **must** know what slot it is executing from, if you only build for slot0 then it will upload to slot1, that is how the code works, so if you load and run from slot1 then it will erase itself

PetrBuchtaNXP:
Yes it does know what slot is is executing from. I built on top of the **blinfo** code, which is already upstream, where the slot is determined in runtime. Of course I had to add a few lines of code for this remapping case but that's all visible in the Zephyr side PR. There is a function **boot_fetch_active_slot_number**() that resolves the active slot in runtime for both the blinfo case and this remapping case. Since the blinfo API was already upstream I didn't have to do any significant refactoring. For MCUmgr the change in my PR is visible [here ](https://github.com/zephyrproject-rtos/zephyr/pull/106683/changes/72baa1c7a631beb31c05e0f0d9cb3288f2421195#diff-9f4626414aafe90942f70ed42999551300d5a4317dc445c38e4ebfb54f6973d7)

PR #2685, opened 2026-04-02
boot: Add support of xip encryption mode for swap

DOAR-Infineon:
Standard `MCUboot` encryption decrypts images during the swap process so that
the `primary slot` always contains plaintext. This works well for internal flash
but prevents execute-in-place (XIP) from external memory, because decrypted
data would be exposed on the external bus.

XIP encryption keeps images encrypted in **all** slots. A hardware crypto engine
(e.g., Infineon SMIF with on-the-fly AES decryption) provides transparent
decryption during code execution. The CPU fetches ciphertext from flash and the
hardware returns plaintext — no software decryption step is needed at runtime.

### How XIP encryption for swap works

1. **Build-time encryption** — Images are encrypted by `imgtool` (or
   `edgeprotecttools`) using AES-CTR **before** signing. The AES-CTR nonce
   follows the edgeprotecttools format:

   ```
   AES-CTR block input = counter_LE32 || nonce[0:12]
   ```

   where `counter_LE32` is the **absolute XIP flash address**
   (`xip_base_address + header_size + block_offset`) encoded as a little-endian
   32-bit integer, and `nonce[0:12]` is the first 12 bytes of the HKDF-derived
   `xip_iv`. The counter is anchored to `xip_base_address` (the primary slot's
   XIP address), so it is not slot-dependent; both the primary and secondary
   slots store identical ciphertext.

2. **Ciphertext signing** — The SHA-256 hash and ECDSA signature are computed
   over the encrypted image (header + ciphertext + protected TLVs). The
   bootloader never performs software decryption during validation — the
   signature covers the exact bytes that reside in flash.

3. **Mandatory signature** — Encrypted XIP images must be signed. The
   bootloader rejects unsigned encrypted images to prevent XOR attacks: an
   attacker who knows plaintext at a given offset could XOR it against the
   ciphertext to recover the keystream. A mandatory signature closes this
   attack vector.

4. **Swap copies raw bytes** — During an upgrade, swap moves encrypted blocks
   between slots without any encryption or decryption. `MCUBOOT_ENC_IMAGES`
   is **not** defined; the data is byte-for-byte identical regardless of which
   slot it resides in.

5. **Post-boot hardware setup** — After `boot_go()` returns, the application
   startup code configures hardware crypto regions for each image. The AES key
   and IV are available in the `boot_rsp` structure:

   ```c
   struct boot_rsp rsp;
   /* ... boot_go(&rsp) ... */

   /* rsp.br_xip_key[4] -- 16-byte AES-128 key */
   /* rsp.br_xip_iv[4]  -- 16-byte IV/nonce     */
   ```

   Keys are cleared from library-internal storage before jumping to the
   application.

6. **Bootloader validation flow** — The `boot_image_check_hook()` validates
   encrypted XIP images without software decryption:
   1. Verify image size fits within slot limits
   2. Compute SHA-256 over raw flash (header + ciphertext + protected TLVs)
   3. Verify hash against `IMAGE_TLV_SHA256`
   4. Verify mandatory ECDSA-P256 signature against `IMAGE_TLV_ECDSA_SIG`
   5. ECIES-P256 key unwrap from `IMAGE_TLV_ENC_EC256` to extract AES key/IV

### Extended ECIES TLV format

XIP images use an extended ECIES TLV (177 bytes) that adds a 32-byte HKDF
salt field beyond the standard ECIES-P256 layout (113 bytes):

```
[ ephemeral_pubkey (65 B) | hmac_tag (32 B) | enc_key (16 B) |
  hkdf_salt (32 B) | reserved (32 B) ]
```

The `hkdf_salt` is a per-image random value used as an HKDF diversifier,
ensuring that every image derives a unique AES key and IV — a critical
requirement for AES-CTR security. (When a non-zero `--xip-base-address` is
supplied, `imgtool` stores it in the salt's final 4 bytes.) The AES key-wrap
IV and the XIP nonce (`xip_iv`) are **derived on-device via HKDF** from the
shared secret and salt; they are not transmitted in the TLV. The final
32-byte field is reserved for future use and must be zero.

### Platform requirements

A platform using XIP encryption must provide or override:

* **`boot_image_check_hook(state, img_index, slot)`** — Performs complete image
  validation including hash and signature verification over ciphertext and
  ECIES key unwrap. A default (weak) implementation is provided by the XIP
  encryption library (`xip_enc_validate.c`), which is sufficient for the
  simulator and software-only builds. Real hardware typically overrides it with
  a strong implementation that performs the ECIES key unwrap inside a secure
  element / hardware crypto engine and adds fault-injection hardening. The
  `state` pointer provides access to boot loader context (slot flash areas,
  secondary offset for swap-offset mode, max image size); it may be `NULL` when
  called outside the normal boot flow (e.g. serial recovery).

  ```c
  fih_ret boot_image_check_hook(struct boot_loader_state *state,
                                int img_index, int slot);
  ```

* **`boot_xip_populate_rsp(img_index, rsp)`** — Called after successful
  validation to copy the extracted AES key and IV into the `boot_rsp`
  structure (`rsp->br_xip_key` / `rsp->br_xip_iv`). The application uses these
  values to configure hardware crypto regions.

  ```c
  void boot_xip_populate_rsp(int img_index, struct boot_rsp *rsp);
  ```

* **Hardware crypto setup** — After `boot_go()` returns, the application
  must configure the hardware crypto engine (e.g., SMIF crypto regions) using
  the key material from `boot_rsp` before jumping to the application image.


docs/encrypted_images.md
```diff
@@ -190,3 +190,121 @@ required keys.
...
+
+Standard `MCUboot` encryption decrypts images during the swap process so that
+the `primary slot` always contains plaintext. This works well for internal flash
+but prevents execute-in-place (XIP) from external memory, because decrypted
+data would be exposed on the external bus.
+
+XIP encryption keeps images encrypted in **all** slots. A hardware crypto engine
+(e.g., Infineon SMIF with on-the-fly AES decryption) provides transparent
+decryption during code execution. The CPU fetches ciphertext from flash and the
+hardware returns plaintext -- no software decryption step is needed at runtime.
```
d3zd3z, line 206: A bit of expansion here I think will help with a security analysis:
* It's probably best to state plainly that in this mode, software never decrypts or encrypts.
* Signatures are over the encrypted image.
* It is _critical_ that every image uses a completely different encryption key. This is a requirement for AES-CTR.

There is a tendency, in the security world, to "mandate" authenticated encryption. I have only partially made an argument with various people at the IETF that this is mostly unworkable for embedded use, and especially XIP.

In this case, it is unlikely we even have control over the particular encryption used. So, key and counter hygeine is very important.

docs/encrypted_images.md
```diff
@@ -190,3 +190,121 @@ required keys.
...
+independently through the hook interface.
+
+### How it works
+
+1. **Build-time encryption** -- Images are encrypted by `imgtool` using
+   AES-CTR, targeted at the primary slot address. The AES-CTR counter is
+   computed as:
+
+   ```
+   counter = (encryption_address + offset) / 16
```
d3zd3z: I'm not clear what these two values are. Below, offset is described as the byte offset within the image. But "encryption address" is not explained. Is this the base address of the first slot? Please make sure to explain it, especially that it is not a property of which slot the image resides in.

boot/bootutil/src/xip_enc/xip_enc_validate.c
```diff
@@ -0,0 +1,216 @@
...
+#include "bootutil/boot_hooks.h"
+#include "bootutil/crypto/sha.h"
+#include "flash_map_backend/flash_map_backend.h"
+#include "xip_enc.h"
+
+/*
+ * Compute SHA-256 hash over header + decrypted payload + protected TLVs.
+ *
+ * Mirrors bootutil_img_hash() logic but uses boot_decrypt_xip() for
+ * payload decryption instead of boot_enc_decrypt().
```
d3zd3z: This, I think is a bit of a fundamental problem here. With swap-based software decryption, the image signature is over the plaintext. This is because for the normal run case, the ciphertext is not available. For XIP, I think it makes much more sense to perform the signature over the ciphertext. This will allow the image to be verified without ever needing to do software decryption (with a goal that `boot_decrypt_xip()` is not present, and that the key is only used to give to the hardware and forgotten).

boot/bootutil/src/xip_enc/xip_enc_validate.c
```diff
@@ -0,0 +1,216 @@
...
+ *   1. ECIES-P256 key unwrap from IMAGE_TLV_ENC_EC256
+ *   2. SHA-256 hash verification against IMAGE_TLV_SHA256
+ *
+ * Non-encrypted images are deferred to upstream (FIH_BOOT_HOOK_REGULAR).
+ *
+ * Platforms needing FIH hardening or hardware crypto should provide
+ * a strong override of this function.
+ */
+__attribute__((weak))
+fih_ret boot_image_check_hook(int img_index, int slot)
```
d3zd3z: As mentioned above, I think this verification will make more sense if we change XIP encryption to sign the ciphertext instead of the plaintext.

But, it is also important to ensure that with encryption, signature verification is mandatory. It shouldn't be possible to build a configuration that has encryption validates the entire ciphertext before loading the encryption key into the hardware.

Without the signature verification, an attacker can trivially XOR parts of the image, which will apply as a XOR of the plaintext as well. Well known areas, such as vectors can easily be manipulated to refer to non-checked code as an attack.

boot/bootutil/src/xip_enc/xip_enc_keys.c
```diff
@@ -0,0 +1,59 @@
...
+    }
+    if (iv != NULL) {
+        (void)memcpy(iv, xip_keys[img_index].iv, XIP_ENC_IV_SIZE);
+    }
+    return 0;
+}
+
+void xip_enc_zeroize(void)
+{
+    (void)memset(xip_keys, 0, sizeof(xip_keys));
```
d3zd3z: There are some various zeroize implementations in the tree, and perhaps this needs to be generalized more. However, beyond this, there doesn't seem to be anything that ever calls this function.

boot/bootutil/src/xip_enc/xip_enc_decrypt.c
```diff
@@ -0,0 +1,85 @@
...
+        (void)memset(key, 0, sizeof(key));
+        return -1;
+    }
+
+    blk_off = off & 0xfu;
+    rc = bootutil_aes_ctr_encrypt(&aes_ctr, nonce, buf, sz, blk_off, buf);
+    bootutil_aes_ctr_drop(&aes_ctr);
+
+    /* Zeroize key from stack */
+    (void)memset(key, 0, sizeof(key));
```
d3zd3z: This needs to use a platform provided zeroize, as compilers are free to optimize away a memset of a stack variable. This applies in all instances throughout the code.

boot/bootutil/src/xip_enc/xip_enc_ecies.c
```diff
@@ -0,0 +1,467 @@
...
+    uint8_t *cp;
+    uint8_t *cpend;
+    int rc = -1;
+
+    /* Validate TLV length */
+    if ((tlv_buf == NULL) || (out_key == NULL) || (out_iv == NULL)) {
+        return -1;
+    }
+
+    if ((tlv_len < ECIES_STD_TLV_SIZE) || (tlv_len > ECIES_EXT_TLV_SIZE)) {
```
d3zd3z: This check will allow a tlv with length anywhere between STD and EXT TLV size. However, the code below will only check if the value is greater than STD, and if so, assume that the extended data is available. If the `tlv_len` is between these values, this will result in that code reading from uninitialized memory.

This check should require the `tlv_len` to be either the STD or the EXT sizes, and not in between.

d3zd3z (changes requested):
A couple of specific comments, notably that I believe we should change to signing the ciphertext and not the plaintext, which avoids having to decrypt the image in software.

A few other things, Zephyr seems to be missing CMake changes to allow this code to be compiled.

It is important to explain the the IV calculation, and for example that either the key itself must be unique per image, or the extended TLV must be used to ensure that the IV is different for each image encrypted.

There are some memory issues, memset on stack variables  is generally optimized away, and some uses of memcmp instead of a constant time comparison.

Also, it isn't clear to me what the 32 bytes of zero padding are after the salt.

I think the missing calls to zeroize concern `prk` and `T`? But, other memset calls need to zeroize. And we need a platform-specific correct zeroize (which can be rather difficult to write).

And, the Zephyr kconfig could have a `depends on !BOOT_ENCRYPT_IMAGE` to detect the misconfiguration early, rather than only after a compile failure.

stindaNXP:
Hi,

This PR gives the impression that it provides generic support for encrypted XIP for other vendors. However, not all implementations are based on AES-CTR, so the image has to be re-encrypted using the platform’s encryption scheme. 
Are you aware of that?

In addition, some implementations cannot set the initial AES-CTR offset value (base address), as it is hardwired to be computed from the beginning of the flash memory. This could be fixed on imgtool side which could provide an argument for initial offset, so the computed cipher could be readable for the implementation.

DOAR-Infineon:
> Hi,
> 
> This PR gives the impression that it provides generic support for encrypted XIP for other vendors. However, not all implementations are based on AES-CTR, so the image has to be re-encrypted using the platform’s encryption scheme. Are you aware of that?
> 
> In addition, some implementations cannot set the initial AES-CTR offset value (base address), as it is hardwired to be computed from the beginning of the flash memory. This could be fixed on imgtool side which could provide an argument for initial offset, so the computed cipher could be readable for the implementation.

Hi, thanks for the review.

  1. This implementation is based on AES-CTR, which is the most common cipher used by XIP-capable flash controllers. The
  design is generic within the AES-CTR family — the nonce format (`counter_LE32 || iv[0:12]`) and per-image key/IV
  derivation via HKDF are not tied to any specific vendor. The architecture is designed to be overridable for platforms with different encryption schemes: boot_image_check_hook(), boot_decrypt_xip(), and boot_xip_populate_rsp() are all __weak functions that platforms can replace with their own validation and hardware setup logic. The AES-CTR-specific code is confined to the xip_enc/ library, which serves as a default implementation.

 2.   The bootloader-side AES-CTR nonce format follows the Infineon edgeprotecttools convention:
 `  AES-CTR block = counter_LE32 || nonce[0:12]`  where the counter is the absolute byte address and the 12-byte nonce is derived via HKDF from the ECIES TLV. The HKDF  salt in the extended TLV also embeds the image base address (last 4 bytes), binding key derivation to the target slot — again matching edgeprotecttools. I'm currently working on aligning imgtool --encrypt-xip with edgeprotecttools as well.

docs/encrypted_images.md
DOAR-Infineon: Updated

boot/zephyr/Kconfig
```diff
@@ -863,6 +861,29 @@ config BOOT_ENCRYPTION_KEY_FILE
 	  e.g. \${CMAKE_CURRENT_LIST_DIR} will allow referencing a file in that directory, these
 	  will be automatically configured by the build system.
 
+config BOOT_ENCRYPT_XIP
```
d3zd3z: My remaining question is why are there two Kconfigs here? That leaves four combinations that could be selected, isn't this just a single configuration?

docs/encrypted_images.md
```diff
@@ -190,3 +190,271 @@ required keys.
...
+* **A signature is mandatory.** Encrypted XIP images must be signed, and the
+  bootloader rejects unsigned encrypted images. Without signature
+  verification an attacker can mount a trivial XOR attack against AES-CTR
+  ciphertext: flipping bits in the ciphertext flips the corresponding bits
+  in the plaintext that the CPU will execute. Well-known regions such as the
+  vector table could be manipulated to redirect execution to attacker-chosen
+  code. The mandatory signature, computed over the ciphertext, closes this
+  attack vector before the AES key is ever loaded into the hardware.
+
+* **Every image must use a unique encryption key and IV.** This is a hard
```
d3zd3z, line 240: Here we document "unique encryption key and IV". The code seems to allow the IV to be entirely zero, but it is unclear to me if that matches this property.  Realistically, I would recommend not allowing the zero-IV case. Is there hardware that needs that?

boot/bootutil/src/bootutil_priv.h
```diff
@@ -197,6 +197,12 @@ _Static_assert(sizeof(boot_img_magic) == BOOT_MAGIC_SZ, "Invalid size for image
 #endif /* MCUBOOT_DIRECT_XIP && MCUBOOT_ENC_IMAGES */
 #endif /* MCUBOOT_DIRECT_XIP || MCUBOOT_RAM_LOAD */
 
+#if defined(MCUBOOT_ENC_IMAGES) && defined(MCUBOOT_ENC_IMAGES_XIP)
+#error "MCUBOOT_ENC_IMAGES and MCUBOOT_ENC_IMAGES_XIP are mutually exclusive. " \
+       "XIP encryption keeps images encrypted in both slots and does not use " \
+       "the standard encrypt/decrypt-during-swap mechanism."
+#endif
+
```
d3zd3z, line 227: What about the combination of `MCUBOOT_DIRECT_XIP` and `MCUBOOT_END_IMAGES_XIP`? does this get rejected somewhere I've missed?

boot/bootutil/include/bootutil/bootutil.h
```diff
@@ -59,6 +59,12 @@ struct boot_rsp {
      */
     uint8_t br_flash_dev_id;
     uint32_t br_image_off;
+#if defined(MCUBOOT_ENC_IMAGES_XIP)
+    /** AES-128 key for post-boot XIP hardware crypto setup. */
+    uint32_t br_xip_key[4];
```
d3zd3z, line 73: What is the lifetime of these values? Is there something that explicitly wipes the keys?

docs/encrypted_images.md
```diff
@@ -190,3 +190,271 @@ required keys.
...
+
+6. **Post-boot hardware setup** -- After `boot_go()` returns, the application
+   startup code configures hardware crypto regions for each image. The AES key
+   and IV are available in the `boot_rsp` structure:
+
+   ```c
+   struct boot_rsp rsp;
+   /* ... boot_go(&rsp) ... */
+
+   /* rsp.xip_key[4] -- 16-byte AES-128 key */
```
d3zd3z: The fields have been renamed in the struct, and that needs to be reflected here.

docs/PORTING.md
```diff
@@ -193,3 +193,158 @@ For reference see [Mbed TLS platform.h](https://tls.mbed.org/api/platform_8h.htm
...
+
+```c
+void boot_xip_populate_rsp(int img_index, struct boot_rsp *rsp);
+```
+
+Called by MCUboot after `fill_rsp()` completes. The platform implementation
+copies the encryption key and initialisation vector (IV) that were extracted
+during validation into the boot response structure:
+
+* `rsp->xip_key` --- the AES key used by the hardware crypto region.
```
d3zd3z: These fields have different names in the real struct.

d3zd3z (changes requested):
In addition to the inline comments, there don't seem to be any tests that test for failures (corrupting an image, and showing validation failing, for example).

docs/encrypted_images.md
DOAR-Infineon: Updated

boot/bootutil/src/xip_enc/xip_enc_validate.c
DOAR-Infineon: Updated

boot/bootutil/src/xip_enc/xip_enc_validate.c
DOAR-Infineon: Updated

boot/bootutil/src/xip_enc/xip_enc_keys.c
DOAR-Infineon: Updated

boot/bootutil/src/xip_enc/xip_enc_decrypt.c
DOAR-Infineon: Updated

boot/bootutil/src/xip_enc/xip_enc_ecies.c
DOAR-Infineon: Updated

boot/zephyr/Kconfig
DOAR-Infineon: Updated

docs/encrypted_images.md
DOAR-Infineon: Updated

boot/bootutil/src/bootutil_priv.h
DOAR-Infineon: Updated

boot/bootutil/include/bootutil/bootutil.h
DOAR-Infineon: Updated

docs/encrypted_images.md
DOAR-Infineon: Updated

docs/PORTING.md
DOAR-Infineon: Updated

docs/encrypted_images.md
DOAR-Infineon: Guard added. Zero IV is rejected now

DOAR-Infineon:
@d3zd3z 
Added negative tests to simulator

DOAR-Infineon:
@d3zd3z ,
please review

sim/mcuboot-sys/build.rs
```diff
@@ -574,13 +575,64 @@ fn main() {
...
+
+        conf.file("csupport/keys.c");
+
+        conf.file("../../ext/tinycrypt/lib/source/utils.c");
+        conf.file("../../ext/tinycrypt/lib/source/sha256.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc_dsa.c");
+        conf.file("../../ext/tinycrypt/lib/source/ecc_platform_specific.c");
+
+        conf.file("../../ext/mbedtls/library/platform_util.c");
```
d3zd3z: There is no longer an `ext/mbedtls` directory, but an `ext/mbedtls-3.6.0` and `ext/mbedtls-4.1.0`. You'll have to figure out which version is being brought into this build and use it explicitly. This is the reason for the failures.

JarmouniA (commented):
general observation: XIP is not exclusive to Ext Flash, so it should be clarified if the added feature could work for internal Flash as well (provided the SoC has the necessary hw decrypt engine). Regarding usage with ext Flash, the feature is really about mem-mapped ext Flash and not XiP specifically, which is just one usecase of mem-mapping, the CPU could be reading a code section to either execute it or just copy it to RAM then execute it, or could be reading data from the concerned Flash section, this also needs to be clarified. MCUboot swap upgrade modes support both XIP and normal RAM execution.

DOAR-Infineon:
> general observation: XIP is not exclusive to Ext Flash, so it should be clarified if the added feature could work for internal Flash as well (provided the SoC has the necessary hw decrypt engine). Regarding usage with ext Flash, the feature is really about mem-mapped ext Flash and not XiP specifically, which is just one usecase of mem-mapping, the CPU could be reading a code section to either execute it or just copy it to RAM then execute it, or could be reading data from the concerned Flash section, this also needs to be clarified. MCUboot swap upgrade modes support both XIP and normal RAM execution.

This implementation is mainly intended to be used with IFX PSE84 and external memory, but also is generic and memory type agnostic - so can be used with internal flash or other NVM.

 The implemented algo doesn't configure the encryption for hardware - it just extracts the encryption key and verifies the image, hardware configuration is mcuboot library caller's (user) responsibility.

JarmouniA:
> > general observation: XIP is not exclusive to Ext Flash, so it should be clarified if the added feature could work for internal Flash as well (provided the SoC has the necessary hw decrypt engine). Regarding usage with ext Flash, the feature is really about mem-mapped ext Flash and not XiP specifically, which is just one usecase of mem-mapping, the CPU could be reading a code section to either execute it or just copy it to RAM then execute it, or could be reading data from the concerned Flash section, this also needs to be clarified. MCUboot swap upgrade modes support both XIP and normal RAM execution.
> 
> This implementation is mainly intended to be used with IFX PSE84 and external memory, but also is generic and memory type agnostic - so can be used with internal flash or other NVM.
> 
> The implemented algo doesn't configure the encryption for hardware - it just extracts the encryption key and verifies the image, hardware configuration is mcuboot library caller's (user) responsibility.

What about the last point: XIP vs mem-mapping

boot/bootutil/src/xip_enc/xip_enc_validate.c
```diff
@@ -0,0 +1,314 @@
...
+            FIH_RET(fih_rc);
+        }
+        {
+            uint32_t img_sz;
+#if defined(MCUBOOT_SWAP_USING_OFFSET)
+            img_sz = sz_it.tlv_end - sz_it.start_off;
+#else
+            img_sz = sz_it.tlv_end;
+#endif
+            if (img_sz > bootutil_max_image_size(state, fap)) {
```
d3zd3z, line 155: This check only happens inside the `state != NULL` case, so when `boot_serial.c` calls this with a null state, this bound check is not made. It likely isn't exploitable, but does overflow a buffer.

d3zd3z (changes requested):
Please also create a `docs/release-notes.d/` entry describing the changes.

Issue #2122, opened 2024-11-15
RFC: Precalculated swap log for supporting devices with bigger write block sizes


de-nordic:
### Where we are

At this point MCUboot, while moving sectors around, logs status of operations with three flags per sector. With devices with write block size of 4b it amounts to 12 bytes per moved sector. This becomes problematic when you want to make MCUboot work with devices that have bigger write block sizes; for example with devices that require 8 bytes this is already 24 bytes, and becomes 48 for devices with 16 byte write block size. When you get to devices that have blocks like 512 byte,  this become hard to handle.

Note: while using "sector" here I mean logical sector of N * erase-block-size of device with the biggest erase-block-size.

### What can we do?

The proposal here is to replace the swap log with information block, layout record, that would state desired state of slot and swap algorithm would just try to restore the desired state from what it has on flash.

The layout record could consist of array of sha of sectors in order they should appear in slot. Note that sha256 is 32 byte long so it is already shorter then log entry for 16 byte write-block-size device.

For example
```
    [ State header ]
    [ sha(sector0) ]
    [ sha(sector1) ]
    ...
    ...
    ...
   [ sha(sectorN) ]
   [ sha(layout record) ]
```

Where `sha` is whatever sha we select (could be configurable, sha256 seems fine), and `sectorX` i logical sector size block on device, within slot.
Each slot will have its own expected layout record, so there would be two such blocks. The layout record sha is just to make sure that the record is complete.

Note that, if we put aside header size, the log size is now just N * sizeof(sha) and is aligned to write-block-size as a whole not per log entry (actually it will be aligned to erase block size, or logical sector).

### Unresolved questions:
1) Where do we store this record?
   Probably storing logs at the end of slot is not a good idea and we should have reserved space outside of slots and the entire slot should be dedicated to application image; problem will be here with devices that have big erase sizes as the advantage of small log record would be killed by using entire large independent erase block size just to store it, and it is better for such devices to have that record inside slot. This could be probably configurable depending on device layout.
2) How to mark test/confirm.
 Generally if we just assume that MCUboot always attempts to restore layout according to the pre-defined log, then just appearance of the log would mean that the swap should happen, and confirmation could be just a removal of the log - this is again problem for devices with big erase blocks, so they would have to mark somehow that they wan to stay with tested layout anyway.
3) Should application carry its own layout record?
 Probably, specifically when signed. Also calculating record on device is risky, but should be evaluated.

[...]

PR #2688, opened 2026-04-04
boot: swap by fingerprint status

DOAR-Infineon:
Based on https://github.com/mcu-tools/mcuboot/issues/2122 implementation idea

# Swap Fingerprint: SHA256-Based Swap Progress Tracking

## Overview

The swap fingerprint feature (`MCUBOOT_SWAP_FINGERPRINT`) replaces per-step
status byte writes during image swap with pre-computed SHA256 fingerprint tables
and post-copy verification.

Instead of writing a progress byte to flash after each state transition (3 writes
per sector in SCRATCH mode), the bootloader:

1. Pre-computes expected sector hashes before the swap starts
2. Stores them in a dedicated flash partition (`FLASH_AREA_IMAGE_FINGERPRINT`)
3. Verifies each copy operation against the tables (post-copy integrity check)
4. On recovery from power failure, scans sector hashes to determine the resume point

## Motivation

The current swap implementations write status bytes to flash after each state
transition. This works but:

- Requires flash writes during the upgrade beyond the actual sector copies
- Status bytes track position only, not content integrity
- Fault injection during a status write can corrupt progress state

The fingerprint approach eliminates runtime status writes, adds content-level
verification of each copy operation, and provides tamper detection.

### Fingerprint Tables

Two SHA256 hash tables are pre-computed and stored in the dedicated fingerprint
partition:

- **Primary table[i]** = SHA256(secondary_sector_data[i]) — what primary[i]
  should contain after the swap completes
- **Secondary table[i]** = SHA256(primary_sector_data[i]) — what secondary[i]
  should contain after the swap completes

Content-only hashing (no address or device binding) allows the same hash to
verify data regardless of its current location — primary, secondary, or scratch.

### Dedicated Flash Partition

Fingerprint tables are stored in `FLASH_AREA_IMAGE_FINGERPRINT` (flash area
ID 6), a dedicated partition separate from both image slots. This partition is
never erased during the swap, eliminating the power failure vulnerability that
would occur if tables were stored in the primary trailer.

Partition layout:

```
+===================================+
| Primary fingerprint table         |
|   (N x 32 bytes)                  |
+-----------------------------------+
| Secondary fingerprint table       |
|   (N x 32 bytes)                  |
+-----------------------------------+
| Table checksum (SHA256, 32 bytes) |
+-----------------------------------+
| Step count (4 bytes + padding)    |
+-----------------------------------+
| MAGIC (16 bytes)                  |
+===================================+
```

Partition size: `BOOT_FINGERPRINT_MAX_ENTRIES * 64 + 32 + BOOT_MAX_ALIGN + 16`
bytes. For `MAX_ENTRIES=128`: 8248 bytes (round up to flash erase boundaries).

### Verified Copies

After each copy operation during the swap, `boot_write_status()` hashes the
destination and compares against the fingerprint table. A mismatch is a fatal
error — the swap is aborted. This replaces the status byte write with an
integrity verification.

For SCRATCH swap, verification per state:

| State | Operation | Verification |
|-------|-----------|-------------|
| STATE_0 | secondary[i] -> scratch | SHA256(scratch) == primary_table[i] |
| STATE_1 | primary[i] -> secondary[i] | SHA256(secondary[i]) == secondary_table[i] |
| STATE_2 | scratch -> primary[i] | SHA256(primary[i]) == primary_table[i] |

### Power Failure Recovery

On recovery, `swap_read_status_bytes()` scans sectors in swap order (last sector
first, matching `swap_run()` iteration) and compares live hashes against stored
fingerprints. For the interrupted sector, it also hashes scratch to determine the
exact resume state.

Two integrity layers protect against init-time power failure:

1. **MAGIC validation** — MAGIC is written last in the fingerprint partition.
   Its presence guarantees all prior writes completed.
2. **Table checksum** — SHA256 of both tables is verified before trusting them.

Recovery state machine (6 cases):

| primary | secondary | scratch | State | Resume |
|---------|-----------|---------|-------|--------|
| match | match | any | Fully swapped | Continue |
| no | no | no | Not started / scratch corrupted | STATE_0 |
| no | no | match | Scratch valid, primary pending | STATE_1 |
| no | match | match | Primary pending, secondary done | STATE_2 |
| no | match | no | Scratch lost after STATE_1 | Error |
| match | no | any | Corruption | Error |


d3zd3z:
I'll point out one discovery that I made when trying to implement this. If a sector ends up containing the same data before and after the operation, it is _important_ to not even try the erase and write operation, just leave it there. Otherwise, you end up in this state where you can't distinguished between: an erase that just barely started, but hasn't robustly changed any data (so might later read with more bits set), the last write that mostly finished, but the data wasn't written fully, and the operation completing.

The other part (that depends on the kind of device) is that if the data page is to be all 0xff, there shouldn't be writes, just leave it erased.

PR #2690, opened 2026-04-09
boot: bootutil: Add a hook for implement image's pre-copying action (encrypted XIP support)

stindaNXP:
This PR adds a generic hook for implementing image pre-copying actions in boot_copy_image(). 

The placement of this hook is ideal for vendors with specific hardware encryption engines who would like to implement encrypted XIP and cannot use SWAP upgrade mode, making OVERWRITE_ONLY upgrade mode the only viable option. Some additional context is available here: https://github.com/mcu-tools/mcuboot/pull/2685#issuecomment-4200588824

Typically, before an image is written (re-encrypted) into the primary slot, the hardware encryption engine needs initialization based on the source image size. Another consideration is that some encryption engines implicitly consume more physical memory, so additional flash sectors must be pre-erased before the update.

This approach is successfully used in our NXP MCUboot fork. For more information see https://github.com/nxp-mcuxpresso/mcuxsdk-examples/blob/release/26.03.00/ota_examples/_doc/encrypted_xip_readme.md

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -962,6 +962,11 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
 #endif
+    rc = BOOT_HOOK_CALL(boot_copy_region_pre_hook, 0, BOOT_CURR_IMG(state),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY), src_size);
+    if (rc != 0) {
+        return rc;
+    }
 
     image_index = BOOT_CURR_IMG(state);
 
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -948,6 +948,13 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
 #endif
+    
+    rc = BOOT_HOOK_CALL(boot_copy_region_pre_hook, 0, BOOT_CURR_IMG(state),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_SECONDARY), src_size);
+    if (rc != 0) {
+        return rc;
+    }
 
     image_index = BOOT_CURR_IMG(state);
 
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -946,11 +946,18 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
 
     (void)bs;
 
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET) || defined(MCUBOOT_IMAGE_ACCESS_HOOKS)
     uint32_t src_size = 0;
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
 #endif
+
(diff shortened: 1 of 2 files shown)


boot/bootutil/src/loader.c
```diff
@@ -962,6 +962,11 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
 #endif
+    rc = BOOT_HOOK_CALL(boot_copy_region_pre_hook, 0, BOOT_CURR_IMG(state),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY), src_size);
```
nordicjm, line 966: doesn't really make sense that this is firing before copying the secondary to the primary, so is supplying with the primary image area

stindaNXP: What do you mean? Look at this line  https://github.com/nxp-upstream/mcuboot/blob/a31f8375f83e5642b20bb83d5f67bd886c9c0bf8/boot/bootutil/src/loader.c#L1068-L1069
Here is the post process hook also supplied with primary image area - I think original intention didn't take multi-image scenario in the account.

nordicjm: yes because the primary is now updated with the secondary, this is before, so should have both primary and secondary because they are different

stindaNXP:
Hi @nordicjm 
I have added a pointer to the secondary flash area and updated the PR. Could you please review?

nordicjm (approved):

boot/bootutil/src/loader.c
```diff
@@ -948,6 +948,13 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
 #endif
+    
+    rc = BOOT_HOOK_CALL(boot_copy_region_pre_hook, 0, BOOT_CURR_IMG(state),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY),
+                        BOOT_IMG_AREA(state, BOOT_SLOT_SECONDARY), src_size);
```
d3zd3z, line 954: `src_size` is only defined with `MCUBOOT_OVERWRITE_ONLY_FAST`, and this only compiling because the args are ignored with `MCUBOOT_IMAGE_ACCESS_HOOKS` off. For example, if `CONFIG_BOOT_BOOTSTRAP=y` and `CONFIG_BOOT_IMAGE_ACCESS_HOOKS=y` it will fail to build.

d3zd3z (changes requested):
loader.c:951 has trailing white space.

The commit text appears to be two squashed commits, neither of which has a subject line followed by a description.

You have a good description in the PR, that should be in the commit text, which is actually preserved in the repo.

boot/bootutil/src/loader.c
stindaNXP: Hi @d3zd3z
gentle ping on this one. I still see only three ways forward here:

Option 1 - Pass the boot loader state to the hook:

`   int boot_copy_region_pre_hook(struct boot_loader_state *state);`

   The state pointer already exposes everything a hook needs (image index, primary/secondary flash areas, and the size via `boot_read_image_size()`). The catch is that struct `boot_loader_state` is defined in the private `bootutil_priv.h`. A forward declaration keeps `boot_hooks.h` itself compiling, but any hook implementation that actually dereferences state would have to pull that private internal header into user code — exposing MCUboot internals through a public API.

Option 2 - Keep the current prototype and widen the compile guard so `src_size ` is declared whenever it can be referenced:
```
   #if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_IMAGE_ACCESS_HOOKS)
       uint32_t src_size = 0;
       rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
       assert(rc == 0);
   #endif
```
This is a minimal fix, keeps the public API limited to public types, and still hands a valid size to the hook.

Option 3 - Guard the entire hook call block
```
#if defined(MCUBOOT_IMAGE_ACCESS_HOOKS)
       {
           uint32_t src_size = 0;
           rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
           assert(rc == 0);
           rc = BOOT_HOOK_CALL(boot_copy_region_pre_hook, 0, ...);
           if (rc != 0) {
               return rc;
           }
       }
   #endif
```
This scopes src_size to exactly the hook usage and makes intent obvious, though it slightly duplicates the read that OVERWRITE_ONLY_FAST also does. Cleaner separation of concerns, marginally more code.

I'd lean toward option 2 or 3, but I'm happy to go with option 1 if you prefer the state-based API. Let me know which you'd prefer.

PR #2693, opened 2026-04-13
Fix encryption

valeriosetti:
- Remove accessing the "ver" field of the RSA structure because this has been removed in TF-PSA-Crypto.
- Add PSA_WANTs for AES and CTR to support encrypted images.


nordicjm (approved):

de-nordic (approved):

lonkaars:
I believe this pull request introduces a regression since the version field is still always present in the private key as specified in RFC3447. When the version field is skipped this results in the version field value instead being read into the public modulus field, causing `rsa_check_context` to fail when attempting to actually use the key.

Since I couldn't find anyone else having issues with encrypted images I wanted to confirm before opening a PR with this fix proposal:
```diff
diff --git a/boot/bootutil/include/bootutil/crypto/rsa.h b/boot/bootutil/include/bootutil/crypto/rsa.h
index 3bc2af73..5bb603e4 100644
--- a/boot/bootutil/include/bootutil/crypto/rsa.h
+++ b/boot/bootutil/include/bootutil/crypto/rsa.h
@@ -234,12 +234,15 @@ bootutil_rsa_parse_private_key(bootutil_rsa_context *ctx, uint8_t **p, uint8_t *
         return -2;
     }
 
-#if !defined(TF_PSA_CRYPTO_VERSION_NUMBER)
+		int dummy;
+		int* version = &dummy;
     /* version - "ver" field was removed from RSA structure starting from TF-PSA-Crypto. */
-    if (mbedtls_asn1_get_int(p, end, &ctx->MBEDTLS_CONTEXT_MEMBER(ver)) != 0) {
+#if !defined(TF_PSA_CRYPTO_VERSION_NUMBER)
+		version = &ctx->MBEDTLS_CONTEXT_MEMBER(ver);
+#endif /* !TF_PSA_CRYPTO_VERSION_NUMBER */
+    if (mbedtls_asn1_get_int(p, end, version) != 0) {
         return -3;
     }
-#endif /* !TF_PSA_CRYPTO_VERSION_NUMBER */
 
     /* Non-optional fields. */
     if ( /* public modulus */
```

valeriosetti:
> When the version field is skipped this results in the version field value instead being read into the public modulus field, causing rsa_check_context to fail when attempting to actually use the key.

@lonkaars yeah, this really looks like a mistake from me, sorry.

> Since I couldn't find anyone else having issues with encrypted images I wanted to confirm before opening a PR with this fix proposal:

IMO the fix you proposed is correct.

PR #2694, opened 2026-04-13
imgtool: fix sign command with no signing key

maulik-arm:
## Summary
Fix imgtool sign so it can create unsigned images when no signing key is provided.

  ## Problem
The command help says imgtool sign can create signed or unsigned images, but invoking it without --key fails before image creation starts.

The failure happens because loaded_keys is set to None when no key is provided, and the code unconditionally accesses the first entry to validate the encryption key type.

  ## Fix
  Guard access to the first signing key so the validation logic only inspects it when a signing key is actually present.


davidvincze:
LGTM

davidvincze (approved):

PR #2698, opened 2026-04-14
bootutil: Fix format string in swap_run function log

MariuszSkamra:
Use the correct unsigned integer format specifier in swap_run logging to avoid format mismatch errors.
"format '%d' expects argument of type 'int', but argument 3 has type 'uint32_t' {aka 'long unsigned int'} [-Werror=format=]"


de-nordic (approved):

MariuszSkamra:
fixed missing signoff

nordicjm (approved):

PR #2704, opened 2026-04-21
imgtool: python version coverage

JPHutchins:
Drop 3.6 and 3.7 support. Update CI with coverage for all supported versions, including those that are EOL (3.8, 3.9), and near EOL (3.10).

Working on imgtool got me nervous about the wide range of python versions. I did find python 3.6 and 3.7 to be unsupported at this time and updated the setup.py accordingly. My opinion is that it's OK since those are extremely far past end of life.

Tested all on local, but also got a good run here: https://github.com/intercreate/mcuboot/actions/runs/24748886181

### DIFF
--- a/scripts/imgtool/keys/general.py
+++ b/scripts/imgtool/keys/general.py
@@ -17,7 +17,7 @@ class FileHandler:
         self.kwargs = kwargs
 
     def __enter__(self):
-        if isinstance(self.file_in, str | bytes | os.PathLike):
+        if isinstance(self.file_in, (str, bytes, os.PathLike)):
             self.file = open(self.file_in, *self.args, **self.kwargs)
         else:
             self.file = self.file_in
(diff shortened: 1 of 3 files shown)


scripts/imgtool/keys/general.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, file, *args, **kwargs):
         self.kwargs = kwargs
 
     def __enter__(self):
-        if isinstance(self.file_in, str | bytes | os.PathLike):
+        if isinstance(self.file_in, (str, bytes, os.PathLike)):
```
de-nordic, line 20: How does the change correspond to version change?

JPHutchins: It is required for <3.10, versions that cannot use PEP604 https://peps.python.org/pep-0604/

de-nordic (approved):

d3zd3z (approved):

PR #2705, opened 2026-04-22
bootutil: add fallthrough to fix implicit fallthrough warning

supperthomas:
Add  /* fallthrough */ annotation in the switch statement within boot_set_next() to explicitly mark the intended fallthrough from BOOT_MAGIC_UNSET to BOOT_MAGIC_GOOD, suppressing the compiler warning.

### DIFF
--- a/boot/bootutil/src/bootutil_public.c
+++ b/boot/bootutil/src/bootutil_public.c
@@ -622,7 +622,7 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
         if (rc != 0) {
             break;
         }
-        /* Pass */
+        /* fallthrough */
 
     case BOOT_MAGIC_GOOD:
         if (confirm) {


supperthomas:
@nordicjm

boot/bootutil/src/bootutil_public.c
```diff
@@ -622,7 +622,7 @@ boot_set_next(const struct flash_area *fa, bool active, bool confirm)
         if (rc != 0) {
             break;
         }
-        /* Pass */
+        /* fallthrough */
```
nordicjm, line 625: not sure I follow this, it appears to be a C23 feature and is done using [] style brackets https://en.cppreference.com/c/language/attributes/fallthrough - not that doing that would be acceptable in MCUboot because of supporting old compilers. Do you have a reference to something this fixes?

supperthomas: https://github.com/mcu-tools/mcuboot/blob/95a6e388b3c3ac2aba7f33a15e36b30f2dbfc6db/boot/bootutil/src/loader.c#L1838-L1842

When I enable the fallthrough compilation warning as an error, an error is reported during compilation at this code segment. I have made corresponding modifications by referring to other loader.c files, which should theoretically comply with the modification conventions of MCUboot. If the old compiler fails to handle this part properly, other code areas would presumably encounter the same issue as well.

supperthomas: @nordicjm So, how to fix for the warning error?

nordicjm: @d3zd3z

d3zd3z: I think it is reasonable to make the fallthrough annotation match the rest of the code, especially since GCC does seem to care about it. I'm not sure a C23 fallthrough annotation is all that useful, as most users aren't using a c23 compiler, and we'd need the comment anyway. Maybe that can be added in the future.

d3zd3z (approved):

nordicjm:
I just don't really see the point, here's what happens when you use `/* Fallthrough */` with clang:
```
blah.c:14:1: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
   14 | case 1:
      | ^
blah.c:14:1: note: insert '__attribute__((fallthrough));' to silence this warning
   14 | case 1:
      | ^
      | __attribute__((fallthrough)); 
blah.c:14:1: note: insert 'break;' to avoid fall-through
   14 | case 1:
      | ^
      | break; 
blah.c:17:1: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
   17 | case 2:
      | ^
blah.c:17:1: note: insert '__attribute__((fallthrough));' to silence this warning
   17 | case 2:
      | ^
      | __attribute__((fallthrough)); 
blah.c:17:1: note: insert 'break;' to avoid fall-through
   17 | case 2:
      | ^
      | break; 
blah.c:20:1: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
   20 | default:
      | ^
blah.c:20:1: note: insert '__attribute__((fallthrough));' to silence this warning
   20 | default:
      | ^
      | __attribute__((fallthrough)); 
blah.c:20:1: note: insert 'break;' to avoid fall-through
   20 | default:
      | ^
      | break; 
4 warnings generated.
```

de-nordic (approved):

PR #2707, opened 2026-04-23
Add LMS signature support

d3zd3z:
Note, this is based on #2713, and includes all of the commits from that PR.  This will be simplified once that PR merges.

Add support for LMS signatures.  This uses Mbed TLS 4.1's newly added support for LMS signatures. Only sha256 w8 h10 is supported as this is the only configuration supported by Mbed TLS.

This adds support for LMS to:
- imgtool, which can now generate keys, and sign and verify images with LMS.
- sim, which also generates keys, and signs and tests images with LMS.
- boot/bootutil, which can verify these signatures.
Currently, only simulator support is present, and no platforms support this yet.

LMS comes with a strong security caveat:
**WARNING**: LMS has stateful keys. A given private key is only able to sign a fixed number of messages (1024 with this configuration). Effectively, a generated private key has 1024 subkeys, each of which is able to sign a single message. If this is violated (for example the private key copied to another machine, or backed up and restored), and more than one different message is signed with the same subkey, the _entire_ private key will be compromised. This generally requires special handling procedures with the signing processes.


OleksandrShkurchenko:
@d3zd3z In case of Post Quantum Algorithms(PQC) (not LMS only) we should consider supporting of Hybrid Mode for signature validation (or we can call it: Double signature mode)  as additional  option.  
Reason: many users still don`t trust existing PQC algorithms and consider ESDSA as reliable at this moment.

d3zd3z:
> @d3zd3z In case of Post Quantum Algorithms(PQC) (not LMS only) we should consider supporting of Hybrid Mode for signature validation (or we can call it: Double signature mode) as additional option. Reason: many users still don`t trust existing PQC algorithms and consider ESDSA as reliable at this moment.

Possibly for ML-DSA as there is concerns about it's newness. LMS does not have these concerns, and other than the challenges with the stateful signing key, and the larger signatures, there is little reason to not just use LMS.

To do hybrid, probably that needs to be done with a TLV mechanism to have _AND_ signatures instead of just _OR_ signatures as it is now.

But, again, I don't think there is any reason to do that with LMS. LMS is arguably better understood than EC.

kkrentz:
I agree that a hybrid mode is questionable here since LMS' security rests on fewer assumptions than ECDSA.

> **WARNING**: LMS has stateful keys. A given private key is only able to sign a fixed number of messages (1024 with this configuration). Effectively, a generated private key has 1024 subkeys, each of which is able to sign a single message. If this is violated (for example the private key copied to another machine, or backed up and restored), and more than one different message is signed with the same subkey, the _entire_ private key will be compromised. This generally requires special handling procedures with the signing processes.

To reduce the risk of key reuse, [this paper](https://arxiv.org/pdf/2106.05577) suggests reusing the version number of the firmware image as the LMS sequence number. In that case, the key does not have to be stateful because the state is somewhat externally managed. I actually prototyped their approach some time ago.

d3zd3z:
> To reduce the risk of key reuse, [this paper](https://arxiv.org/pdf/2106.05577) suggests reusing the version number of the firmware image as the LMS sequence number. In that case, the key does not have to be stateful because the state is somewhat externally managed. I actually prototyped their approach some time ago.

One of the challenges with that is that it places constraints on the version numbers that can be used, and has additional questions as well (these two things aren't really correlated at all), the there might be different major version bumps, whereas a simple sequence number would kind of imply just a sequence.

The issue is very real, though, and maybe there is value in encoding the key number more clearly in the image (I'm not actually sure if you can deduce it from the signature itself). But, I think we want to be careful in terms how of how much we try to impose on the signing side and infrastructure.

kkrentz:
> The issue is very real, though, and maybe there is value in encoding the key number more clearly in the image (I'm not actually sure if you can deduce it from the signature itself). But, I think we want to be careful in terms how of how much we try to impose on the signing side and infrastructure.

RFC 8554 calls an LMS sequence number "q". q comes at the beginning of an LMS signature. So, one could at least validate that a new version also has a new q.

Stephane-LeRoy:
Besides LMS signature, is there plan to support a PQC algorithm to distribute the symmetric encryption key? 
If so, which one?
To my understanding, for PQC compliance, both are required.

d3zd3z:
> Besides LMS signature, is there plan to support a PQC algorithm to distribute the symmetric encryption key? If so, which one? To my understanding, for PQC compliance, both are required.

This change only addresses the LMS signatures, primarily because that functionality has been added to mbed TLS. Other aspects will be addresses as the implementations settle in mbed TLS.

MCUboot can be configured (and is commonly used) with just signatures and no encryption, so this change makes sense in isolation.

PR #2711, opened 2026-04-24
bootutil: mynewt: Fix boot_serial unit tests

sjanc:
When in unittest we shall include all sources.

### DIFF
--- a/boot/bootutil/pkg.yml
+++ b/boot/bootutil/pkg.yml
@@ -47,7 +47,7 @@ pkg.ign_files:
     - "ed25519_psa.c"       # Currently no PSA for mynewet
     - "encrypted_psa.c"
 
-pkg.source_files.!BOOT_LOADER:
+pkg.source_files.'!BOOT_LOADER && !SELFTEST':
     - "src/bootutil_public.c"
 
 pkg.deps.BOOTUTIL_USE_MBED_TLS:


sjanc:
@m-gorecki FYI

d3zd3z:
For example, here is a more detailed commit text, feel free to use and/or adapt.
```
  bootutil: mynewt: Fix boot_serial unit tests

  Commit e0152095 ("bootutil: mynewt: Improve BOOTUTIL configuration")
  added a pkg.source_files.!BOOT_LOADER override so that applications
  depending on bootutil only build bootutil_public.c, avoiding the
  need to set bootloader-only syscfgs for plain application builds.

  This override also fires when the boot_serial test package
  (pkg.type: unittest) is built, since BOOT_LOADER is not defined in
  that case. The result is that the unittest binary is linked against
  only bootutil_public.c and fails to resolve the rest of the
  bootutil symbols it exercises.

  Restrict the override to non-SELFTEST builds so unit tests get the
  full bootutil source list while application builds remain
  unaffected.

  Signed-off-by: Szymon Janc <szymon.janc@codecoup.pl>
```

d3zd3z (approved):

sjanc:
gentle ping :)

d3zd3z (approved):

MariuszSkamra (approved):

de-nordic (approved):

PR #2712, opened 2026-04-24
Mynewt CI update

sjanc:


sjanc:
CI failure doesn't seem to be related to this change

d3zd3z:
The CI failures aren't related.

Some suggested commit text, although I'm not completely sure the reason for the 48K change, but it'd be good to capture:
```
  Commit 1 — b78a15aa ci: mynewt: Update actions

  ci: mynewt: modernize toolchain and inline workflow

  Replace the hand-rolled ARM toolchain install (wget of a 2017
  GCC 7 release) with carlosperate/arm-none-eabi-gcc-action@v1
  pinned to 15.2.Rel1, and install newt via "go install" instead
  of cloning and building newt from source. Both removals shed
  ~110 lines of bootstrap shell that only existed to recreate
  what the upstream actions now provide.

  Inline the install and run logic into the workflow itself and
  delete the ci/mynewt_install.sh and ci/mynewt_run.sh wrappers.

  After "newt upgrade", drop the PR checkout into repos/mcuboot
  via "git clone ..", so the build exercises the mcuboot revision
  under review rather than whatever upstream tag newt would
  otherwise resolve. Add mcuboot to project.repositories.allowed
  in ci/mynewt_project.yml so newt accepts the local repo.

  The redundant "Signed commit check" step is dropped here; DCO
  is already enforced by the sim, fih_tests, and espressif
  workflows.

  Signed-off-by: Szymon Janc <...>

  ---
  Commit 2 — e6367972 ci: mynewt: Fix RSA targets build

  ci: mynewt: enlarge bootloader slot for rsa_rsaoaep

  The rsa_rsaoaep target no longer fits in the default BSP
  bootloader flash region after the toolchain bump to GCC 15.2,
  which produces a larger image than the previous GCC 7 build.

  Set BSP_BOOTLOADER_48K=1 in the target's syscfg.yml so the
  BSP reserves a 48K bootloader slot and the build links cleanly
  again.

  Signed-off-by: Szymon Janc <...>
```

d3zd3z (approved):
Looks good, but the CI failures are going to be annoying. Hopefully we manage to fix overall CI.

de-nordic (approved):

PR #2714, opened 2026-04-24
Prepare for TF-M v2.3

valeriosetti:
Apply changes that are required for TF-M v2.3:

- use a HEAD reference after v2.4.0, which is the one [being pointed to TF-M](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/refs/tags/TF-Mv2.3.0/config/config_base.cmake#43)
- apply TF-M's [extra patch](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/refs/tags/TF-Mv2.3.0/lib/ext/mcuboot/0001-bootutil-Parse-key-ID-for-built-in-keys.patch)
- add some extra fix which is required to resolve a build issue in Zephyr and to make the CI passing in this PR

This is being tested from https://github.com/zephyrproject-rtos/zephyr/pull/107373


d3zd3z:
This is failing because the committer (@valeriosetti ) also needs to add a Signed-off-by lines, per the DCO.

d3zd3z (changes requested):
The assert should be an explicit check.

Also, there are no tests for this functionality. I realize that the key management here is for out of tree code, so it's kind of hard to test without inventing an implementation for the simulator. But at least something to consider.

valeriosetti:
> This is failing because the committer (@valeriosetti ) also needs to add a Signed-off-by lines, per the DCO.

Done. Thanks for the notification

valeriosetti:
The CI is failing here because:
```
/usr/lib/gcc/arm-none-eabi/10.3.1/../../../arm-none-eabi/bin/ld: ext/mcuboot/bootutil/libbootutil.a(bootutil_find_key.o): in function `bootutil_find_key':
/root/work/tfm/mcuboot/boot/bootutil/src/bootutil_find_key.c:66: undefined reference to `boot_verify_key_id_for_image'
```

Unfortunately `boot_verify_key_id_for_image` is defined in TF-M only (@davidvincze ) so I need to figure out how to solve this.

valeriosetti:
@d3zd3z can you please help me understand why the CI is hanging on `FIH hardening` tests?

boot/bootutil/include/bootutil/image.h
```diff
@@ -98,6 +98,7 @@ extern "C" {
  */
 #define IMAGE_TLV_KEYHASH           0x01    /* hash of the public key */
 #define IMAGE_TLV_PUBKEY            0x02    /* public key */
+#define IMAGE_TLV_KEYID             0x03    /* Signing key ID */
```
nordicjm, line 101: needs @d3zd3z and @de-nordic to review

boot/bootutil/src/bootutil_find_key.c
```diff
@@ -51,34 +51,34 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #ifdef EXPECTED_SIG_TLV
 #if !defined(MCUBOOT_BYPASS_KEY_MATCH)
 /* Find functions are only needed when key is checked first */
-#if !defined(MCUBOOT_BUILTIN_KEY)
-#if !defined(MCUBOOT_HW_KEY)
-int bootutil_find_key(uint8_t *keyhash, uint8_t keyhash_len)
+#if defined(MCUBOOT_BUILTIN_KEY)
+/* This is a weak stub function meant to be overridden by some strong
+ * implementation (ex: see TF-M). */
+__attribute__((weak)) fih_ret boot_verify_key_id_for_image(uint8_t image_index, uint32_t key_id)
```
nordicjm: MCUboot does not nor should it use weak functions, there is a whole method of doing things with hooks, please check/update, and returning failure just breaks all existing users requiring them to make a change they shouldn't have to

nordicjm (changes requested):

boot/bootutil/src/bootutil_find_key.c
valeriosetti: Does this address your comment https://github.com/mcu-tools/mcuboot/pull/2714/commits/67979cbfcdd7ea03e52e5f07fac869d04c70e8e6?
If so I will update the history of the PR to make it cleaner

Edit: I just tried this in https://github.com/zephyrproject-rtos/zephyr/pull/107373 and based on the definition of `#define MCUBOOT_VERIFY_KEY_ID_FOR_IMAGE_HOOKS` in TF-M I can still make the board to boot properly.

valeriosetti:
Following the discussion I had with @de-nordic @nordicjm I understood that this functionality is not ready yet to be upstreamed. Therefore we're ignoring the `IMAGE_TLV_KEYID` feature in Zephyr for now until it get to an acceptable state.
This PR can be closed as no more required for https://github.com/zephyrproject-rtos/zephyr/pull/107373

PR #2718, opened 2026-05-04
boot: use flash_area_flatten() for no-erase devices

jaz1-nordic:
For devices that report no explicit-erase capability, MCUboot currently scrambles the affected region by issuing a flash_area_write() of erase_value for every alignment-sized chunk. On RAM-like backings (RRAM/MRAM/FRAM/auto-erase NOR) this collapses to one tiny write per 4..16 byte chunk, each forcing the underlying driver to commit its hardware write buffer. Scrambling a 1 MiB slot during a downgrade prevention pass costs millions of buffer commits and seconds of wall time on nRF54L RRAM.

### DIFF
--- a/boot/bootutil/src/bootutil_area.c
+++ b/boot/bootutil/src/bootutil_area.c
@@ -37,6 +37,24 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
+#ifndef MIN
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+#endif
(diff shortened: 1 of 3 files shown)


nordicjm (changes requested):
this is not valid, the actual fix is to adjust the erase size in dts using an overlay, then pair with https://github.com/zephyrproject-rtos/zephyr/pull/105313

jaz1-nordic:
Dropped.

PR #2513, opened 2025-10-22
WIP: Update FIH docker to Ubuntu Plucky (25.04)

davidvincze:
This distribution includes a newer version (14.2) of the GNU Arm toolchain which is in line with [TF-M's CI environment](https://www.trustedfirmware.org/projects/open-ci).

PR #2721, opened 2026-05-05
Two updates for fih

d3zd3z:
This updates FIH, combining #2513, updating the underlying Ubuntu version to 'plucky', along with my patch to add LMS support to the python packages included in the docker image. This will make CI easier for the FIH tests, and use a version better matched by TF-M.


nordicjm (approved):

d3zd3z:
As far as I can tell, FIH is failing because CI is running off of the old docker image. Two options:
- Push a test FIH docker image through to see if CI passes.
- Conditionalize the script enough so that it will work with either.

de-nordic (approved):

PR #2723, opened 2026-05-08
boot: zephyr: Add PSOC Edge case in flash_map_extended.c

lauracarlesso:
Add the correct selection for FLASH_DEVICE for the Infineon PSOC Edge family of devices.

For PSOC Edge devices, the MCUBoot gets placed in RRAM but the slots are placed in external flash. Therefore the flash node used to calculate the base address for the jump to slot needs to be based on the chosen_flash_controller node.

### DIFF
--- a/boot/zephyr/flash_map_extended.c
+++ b/boot/zephyr/flash_map_extended.c
@@ -46,6 +46,11 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
 #define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_PARENT(FLASH_DEVICE_NODE), 1)
 
+#elif (defined(CONFIG_SOC_SERIES_PSE84) && DT_HAS_CHOSEN(zephyr_flash_controller))
+#define FLASH_DEVICE_ID SOC_FLASH_0_ID
+#define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
+#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(FLASH_DEVICE_NODE, 1)
+
 #elif (!defined(CONFIG_XTENSA) && DT_HAS_CHOSEN(zephyr_flash_controller))
 #define FLASH_DEVICE_ID SOC_FLASH_0_ID
 #define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS


boot/zephyr/flash_map_extended.c
```diff
@@ -46,6 +46,11 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
 #define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_PARENT(FLASH_DEVICE_NODE), 1)
 
+#elif (defined(CONFIG_SOC_SERIES_PSE84) && DT_HAS_CHOSEN(zephyr_flash_controller))
+#define FLASH_DEVICE_ID SOC_FLASH_0_ID
+#define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
+#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(FLASH_DEVICE_NODE, 1)
+
```
nordicjm, line 53: this makes no sense though, it should use mapped partition, as I said here and you said updated and was no longer needed https://github.com/mcu-tools/mcuboot/pull/2673#issuecomment-4132220134

nordicjm (changes requested):

boot/zephyr/flash_map_extended.c
lauracarlesso: the closed PR referenced was opened during the first attempt of integrating the MCUBoot support for the Edge boards, in that case all images (mcuboot + slots) were placed in the external flash but we were not using mapped partitions and I was resorting to updating (poorly) the base address for "jump to slot" purposes while maintaining it as is for all other use cases. However that implementation was discarded in favor of a new one.
In the new implementation ( https://github.com/zephyrproject-rtos/zephyr/pull/108938 ) the MCUBoot image is placed in internal RRAM (using mapped partition logic) and the slot images are placed in external flash (using mapped partition logic again). 
Without the changes in this PR the mcuboot properly verifies the images in the external memory (our serial memory uses 0x0 to EXT_FLASH_SIZE addresses for its operations) but at the "jump to slot" moment it tries to use its own flash base (rram base address) as the starting point to which adding the offset, instead of using the external flash base address. 
I have not been able to find any different way to handle this other than the changes suggested here, is there a different approach I could be using?

maass-hamburg:
Probably no longer needed due to https://github.com/mcu-tools/mcuboot/pull/2725#event-25446065570

lauracarlesso:
> Probably no longer needed due to [#2725 (comment)](https://github.com/mcu-tools/mcuboot/pull/2725#event-25446065570)

I was not aware of these changes being in progress, I'll retest my integration with these and verify this is no longer needed.

lauracarlesso:
Retested MCUBoot integration with latest changes in #2725 
All working as expected without need for these updates so closing this PR.
I guess my next question is when will these updates be brought into Zehpyr's mcuboot fork?

maass-hamburg:
> Retested MCUBoot integration with latest changes in #2725 All working as expected without need for these updates so closing this PR. I guess my next question is when will these updates be brought into Zehpyr's mcuboot fork?

Already in and merged with https://github.com/zephyrproject-rtos/zephyr/pull/108960

PR #2725, opened 2026-05-08
boot: zephyr: flash_map_extended: use PARTITION_* macros and remove SoC specific code

maass-hamburg:
use PARTITION_* macros and implement getting
the device id and the flash base based on
PARTITION_* macros. With this we no longer
need to have SoC specific code here.

Also removes `flash_area_sector_from_off()` as it is no longer needed.

### DIFF
--- a/boot/zephyr/flash_map_extended.c
+++ b/boot/zephyr/flash_map_extended.c
@@ -21,58 +21,97 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
-#if defined(CONFIG_STM32_MEMMAP)
-/* MEMORY MAPPED for XiP on external NOR flash takes the sspi-nor or ospi-nor or qspi-nor device */
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#if DT_NODE_HAS_STATUS(DT_INST(0, st_stm32_xspi_nor), okay)
-#define DT_DRV_COMPAT st_stm32_xspi_nor
-#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)
-#elif DT_NODE_HAS_STATUS(DT_INST(0, st_stm32_ospi_nor), okay)
-#define DT_DRV_COMPAT st_stm32_ospi_nor
-#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)
-#elif DT_NODE_HAS_STATUS(DT_INST(0, st_stm32_qspi_nor), okay)
-#define DT_DRV_COMPAT st_stm32_qspi_nor
-#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_INST_PARENT(0), 1)
-#else
-#error "FLASH_DEVICE_NODE could not be determined"
-#endif
+#define FLASH_PARTITION_ADDRESS(partition) (PARTITION_ADDRESS(partition) - PARTITION_OFFSET(partition))
 
-#elif (DT_NODE_HAS_COMPAT(DT_PARENT(DT_CHOSEN(zephyr_flash_controller)),       \
-			  nxp_imx_flexspi))
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
-#define FLASH_DEVICE_BASE DT_REG_ADDR_BY_IDX(DT_PARENT(FLASH_DEVICE_NODE), 1)
-
-#elif (!defined(CONFIG_XTENSA) && DT_HAS_CHOSEN(zephyr_flash_controller))
-#define FLASH_DEVICE_ID SOC_FLASH_0_ID
-#define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS
-
-#elif (defined(CONFIG_XTENSA) && DT_NODE_EXISTS(DT_INST(0, jedec_spi_nor)))
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#define FLASH_DEVICE_BASE 0
-
-#elif defined(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
-
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#define FLASH_DEVICE_BASE 0
-
-#elif (defined(CONFIG_SOC_SERIES_NRF54H) && DT_HAS_CHOSEN(zephyr_flash))
-
-#define FLASH_DEVICE_ID SOC_FLASH_0_ID
-#define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS
-
-#else
-#error "FLASH_DEVICE_ID could not be determined"
-#endif
+#define FLASH_PARTITION_DEV_UNIQUE(n, m) (PARTITION_EXISTS(n) && \
+                                          !(DT_SAME_NODE(PARTITION_MTD(n), PARTITION_MTD(m))))
 
 int flash_device_base(uint8_t fd_id, uintptr_t *ret)
 {
-    if (fd_id != FLASH_DEVICE_ID) {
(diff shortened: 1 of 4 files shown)


boot/zephyr/flash_map_extended.c
```diff
@@ -21,58 +21,55 @@
...
-
-#elif defined(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
-
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#define FLASH_DEVICE_BASE 0
-
-#elif (defined(CONFIG_SOC_SERIES_NRF54H) && DT_HAS_CHOSEN(zephyr_flash))
+#define FLASH_DEVICE_BASE (FLASH_PARTITION_ADDRESS(slot0_partition))
 
+#if IN_RANGE(CONFIG_FLASH_BASE_ADDRESS, FLASH_DEVICE_BASE, PARTITION_ADDRESS(slot0_partition))
```
nordicjm: this seems like a false positive, e.g. imagine a device with internal and external flash on SPI bus where internal flash is at 0x0, external flash would _also_ be at 0x0, so slot0 and slot1 would clash. Should instead check the node, see https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/mgmt/mcumgr/grp/img_mgmt/src/img_mgmt.c#L47 for examples. Also flash base address is not valid in future anyhow with mapped partitions the PARTITION_* macros should be used

maass-hamburg: as the exact value of that is not relevant, only `flash_device_base()` and `flash_area_get_device_id()` have to work with it, we could remove that right?

JarmouniA:
Cc @erwango @sylvioalves @JiafeiPan

boot/zephyr/flash_map_extended.c
```diff
@@ -21,67 +21,47 @@
...
-#define FLASH_DEVICE_NODE DT_INST(0, jedec_spi_nor)
-
-#elif defined(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
-
-#define FLASH_DEVICE_ID SPI_FLASH_0_ID
-#define FLASH_DEVICE_BASE 0
-#define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash_controller)
-
-#elif (defined(CONFIG_SOC_SERIES_NRF54H) && DT_HAS_CHOSEN(zephyr_flash))
+#define FLASH_PARTITION_ADDRESS(partition) (PARTITION_ADDRESS(partition) - PARTITION_OFFSET(partition))
```
sylvioalves, line 24: Won't this make the bootloader jump to address 0 + image_off instead of the XIP window on memmap targets?

boot/zephyr/flash_map_extended.c
```diff
@@ -163,26 +143,40 @@ int flash_area_id_from_direct_image(int image_id)
...
-    sector->fs_size = page.size;
-
-    return rc;
-}
-
-uint8_t flash_area_get_device_id(const struct flash_area *fa)
-{
-    (void)fa;
-    return FLASH_DEVICE_ID;
+    return 0U;
```
sylvioalves, line 281: This falls thru a valid return 0 value (slot0-partition). Could the fallback log and return a sentinel like UINT8_MAX so flash_device_base() reports -EINVAL instead of silently producing a wrong address?

boot/zephyr/flash_map_extended.c
maass-hamburg: PARTITION_ADDRESS returns the memory mapped address of that partition. It worked fine for the `stm32h573i_dk` and `stm32h573i_dk/stm32h573xx/ext_flash_app` target. There the internal flash (target: `stm32h573i_dk`) already uses `zephyr,mapped-partition`, while the external flash (target: `stm32h573i_dk/stm32h573xx/ext_flash_app`) still uses `fixed-partitions`

boot/zephyr/flash_map_extended.c
maass-hamburg: What do you think about using an assert instead?  Because it is not something that will happen "randomly", it would be a configuration problem

sylvioalves: Sounds good.

boot/zephyr/flash_map_extended.c
```diff
@@ -21,67 +21,47 @@
...
-#define FLASH_DEVICE_BASE CONFIG_FLASH_BASE_ADDRESS
-#define FLASH_DEVICE_NODE DT_CHOSEN(zephyr_flash)
-
-#else
-#error "FLASH_DEVICE_ID could not be determined"
-#endif
-
-static const struct device *flash_dev = DEVICE_DT_GET(FLASH_DEVICE_NODE);
+#define FLASH_PARTITION_DEV_UNIQUE(n, m) (PARTITION_EXISTS(n) && \
+                                          (FLASH_PARTITION_ADDRESS(n) != FLASH_PARTITION_ADDRESS(m)))
```
nordicjm: this needs to check if the flash controller node is the same or not as per other comment

boot/zephyr/flash_map_extended.c
```diff
@@ -21,67 +21,47 @@
...
+        break;
+#endif
+#if FLASH_PARTITION_DEV_UNIQUE(slot4_partition, slot3_partition)
+    case 4U:
+        *ret = FLASH_PARTITION_ADDRESS(slot4_partition);
+        break;
+#endif
+#if FLASH_PARTITION_DEV_UNIQUE(slot5_partition, slot4_partition)
+    case 5U:
+        *ret = FLASH_PARTITION_ADDRESS(slot5_partition);
```
nordicjm, line 58: needs to go up to `slot15_partition`

boot/zephyr/flash_map_extended.c
```diff
@@ -163,26 +143,42 @@ int flash_area_id_from_direct_image(int image_id)
...
-    struct flash_pages_info page;
+    const struct device *dev = flash_area_get_device(fa);
 
-    rc = flash_get_page_info_by_offs(flash_dev, off, &page);
-    if (rc) {
-        return rc;
+    if (dev == PARTITION_DEVICE(slot0_partition)) {
+        return 0U;
     }
+    #if FLASH_PARTITION_DEV_UNIQUE(slot1_partition, slot0_partition)
```
nordicjm:
```suggestion
#if FLASH_PARTITION_DEV_UNIQUE(slot1_partition, slot0_partition)
```
no indent on these

boot/zephyr/flash_map_extended.c
```diff
@@ -163,26 +143,42 @@ int flash_area_id_from_direct_image(int image_id)
...
+        return 3U;
+    }
+    #endif
+    #if FLASH_PARTITION_DEV_UNIQUE(slot4_partition, slot3_partition)
+    if (dev == PARTITION_DEVICE(slot4_partition)) {
+        return 4U;
+    }
+    #endif
+    #if FLASH_PARTITION_DEV_UNIQUE(slot5_partition, slot4_partition)
+    if (dev == PARTITION_DEVICE(slot5_partition)) {
```
nordicjm, line 224: likewise up to slot15

boot/zephyr/flash_map_extended.c
```diff
@@ -163,26 +143,42 @@ int flash_area_id_from_direct_image(int image_id)
...
-
-    return rc;
-}
+    __ASSERT(0, "invalid flash area device");
 
-uint8_t flash_area_get_device_id(const struct flash_area *fa)
-{
-    (void)fa;
-    return FLASH_DEVICE_ID;
+    return 0U;
```
nordicjm, line 281: should probably return non-zero? i.e. with asserts disabled, the device should not boot if this is the case

maass-hamburg: we should never reach this and if we do mcuboot would be broken. And by having the return value the same as the first as the one of the flash dev of slot 1, the compiler can optimize it away.

maass-hamburg: I mean i could change that to UINT8_MAX, but then the image would be slightly biger

nordicjm (approved):
very nice optimisation! Thanks

de-nordic (approved):

PR #2728, opened 2026-05-12
boot: zephyr: remove selection of CONFIG_PSA_CRYPTO_CLIENT

tomi-font:
The Kconfig option is removed in Zephyr.

Additionally, clean up the related selects and replace them by just CONFIG_PSA_CRYPTO which is the only option that should be enabled manually when requesting PSA Crypto support.


nordicjm (approved):

tomi-font:
@d3zd3z or @davidvincze if you could give this tiny PR a quick review, would be much appreciated!

de-nordic (approved):

PR #2732, opened 2026-05-14
bootutil: Use mbedtls/private paths for TF-PSA-Crypto

krish2718:
MCUboot Mbed TLS crypto wrappers and simulator custom crypto helpers included legacy headers as mbedtls/<alg>.h. TF-PSA-Crypto now exposes these declarations only under mbedtls/private/.

Assisted-by: Cursor: Auto

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/common.h
+++ b/boot/bootutil/include/bootutil/crypto/common.h
@@ -7,6 +7,21 @@
 #ifndef __BOOTUTIL_CRYPTO_COMMON_H__
 #define __BOOTUTIL_CRYPTO_COMMON_H__
 
+/*
+ * TF-PSA-Crypto keeps several legacy crypto declarations under the
+ * mbedtls private include tree (e.g. mbedtls/private/sha256.h).
+ * Classic Mbed TLS exposes the same APIs under mbedtls/sha256.h.
+ * Unless the build forces MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR to 0 or 1,
+ * detect at preprocess time whether the private sha256 header exists.
+ */
+#if !defined(MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR)
+#if defined(__has_include) && __has_include(<mbedtls/private/sha256.h>)
+#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 1
+#else
+#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 0
+#endif
+#endif
+
 /* The check below can be performed even for those cases
  * where MCUBOOT_USE_MBED_TLS has not been defined
  */
--- a/boot/bootutil/include/bootutil/crypto/ecdh_p256.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdh_p256.h
@@ -19,9 +19,13 @@
 #endif
 
 #if defined(MCUBOOT_USE_MBED_TLS)
-    #include <mbedtls/ecp.h>
-    #include <mbedtls/ecdh.h>
-    #define EC256_PUBK_LEN (65)
+#include "bootutil/crypto/common.h"
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
+#include <mbedtls/private/ecp.h>
+#else
+#include <mbedtls/ecp.h>
+#endif
+#define EC256_PUBK_LEN (65)
 #endif /* MCUBOOT_USE_MBED_TLS */
 
 #if defined(MCUBOOT_USE_TINYCRYPT)
@@ -75,6 +79,35 @@ static inline int bootutil_ecdh_p256_shared_secret(bootutil_ecdh_p256_context *c
 #if MBEDTLS_VERSION_NUMBER >= 0x03000000
 static int fake_rng(void *p_rng, unsigned char *output, size_t len);
 #endif
+
+static inline int bootutil_ecdh_compute_shared(mbedtls_ecp_group *grp, mbedtls_mpi *z,
+                                               const mbedtls_ecp_point *Q,
+                                               const mbedtls_mpi *d,
+                                               int (*f_rng)(void *, unsigned char *, size_t),
+                                               void *p_rng)
+{
+    int ret;
+    mbedtls_ecp_point P;
+
+    mbedtls_ecp_point_init(&P);
+
+    ret = mbedtls_ecp_mul(grp, &P, d, Q, f_rng, p_rng);
+    if (ret != 0) {
+        goto cleanup;
+    }
+
+    if (mbedtls_ecp_is_zero(&P) != 0) {
+        ret = MBEDTLS_ERR_ECP_BAD_INPUT_DATA;
+        goto cleanup;
+    }
+
+    ret = mbedtls_mpi_copy(z, &P.MBEDTLS_CONTEXT_MEMBER(X));
+
+cleanup:
+    mbedtls_ecp_point_free(&P);
+
+    return ret;
+}
 
 typedef struct bootutil_ecdh_p256_context {
     mbedtls_ecp_group grp;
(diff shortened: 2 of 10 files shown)


d3zd3z (commented):
This will probably take a bit to get to be "right". One thing I'm curious about (I might investigate this next week) is if this change is sufficient to get everything to pass against mbed TLS 4.1. Change #2713 ends up bringing in both 4.1 and 3.6 because of this issue.

My other concern is how can we encourage the use of the private headers to "go away" at some point. We definitely want to make sure we aren't introducing new instances.

krish2718:
> This will probably take a bit to get to be "right". One thing I'm curious about (I might investigate this next week) is if this change is sufficient to get everything to pass against mbed TLS 4.1. Change #2713 ends up bringing in both 4.1 and 3.6 because of this issue

Ah I see, so, if we are maintaining both 3.6 and 4.1 version, then can we absorb this change in 4.1 specific code?

> My other concern is how can we encourage the use of the private headers to "go away" at some point. We definitely want to make sure we aren't introducing new instances.

That's actually one of the reasons for this PR (and associated ones), the current Zephyr copies these headers and drops "private", so, the users would have no idea that they are using headers that could be removed. And as for the actual deletion, it depends on migrating to PSA only versions and that might take a while esp. if the ecosystem continues to use these deprecated algos.

krish2718:
@d3zd3z The CI passed, is that enough to merge this? This is the only thing blocking https://github.com/zephyrproject-rtos/zephyr/pull/108979/ merge.

krish2718:
> @d3zd3z The CI passed, is that enough to merge this? This is the only thing blocking [zephyrproject-rtos/zephyr#108979](https://github.com/zephyrproject-rtos/zephyr/pull/108979) merge.

Gentle reminder @d3zd3z

de-nordic (approved):

boot/bootutil/include/bootutil/crypto/ecdh_p256.h
```diff
@@ -19,8 +19,13 @@
...
 #if defined(MCUBOOT_USE_MBED_TLS)
-    #include <mbedtls/ecp.h>
-    #include <mbedtls/ecdh.h>
+#include "bootutil/crypto/common.h"
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
+#include <mbedtls/private/ecp.h>
+#else
+#include <mbedtls/ecp.h>
+#endif
+#include <mbedtls/ecdh.h>
```
tomi-font: This header doesn't exist anymore?

boot/bootutil/include/bootutil/crypto/sha.h
```diff
@@ -62,11 +62,21 @@
...
+#else
 #include <mbedtls/sha512.h>
+#endif
+#else
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
+#include <mbedtls/private/sha256.h>
 #else
 #include <mbedtls/sha256.h>
 #endif
+#endif
```
tomi-font, line 79: Combine the two ifs?

sim/mcuboot-sys/csupport/custom_crypto/sim_custom_ecdh_p256.h
```diff
@@ -15,10 +15,16 @@
...
 #include <mbedtls/build_info.h>
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
+#include <mbedtls/private/ecp.h>
+#include <mbedtls/private/bignum.h>
+#else
 #include <mbedtls/ecp.h>
-#include <mbedtls/ecdh.h>
 #include <mbedtls/bignum.h>
+#endif
+#include <mbedtls/ecdh.h>
```
tomi-font: same remark (`ecdh.h` gone)

boot/bootutil/include/bootutil/crypto/common.h
```diff
@@ -7,6 +7,24 @@
...
+#if defined(MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR)
+#elif defined(__has_include)
+#if __has_include(<mbedtls/private/sha256.h>)
+#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 1
+#else
+#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 0
+#endif
+#else
+#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 0
+#endif
```
tomi-font: Why not simply
```suggestion
#if !defined(MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR)
#if defined(__has_include) && __has_include(<mbedtls/private/sha256.h>)
#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 1
#else
#define MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR 0
#endif
#endif
```
?

tomi-font (commented):
looks okay to me, @valeriosetti please give this a review too

tomi-font:
@krish2718 build error https://github.com/mcu-tools/mcuboot/actions/runs/27676064570/job/81869310456?pr=2732

krish2718:
> @krish2718 build error https://github.com/mcu-tools/mcuboot/actions/runs/27676064570/job/81869310456?pr=2732

Yes, looking into it. Thanks.

valeriosetti:
@krish2718 the PR looks good to me, but I have a question: whenever this PR and https://github.com/zephyrproject-rtos/zephyr/pull/108979 go in, do you think it should be possible to have a follow-up PR here to remove the `MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR` logic? That would help in keeping things a bit clearer.

krish2718:
> @krish2718 the PR looks good to me, but I have a question: whenever this PR and [zephyrproject-rtos/zephyr#108979](https://github.com/zephyrproject-rtos/zephyr/pull/108979) go in, do you think it should be possible to have a follow-up PR here to remove the `MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR` logic? That would help in keeping things a bit clearer.

IIUC, are you suggesting to always use the `private` and remove the Kconfig? Wouldn't that need the MCUBoot should also use 4.1.0 always?

krish2718:
The SIM failure looks like some rust dependency issue, can someone re-run?

```
error: failed to get `regex-syntax` as a dependency of package `regex v1.5.5`
    ... which satisfies dependency `regex = "^1.4.0"` (locked to 1.5.5) of package `docopt v1.1.1`
    ... which satisfies dependency `docopt = "^1.1.0"` (locked to 1.1.1) of package `bootsim v0.1.0 (/home/runner/work/mcuboot/mcuboot/sim)`
```

krish2718:
> The SIM failure looks like some rust dependency issue, can someone re-run?
> 
> ```
> error: failed to get `regex-syntax` as a dependency of package `regex v1.5.5`
>     ... which satisfies dependency `regex = "^1.4.0"` (locked to 1.5.5) of package `docopt v1.1.1`
>     ... which satisfies dependency `docopt = "^1.1.0"` (locked to 1.1.1) of package `bootsim v0.1.0 (/home/runner/work/mcuboot/mcuboot/sim)`
> ```

We can re-run and leave it or cursor is suggesting below fix, if this is fine I can push a commit with this fix?

```
diff --git a/ci/sim_install.sh b/ci/sim_install.sh
index 8ec96cf4..3fd43160 100755
--- a/ci/sim_install.sh
+++ b/ci/sim_install.sh
@@ -14,6 +14,30 @@
 
 pip3 install --user -r ci/requirements.txt
 
-pushd sim && cargo fetch
+# crates.io downloads occasionally fail in CI with partial transfers.
+CARGO_NET_RETRY="${CARGO_NET_RETRY:-10}"
+export CARGO_NET_RETRY
+
+cargo_fetch() {
+    local attempt max_attempts delay
+
+    max_attempts="${CARGO_FETCH_RETRIES:-5}"
+    delay=5
+    attempt=1
+    while [ "${attempt}" -le "${max_attempts}" ]; do
+        if cargo fetch; then
+            return 0
+        fi
+        if [ "${attempt}" -eq "${max_attempts}" ]; then
+            return 1
+        fi
+        echo "cargo fetch failed (attempt ${attempt}/${max_attempts}), retrying..."
+        sleep "${delay}"
+        delay=$((delay * 2))
+        attempt=$((attempt + 1))
+    done
+}
+
+pushd sim && cargo_fetch
 [[ $? -ne 0 ]] && exit 1
 popd
```

valeriosetti:
> > @krish2718 the PR looks good to me, but I have a question: whenever this PR and [zephyrproject-rtos/zephyr#108979](https://github.com/zephyrproject-rtos/zephyr/pull/108979) go in, do you think it should be possible to have a follow-up PR here to remove the `MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR` logic? That would help in keeping things a bit clearer.
> 
> IIUC, are you suggesting to always use the `private` and remove the Kconfig? Wouldn't that need the MCUBoot should also use 4.1.0 always?

Well, from Zephyr point of view that would work... but I guess that there might be other users as well which might not have transitioned yet to Mbed TLS 4. Ok, let's keep things as they are then

valeriosetti (approved):

nordicjm (approved):

PR #2734, opened 2026-05-14
boot: clear AES key material and plaintext from the in-place decrypt path

d3zd3z:
## What

The serial-recovery in-place decrypt path keeps AES key material and decrypted image plaintext in stack-local buffers and crypto contexts, but did not consistently clear them once they were no longer needed. This series makes that path (and the Tinycrypt AES-CTR backend) clear sensitive data on every exit.

## Changes

Four commits:

1. **Add `bootutil_wipe_memory()` secure-zero helper.** `loader.c` already had a file-local `like_mbedtls_zeroize()` that writes zeros through a `volatile` pointer so the compiler cannot drop the store as a dead store. This commit promotes it to `bootutil_misc.c` (declared in `bootutil_priv.h`), renamed without the provenance baggage, and updates `loader.c` to use it. No behavioural change.

2. **Clear AES key material from the stack after in-place decrypt.** `decrypt_image_inplace()` and `boot_image_validate_encrypted()` keep the raw AES key in a local `struct boot_status` (`_bs.enckey[]`) and did not clear it on exit. `decrypt_image_inplace()` also returned directly from the per-sector loop on a decrypt failure, bypassing the `total_out:` cleanup entirely. Route every exit through a common cleanup label and clear `_bs` there with `bootutil_wipe_memory()`. The early-`FIH_RET` → `goto` change is a correctness fix in its own right.

3. **Clear decrypted plaintext from the in-place decrypt buffer.** `decrypt_region_inplace()` decrypts each flash sector into a stack buffer; on return that buffer still held a chunk of decrypted image plaintext. Restructure the function to a single cleanup point and clear the buffer there.

4. **Clear the AES key schedule in the Tinycrypt `aes_ctr_drop`.** `bootutil_aes_ctr_drop()` is the hook for releasing an AES-CTR context. The mbedTLS and PSA backends free/destroy the underlying key state, but the Tinycrypt implementation was an empty stub, leaving the expanded key schedule behind. Clear the context so `boot_enc_drop()` / `boot_state_clear()` wipe it for this backend too.

## Notes

Hardening/cleanup. The bulk of the change is in `boot_serial_encryption.c`, which the Rust simulator does not compile (it is a Zephyr/Mynewt-only file), so commits 2 and 3 are reviewed at the code level rather than exercised by the sim suite. Commits 1 and 4 are exercised by `cargo test` with the encryption features enabled.

## Testing

`cd sim && cargo test --features enc-rsa` and `--features enc-kw` pass (25 tests each).

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
@@ -13,6 +13,7 @@
 #include <string.h>
 #include <stdint.h>
 #include "mcuboot_config/mcuboot_config.h"
+#include "bootutil/bootutil.h"
 #include "bootutil/enc_key_public.h"
 
 #include <tinycrypt/aes.h>
@@ -37,7 +38,10 @@ static inline void bootutil_aes_ctr_init(bootutil_aes_ctr_context *ctx)
 
 static inline void bootutil_aes_ctr_drop(bootutil_aes_ctr_context *ctx)
 {
-    (void)ctx;
+    /* Tinycrypt has no aes_free()-style call; the expanded key schedule
+     * lives entirely in ctx, so wipe it directly.
+     */
+    bootutil_wipe_memory(ctx, sizeof(*ctx));
 }
 
 static inline int bootutil_aes_ctr_set_key(bootutil_aes_ctr_context *ctx, const uint8_t *k)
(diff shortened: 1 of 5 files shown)


de-nordic (approved):

boot/bootutil/include/bootutil/crypto/aes_ctr_tinycrypt.h
```diff
@@ -37,7 +37,13 @@ static inline void bootutil_aes_ctr_init(bootutil_aes_ctr_context *ctx)
 
 static inline void bootutil_aes_ctr_drop(bootutil_aes_ctr_context *ctx)
 {
-    (void)ctx;
+    /* Clear the expanded AES key schedule. Written through a volatile
+     * pointer so the compiler cannot drop it as a dead store.
+     */
+    volatile uint8_t *p = (volatile uint8_t *)ctx;
```
nordicjm: why not call the bootutil function added in the other commit from this file instead of having a duplicate function?

d3zd3z: Fixed, I moved the wipe function to bootutil/bootutil.h so that it can be used from inline.

nordicjm (approved):

PR #2735, opened 2026-05-14
boot: bootutil: fault-injection hardening for RSA-PSS verification

d3zd3z:
## Summary

This series adds fault-injection hardening around the RSA-PSS signature verification in `bootutil`, in line with the FIH patterns already used elsewhere in the bootloader.

## Changes

**1. Harden the non-PSA RSA-PSS verification path**

`bootutil_cmp_rsasig()` previously invoked the public-key modular exponentiation as a plain, unwrapped call with no hardening around it. Two defensive measures are added:

- The `bootutil_rsa_public()` call is routed through a thin `FIH` wrapper invoked via `FIH_CALL`, so a skipped call is caught by control-flow integrity counting, consistent with the rest of `bootutil`.
- After the exponentiation, the signature is rejected if the output buffer is identical to the input. A correct RSA public-key operation does not produce that result. A dedicated `fih_ret` variable is used for this comparison so a later skipped check cannot leave a stale success value behind.

**2. Optional redundant RSA-PSS verification for the PSA crypto path**

With the PSA crypto backend the verification is a single `psa_verify_hash()` call and the modular exponentiation runs inside the PSA implementation, where it is not reachable from MCUboot. A new optional countermeasure runs the whole RSA-PSS verification a second time and requires both attempts to succeed.

It is gated by a new Kconfig option, `BOOT_RSA_PSA_DOUBLE_VERIFY`, because the extra verification roughly doubles the RSA verify time and different users will want different tradeoffs. The option defaults to enabled whenever a fault injection hardening profile is active, and disabled otherwise. Ports that do not define `MCUBOOT_RSA_PSA_DOUBLE_VERIFY` keep the previous single-verification behaviour. Hardening of the modular exponentiation itself remains the responsibility of the PSA crypto implementation.

## Testing

- `cargo test --features sig-rsa` and `--features sig-rsa3072`: all pass; legitimately signed images still validate.
- The non-PSA path compiles cleanly under all four FIH profiles (`OFF`/`LOW`/`MEDIUM`/`HIGH`).
- The PSA path compiles cleanly with and without `MCUBOOT_RSA_PSA_DOUBLE_VERIFY`.
- Negative check: temporarily forcing the exponentiation output to equal its input causes `bootutil_cmp_rsasig()` to reject the image, confirming the new check fires.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -303,6 +303,25 @@ config BOOT_RSA_PSA
 
 endchoice
 
+config BOOT_RSA_PSA_DOUBLE_VERIFY
+	bool "Verify RSA signatures twice (fault injection hardening)"
+	depends on BOOT_RSA_PSA
+	default y if BOOT_FIH_PROFILE_MEDIUM || BOOT_FIH_PROFILE_HIGH
+	help
+	  With the PSA crypto backend the RSA modular exponentiation is
+	  performed inside the PSA implementation and cannot be checked
+	  directly by MCUboot. When this option is enabled the whole
+	  RSA-PSS verification is performed a second time and both attempts
+	  must succeed, providing a fault injection countermeasure at the
+	  cost of doubling the (relatively expensive) RSA verification time.
+
+	  This is the same redundancy strategy as the MEDIUM profile's
+	  double variables, applied to the one operation FIH's control-flow
+	  checks cannot see into, so it defaults to enabled at the MEDIUM
+	  and HIGH profiles. The LOW profile stays control-flow only; enable
+	  this explicitly there if the boot time cost is acceptable. Disable
+	  it at any level if the cost outweighs the added protection.
+
 endif # BOOT_SIGNATURE_TYPE_RSA
 
 config BOOT_SIGNATURE_TYPE_ECDSA_P256
--- a/boot/bootutil/src/image_rsa.c
+++ b/boot/bootutil/src/image_rsa.c
@@ -115,6 +115,26 @@ pss_mgf1(uint8_t *mask, const uint8_t *hash)
 }
 
 /*
+ * FIH wrapper around bootutil_rsa_public() so the modular exponentiation
+ * call crosses a FIH_CALL boundary and is covered by control-flow
+ * integrity counting, consistent with the rest of bootutil. This guards
+ * against the whole call being skipped; the result check in the caller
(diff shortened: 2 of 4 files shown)


boot/zephyr/Kconfig
```diff
@@ -303,6 +303,21 @@ config BOOT_RSA_PSA
 
 endchoice
 
+config BOOT_RSA_PSA_DOUBLE_VERIFY
+	bool "Verify RSA signatures twice (fault injection hardening)"
+	depends on BOOT_RSA_PSA
+	default y if !BOOT_FIH_PROFILE_OFF
```
nordicjm: only question would be on if this should be the default if FIH is enabled in general, or if it should only be enabled by default if a certain FIH level is used e.g. medium/high? I.e. does low need this? Does medium need this?

d3zd3z: ~~I'm actually not sure there is a definitive answer to this. The different FIH levels cost a bit, this costs a lot, and perhaps should only be explicitly enabled, but I'm just not sure how much people would notice it.~~

d3zd3z: > only question would be on if this should be the default if FIH is enabled in general, or if it should only be enabled by default if a certain FIH level is used e.g. medium/high? I.e. does low need this? Does medium need this?

Good point, and thinking about it more, I don't think LOW should pull this in by default. The profiles form a deliberate ladder: LOW is control-flow integrity only, MEDIUM adds double variables (redundancy to catch corruption), and HIGH adds random delays. This double-verify is really the same redundancy idea as MEDIUM's double variables, just applied to the one operation — the modexp inside PSA — that the control-flow counting structurally can't see into. So it pairs naturally with MEDIUM rather than sitting on the cheapest rung.

There's also a cost argument: someone who selects LOW is signaling cost-sensitivity, and doubling the RSA verify (tens of ms for RSA-3072 on an unaccelerated core) is exactly the kind of regression they'd notice and probably not expect from "moderate hardening."

I've changed the default to:

```
default y if BOOT_FIH_PROFILE_MEDIUM || BOOT_FIH_PROFILE_HIGH
```

and updated the help text to explain the reasoning and note that LOW users can still opt in explicitly. Let me know if you'd rather it be MEDIUM+ vs. HIGH-only.

nordicjm: medium seems fine but 2nd and 3rd commits should be sqaushed

PR #2739, opened 2026-05-21
scripts: imgtool: Update regex pattern

ozersa:
[A-f] regex mask search non-numeric characters below
image

For clarity it would be better if it only search exact pattern.

Note: It is not critical change, just improvement for clarity, because even if non-numeric chars has been found with existing regex pattern while converting them to hex with
[bytes.fromhex()](https://github.com/mcu-tools/mcuboot/blob/main/scripts/imgtool/image.py#L269) function, it will fails.

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -264,12 +264,12 @@ def tlv_matches_key_type(tlv_type, key):
 
 def parse_uuid(namespace, value):
     # Check if UUID is in the RAW format (12345678-1234-5678-1234-567812345678)
-    uuid_re = r'[0-9A-f]{8}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{12}'
-    if re.match(uuid_re, value):
+    uuid_re = r'[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
+    if re.fullmatch(uuid_re, value):
         uuid_bytes = bytes.fromhex(value.replace('-', ''))
 
     # Check if UUID is in the RAW HEX format (12345678123456781234567812345678)
-    elif re.match(r'[0-9A-f]{32}', value):
+    elif re.fullmatch(r'[0-9a-fA-F]{32}', value):
         uuid_bytes = bytes.fromhex(value)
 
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

scripts/imgtool/image.py
```diff
@@ -264,12 +264,12 @@ def tlv_matches_key_type(tlv_type, key):
 
 def parse_uuid(namespace, value):
     # Check if UUID is in the RAW format (12345678-1234-5678-1234-567812345678)
-    uuid_re = r'[0-9A-f]{8}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{4}-[0-9A-f]{12}'
+    uuid_re = r'[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
```
d3zd3z, line 267: Definitely fixes the but, but without a '$', this will match valid UUIDs with garbage at the end of them. Do you mind adding that as well?

scripts/imgtool/image.py
```diff
@@ -264,12 +264,12 @@ def tlv_matches_key_type(tlv_type, key):
...
     if re.match(uuid_re, value):
         uuid_bytes = bytes.fromhex(value.replace('-', ''))
 
     # Check if UUID is in the RAW HEX format (12345678123456781234567812345678)
-    elif re.match(r'[0-9A-f]{32}', value):
+    elif re.match(r'[0-9a-fA-F]{32}', value):
```
d3zd3z: Likewise this as well, garbage is accepted after the 32 characters.

d3zd3z (changes requested):

scripts/imgtool/image.py
d3zd3z: Alternatively, you can just change `re.match` to `re.fullmatch` in both cases.

ozersa: Thanks, as per of your suggestion used .fullmatch function.

scripts/imgtool/image.py
ozersa: Replaced with .fullmatch function as https://github.com/mcu-tools/mcuboot/pull/2739#discussion_r3354241188

de-nordic (approved):

nordicjm:
Merging as changes made and reviewer has failed to re-review

PR #2741, opened 2026-05-21
boot: bootutil: introduce boot_load_image_to_sram hook

danieldegrasse:
This hook allows to define a custom way of handling loading an image to SRAM. This can be useful if a system needs to override image loading routines (for example, if a security enclave needs to perform loading and validation)

### DIFF
--- a/boot/bootutil/include/bootutil/boot_hooks.h
+++ b/boot/bootutil/include/bootutil/boot_hooks.h
@@ -225,6 +225,21 @@ int boot_img_install_stat_hook(int image_index, int slot,
  */
 int boot_reset_request_hook(bool force);
 
+/** Hook provides a way for the custom image loading and validation.
+ *
+ * @param img_index the index of the image pair
+ * @param slot slot number
+ * @param img_dst image destination address to be populated
+ * @param img_sz image size to be populated
+ *
+ * @retval 0: success, mcuboot will follow normal code execution flow after
+ *            execution of this call.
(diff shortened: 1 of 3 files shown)


boot/bootutil/include/bootutil/boot_hooks.h
```diff
@@ -225,6 +225,21 @@ int boot_img_install_stat_hook(int image_index, int slot,
...
+ * @param img_dst image destination address to be populated
+ * @param img_sz image size to be populated
+ *
+ * @retval 0: success, mcuboot will follow normal code execution flow after
+ *            execution of this call.
+ *         non-zero: an error, will be transferred as part of command response
+ *            as "rc" entry.
+ */
+int boot_load_image_to_sram_hook(int image_index, int slot,
+                                 uint32_t *img_dst, uint32_t *img_sz);
```
nordicjm, line 241: this doesn't make sense vs what `boot_load_image_to_sram` has?

danieldegrasse: Sorry, not 100% clear here. The intention of this call is that we can hook the SRAM load process to replace the decryption and memory copy routines, for cases where we have custom hardware that handles those routines. Are you saying that the documentation here isn't clear, or that the hook should behave differently?

nordicjm: This should have the same function arguments that the normal function has to allow it to do a variety of different things. Also not sure how this would work loading since you have no information from the image?

PR #2743, opened 2026-05-28
boot: add delta DFU support

hlord2000:
## Summary

Adds overwrite-only delta DFU support.

A signed reversible delta image in the secondary slot can patch the current primary image. MCUboot verifies the expected base, recovers interrupted updates using the stored old bytes, and validates the reconstructed target. Unconfirmed test updates can be reverted.

## Changes

- Add `imgtool` support with `--delta-base` and `--delta-block-size`.
- Generate reversible deltas containing both new and old bytes.
- Add `bootutil` support for validation, application, and reset-safe recovery.
- Defer security-counter updates for test images until confirmation.
- Add Zephyr configuration, documentation, simulator tests, and CI coverage.

## Testing

- `uv run pytest scripts/tests/test_delta.py`
- Simulator tests for unsigned, ECDSA-signed, and rollback-protected configurations.


d3zd3z (changes requested):
So a few high level things:
- First to be clear: this is a non-robust upgrade, much worse than overwrite only. Whereas overwrite only will effectively restart the operation, this doesn't appear to be recoverable if power is lost, effectively bricking the device. Correct me if my understand is not right, but that isn't really a mode we want to have in MCUboot.
- Second, it isn't tested (which would point out that it doesn't work with power loss).

So, in order to get this through, it is going to need to be added to the simulator for testing. And, more importantly the robustness needs to be addressed. There either needs to be an argument for a use case where bricking a device on an upgrade interruption is acceptable, or the code needs to be made robust (it needs to be able to recover, and continue the upgrade when interrupted at any point.

hlord2000:
> So a few high level things:
> 
> * First to be clear: this is a non-robust upgrade, much worse than overwrite only. Whereas overwrite only will effectively restart the operation, this doesn't appear to be recoverable if power is lost, effectively bricking the device. Correct me if my understand is not right, but that isn't really a mode we want to have in MCUboot.
> * Second, it isn't tested (which would point out that it doesn't work with power loss).
> 
> So, in order to get this through, it is going to need to be added to the simulator for testing. And, more importantly the robustness needs to be addressed. There either needs to be an argument for a use case where bricking a device on an upgrade interruption is acceptable, or the code needs to be made robust (it needs to be able to recover, and continue the upgrade when interrupted at any point.

Thanks for the feedback!

On your first point: yes, there is a non-robust upgrade path. This is what I started with, just overwriting the existing application w/o maintaining a diff of delta->current firmware. Gated behind "MCUBOOT_DELTA_DFU_RESTORE" is a flow where this diff is generated before anything is overwritten. I'll remove the non-robust upgrade path, that's fine as this still represents significant NVM savings for mature applications. 

I'll work on getting this working w/ the simulator.

d3zd3z (commented):
This looks good, thank you for the work. Just a few questions:
- First, this will need to be rebased to be able to merge.
- Does the security counter properly update when the reversible delta is confirmed.
- Please update the description, which still describes `--delta-revertible` as optional, and doesn't mention the robustness rework.
- The per-record bounds are validate in a pass separate from where records are re-read and applied. A flash change between the two passes could smuggle an out-of-range offset/size. This isn't exploitable remotely, and I think this requires physically changing the flash between verified flash, so probably isn't covered by our threat model, but something to keep in mind.
- Please update `docs/design.md` to describe the new flag and TLFv.
- It might be worth mentioning in `docs/delta_dfu.md` that since each changed region costs 2x bytes, a delta that touches scattered regions could exceed a plain overwrite in size.

d3zd3z (commented):
From my perspective, the only thing remaining is to update the description of the pr to match what the change now does.

hlord2000:
> From my perspective, the only thing remaining is to update the description of the pr to match what the change now does.

Just updated

boot/bootutil/include/bootutil/image.h
```diff
@@ -134,6 +135,8 @@ extern "C" {
 #define IMAGE_TLV_COMP_DEC_SIZE     0x73    /* Compressed decrypted image size */
 #define IMAGE_TLV_UUID_VID          0x74    /* Vendor unique identifier */
 #define IMAGE_TLV_UUID_CID          0x75    /* Device class unique identifier */
+#define IMAGE_TLV_DELTA_BASE_SHA    0x76    /* SHA of image the delta applies to */
+#define IMAGE_TLV_DELTA_TARGET_SHA  0x77    /* SHA of image produced by the delta */
```
nordicjm: we already have TLVs for output and input SHA, see decomp, I dont think we should add more when we can just reuse those (they can be renamed)

docs/delta_dfu.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+The header is followed by `record_count` records:
+
+```c
+struct boot_delta_record {
+    uint32_t offset;
+    uint32_t size;
+    uint8_t  new_data[size];
+    uint8_t  old_data[size];
```
nordicjm: 1 space

nordicjm (commented):
incredibly quick scan

d3zd3z (approved):

d3zd3z:
Claude suggests a fix: "Rebase onto current main, and in the delta path change the new `image_largest_trailer(dev, areadesc, &slots[slot_ind])` call to pass slots, matching the updated signature (the function now looks at both slots internally to size the swap-scratch trailer). Worth a sentence asking them to check that the both-slots estimate is still what their max(trailer, delta-recovery-trailer) computation wants, but mechanically it's a one-line fix plus rebase."

hlord2000:
> Claude suggests a fix: "Rebase onto current main, and in the delta path change the new `image_largest_trailer(dev, areadesc, &slots[slot_ind])` call to pass slots, matching the updated signature (the function now looks at both slots internally to size the swap-scratch trailer). Worth a sentence asking them to check that the both-slots estimate is still what their max(trailer, delta-recovery-trailer) computation wants, but mechanically it's a one-line fix plus rebase."

Addressed this + a CI failure :D

PR #2749, opened 2026-06-01
boot: zephyr: Process all slots for image details

nordicjm:
Adds support for processing all slots to work out the amount of sectors needed to swap an image. Also provides the details of other slots via compilation definitions to allow for checking of image details


boot/zephyr/CMakeLists.txt
```diff
@@ -476,62 +476,93 @@ function(dt_get_nvm_device node nvm_device)
...
-if(CONFIG_BOOT_MAX_IMG_SECTORS_AUTO)
-  if(NOT DEFINED slot0_size)
-    message(WARNING "Unable to determine size of slot0 partition, cannot calculate minimum sector usage")
-  elseif(NOT DEFINED erase_size_slot0)
-    message(WARNING "Unable to determine erase size of slot0 partition, cannot calculate minimum sector usage")
-  else()
-    math(EXPR slot_min_sectors "${slot0_size} / ${erase_size_slot0}")
+  if(CONFIG_BOOT_MAX_IMG_SECTORS_AUTO)
+    if(DEFINED slot0_size AND DEFINED erase_size_slot0)
+      math(EXPR slot_min_sectors "${slot0_size} / ${erase_size_slot0}")
```
tomchy, line 507: You may consider using the ceiling operation here (and for secondary slots) by adding an almost erase block size:

```suggestion
      math(EXPR slot_min_sectors "(${slot0_size} +  ${erase_size_slot0} - 1) / ${erase_size_slot0}")
```

nordicjm: not sure I see the point since the slot size will be a multiple of the erase size (either that or the dts is invalid)

de-nordic (approved):

PR #2751, opened 2026-06-03
boot: zephyr: enter USB DFU indefinitely on retention boot mode

benner:
Adds `BOOT_USB_DFU_BOOT_MODE`, an independent USB DFU entrance method: an
application sets Zephyr's boot mode retention flag and reboots, and the
bootloader stays in USB DFU indefinitely. It mirrors `BOOT_SERIAL_BOOT_MODE`
for serial recovery.

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -226,7 +226,8 @@ int main(void)
 {
     struct boot_rsp rsp;
     int rc;
-#if defined(CONFIG_BOOT_USB_DFU_GPIO) || defined(CONFIG_BOOT_USB_DFU_WAIT)
+#if defined(CONFIG_BOOT_USB_DFU_GPIO) || defined(CONFIG_BOOT_USB_DFU_WAIT) || \
+    defined(CONFIG_BOOT_USB_DFU_BOOT_MODE)
     bool usb_dfu_requested = false;
     bool usb_dfu_forever = false;
 #endif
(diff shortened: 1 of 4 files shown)


boot/zephyr/main.c
```diff
@@ -597,11 +597,21 @@ int main(void)
...
-            wait_for_usb_dfu(K_MSEC(CONFIG_BOOT_USB_DFU_WAIT_DELAY_MS));
-            BOOT_LOG_INF("USB DFU wait time elapsed");
-            mcuboot_status_change(MCUBOOT_STATUS_USB_DFU_TIMED_OUT);
+#if defined(CONFIG_BOOT_USB_DFU_BOOT_MODE)
+            if (io_detect_boot_mode()) {
+                BOOT_LOG_DBG("Staying in USB DFU");
+                mcuboot_status_change(MCUBOOT_STATUS_USB_DFU_WAITING);
+                wait_for_usb_dfu(K_FOREVER);
+            } else
+#endif
```
nordicjm: to me it makes sense to have this as a _new_ mode, not having to have it as part of waiting. For that the Kconfig choice should be changed to have options, like serial recovery does

benner: reworked after https://github.com/mcu-tools/mcuboot/pull/2761 merge.

gabenner:
rebased from newest main

PR #2752, opened 2026-06-03
boot: zephyr: arch separation for do_boot

almir-okato:
This PR separates arch specific booting process to reduce `ifdefs` and improve organization on `boot/zephyr/main.c`.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -73,6 +73,18 @@ target_sources(app PRIVATE
   watchdog.c
 )
 
+if(CONFIG_ARM)
+  target_sources(app PRIVATE arch/arm.c)
+elseif(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
+  target_sources(app PRIVATE arch/esp32.c)
+elseif(CONFIG_XTENSA)
+  target_sources(app PRIVATE arch/xtensa.c)
+elseif(CONFIG_ARC)
+  target_sources(app PRIVATE arch/arc.c)
+else()
+  target_sources(app PRIVATE arch/default.c)
+endif()
+
 if(CONFIG_BOOT_SHARE_BACKEND_RETENTION)
   zephyr_sources(shared_data.c)
 endif()
--- a/boot/zephyr/arch/arm.c
+++ b/boot/zephyr/arch/arm.c
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) 2012-2014 Wind River Systems, Inc.
+ * Copyright (c) 2020 Arm Limited
+ * Copyright (c) 2021-2023 Nordic Semiconductor ASA
+ * Copyright (c) 2025 Aerlync Labs Inc.
+ * Copyright (c) 2025 Siemens Mobility GmbH
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/drivers/flash.h>
+#include <zephyr/drivers/timer/system_timer.h>
+#include <zephyr/usb/usb_device.h>
+#include <zephyr/linker/linker-defs.h>
+#include <soc.h>
+
+#if defined(CONFIG_BOOT_DISABLE_CACHES)
+#include <zephyr/cache.h>
+#endif
+
+#if defined(CONFIG_CPU_CORTEX_M)
+#include <cmsis_core.h>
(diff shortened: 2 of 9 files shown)


boot/zephyr/arch/arm.c
```diff
@@ -0,0 +1,209 @@
...
+ * Copyright (c) 2025 Siemens Mobility GmbH
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/drivers/flash.h>
+#include <zephyr/drivers/timer/system_timer.h>
+#include <zephyr/usb/usb_device.h>
```
sylvioalves, line 15: Check if the below is missing:
`#include <zephyr/linker/linker-defs.h>`

sylvioalves:
Regarding the new `arch` folder: I think it mixes architecture with soc/vendor. Perhaps it could be split in arch and soc folder's? Don't need to accept this right away. Would be good to have additional feedback from other reviewers.

```c 
boot/zephyr/
├── arch/                  <- architecture-keyed ONLY
│   ├── arm.c              (CONFIG_ARM)
│   ├── arc.c              (CONFIG_ARC)
│   ├── xtensa.c           (CONFIG_XTENSA)   generic Xtensa SRAM-copy path
│   └── default.c          (else: x86, etc.)
│
├── socs/                  <- vendor/SoC-keyed (already exists for .conf/.overlay)
│   ├── espressif/
│   │   └── do_boot.c      (CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)  <- was arch/esp32.c
│   ├── esp32_procpu.conf        (existing files unchanged)
│   ├── esp32c3.conf
│   └── ... (all existing .conf/.overlay stay put)
│
├── include/
│   └── do_boot.h
└── main.c
```

almir-okato:
> Regarding the new `arch` folder: I think it mixes architecture with soc/vendor. Perhaps it could be split in arch and soc folder's? Don't need to accept this right away. Would be good to have additional feedback from other reviewers.
> 
> ```c
> boot/zephyr/
> ├── arch/                  <- architecture-keyed ONLY
> │   ├── arm.c              (CONFIG_ARM)
> │   ├── arc.c              (CONFIG_ARC)
> │   ├── xtensa.c           (CONFIG_XTENSA)   generic Xtensa SRAM-copy path
> │   └── default.c          (else: x86, etc.)
> │
> ├── socs/                  <- vendor/SoC-keyed (already exists for .conf/.overlay)
> │   ├── espressif/
> │   │   └── do_boot.c      (CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)  <- was arch/esp32.c
> │   ├── esp32_procpu.conf        (existing files unchanged)
> │   ├── esp32c3.conf
> │   └── ... (all existing .conf/.overlay stay put)
> │
> ├── include/
> │   └── do_boot.h
> └── main.c
> ```

Ok, I changed the PR to "Ready for review", so we may get more feedback on the reorganization changes.

As there was few variations of `do_boot` in `main.c`, I tried to keep them in the same "organization level".

@d3zd3z could you please take a look?

d3zd3z (approved):
Other than the issues already brought up, this looks good.

sylvioalves:
@nordicjm would you take a look? Thanks.

boot/zephyr/CMakeLists.txt
```diff
@@ -73,6 +73,18 @@ target_sources(app PRIVATE
...
+if(CONFIG_ARM)
+  target_sources(app PRIVATE arch/arm.c)
+elseif(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
+  target_sources(app PRIVATE arch/esp32.c)
+elseif(CONFIG_XTENSA)
+  target_sources(app PRIVATE arch/xtensa.c)
+elseif(CONFIG_ARC)
+  target_sources(app PRIVATE arch/arc.c)
+else()
+  target_sources(app PRIVATE arch/default.c)
```
marekmatej, line 85: The “Default” case as the architecture feels a bit weird, but I guess it is the price for splitting the common code.

nordicjm (approved):

marekmatej (approved):
Left a comment, but I guess it could be fixed in follow-ups.

boot/zephyr/CMakeLists.txt
almir-okato: `main.c` already had the "default" path (the `else` case) for `do_boot`, so I just kept that logic.

PR #2755, opened 2026-06-08
boot: boot_serial: add raw (non-console) SMP recovery protocol

JPHutchins:
~~I'm interested in exploring appetite for this, so here's a draft.~~ This can yield a ~30% transport speed boost (eliminates base64 overhead) and smaller mcuboot program size. A few thoughts:

* The timeout approach used by Zephyr is easy to implement in MCUBoot, so it's included
  * A more aggressive timeout may be appropriate, OK to leave up to user though
* It's great for USB, where we already have CRC on the line, no notes!
* It makes me nervous over a raw UART
  * If you consider that it's mostly used for image upload, this is mitigated somewhat by the signing verification and sha check (though I don't recall whether or not we have the sha check at end of upload enabled or not)
  * I've considered a follow-ups both here and over at Zephyr:
    * kconfig to enable flags repurposed as CRC8. The problem is that we really need 16-bit CRC to get meaningful protection of large packets
    * kconfig to enable framing WITHOUT base64
      * COBS + CRC16 trailer
      * magic, length, <raw packet>, CRC16 (8 bytes overhead if we use 32-bit magic, for example)
    > But all of this feels bad; so many options, how would users understand what to pick?
    
 Raw commit message follows:

Add the BOOT_SERIAL_RAW_PROTOCOL option, which exchanges SMP packets
over the serial port as raw binary instead of the SMP-over-console
encoding: no base64, no two-byte length prefix, no CRC16 and no
SHELL_NLIP framing. Packet boundaries are derived from the length field
of the SMP header, so the encoding is transport-agnostic (UART, CDC ACM,
CAN, ...) and only requires a binary-capable link. The mcumgr client
must be configured to use a matching raw transport.

This mirrors Zephyr's CONFIG_UART_MCUMGR_RAW_PROTOCOL. The two encodings
are mutually exclusive and selected at build time; in raw mode the
base64 and CRC code and libraries are dropped, saving roughly 1 KiB of
flash and 1 KiB of RAM and increasing transfer speed.

Because the raw encoding has no CRC or framing to detect a truncated
packet, also add BOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT (enabled by
default with the raw protocol): a partially received packet is dropped
after BOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT_MS with no new data, so a
stalled transfer cannot wedge recovery. The timeout is an inactivity
window that restarts whenever more data arrives, so it never expires a
transfer that is still making progress. This mirrors Zephyr's
CONFIG_MCUMGR_TRANSPORT_RAW_UART_INPUT_TIMEOUT.

A serial_recovery_raw.conf fragment and a matching Twister build test
(sample.bootloader.mcuboot.serial_recovery_raw) are added so the raw
configuration is built in CI.

Signed-off-by: JP Hutchins <jp@intercreate.io>
Assisted-By: Claude:opus-4.8


nordicjm:
> I'm interested in exploring appetite for this, so here's a draft. Reminder that this can yield a ~30% transport speed boost (eliminates base64 overhead) and smaller mcuboot program size. A few thoughts:

Not checked code but is very much welcome! Arguably there isn't really much point in SMP over console for MCUboot given that you cannot use it alongside the shell anyway so this seems like a better usage method

JPHutchins:
> > I'm interested in exploring appetite for this, so here's a draft. Reminder that this can yield a ~30% transport speed boost (eliminates base64 overhead) and smaller mcuboot program size. A few thoughts:
> 
> Not checked code but is very much welcome! Arguably there isn't really much point in SMP over console for MCUboot given that you cannot use it alongside the shell anyway so this seems like a better usage method

OK, I did a review pass and marked ready for review. FYI, we're able to test mcuboot + zephyr including DFU with this new QEMU fixture system over at smpclient, so I'll be adding coverage there: https://github.com/intercreate/smpclient/actions/runs/27441392305/job/81116055485

boot/boot_serial/src/boot_serial.c
```diff
@@ -1374,13 +1376,7 @@ static void
 boot_serial_output(void)
 {
     char *data;
-    int len, out;
-    uint16_t crc;
-    uint16_t totlen;
-    char pkt_cont[2] = { SHELL_NLIP_DATA_START1, SHELL_NLIP_DATA_START2 };
-    char pkt_start[2] = { SHELL_NLIP_PKT_START1, SHELL_NLIP_PKT_START2 };
-    char buf[BOOT_SERIAL_OUT_MAX + sizeof(*bs_hdr) + sizeof(crc) + sizeof(totlen)];
-    char encoded_buf[BASE64_ENCODE_SIZE(sizeof(buf))];
```
nordicjm, line 1417: variable declarations should be at top of scope so these should remain but #ifdef here so the raw mode doesn't have them

boot/boot_serial/src/boot_serial.c
```diff
@@ -1573,7 +1654,15 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
             boot_serial_input(&dec_buf[2], dec_off - 2);
         }
         off = 0;
+#endif /* MCUBOOT_SERIAL_RAW_PROTOCOL */
 check_timeout:
+#ifdef MCUBOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT
+        if (off > 0 && (k_uptime_get_32() - raw_input_start) >=
+                       MCUBOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT_MS) {
```
nordicjm: k_uptime_get is a zephyr function, you cannot use zephyr functions in common code

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -46,6 +46,41 @@ config BOOT_SERIAL_CDC_ACM
 
 endchoice
 
+config BOOT_SERIAL_RAW_PROTOCOL
+	bool "Raw (SMP without framing and encoding) protocol"
+	help
+	  If y, SMP (mcumgr) commands are exchanged over the serial port as raw SMP
```
nordicjm: *MCUmgr

nordicjm (changes requested):

boot/boot_serial/src/boot_serial.c
JPHutchins: I've added uptime.c/h for this, LMK if that's the right way to add to the OSAL.

boot/zephyr/Kconfig.serial_recovery
JPHutchins: fixed

boot/boot_serial/src/boot_serial.c
JPHutchins: fixed

boot/boot_serial/src/boot_serial.c
```diff
@@ -1503,6 +1514,48 @@ boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 
     return 1;
 }
+#endif /* !MCUBOOT_SERIAL_RAW_PROTOCOL */
+
+#ifdef MCUBOOT_SERIAL_RAW_PROTOCOL
+/* Dispatch every complete raw SMP packet at the front of "buf" (length from the
+ * SMP header, capacity "max_input") via boot_serial_input(), keeping any
+ * trailing bytes; returns the count of unconsumed bytes left in "buf". */
```
nordicjm: nit. `*/` on newline

boot/boot_serial/src/boot_serial.c
```diff
@@ -1550,6 +1616,22 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
...
+         * packets have been received; a single read may carry several complete
+         * packets (or a packet plus the start of the next). Dispatch all
+         * complete packets and keep any trailing bytes for the next read.
+         */
+#ifdef MCUBOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT
+        /*
+         * New data arrived; restart the inactivity window so the timeout below
+         * fires only after a partial packet has stalled, not mid-transfer.
+         */
+        raw_input_start = mcuboot_uptime_get_ms_32();
```
nordicjm: prefix of this function should not be `mcuboot_`, have it as e.g. `os_`

boot/zephyr/include/uptime.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2026 Intercreate, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+
```
nordicjm: this can then go into os.h

boot/zephyr/uptime.c
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2026 Intercreate, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <uptime.h>
+#include <zephyr/kernel.h>
+
+__weak uint32_t mcuboot_uptime_get_ms_32(void)
```
nordicjm: no need for weak on this, this is the correct function

nordicjm (commented):
Some small bits need changing but looks good!

boot/zephyr/uptime.c
JPHutchins: fixed, and changed to `os_uptime_get_ms_32()`, moved to os.c, and remove uptime.c

boot/zephyr/include/uptime.h
JPHutchins: Moved this to os.h, and removed uptime.h.

boot/boot_serial/src/boot_serial.c
JPHutchins: fixed

boot/boot_serial/src/boot_serial.c
JPHutchins: fixed

nordicjm (approved):

de-nordic (approved):

PR #2758, opened 2026-06-12
espressif: ESP32-P4 initial support

almir-okato:
Add initial support for ESP32-P4.


boot/espressif/port/include/esp32p4/memory.h
```diff
@@ -0,0 +1,36 @@
...
+        (BOOTLOADER_IRAM_LOADER_SEG_START - BOOTLOADER_DRAM_LOADER_SEG_LEN)
+
+#define BOOTLOADER_IRAM_SEG_LEN             0xF000
+#define BOOTLOADER_IRAM_SEG_START \
+        (BOOTLOADER_DRAM_LOADER_SEG_START - BOOTLOADER_IRAM_SEG_LEN)
+
+#define BOOTLOADER_DRAM_SEG_LEN             0xB000
+#define BOOTLOADER_DRAM_SEG_START \
+        (BOOTLOADER_IRAM_SEG_START - BOOTLOADER_DRAM_SEG_LEN)
+
```
sylvioalves: Fix this missing newline

sylvioalves (approved):

almir-okato:
@d3zd3z could you please take a look?

boot/espressif/ci_configs/esp32p4-secureboot.conf
```diff
@@ -0,0 +1,15 @@
...
+
+# ATTENTION:
+# This configuration file targets the building for CI environment and contains
+# a set of definitions to resemble a bootloader image for RELEASE environment.
+# Running the generated firmware image may result in irreversible operations
+# to the chip!
+
+CONFIG_SECURE_SIGNED_APPS_RSA_SCHEME=1
+# ESP32-P4 CI overlay: skip flash encryption and downgrade prevention
+CONFIG_SECURE_FLASH_ENC_ENABLED=0
```
d3zd3z: I don't think these do what you are expecting. They'll get defined with `#define` and the code checks with ifdef, so they'll end up set, rather than unset.

.github/workflows/espressif.yaml
```diff
@@ -18,10 +18,13 @@ jobs:
...
     strategy:
       matrix:
-        targets: [esp32, esp32s2, esp32s3, esp32c2, esp32c3, esp32c5, esp32c6, esp32c61, esp32h2]
+        targets: [esp32, esp32s2, esp32s3, esp32c2, esp32c3, esp32c5, esp32c6, esp32c61, esp32h2, esp32p4]
         features:
         - "secureboot-sign-rsa2048,secureboot-sign-rsa3072,secureboot-sign-ec256,secureboot-sign-ed25519"
         - "serialrecovery"
+        exclude:
+          - targets: esp32p4
+            features: "secureboot-sign-rsa2048,secureboot-sign-rsa3072,secureboot-sign-ec256,secureboot-sign-ed25519"
```
d3zd3z: Are these exclusions inherent to this device (not enough memory?). That makes sense, as long as this isn't just something that isn't working yet.

d3zd3z: Perhaps these aren't working because of the `=0` issues below.

d3zd3z (changes requested):
One issue, that I think is probably the reason you had to disable some tests.

boot/espressif/ci_configs/esp32p4-secureboot.conf
d3zd3z: I think you want `=n` on these (and `=y` on the boolean ones).

almir-okato: I removed those, so far the building for the configuration works. But thanks for noticing the `=0` issue.

almir-okato:
Hi @d3zd3z , could you please take a look again?

fdcavalcanti (approved):

eren-terzioglu (approved):

d3zd3z (approved):

PR #2760, opened 2026-06-15
boot/zephyr: serial_adapter: batch UART FIFO reads in ISR

Solar936:
Replace the single-byte `uart_fifo_read()` loop in `boot_uart_fifo_callback()` with a configurable batch read. Within each batch, `memchr()` locates the line delimiter and `memcpy()` copies data into the line buffer, reducing per-byte loop overhead compared to the previous approach.

A new Kconfig option `BOOT_SERIAL_UART_RX_BATCH_SIZE` (default 512, range 1-1024) is added to control the read batch size.

The behavior for overlong lines (truncation + CRC failure) is preserved from the original implementation.

### DIFF
--- a/boot/zephyr/Kconfig.serial_recovery
+++ b/boot/zephyr/Kconfig.serial_recovery
@@ -137,6 +137,15 @@ config BOOT_SERIAL_MAX_RECEIVE_SIZE
 	  bound keeps the buffer large enough to hold one fragment so that a
 	  misconfiguration fails here rather than silently dropping frames.
 
+config BOOT_SERIAL_UART_RX_BATCH_SIZE
+	int "UART RX batch read size"
+	default 512
+	range 1 1024
+	help
+	  Number of bytes read from the UART FIFO per uart_fifo_read() call
+	  in the serial recovery ISR. Larger values reduce per-call overhead
+	  at the cost of a slightly larger static buffer.
+
 config BOOT_ERASE_PROGRESSIVELY
 	bool "Erase flash progressively when receiving new firmware"
 	default y if SOC_FAMILY_NORDIC_NRF || SOC_FAMILY_NXP_IMXRT
--- a/boot/zephyr/serial_adapter.c
+++ b/boot/zephyr/serial_adapter.c
@@ -62,6 +62,9 @@ static sys_slist_t avail_queue;
 static sys_slist_t lines_queue;
 
 static uint16_t cur;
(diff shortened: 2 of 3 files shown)


Solar936:
@d3zd3z @de-nordic @nordicjm

Friendly ping. The PR has been rebased onto current `main`, the conflict is resolved, and the history is cleaned back to a single commit. It should be ready for review now. Thanks!

Solar936:
Performance comparison under the same test conditions:

- UART baud rate: 3 Mbps
- CPU clock: 44 MHz
- Host sends 2880 bytes in one burst

Before this change:
- single-byte `uart_fifo_read()`
- ISR time spent receiving/processing: about 23 ms

After this change:
- batch read with `uart_fifo_read(..., 512)`
- total ISR time spent receiving/processing: about 1 ms

So in this setup, the ISR time dropped from ~23 ms to ~1 ms.

I also checked the code size impact in mcuboot.

Compared with the previous build, the text size increase is only +192 bytes. The standalone `memchr()` implementation contributes 36 bytes of text, and the remaining increase comes from the added batch RX control logic itself.

So the code-size overhead in text is quite small.

nordicjm (commented):
So, some issues whilst testing this:
```
/tmp/aa/bootloader/mcuboot/boot/zephyr/serial_adapter.c: In function 'boot_uart_fifo_callback':
/tmp/aa/bootloader/mcuboot/boot/zephyr/serial_adapter.c:188:17: warning: unused variable 'byte' [-Wunused-variable]
  188 |         uint8_t byte;
      |                 ^~~~
```
and:
```
/tmp/aa/bootloader/mcuboot/boot/zephyr/serial_adapter.c:65:16: warning: 'rx_batch_buf' defined but not used [-Wunused-variable]
   65 | static uint8_t rx_batch_buf[CONFIG_BOOT_SERIAL_UART_RX_BATCH_SIZE];
      |                ^~~~~~~~~~~~
```
~~Also with raw UART transport depending upon what data you send it appears you can get this into a state whereby it is completely broken and does not work anymore~~ seems I can't reproduce this anymore

Solar936:
Thanks for catching this. I've fixed the warning by scoping the variables to the relevant build paths only:
- `byte` is now declared only in the raw protocol path
- `rx_batch_buf` is now declared only in the non-raw path

This keeps the behavior unchanged and removes the unused-variable warnings in both configurations.

nordicjm (approved):

de-nordic (approved):

PR #2762, opened 2026-06-17
ci: fih: Use official TF-M github repos

nordicjm:
Uses the official repos and also gets tf-m-extras to avoid still continual random clone failures when using the TF-M git server, which started having similar issues over a year ago


adeaarm:
Can you check why `tf-m-extras` is still being downloaded with the change proposed here? @davidvincze @mdalzellarm

ci/fih_test_docker/execute_test.sh
```diff
@@ -44,6 +44,7 @@ cmake -S $TFM_TESTS_PATH/tests_reg/spe \
     -B $TFM_SPE_BUILD_PATH \
     -DTFM_PLATFORM=arm/mps2/an521 \
     -DCONFIG_TFM_SOURCE_PATH=$TFM_PATH \
+    -DCONFIG_TFM_EXTRAS_SOURCE_PATH=$TFM_EXTRAS_PATH \
```
davidvincze:
```suggestion
    -DTFM_EXTRAS_REPO_PATH=$TFM_EXTRAS_PATH \
```

davidvincze: I was 1 minute late. :)

nordicjm: Seems to have fixed it, thanks!

de-nordic (approved):

PR #2763, opened 2026-06-17
Fixed compiler warnings in boot_serial

### LINKED ISSUE #2753: Compiler warnings
Following the recent move to GCC 15.2.Rel1 in .NET nanoFramework pipelines, the following warnings surfaced:

```text
D:/a/1/s/build/_deps/mcuboot-src/boot/boot_serial/src/boot_serial.c:282:46: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'uint32_t' {aka 'long unsigned int'} [-Wformat=]
  282 |       snprintf(&dst[len], (maxlen - len), ".%u", ver->iv_build_num);
      |                                             ~^   ~~~~~~~~~~~~~~~~~
      |                                              |      |
      |                                              |      uint32_t {aka long unsigned int}
      |                                              unsigned int
      |                                             %lu
```

and

```text
D:/a/1/s/build/_deps/mcuboot-src/boot/boot_serial/src/boot_serial.c:1478:12: warning: comparison of unsigned expression in '< 0' is always false [-Wtype-limits]
 1478 |     if (rc < 0) {
      |            ^
```

Would you accept a PR fixing these, or would you prefer to handle this internally?

josesimoes:
- Switched the build-number format from %u to %" PRIu32 " for uint32_t.
- Made rc type `size_t` only for Zephyr/ESP base64 APIs, and `int` for the fallback decoder.

This last one may worth some thinking (from the code owner?) as it looks a bit off having to use a var with different types depending on the platform...

Closes #2753

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -278,7 +278,7 @@ bs_list_img_ver(char *dst, int maxlen, struct image_version *ver)
                   (uint16_t)ver->iv_minor, ver->iv_revision);
 
    if (ver->iv_build_num != 0 && len > 0 && len < maxlen) {
-      snprintf(&dst[len], (maxlen - len), ".%u", ver->iv_build_num);
+      snprintf(&dst[len], (maxlen - len), ".%" PRIu32, ver->iv_build_num);
    }
 }
 #endif /* !MCUBOOT_USE_SNPRINTF */
@@ -1497,7 +1497,11 @@ boot_serial_output(void)
 static int
 boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 {
+#if defined(__ZEPHYR__) || defined(__ESPRESSIF__)
     size_t rc;
+#else
+    int rc;
+#endif
     uint16_t crc;
     uint16_t len;
 
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -1497,27 +1497,28 @@ boot_serial_output(void)
 static int
 boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 {
-#if defined(__ZEPHYR__) || defined(__ESPRESSIF__)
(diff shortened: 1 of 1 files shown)


boot/boot_serial/src/boot_serial.c
```diff
@@ -1485,7 +1485,11 @@ boot_serial_output(void)
 static int
 boot_serial_in_dec(char *in, int inlen, char *out, int *out_off, int maxout)
 {
+#if defined(__ZEPHYR__) || defined(__ESPRESSIF__)
     size_t rc;
+#else
+    int rc;
+#endif
```
nordicjm: I don't really follow this, int and size_t will be the same size, this isn't actually doing anything?

de-nordic: `int err` is already locally defined in below ifdefs, the `rc` definition could be moved there too.
Then intermediate variable of int type could be used to complete arithmetic at line `*out_off += rc;`.

josesimoes: This is not about size, rather fixing the warning about comparing unsigned expression in '< 0' which was always false. Please take a look at the linked issue.

nordicjm: replace `int` with `unsigned int`? What am I missing

josesimoes: Replacing the generic-path int rc with unsigned int would not work as-is. In the non-Zephyr/non-ESPRESSIF path, base64_decode() signals failure with a negative return value. Line 1513. rc < 0 would always be false, so decode errors would no longer be caught. Which puts us back in the initial concern of this issue. rc has to stay signed. Or the code has to be further reworked.

nordicjm (approved):

boot/boot_serial/src/boot_serial.c
josesimoes: Done!

josesimoes:
Just noticed that I've deleted this work branch unintentionally. Apologies.

PR #2771, opened 2026-06-19
sim: Fix STATUS_ACCESS_VIOLATION on Windows x86_64 

josesimoes:
The C `jmp_buf` embedded in this FFI struct needs 16-byte alignment under mingw's setjmp (saves XMM registers with movaps). The C compiler aligns the struct to 16 to satisfy that, but Rust's repr(C) only picked the 8-byte alignment of [u64; 48], so the field could land 8-byte aligned and setjmp faulted with STATUS_ACCESS_VIOLATION.

### DIFF
--- a/sim/mcuboot-sys/src/api.rs
+++ b/sim/mcuboot-sys/src/api.rs
@@ -89,7 +89,10 @@ impl Default for FlashContext {
     }
 }
 
-#[repr(C)]
+// Windows (both mingw and msvc) requires 16-byte alignment for jmp_buf.
+// Always align to 16 bytes; the few extra bytes wasted on other platforms
+// are immaterial for a struct instantiated a handful of times per test.
+#[repr(C, align(16))]
 #[derive(Debug)]
 pub struct CSimContext {
     pub flash_counter: libc::c_int,
@@ -101,6 +104,10 @@ pub struct CSimContext {
     // The size below is enough to store data on a x86_64 machine.
     pub boot_jmpbuf: [u64; 48],
 }
+
+// Guard against a future field change shifting boot_jmpbuf off a
+// 16-byte boundary (struct alignment alone doesn't guarantee this).
+const _: () = assert!(mem::offset_of!(CSimContext, boot_jmpbuf) % 16 == 0);
 
 impl Default for CSimContext {
     fn default() -> Self {


josesimoes:
Pinging @d3zd3z in case this slipped through the cracks...

sim/mcuboot-sys/src/api.rs
```diff
@@ -89,7 +89,9 @@ impl Default for FlashContext {
     }
 }
 
-#[repr(C)]
+// On x86_64-pc-windows-gnu mingw's requires 16-byte alignment
+#[cfg_attr(all(target_os = "windows", target_env = "gnu"), repr(C, align(16)))]
```
d3zd3z: This misses windows-msvc. But, why not just remove the conditionals and make this just `#[repr(C, align(16)))]`. It's a single, fairly large struct instantiated a few times per test, the extra 8 bytes occasionally wasted on other platforms shouldn't hurt anything.

Since this isn't tested by CI on windows, some reordering or change of the fields could silently re-introduce misalignment.  Perhaps:

```rust
const _: () = assert!(align_of::<CSimContext)() >= 16);
```

after the declaration should both document, and check at compile time.

d3zd3z (changes requested):

sim/mcuboot-sys/src/api.rs
josesimoes: Done!
Is it worth it to add a Windows machine to the pipeline?

sim/mcuboot-sys/src/api.rs
```diff
@@ -102,6 +105,10 @@ pub struct CSimContext {
     pub boot_jmpbuf: [u64; 48],
 }
 
+// Guard against a future field reordering/addition silently dropping the
+// alignment below what Windows' jmp_buf requires.
+const _: () = assert!(mem::align_of::<CSimContext>() >= 16);
```
d3zd3z: So, this is apparently not asserting the right thing, sorry, my bad.

This might help with the failure, and actually checks what we are supposed to be checking.
```rust
const _: () = assert!(mem::offset_of!(CSimContext, boot_jmpbuf) % 16 == 0);
```

josesimoes: Got it! New assert commited.

josesimoes:
@d3zd3z anything missing here?

d3zd3z (commented):
This looks good. The only thing I'd say remains would be to just squash the three commits into a single commit. This will help with future bisecting.

josesimoes:
@d3zd3z done. And I've also rebased on main.

d3zd3z (approved):

PR #2778, opened 2026-07-03
boot: bootutil: Fix aes.h include path

nordicjm:
Fixes this include path on newer versions of MBEDTLS

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
@@ -13,8 +13,13 @@
 #include <string.h>
 #include <stdint.h>
 #include "mcuboot_config/mcuboot_config.h"
+#include "bootutil/crypto/common.h"
 #include "bootutil/enc_key_public.h"
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
+#include <mbedtls/private/aes.h>
+#else
 #include <mbedtls/aes.h>
+#endif
 
 #define BOOT_ENC_BLOCK_SIZE (16)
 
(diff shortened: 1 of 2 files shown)


boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
```diff
@@ -14,7 +14,11 @@
 #include <stdint.h>
 #include "mcuboot_config/mcuboot_config.h"
 #include "bootutil/enc_key_public.h"
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
```
d3zd3z, line 18: I don't think this will work unless you `#include "bootutil/crypto/common.h"` before this. Search for other uses of this macro.

boot/bootutil/include/bootutil/crypto/aes_kw.h
```diff
@@ -18,7 +18,11 @@
 #endif
 
 #if defined(MCUBOOT_USE_MBED_TLS)
+#if MCUBOOT_MBEDTLS_CRYPTO_IN_PRIVATE_SUBDIR
```
d3zd3z, line 23: Same here, needs common.h.

d3zd3z (changes requested):

boot/bootutil/include/bootutil/crypto/aes_kw.h
nordicjm: Fixed

boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
nordicjm: Fixed

d3zd3z (approved):

PR #2784, opened 2026-07-07
boot: bootutil: loader: Fix bootstrap with swap using offset

nordicjm:
Fixes an issue when bootstrap is used to load an image in swap
    using offset mode whereby the copy size was not reduced to account
    for the swap sector

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -934,7 +934,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     const struct flash_area *fap_secondary_slot;
     uint8_t image_index;
 
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
     uint32_t sector;
     uint32_t trailer_sz;
     uint32_t off;
@@ -943,7 +943,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
 
     (void)bs;
 
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
     uint32_t src_size = 0;
     rc = boot_read_image_size(state, BOOT_SLOT_SECONDARY, &src_size);
     assert(rc == 0);
@@ -966,7 +966,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
         rc = boot_erase_region(fap_primary_slot, size, this_size, false);
         assert(rc == 0);
 
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
         if ((size + this_size) >= src_size) {
             size += src_size - size;
             size += BOOT_WRITE_SZ(state) - (size % BOOT_WRITE_SZ(state));
@@ -977,28 +977,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
         size += this_size;
     }
 
-#if defined(MCUBOOT_SWAP_USING_MOVE)
-    /* When using MCUBOOT_SWAP_USING_MOVE, primary region is larger then the secondary region
-     * Optimal region configuration: # useful regions in primary region = # regions in secondary region + 1
-     * This means that we have to use the size of the secondary region (so without the swap sector)
-     */
-    sect_count = boot_img_num_sectors(state, BOOT_SLOT_SECONDARY);
-    for (sect = 0, size = 0; sect < sect_count; sect++) {
-        this_size = boot_img_sector_size(state, BOOT_SLOT_SECONDARY, sect);
-
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
-        if ((size + this_size) >= src_size) {
-            size += src_size - size;
-            size += BOOT_WRITE_SZ(state) - (size % BOOT_WRITE_SZ(state));
-            break;
-        }
-#endif
-
-        size += this_size;
-    }
-#endif
-
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
     trailer_sz = boot_trailer_sz(BOOT_WRITE_SZ(state));
     sector = boot_img_num_sectors(state, BOOT_SLOT_PRIMARY) - 1;
     sz = 0;
@@ -1039,7 +1018,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
         return rc;
     }
 
-#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST) || defined(MCUBOOT_SWAP_USING_MOVE) || defined(MCUBOOT_SWAP_USING_OFFSET)
     rc = boot_write_magic(fap_primary_slot);
     if (rc != 0) {
         return rc;
@@ -1065,7 +1044,7 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     }
 #endif /* MCUBOOT_HW_ROLLBACK_PROT */
 
-#ifndef MCUBOOT_OVERWRITE_ONLY_KEEP_BACKUP
+#if !defined(MCUBOOT_OVERWRITE_ONLY_KEEP_BACKUP) && !defined(MCUBOOT_SWAP_USING_MOVE) && !defined(MCUBOOT_SWAP_USING_OFFSET)
     /*
      * Erases header and trailer. The trailer is erased because when a new
      * image is written without a trailer as is the case when using newt, the
--- a/docs/release-notes.d/bootstrap-swap-offset.md
+++ b/docs/release-notes.d/bootstrap-swap-offset.md
@@ -1,0 +1,4 @@
+- Fixed various issues with bootstrap in swap modes whereby it
+  could read from beyond the flash area, erase (or not) erase
+  sectors, and whereby it would pointlessly copy the entire
+  slot instead of just copying the image data.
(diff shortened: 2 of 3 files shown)


boot/bootutil/src/loader.c
```diff
@@ -1027,6 +1027,11 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     }
 #endif
 
+#if defined(MCUBOOT_SWAP_USING_OFFSET)
+    /* Reduce copy size after erase to ensure whole of primary image is erased */
+    size -= boot_img_sector_size(state, BOOT_SLOT_SECONDARY, 0);
```
d3zd3z: This is assuming the size is a multiple of the sector size here. When less, this subtracts too much. I believe this is the cause of the CI failure.

boot/bootutil/src/loader.c
d3zd3z: Note how this is calculated here.

d3zd3z (changes requested):
Currently, enabling bootstrap forces `OVERWRITE_ONLY_FAST` in `build.rs` in the sim, I don't think this fix will even get tested by the sim. It does test the path where this breaks, though.

d3zd3z:
Some notes from Claude that might be helpful for addressing the failures:

# PR #2784 — what the CI failures actually are

Short version: your `build.rs` change is right, and it does what it was asked to do. But it
compiles a code path — `MCUBOOT_BOOTSTRAP` **without** `MCUBOOT_OVERWRITE_ONLY_FAST` — that the
simulator has never compiled before, and that path has two bugs in it. Neither bug is caused by
your `loader.c` change. They were always there; the old `build.rs` was hiding them.

There are **two** failures, not one. CI only reports one of them, because the other is masked by
a fail-fast cancellation.

---

## Background: why the sim never tested this before

The old code was:

```rust
if bootstrap {
    conf.conf.define("MCUBOOT_BOOTSTRAP", None);
    conf.conf.define("MCUBOOT_OVERWRITE_ONLY_FAST", None);   // unconditional
}
```

So every sim build with `bootstrap` also got `OVERWRITE_ONLY_FAST`.

That combination cannot happen on real hardware. In Zephyr, `MCUBOOT_OVERWRITE_ONLY_FAST` comes
only from `CONFIG_BOOT_UPGRADE_ONLY`:

```c
/* boot/zephyr/include/mcuboot_config/mcuboot_config.h:82-85 */
#ifdef CONFIG_BOOT_UPGRADE_ONLY
#define MCUBOOT_OVERWRITE_ONLY
#define MCUBOOT_OVERWRITE_ONLY_FAST
#endif
```

and `BOOT_UPGRADE_ONLY`, `BOOT_SWAP_USING_MOVE` and `BOOT_SWAP_USING_OFFSET` are mutually
exclusive members of `choice BOOT_IMAGE_UPGRADE_MODE` (`boot/zephyr/Kconfig:597-684`).
`CONFIG_BOOT_BOOTSTRAP` (`Kconfig:710-715`) has no `depends on`, so it combines freely with any
of them.

**So on real hardware, `bootstrap` + any swap mode always means no `OVERWRITE_ONLY_FAST`.** That
is precisely the path the sim never built. Your `build.rs` change fixes that divergence — this
is the most valuable part of the PR, and it's why the failures showed up.

---

## Failure 1 — `swap-move bootstrap`: write to an unerased trailer

This is the one CI reports:

```
environment (sig-ecdsa-psa,sig-ecdsa-psa sig-p384,sig-ecdsa-psa swap-move bootstrap max-align-16)
thread 'bootstrap' panicked at sim/simflash/src/lib.rs:209:
Write to unerased location at 0x3ffe0
```

Reproduce:

```sh
cd sim
MCUBOOT_SKIP_SLOW_TESTS=true \
  cargo test --features sig-ecdsa-psa,swap-move,bootstrap,max-align-16 bootstrap
```

`0x3ffe0` is the **`image_ok` field of the primary slot's trailer** (slot ends at `0x40000`,
`BOOT_MAX_ALIGN=16` puts `image_ok` two aligns before the end).

Walking `boot_copy_image()` in `boot/bootutil/src/loader.c` with swap-move and no
`OVERWRITE_ONLY_FAST`:

1. **Lines 964-978** — the primary erase loop runs to completion (its `OVERWRITE_ONLY_FAST`
   early-break is compiled out), so the whole primary slot, trailer included, is erased.

2. **Lines 980-999** — the `SWAP_USING_MOVE` block recomputes `size` by summing the secondary
   sectors. Its `OVERWRITE_ONLY_FAST` early-break is *also* compiled out, so `size` ends up as
   the **full secondary slot size**, never truncated to the actual image size.

3. **Line 1041** — `boot_copy_region(..., 0, 0, size)` copies that full slot size into the
   primary. This writes across the primary trailer region, including `0x3ffe0`. First write.

4. **Lines 1047-1052** — `boot_write_magic()` and the trailer re-erase above it are both
   `#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)`. Compiled out. So nothing re-erases the trailer.

5. **Line 1276** — back in `boot_perform_update()`, bootstrap is a permanent install, so
   `swap_set_image_ok()` writes `image_ok` at `0x3ffe0`. Second write, to an unerased location.
   Panic.

In the old fast path, step 2 truncated `size` to the image size (so the copy never reached the
trailer) *and* step 4 re-erased the trailer before writing magic. Removing `OVERWRITE_ONLY_FAST`
removed both protections at once.

### This is not your `loader.c` change

Your hunk is inside `#if defined(MCUBOOT_SWAP_USING_OFFSET)`, and this build is swap-**move**, so
it isn't even compiled. Confirmed empirically — with `main`'s `loader.c` and your `build.rs`, the
same panic occurs:

```
main's loader.c + your build.rs, swap-move  → Write to unerased location at 0x3ffe0
main's loader.c + your build.rs, swap-offset → Write to unerased location at 0x3ffe8
```

and restoring the unconditional `OVERWRITE_ONLY_FAST` define makes both pass again on `main`'s
`loader.c`. The bug is pre-existing; `build.rs` only made it visible.

---

## Failure 2 — `swap-offset bootstrap`: trailer never finalized

**This one does not appear as a failure in CI.** `.github/workflows/sim.yaml:32` is a three-part
matrix line:

```
sig-rsa enc-rsa validate-primary-slot,
swap-move enc-rsa sig-rsa validate-primary-slot bootstrap,
swap-offset enc-rsa sig-rsa validate-primary-slot bootstrap
```

GitHub shows it as **CANCELLED** (fail-fast collateral from Failure 1), which reads as harmless.
It isn't. Run the third element by hand:

```sh
cd sim
MCUBOOT_SKIP_SLOW_TESTS=true \
  cargo test --features swap-offset,enc-rsa,sig-rsa,validate-primary-slot,bootstrap bootstrap
```

```
test bootstrap ... FAILED
assertion failed: !image.run_bootstrap()
[WARN] Mismatched trailer for the primary slot
```

`run_bootstrap()` (`sim/src/image.rs:690-719`) checks three things. Note which one fails:

1. `boot_go()` succeeds — **passes**.
2. `verify_images()` — the primary slot content matches the image byte-for-byte — **passes**.
   Your copy-length fix works. The image lands correctly.
3. `verify_trailers(..., BOOT_MAGIC_GOOD, BOOT_FLAG_SET, BOOT_FLAG_SET)` — **fails**.

The primary trailer never gets `magic` / `image_ok` / `copy_done` set. Same root cause as step 4
above: `boot_write_magic(fap_primary_slot)` at `loader.c:1047-1052` is `OVERWRITE_ONLY_FAST`-only
and is now compiled out.

Your fix shortens the copy by one sector, so unlike swap-move it no longer clobbers the trailer
(that's why you get a clean assertion instead of the unerased-write panic). But nothing writes a
valid trailer afterwards.

**So the configuration this PR exists to fix still does not work end to end.** The over-read is
gone; the finalization is missing.

---

## The shape of the fix

Both failures come from the same place: `boot_copy_image()` has trailer handling that only exists
under `OVERWRITE_ONLY_FAST`.

```c
#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
    /* re-erase the trailer region  — loader.c:1001-1013 */
#endif
    ...copy...
#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
    rc = boot_write_magic(fap_primary_slot);   /* loader.c:1047-1052 */
#endif
```

For bootstrap under a swap mode, one or both of these needs a home outside that guard — or the
copy length needs to stop short of the trailer so the normal swap finalize path can write it
cleanly.

Worth deciding before writing code:

- **Does bootstrap + swap-offset work on your hardware today?** If it does, then the sim's
  `verify_trailers` expectation may be what's wrong for a non-overwrite bootstrap, and the loader
  is fine. If it doesn't, the loader needs the fix above. Only you can say which.
- **Should `build.rs` be split into its own PR?** It stands on its own — it corrects a real
  sim/Zephyr divergence — and it surfaces two bugs unrelated to the copy length. Landing it
  separately would let the `loader.c` fix be judged on its own, and give the trailer bug its own
  changelog entry.

---

## Two smaller notes on `loader.c`

**The subtraction assumes `size` is a sector multiple.** Under `OVERWRITE_ONLY_FAST` the erase
loop's early break (`loader.c:969-975`) leaves `size` at `round_up(src_size, WRITE_SZ)` — the
*image* size, not the slot size. Subtracting a full sector then truncates the image, and for an
image smaller than one sector, `size` is `size_t` and the subtraction underflows.

This is currently unreachable: Zephyr's Kconfig `choice` forbids offset+`UPGRADE_ONLY`; Espressif
never defines `OVERWRITE_ONLY_FAST` (`boot/espressif/include/mcuboot_config/mcuboot_config.h:53-64`,
commented out); and post-PR `build.rs:79-81` excludes it in the sim. Your own `build.rs` change is
what closes it. Mentioned only so it's a conscious decision rather than an accident.

**Consider recomputing `size` rather than subtracting.** The `SWAP_USING_MOVE` block directly above
(`loader.c:980-999`) recomputes `size` by summing the secondary sectors, with the fast-path break
re-applied inside the loop. Mirroring that for offset mode would sidestep the above and wouldn't
hard-code "subtract exactly sector 0's size."

One real difference to think about: `size -= S` yields `primary_size − S`, while a secondary-sector
recompute yields `secondary_size − S`. `boot_slots_compatible()` (`swap_offset.c:316-317`) permits
both `num_pri == num_sec` and `num_pri + 1 == num_sec`. In the equal case the two agree; in the
`num_sec = num_pri + 1` case they differ by a sector, and the subtraction copies one sector fewer
than the secondary holds. Whether that sector can contain image data is worth confirming.

---

## Reproduction notes

All runs above were done in a checkout at the PR tip with `MCUBOOT_SKIP_SLOW_TESTS=true`.

One trap when producing a baseline: don't build `origin/main` by checking out just its
`sim/mcuboot-sys/build.rs` into a newer tree. `main`'s `build.rs` needs a `cmake` crate that isn't
in this tip's lockfile, and the build fails with `error[E0463]: can't find crate for 'cmake'` — the
test never runs, and the empty output looks like a pass. To baseline, either build `main` fully, or
keep the tip's tree and just neutralize the guard:

```rust
if bootstrap {
    conf.conf.define("MCUBOOT_BOOTSTRAP", None);
    if true {                              // was: if !swap_offset && !swap_move
        conf.conf.define("MCUBOOT_OVERWRITE_ONLY_FAST", None);
    }
}
```

With that, both bootstrap tests pass on `main`'s `loader.c` — which is the proof that the failures
are unmasked, not introduced.

d3zd3z (approved):
My concerns have been addressed. Looks good.

de-nordic (approved):

PR #2785, opened 2026-07-09
Logical sectors

nordicjm:
Duplicate of https://github.com/mcu-tools/mcuboot/pull/2352

Original PR text:

So far tested on nrf52840 with internal memory only.
Normal build from main, with hardware pages/sectors

```
Memory region         Used Size  Region Size  %age Used
           FLASH:       32716 B        48 KB     66.56%
             RAM:       22464 B       256 KB      8.57%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors ( -DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_FLASH_PAGE_LAYOUT=n)
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       32188 B        48 KB     65.49%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors validation build (-DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_MCUBOOT_VERIFY_LOGICAL_SECTORS=y)
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       33452 B        48 KB     68.06%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```

PR #2352, opened 2025-06-23
Logical sectors

de-nordic:
So far tested on nrf52840 with internal memory only.
Normal build from main, with hardware pages/sectors

```
Memory region         Used Size  Region Size  %age Used
           FLASH:       32716 B        48 KB     66.56%
             RAM:       22464 B       256 KB      8.57%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors ( -DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_FLASH_PAGE_LAYOUT=n)
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       32188 B        48 KB     65.49%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors validation build (-DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_MCUBOOT_VERIFY_LOGICAL_SECTORS=y)
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       33452 B        48 KB     68.06%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```

PR #2786, opened 2026-07-09
Logical sectors

### LINKED ISSUE #2314: Move to "software defined page size"
Remove per-sector query of page size, and make all of mcuboot use a software defined page size. This should allow improvements with variable page-size blocks, as well as allowing swap-move to work on all device, and better support for external devices with differing page sizes.

- [ ] Single query for erase and write sizes, consistent across entire partition.
- [ ] Definable in DT (on Zephyr) to override real sizes.
- [ ] Testing

nordicjm:
Duplicate of https://github.com/mcu-tools/mcuboot/pull/2352
Duplicate of https://github.com/mcu-tools/mcuboot/pull/2785 but done on my account so I can rebase
Fixes #2314

Original PR text:

So far tested on nrf52840 with internal memory only.
Normal build from main, with hardware pages/sectors

```
           FLASH:       32716 B        48 KB     66.56%
             RAM:       22464 B       256 KB      8.57%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors ( -DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_FLASH_PAGE_LAYOUT=n)
```
           FLASH:       32188 B        48 KB     65.49%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Logical sectors validation build (-DCONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE=4096 -DCONFIG_MCUBOOT_VERIFY_LOGICAL_SECTORS=y)
```
           FLASH:       33452 B        48 KB     68.06%
             RAM:       20544 B       256 KB      7.84%
        IDT_LIST:          0 GB        32 KB      0.00%
```


boot/bootutil/src/bootutil_area.c
```diff
@@ -128,23 +127,38 @@ boot_header_scramble_off_sz(const struct flash_area *fa, int slot, size_t *off,
...
         /* For device requiring erase align to erase unit */
+        struct flash_sector sector;
+
         ret = flash_area_get_sector(fa, loff, &sector);
         if (ret < 0) {
             return ret;
         }
 
         *size = flash_sector_get_size(&sector);
+#endif
```
josesimoes, line 161: The logical-sector branch skips bounds validation present in the physical path.
The non-logical branch validates loff/erase-unit placement via flash_area_get_sector(), which returns -ERANGE if the offset lies outside the flash area. The new MCUBOOT_LOGICAL_SECTOR_SIZE branch hardcodes loff/*size without any check that MCUBOOT_LOGICAL_SECTOR_SIZE (or 2 * MCUBOOT_LOGICAL_SECTOR_SIZE for the swap-offset secondary-slot case) actually fits within fa. On a misconfigured device (logical sector size larger than the flash area, or area not a multiple of it), this can silently compute an out-of-bounds header offset/size instead of failing safely as I believe it was happening before.
If device-geometry compatibility is expected to be validated earlier (during sector-buffer setup?) we should make sure that guard runs for all bootutil consumers of this function.

boot/bootutil/src/bootutil_misc.c
```diff
@@ -472,6 +476,139 @@ boot_initialize_area(struct boot_loader_state *state, int flash_area)
...
+#if MCUBOOT_SWAP_USING_SCRATCH
+    } else if (flash_area == FLASH_AREA_IMAGE_SCRATCH) {
+        area_size = flash_area_get_size(state->scratch.area);
+        out_num_sectors = &state->scratch.num_sectors;
+#endif
+    } else {
+        return BOOT_EFLASH;
+    }
+
+    *out_num_sectors = area_size / MCUBOOT_LOGICAL_SECTOR_SIZE;
```
josesimoes, line 606: *out_num_sectors = area_size / MCUBOOT_LOGICAL_SECTOR_SIZE; truncates any trailing partial sector without any check, unless MCUBOOT_VERIFY_LOGICAL_SECTORS is enabled (which the Kconfig/docs say should be disabled for production once validated). A cheap unconditional assert here would catch a misconfigured slot size even when full runtime verification is compiled out, at zero cost.

```suggestion
    `assert(area_size % MCUBOOT_LOGICAL_SECTOR_SIZE == 0);` 
    *out_num_sectors = area_size / MCUBOOT_LOGICAL_SECTOR_SIZE;
```

boot/zephyr/flash_map_extended.c
```diff
@@ -308,3 +309,19 @@ int flash_area_get_sector(const struct flash_area *fap, off_t off,
...
+                     (long)off, fap);
+        return -ERANGE;
+    }
+
+    fsp->fs_off = off & ~(CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE - 1);
+    fsp->fs_size = CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE;
+
+    return 0;
+}
+#endif
```
josesimoes, line 333: fsp->fs_off = off & ~(CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE - 1); assumes CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE is a power of two; if the Kconfig option ever allows a non-power-of-two value, this silently computes a wrong offset instead of failing. Additionally, fs_size is unconditionally set to the full logical sector size even for the last sector in an area whose size isn't an exact multiple of CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE, which can report a sector extending past the end of the flash area. Compare with the physical-page branch above, which derives real page boundaries from flash_get_page_info_by_offs() and cannot overrun the area.
We should make sure the Kconfig definition for CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE enforces a power-of-two range, and that upstream area-size/logical-sector compatibility validation actually prevents this overrun case from being reached in practice.

nordicjm: you cannot do such a thing in Kconfig, though you can assert in the code

boot/zephyr/flash_map_extended.c
```diff
@@ -308,3 +309,19 @@ int flash_area_get_sector(const struct flash_area *fap, off_t off,
 
     return rc;
 }
+#else
+int flash_area_get_sector(const struct flash_area *fap, off_t off,
```
josesimoes, line 319: Something like this?

```suggestion
BUILD_ASSERT((CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE &
              (CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE - 1)) == 0,
             "CONFIG_MCUBOOT_LOGICAL_SECTOR_SIZE must be a power of two");

int flash_area_get_sector(const struct flash_area *fap, off_t off,
```

boot/zephyr/flash_map_extended.c
josesimoes: Understood. Please take a look at the suggestion above.

boot/bootutil/src/bootutil_area.c
nordicjm: it would still fail though, `boot_header_scramble_off_sz` is called then `boot_scramble_region` is called which will return an error if it is beyond the flash area

boot/bootutil/src/bootutil_misc.c
nordicjm: added check

boot/zephyr/flash_map_extended.c
nordicjm: has been added

boot/bootutil/src/bootutil_area.c
josesimoes: understood

josesimoes (approved):
LGTM! 

PS: for some reason GH doens't let me resolve the conversation. Please do it at your end.

de-nordic (approved):
Looks OK. Thanks for dragging it to the finish line.

d3zd3z (approved):

PR #2789, opened 2026-07-14
boot: bootutil: Change assert() calls to ASSERT()

nordicjm:
MCUboot has an assert handler macro which is used from the simulator, align usage of assert calls to match it

### DIFF
--- a/boot/bootutil/src/bootutil_loader.c
+++ b/boot/bootutil/src/bootutil_loader.c
@@ -56,7 +56,7 @@ boot_check_header_erased(struct boot_loader_state *state, int slot)
     struct image_header *hdr;
 
     fap = BOOT_IMG_AREA(state, slot);
-    assert(fap != NULL);
+    ASSERT(fap != NULL);
 
     hdr = boot_img_hdr(state, slot);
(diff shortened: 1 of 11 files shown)


d3zd3z (approved):
One problem with `assert` is that it can be compiled out. For security, that can really never be the case. Does ASSERT have this property?  Otherwise, looks good.

nordicjm:
> One problem with `assert` is that it can be compiled out. For security, that can really never be the case. Does ASSERT have this property? Otherwise, looks good.

It's a macro you defined :) so there is this for all targets: https://github.com/mcu-tools/mcuboot/blob/74f4d64172601f14d1dc7e3f98a187fba66a56cb/boot/bootutil/include/bootutil/bootutil_public.h#L145 which does also allow using a custom assert, and the simulator uses this: https://github.com/mcu-tools/mcuboot/blob/74f4d64172601f14d1dc7e3f98a187fba66a56cb/sim/mcuboot-sys/csupport/mcuboot_config/mcuboot_assert.h#L25 which is important for the simulator as it needs to be able to _catch_ asserts for some tests rather than the rust executable dying at that point due to the assert() call, so for simulator this fixes this, for all other platforms there is no difference at all (unless they want a custom assert handler, in which case this would fix that also)

PR #2791, opened 2026-07-15
bootutil: crypto: Add PSA crypto implementation for AES-KW

tpambor:
Add a PSA crypto implementation for AES-KW. This implementation uses the PSA Crypto API to perform AES key unwrapping operations.

Support for key wrapping was added in version 1.4 of the PSA Crypto API specification.

This implementation has been tested with TF-M, which also provides a implementation in its [thin_psa_crypto_core](https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/main/lib/ext/thin-psa-crypto-core/thin_psa_crypto_core.c) layer.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_kw.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_kw.h
@@ -2,9 +2,9 @@
  * This module provides a thin abstraction over some of the crypto
  * primitives to make it easier to swap out the used crypto library.
  *
- * At this point, there are two choices: MCUBOOT_USE_MBED_TLS, or
- * MCUBOOT_USE_TINYCRYPT.  It is a compile error there is not exactly
- * one of these defined.
+ * At this point, there are three choices: MCUBOOT_USE_MBED_TLS,
+ * MCUBOOT_USE_TINYCRYPT, or MCUBOOT_USE_PSA_CRYPTO.  It is a compile
+ * error if there is not exactly one of these defined.
  */
 
 #ifndef __BOOTUTIL_CRYPTO_AES_KW_H_
@@ -13,8 +13,9 @@
 #include "mcuboot_config/mcuboot_config.h"
 
 #if (defined(MCUBOOT_USE_MBED_TLS) + \
+     defined(MCUBOOT_USE_PSA_CRYPTO) + \
      defined(MCUBOOT_USE_TINYCRYPT)) != 1
-    #error "One crypto backend must be defined: either MBED_TLS or TINYCRYPT"
+    #error "One crypto backend must be defined: either MBED_TLS or PSA or TINYCRYPT"
 #endif
 
 #if defined(MCUBOOT_USE_MBED_TLS)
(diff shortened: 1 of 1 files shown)


boot/bootutil/include/bootutil/crypto/aes_kw.h
```diff
@@ -66,6 +71,69 @@ static inline int bootutil_aes_kw_unwrap(bootutil_aes_kw_context *ctx, const uin
...
+    psa_key_id_t unwrapped_key_id = PSA_KEY_ID_NULL;
+    psa_status_t status;
+    size_t exported_length;
+
+    psa_set_key_type(&attributes, PSA_KEY_TYPE_AES);
+    psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_EXPORT);
+    psa_set_key_algorithm(&attributes, PSA_ALG_CTR);
+    psa_set_key_bits(&attributes, key_len * 8);
+
+    status = psa_unwrap_key(&attributes, ctx->key_id, PSA_ALG_ECB_NO_PADDING,
```
valeriosetti, line 121: As far as I can tell `psa_unwrap_key` is only implemented in the `thin_psa_crypto_core.c` you mentioned and that's only built when MCUboot is built as part of the TF-M build. TBH I don't know how common it is to encrypt images in devices which are not also using TF-M, but if that's possible shouldn't we add some guard to ensure that this block is not compiled when MCUboot is built without TF-M? Wdyt?

tpambor: No strong opinion here, but I don't think it is necessary as one can select PSA or another crypto implementation and whether to use KW or ECIES, etc. tf-psa-crypto currently does not implement it, but there are other PSA implementations like Oberon PSA that also have support for AES-KW (https://github.com/nrfconnect/sdk-oberon-psa-crypto/blob/87917371cc6f3883dbdf0c7751c48a358ab8920a/core/psa_crypto.c#L6041)

valeriosetti (approved):

boot/bootutil/include/bootutil/crypto/aes_kw.h
davidvincze: It's part of the PSA Crypto API specification since version 1.4.0:
https://arm-software.github.io/psa-api/crypto/

tpambor: Yes, unfortunately the API in `thin_psa_crypto_core.c` is a bit different from the specification in 1.4.0, so that this PR is currently is not compatible with 1.4.0. I have also submitted AES-KW support to tf-psa-crypto. Once this is merged it will be easier to test this PR. https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/863

davidvincze: Yes, currently `thin_psa_crypto_core.c` relies on `mbedtls_nist_kw_unwrap()` for unwrapping.

davidvincze (approved):

nordicjm (approved):

PR #2793, opened 2026-07-17
sim: Test power loss tearing the interrupted flash write

neilberkman:
The simulator's power-loss testing interrupts `boot_go` between whole flash operations: when the flash counter expires, `flash_area_write` in `sim/mcuboot-sys/csupport/run.c` returns via longjmp before calling the simulated flash write. Every simulated power cut is therefore write-atomic, so states left by a write torn partway through are unreachable. That is why the revert interruption sweep did not catch #1966: that failure needs power loss inside `boot_write_magic` during swap-move revert fixup, leaving a partially written magic that reads back as `BOOT_MAGIC_BAD` and matches no swap table row.

This PR adds a torn-write mode:

- When the interruption counter expires in torn mode, the write programs all but the last byte of its payload before the simulated power loss. Write alignment is a driver interface requirement, not a power-fail atomicity guarantee, so the tear point falls inside the final write unit regardless of `max-align`.
- The simulated flash marks torn bytes as retryable: one subsequent write of the same values is allowed, since recovery replays the interrupted operation. The write-once discipline stays enforced everywhere else, and unit tests cover the tear, the single replay, a mismatched replay, and erase restoring normal behavior.
- `run_revert_with_torn_writes` sweeps the revert path: perform a test upgrade, count the flash operations of an uninterrupted revert, then for each operation clone the post-upgrade state, tear that write during revert, and verify the device recovers to the original image with clean trailers. Gated to swap-using-move.

Validation:

- On current main, both the existing `revert_with_fails` sweep and the new torn-write sweep pass (`--features swap-move`, and `swap-move,max-align-32`).
- With the one-word fix from 4f393563 reverted (`BOOT_MAGIC_ANY` back to `BOOT_MAGIC_UNSET` in the revert row), the new sweep fails at the fixup_revert magic write, including under `max-align-32`, while the existing whole-operation sweep still passes. That reproduces #1966 offline in the simulator.

Follow-up candidates, intentionally out of scope: torn erases, multiple power cuts in one recovery chain, and non-prefix corruption patterns.

Context: this addresses the simulator question raised in the review of #2648.

### DIFF
--- a/docs/release-notes.d/sim-torn-writes.md
+++ b/docs/release-notes.d/sim-torn-writes.md
@@ -1,0 +1,5 @@
+- Added a torn-write mode to the simulator's power-loss testing: the
+  interrupted flash write can now be left partially programmed instead
+  of always aborting atomically, and a new revert sweep uses it to
+  cover interrupted-write states such as a partially written trailer
+  magic.
--- a/sim/mcuboot-sys/csupport/run.c
+++ b/sim/mcuboot-sys/csupport/run.c
@@ -48,6 +48,8 @@ extern int sim_flash_read(uint8_t flash_id, uint32_t offset, uint8_t *dest,
         uint32_t size);
 extern int sim_flash_write(uint8_t flash_id, uint32_t offset, const uint8_t *src,
         uint32_t size);
+extern int sim_flash_write_torn(uint8_t flash_id, uint32_t offset,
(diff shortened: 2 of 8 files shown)


neilberkman:
Folding this into #2648, which now carries both the simulator torn-write coverage and the CI canary as one change.

PR #2797, opened 2026-07-21
boot: imgtool: Extend align to 128B

URNHere:
I wanted to use the latest mcuboot upstream version together building it with Zephyr.
We are using a Renesas RA6M3 MCU, which has a controller which can write the flash only 128Bytes at a time.
I extended the aligment and max-alignment in the imgtool signer to allow the app to be signed by west, and the image gets correctly signed.

For what concerns bootloader support, I managed to make the boot work on a custom board we have by setting CONFIG_BOOT_SERIAL_UNALIGNED_BUFFER_SIZE=128.
CONFIG_BOOT_MAX_IMG_SECTORS_AUTO=n
CONFIG_BOOT_MAX_IMG_SECTORS=14
As the renesas drivers do not support the soc-nv-flash compatible yet and the mapped-partitions but AFAIK there is some work in progress here:
https://github.com/zephyrproject-rtos/zephyr/issues/110365
https://github.com/zephyrproject-rtos/zephyr/pull/109125


scripts/imgtool/main.py
```diff
@@ -473,11 +473,13 @@ def convert(self, value, param, ctx):
...
               'keyword to automatically generate it from the image version.')
 @click.option('-v', '--version', callback=validate_version,  required=True)
-@click.option('--align', type=click.Choice(['1', '2', '4', '8', '16', '32']),
+@click.option('--align', type=click.Choice(['1', '2', '4', '8', '16', '32',
+              '64', '128']),
               default='1',
               required=False,
               help='Alignment used by swap update modes.')
-@click.option('--max-align', type=click.Choice(['8', '16', '32']),
+@click.option('--max-align', type=click.Choice(['1', '2', '4', '8', '16', '32',
```
d3zd3z: This adds 1, 2 and 4, is this intentional? Should probably mention this in the PR/commit text if so. Has it been tested with these values?

d3zd3z (changes requested):
This doesn't appear to test the larger alignment in the simulator.

Also, the `docs/design.md` needs to be updated to include the new alignments possible.

scripts/imgtool/main.py
URNHere: Was a mistake, thanks for noticing!

URNHere:
Thanks for the review.
I followed what has been done in #2564. 
Didn't edit the sim.yaml yet (maybe you have some advice before I do).
When I run cargo test --features "max-align-128" or --features "max-align-64", the only test that fails is oversized_secondary_slot and only for Stm32f4SpiFlash (if I comment it out from ALL_DEVICES all tests pass). 
Do you have any hints?
First time contributing, and not so proficient with rust, let me know if you have some advice.

sim/Cargo.toml
```diff
@@ -36,6 +36,8 @@ direct-xip = ["mcuboot-sys/direct-xip"]
 downgrade-prevention = ["mcuboot-sys/downgrade-prevention"]
 max-align-16 = ["mcuboot-sys/max-align-16"]
 max-align-32 = ["mcuboot-sys/max-align-32"]
+max-align-64 = ["mcuboot-sys/max-align-64"]
```
d3zd3z, line 39: This part looks good, adding it to the simulator. But, you'll also need to add some stanzas to `.github/workflows/sim.yaml` so that they actually get run.

Trying to run some simple cases locally, these changes didn't appear to work, so that will have to be figured out.

d3zd3z (changes requested):
Please ensure the simulator support is also then tested. Also, the docs still need to be updated to include the new alignments.

sim/Cargo.toml
URNHere: Ok, thanks. I'll work on the docs too.
Which stanzas do you advice to add? Shall I replicate the ones we are already testing for max-align-32?

This is how I tested the changes locally, just to make sure I am doing it correctly:
I ran a `cargo clean` first, followed by a `cargo test --features "max-align-128"`, and this is what I get:
```
running 25 tests
test dependency_combos ... ok
[2026-07-23T14:38:24Z ERROR bootsim::image] Expected an upgrade failure and primary slot left untouched
test oversized_secondary_slot ... FAILED
test ram_load_corrupt_higher_version_image ... ok
test oversized_bootstrap ... ok
test hw_prot_missing_security_cnt ... ok
test ram_load_out_of_bounds ... ok
test hw_prot_failed_security_cnt_check ... ok
test ram_load_missing_header_flag ... ok
test ram_load_failed_validation ... ok
test direct_xip_first ... ok
test ram_load_from_flash ... ok
test ram_load_first ... ok
test bootstrap ... ok
test norevert_newimage ... ok
test ram_load_split ... ok
test secondary_trailer_leftover ... ok
test downgrade_prevention ... ok
test bad_secondary_slot ... ok
test status_write_fails_complete ... ok
test perm_with_random_fails ... ok
test norevert ... ok
test status_write_fails_with_reset ... ok
test basic_revert ... ok
test perm_with_fails ... ok
test revert_with_fails ... ok

failures:

---- oversized_secondary_slot stdout ----

thread 'oversized_secondary_slot' (440535) panicked at sim/tests/core.rs:62:1:
assertion failed: !image.run_fail_upgrade_primary_intact()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    oversized_secondary_slot

test result: FAILED. 24 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 32.01s
```

By digging into it a little deeper, it seems that the only device that fails the test is Stm32f4SpiFlash (if I comment it out of ALL_DEVICES all tests pass).
I can investigate it further, do you have any advice?

jeremydick:
https://github.com/mcu-tools/mcuboot/pull/2801 will handle setting CONFIG_BOOT_SERIAL_UNALIGNED_BUFFER_SIZE to 128 for all renesas RA family of MCUs

nordicjm:
Would suggest reading comments on https://github.com/mcu-tools/mcuboot/pull/2440 if your write size is 128 bytes then really it doesn't make much sense to use a swap mode since your overhead is so large it destroys any real swap use case

URNHere:
Thanks @nordicjm for the review. Still new to the project, and I missed that closed PR.
So, no swap modes due to the overhead it brings regarding flash size (I'll remove that assert line change).
However, I am missing one point to fully understand: #2440 was closed, but if I want to sign an image with Imgtool using --overwrite-only for the Renesas RA family I still need the modifications to allow 128B as max-align (since it is used by Imgtool even in that case), am I correct or am I missing something?

jeremydick:
I'm not understanding the reason for the closure of https://github.com/mcu-tools/mcuboot/pull/2440 either. I believe swap is still usable. The write block size is 128 bytes, but the erase block or sector size is larger (a combination of 8-8KiB and 30-32KiB blocks on the RA6M3, other RA6 family have similar). If you treat it all as 32KiB sectors, only 6KiB is required for the trailer, rather than 1.5MiB. Treating it all as 8KiB sectors requires a 24KiB trailer. Not great, but not unusable.

URNHere:
Yes, also looking back at the CMakeLists.txt we have:
```
 if(CONFIG_BOOT_MAX_IMG_SECTORS_AUTO)
    if(DEFINED slot0_size AND DEFINED erase_size_slot0)
      math(EXPR slot_min_sectors "${slot0_size} / ${erase_size_slot0}")

      if(${slot_min_sectors} GREATER ${auto_min_sectors})
        set(auto_min_sectors ${slot_min_sectors})

        if(${image} EQUAL 0)
          set(image_0_min_sectors ${slot_min_sectors})
        endif()
      endif()
```
In my case we have:
slot0_size = 448KB, erase_size_slot0 = 32KB (.dts of the RA6M3 [here ](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/renesas/ra/ra6/r7fa6m3af2cbg.dtsi)), as I reserved the small 8K sectors to MCUBoot, and all the big 32K to the app) , which yields to slot_min_sectors= 14.
```
Then we have:
 if(CONFIG_BOOT_SWAP_USING_SCRATCH OR CONFIG_BOOT_SWAP_USING_MOVE)
      if(CONFIG_BOOT_MAX_IMG_SECTORS_AUTO AND DEFINED image_0_min_sectors AND "${image_0_min_sectors}" GREATER "0")
        math(EXPR boot_status_data_size "${image_0_min_sectors} * (3 * ${write_size})")
```
And:
`      math(EXPR boot_swap_data_size "${max_align_size} * 4")`

boot_status_data_size = 14 * 3 * 128 = 5376 B
boot_swap_data_size  = 128*4 = 512 B

In the end: 
`    math(EXPR trailer_size "${key_size} + ${boot_magic_size} + ${boot_swap_data_size} + ${boot_status_data_size}")`
trailer_size = 256 (considering RSA) + 128 (magic is aligned to 128, which is the write size) + 5376 + 512 = 6272 B, which needs a whole 32KB sector, because it has to be aligned to the erase size. Not so great, but definitely far from the 1.5MB depicted in #2440. 

@nordicjm You're far more expert than me, can you tell me please if we miss something?

P.S. I found out how it manages to sign in overwrite only mode, it sets align to 1 [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/mcuboot.cmake)

nordicjm:
> Thanks @nordicjm for the review. Still new to the project, and I missed that closed PR. So, no swap modes due to the overhead it brings regarding flash size (I'll remove that assert line change). However, I am missing one point to fully understand: #2440 was closed, but if I want to sign an image with Imgtool using --overwrite-only for the Renesas RA family I still need the modifications to allow 128B as max-align (since it is used by Imgtool even in that case), am I correct or am I missing something?

With overwrite only, the max align is ignored since you have nothing extra to write, so you don't need any changes

nordicjm:
> I'm not understanding the reason for the closure of #2440 either. I believe swap is still usable. The write block size is 128 bytes, but the erase block or sector size is larger (a combination of 8-8KiB and 30-32KiB blocks on the RA6M3, other RA6 family have similar). If you treat it all as 32KiB sectors, only 6KiB is required for the trailer, rather than 1.5MiB. Treating it all as 8KiB sectors requires a 24KiB trailer. Not great, but not unusable.

The comment gives the figures it uses which was based on a random image size of 512KiB and assuming that the erase and write block sizes were the same (the user did not specify an erase size)

URNHere:
Ok thanks for all the clarifications.
Updated the sim.yaml to include some testing for the new aligments and the docs, in case you want to go ahead and give it a test.
I had to disable the oversized_secondary_slot because it was failing only on one device (stm32f4ExternalSpi).
@nordicjm  and @d3zd3z  if you consider it something useful to add we can go ahead, otherwise we can close the PR, up to you.

docs/imgtool.md
```diff
@@ -117,10 +117,10 @@ primary slot and adds a header and trailer that the bootloader is expecting:
       --public-key-format [hash|full]
                                       In what format to add the public key to the
                                       image manifest: full key or hash of the key.
-      --max-align [8|16|32]           Maximum flash alignment. Set if flash
+      --max-align [8|16|32|64|128]           Maximum flash alignment. Set if flash
```
d3zd3z: This messes up the alignment. Can you regenerate this with `--help` to get the exact output.

URNHere:
Should be signed-off now, sorry about that

d3zd3z (changes requested):
I think this is in reasonably good shape. I think, aside from this, we probably have an issue with the CI test for each PR getting too long, but that is something to address separately. There is a formatting issue in the docs, which should be fixable by pasting in the actual output of `--help` instead of  editing that output.

sim/Cargo.toml
d3zd3z: Give me a few minutes, I've made a PR that I believe fixes the issue for Stm32f4SpiFlash. I'll put a reference here, and you can try that on your machine to see if it addresses the issue.

d3zd3z:
I have confirmed that #2808 fixes the issue with the Stm32f4SpiFlash, and the combined result passes CI without needing to disable at specific part.

URNHere:
Great, I'll remove the suppression, thanks for your help!
I'll let the CI finish a full round of tests to assess if there is anything else to do before pushing another time.

d3zd3z:
We'll need to get 2808 merged before I'd expect CI to pass. You can also pull 2808 into your tree, rebase on top of that, and set the merge base of this PR to my branch, but it might be easier to just get the 2 reviews on 2808 and get it merged, and you can then just rebase on main.

URNHere:
Ok, I'll keep an eye out for #2808 being merged, I'll rebase and remove that suppression before pushing again.
Thanks for your help! Highly appreciated!

URNHere:
Rebased on top of main which now features the commits contained in #2808 and edited the second commit to remove the suppression tests/core.rs not in the modified files  list anymore.

d3zd3z (approved):
My concerns have been addressed. I'll let others chime in on the utility of large align and swap, but I don't see any more problems with the code.

URNHere:
Ok thanks @d3zd3z! Sorry for the late response.
To me it would be useful to use the swap modes even if it leads to some size overhead.
Also, for instance the RA6M3 has 1MB or 2MB of size, so we have plenty of space.
What do you guys think about it? @nordicjm @davidvincze 
@jeremydick

jeremydick:
Yes, I also believe the swap mode will be useful, even with the overhead

ryantra:
Here is some independent data in case it is useful since we have been running this on ra6e1- (128-byte code-flash write)

When applied these three hunks on top of v2.4.0 and running `swap-using-offset` at `BOOT_MAX_ALIGN=128`.

On whether swap is worth it at a 128-byte write size: the trailer cost is bounded by the erase sector, not the write unit. On our 8x32K-sector, 256K slots it comes to about 2.8 KB, leaving the application 196458 B of the 262144 B slot. Renesas ship swap at this alignment themselves in the FSP MCUboot port (app note R11AN0497).

- Sim: 25/25 at `max-align-128` for swap-scratch/move/offset, power-cut cases included. `oversized_secondary_slot` failed at align >= 64 until we backported the #2808 fix.
- Hardware: a correctly signed image that never confirms reverts at one reset, interrupted swaps resume, and real supply collapses swept across the validate+swap window all came back with the update installed. No bricks.

Happy to try other configurations on this part if that helps. One thing to flag: our copy of the assert also requires a power of two, which yours does not harmless, but not identical.

URNHere:
Thanks @ryantra for the testing and glad it works on your board too.
 I am happy there is interest in this feature.
I know that Renesas ships MCUBoot with FSP, but it wasn't the latest version last time I checked (it was 2.1.0 if I am not wrong).
The main goal of this PR is to be able to both have the latest version and be able to use it with Zephyr's sysbuild, without having to resort to Renesas's FSP.
@d3zd3z @davidvincze since there is some interest in this, is there any chance we can get a second review, to get a second approval, and if all is well merge it?
Thanks

tannewt:
I'm interested in this too!

URNHere:
Hi all,
Quick update:

- Renesas merged [#109125](https://github.com/zephyrproject-rtos/zephyr/pull/109125) in Zephyr, allowing to use mapped partitions, so CONFIG_BOOT_MAX_IMG_SECTORS_AUTO now works as expected. 

- Also, [#114429](https://github.com/zephyrproject-rtos/zephyr/pull/114429) in Zephyr got merged.

- The last piece to get it working out of the box is merging this PR together with #2801.

I gave it try this morning by applying the commit in #2801 on top of this branch and we don't need to change  configs manually anymore to make mcuboot with the new align and swap modes work on the Renesas RA family.
 @d3zd3z @nordicjm can you help us in some way to get it merged?

Issue #2798, opened 2026-07-21
Restructure fault injection handling code to facilitate reuse elsewhere


UEWBot:
Fault injection handling (FIH) is a useful subset of the mcuboot code, and mcuboot provides a good implementation of it. It would be useful to be able to use this code elsewhere without pulling in the whole of mcuboot. In particular, TF-M currently contains a copy of the FIH code, with all the corresponding issues of the two copies getting out of sync and bugfixes/enhancements that get applied to one not always making it into the other. It would also be nice to add FIH to crypto libraries, which would be easier if the mcuboot FIH code was easily separable. There may well be other projects that would benefit from a stand-alone FIH library.

The obvious way to do so is to split the FIH code into a separate repo, and to include it as a git submodule in mcuboot. mcuboot already includes several submodules, so this would be consistent with the current structure, but would enable other projects to pull just the FIH repo.

I propose a 3-step process:
1. Move the FIH files into their own directory. This would relocate three files (fault_inject_hardening.h, fault_injection_hardening.c, and fault_injection_hardening_delay_rng.h) and would change includes in other file to correspond. Note that the two rng implementations (fault_injection_hardening_delay_rng_mbedtls.c and fault_injection_hardening_delay_rng_psa.c) would not be moved.
2. Change how the FIH profile is specified. Today, fault_inject_hardening.h includes mcuboot_config.h to get at MCUBOOT_FIH_PROFILE_* which specifies which parts of the FIH code are enabled. If FIH is to be usable outside of mcuboot, this approach doesn't make sense. Options here are a separate config file (which would usually only contain a single define) or to specify the FIH config on the command line. It may make sense to alter the "default to FIH off" behaviour of FIH in favour of requiring that a level be specified so as to ensure that FIH isn't inadvertently disabled in any users by this change.
3. Move the new FIH directory to its own repo, and include it into mcuboot as a submodule.

PR #2799, opened 2026-07-23
Separate fih

UEWBot:
Move the fault injection hardening code into its own directory, so that it can be moved into a separate repo.

Issue #2798


nordicjm:
afaik TF-M wanted to do something similar @davidvincze might know about that

PR #2816, opened 2026-08-04
boot: bootutil: Remove excess flash_area_close function calls

### LINKED ISSUE #2815: Flash area incorrectly closed
https://github.com/mcu-tools/mcuboot/blob/7ad67106c3253d03ae4bd8ab48e6bdf7bc46f43e/boot/bootutil/src/swap_offset.c#L794

The function boot_read_image_size calls flash_area_close at the end, but does not have a corresponding open. It expects the area to already be open by a higher level caller. Closing it then causes issues for the following processing that also expects the area to be open. Likely not noticed because many implementations implement close as a nop.

nordicjm:
Removes calling this function where it should not be called, as MCUboot cleans up all open slots when it finishes execution

Fixes #2815

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1207,7 +1207,6 @@ boot_swap_image(struct boot_loader_state *state, struct boot_status *bs)
             }
         }
 #endif
-        flash_area_close(fap);
     }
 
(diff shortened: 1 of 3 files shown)


de-nordic (approved):

nordicjm:
@jwildp4 try this

PR #2818, opened 2026-08-04
boot: bootutil: bounds-check PSA ECDSA signature parse

d3zd3z:
`parse_signature_from_rfc5480_encoding()` walks the DER encoding of an ECDSA
signature using length bytes taken from the buffer itself, without knowing how
many bytes that buffer holds — `bootutil_ecdsa_verify()` discards its `slen`
argument outright. The `r` length at `sig[3]`, the `s` length derived from it,
and the `memcpy` source offsets computed from both are all used unchecked, so a
malformed signature causes reads past the end of the caller's signature buffer.

The size cannot be inferred from the encoding either, since the lengths that
drive the walk come from the buffer contents rather than from `slen`.

This passes the length down and validates the encoding against it before any of
the contents are used:

- reject a buffer too short for the smallest `SEQUENCE` of two `INTEGER`s;
- check the `SEQUENCE` and `INTEGER` tags the old code only assumed;
- require the outer length and each element length to fit in the bytes that
  remain.

Lengths are compared against the remaining count rather than added to an offset,
so none of the arithmetic can wrap. A rejected encoding returns
`PSA_ERROR_INVALID_SIGNATURE` and `psa_verify_hash()` is not called.

Deriving both element offsets from the start of the buffer also removes the
odd-length fixup the old code needed to keep its running pointer in step. That
fixup only landed on the correct `s` length byte when `r_len` was
`num_of_curve_bytes` or one more — the two encodings a conforming signer
produces. For any other `r_len` the old code read the `s` length from the wrong
offset.

Only the PSA backend is affected. The TinyCrypt and mbedTLS ECDSA paths use
`bootutil_decode_sig()`, which is already bounded by an end pointer, and are
untouched.

### Testing

Coverage of the changed function was confirmed by deliberately introducing a
syntax error in it: both PSA configurations fail to build, and `sig-ecdsa` (the
TinyCrypt path) builds fine as a negative control.

- `cargo test --features sig-ecdsa-psa` — 25/25
- `cargo test --features "mbedtls-v4 sig-p384"` — 25/25
- `cargo test --features sig-ecdsa` — 25/25

Note that the simulator only emits well-formed maximum-length DER, so the accept
path is exercised right at the new bounds, but the rejection branches are not
reached at runtime.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/ecdsa.h
+++ b/boot/bootutil/include/bootutil/crypto/ecdsa.h
@@ -343,48 +343,87 @@ static inline void get_public_key_from_rfc5280_encoding(uint8_t **p, size_t *siz
  * (r,s) of contiguous bytes
  *
  * \param[in]  sig               Pointer to a buffer containing the encoded signature
+ * \param[in]  sig_len           Size in bytes of the buffer pointed to by \p sig
  * \param[in] num_of_curve_bytes The required number of bytes for r and s
  * \param[out] r_s_pair          Buffer containing the (r,s) pair extracted. It's caller
  *                               responsibility to ensure the buffer is big enough to
  *                               hold the parsed (r,s) pair.
+ *
+ * \return PSA_SUCCESS on success, PSA_ERROR_INVALID_SIGNATURE if the encoding is
+ *         malformed or describes elements that do not fit in \p sig_len bytes.
  */
-static void parse_signature_from_rfc5480_encoding(const uint8_t *sig,
-                                                  size_t num_of_curve_bytes,
-                                                  uint8_t *r_s_pair)
+static int parse_signature_from_rfc5480_encoding(const uint8_t *sig,
+                                                 size_t sig_len,
+                                                 size_t num_of_curve_bytes,
+                                                 uint8_t *r_s_pair)
 {
-    const uint8_t *sig_ptr = NULL;
+    size_t seq_end, r_len, s_len, s_off;
+
+    /* The shortest encoding this parser can be handed is a SEQUENCE holding
+     * two one byte INTEGERs, i.e. 30 06 02 01 rr 02 01 ss
+     */
+    if (sig_len < 8) {
+        return (int)PSA_ERROR_INVALID_SIGNATURE;
+    }
+
+    /* sig[0] == 0x30, sig[1] == <length>, sig[2] == 0x02 */
+    if ((sig[0] != 0x30) || (sig[2] != 0x02)) {
+        return (int)PSA_ERROR_INVALID_SIGNATURE;
+    }
+
+    /* The contents of the SEQUENCE must be long enough to hold the two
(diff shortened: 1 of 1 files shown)


boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -343,48 +343,87 @@ static inline void get_public_key_from_rfc5280_encoding(uint8_t **p, size_t *siz
...
+    /* The contents of the SEQUENCE must be long enough to hold the two
+     * INTEGERs and must fit in the buffer. Only the ASN.1 short form length is
+     * handled, which is all the signatures of the supported curves need. Every
+     * length below is compared against the number of bytes that remain, rather
+     * than added to an offset, so that none of the arithmetic can wrap around.
+     */
+    if ((sig[1] < 6) || (sig[1] > sig_len - 2)) {
+        return (int)PSA_ERROR_INVALID_SIGNATURE;
+    }
+    seq_end = 2 + (size_t)sig[1];
```
de-nordic, line 383: Addition of 2 or why length stored at sig[1] has to be <= 253 is not explained either. The entire comment is describing nothing that happens below it.

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -343,48 +343,87 @@ static inline void get_public_key_from_rfc5280_encoding(uint8_t **p, size_t *siz
...
+
+    /* The contents of r start at sig[4], and must leave room for the two
+     * header bytes of s
+     */
+    r_len = sig[3];
```
de-nordic, line 388: OK, how is the sig[3] related to sig[4]. from the comment? sig[4] is not used till line 414, i do not count line 410, because there it is offset by r_len (sig[3]  and - num_of_curve_bytes.
The comment describes nothing that happens below it, at least till line 398? But there is s_off which is `s_off` used for offsetting source, and the comment sound more like we are constructing some buffer and need to have some extra 2 bytes for something?

boot/bootutil/include/bootutil/crypto/ecdsa.h
```diff
@@ -343,48 +343,87 @@ static inline void get_public_key_from_rfc5280_encoding(uint8_t **p, size_t *siz
...
+        return (int)PSA_ERROR_INVALID_SIGNATURE;
+    }
+
+    /* The INTEGER holding s follows the contents of r */
+    if (sig[4 + r_len] != 0x02) {
+        return (int)PSA_ERROR_INVALID_SIGNATURE;
+    }
+    s_len = sig[4 + r_len + 1];
+    s_off = 4 + r_len + 2;
+    if (s_len > seq_end - s_off) {
```
de-nordic, line 399: Here is the part where we start to play with `s`, and it is missing description what is happening here. Yeah, I have figured out, without a comment.
It is like we are missing a description of the input buffer and how certain fields relate to the stream sequence in it, but the comments, like in line 385 and 374, do not help; they look like randomly placed.

Issue #2810, opened 2026-07-31
boot_serial: direct-image serial-recovery upload to a secondary slot is broken under MCUBOOT_SWAP_USING_OFFSET


josesimoes:
### Context

We're adding MCUboot support to [.NET nanoFramework](https://github.com/nanoframework/nf-interpreter) as the
bootloader for our managed-code runtime (nanoCLR). Our target config:

- 2 images: image 0 = nanoCLR firmware, image 1 = a "deployment" area holding the managed
  application the CLR loads at runtime.
- `MCUBOOT_SWAP_USING_OFFSET` - primary slots in internal MCU flash, secondary slots on an
  external SPI NOR flash.
- `MCUBOOT_LOGICAL_SECTOR_SIZE=0x8000` (32 KB), since the external flash's physical erase
  granularity (4 KB) doesn't line up with the internal flash's much larger physical sectors.
- `MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD` - serial-recovery (SMP over USB CDC) uploads target a
  named image/slot directly.

While validating this, an image uploaded to a secondary slot via serial recovery was silently
never picked up by MCUboot on the next reset - `boot_swap_type_multi()` reported `Swap type:
none` and the secondary slot's trailer read `magic=unset`, as if nothing had ever been written
there, despite the upload completing successfully over SMP.

Root-causing this against `boot_serial.c` turned up **two independent bugs**, both inside
`bs_upload()`, and both needed to make this exact configuration work.

### Bug (a) - secondary-slot offset ignores `MCUBOOT_LOGICAL_SECTOR_SIZE`

Under `MCUBOOT_SWAP_USING_OFFSET`, the image in the secondary slot starts one sector in.
`bs_upload()` derives that offset from the first physical sector reported by
`flash_area_get_sectors()`:

```c
rc = flash_area_get_sectors(fap->fa_id, &num_sectors, &sector_data);
...
start_off = sector_data.fs_size;
```

This is exactly the offset `MCUBOOT_LOGICAL_SECTOR_SIZE` was introduced to change. Once logical
sectors are configured, the rest of the swap machinery no longer works in physical sectors -
`boot_img_sector_size()`/`boot_img_sector_off()` return the logical size unconditionally, and
`boot_header_scramble_off_sz()` (`bootutil_area.c`) uses `MCUBOOT_LOGICAL_SECTOR_SIZE` for this
[...]

PR #2821, opened 2026-08-05
boot_serial: detect the secondary slot by flash area id, not direct-image id

josesimoes:
Under MCUBOOT_SWAP_USING_OFFSET, bs_upload() has to know whether a direct image upload targets a secondary slot, because only then does the image start one sector into the slot.  It decided this from the id's parity:

    img_num > 0 && (img_num % BOOT_NUM_SLOTS) == 0

which holds only for the numbering used by the Zephyr port, where 0 and 1 are image 0's primary, 2 is image 0's secondary, 3 is image 1's primary, and so on.

That numbering is not part of the interface.  flash_area_id_from_direct_ image() is supplied by the port, so a port is free to order the ids differently -- and one that does gets a silently wrong offset for some slots while others happen to work, which is awkward to track down.

Use the information the port has already given us: flash_area_open() has resolved the id to a flash area, so compare that area against the port's own FLASH_AREA_IMAGE_SECONDARY() for each image.  This is correct for any numbering and needs no new port API.  BOOT_DIRECT_UPLOAD_SECONDARY_SLOT_ ID_REMAINDER has no remaining users and is removed.

No functional change for ports that follow the Zephyr numbering.

Addresses #2810

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -158,7 +158,6 @@ BOOT_LOG_MODULE_DECLARE(mcuboot);
 #endif
 
 #define SWAP_USING_OFFSET_SECTOR_UPDATE_BEGIN 1
-#define BOOT_DIRECT_UPLOAD_SECONDARY_SLOT_ID_REMAINDER 0
 
 static char in_buf[MCUBOOT_SERIAL_MAX_RECEIVE_SIZE + 1];
 #ifndef MCUBOOT_SERIAL_RAW_PROTOCOL
@@ -993,6 +992,7 @@ bs_upload(char *buf, int len)
 #if defined(MCUBOOT_SWAP_USING_OFFSET) && defined(MCUBOOT_SERIAL_DIRECT_IMAGE_UPLOAD)
         uint32_t num_sectors = SWAP_USING_OFFSET_SECTOR_UPDATE_BEGIN;
         struct flash_sector sector_data;
+        bool is_secondary = false;
 #endif
 
         curr_off = 0;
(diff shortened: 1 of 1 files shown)


nordicjm:
> which holds only for the numbering used by the Zephyr port, where 0 and 1 are image 0's primary, 2 is image 0's secondary, 3 is image 1's primary, and so on.

No it's part of the MCUmgr protocol...

josesimoes:
🤔 Allow me to push back on this... 
If this is part of MCUmgr protocol, why is it `flash_area_id_from_direct_image()` is declared as `extern` in boot\boot_serial\src\boot_serial_priv.h and the only implementation of it lives at boot\zephyr\flash_map_extended.c (under Zephyr)? 
I would say that the declaration itself suggests that provinding the image ID its a port hook. With Zephyr being the sole in-tree provider.
Which makes sense to me as a port genuinely could return different flash areas, different partition layouts, whatever its hardware needs.
The numbering scheme (which id means which image+slot) isn't something the hook is free to redefine. THat comes from `bs_slot_info()`.
The port's job is to translate that fixed id into the right flash area for its own layout, which is exactly what Zephyr's switch in flash_area_id_from_direct_image() does. 
Not doing so will cause an internal inconsistency between what the device advertises and what it does. Which is what our testing with the original code uncovered.

nordicjm:
Slot and image numbering is part of MCUboot and MCUmgr, that does not mean that _zephyr_ or another RTOS has to use the same mapping internally, e.g. you might have 7 partitions on a device and slot 0 and slot 1 are partitions IDs 5 and 6 internally in your RTOS, but to MCUboot and MCUmgr those are slot 0 and slot 1, primary and secondary slots of image 0

josesimoes:
Got it. Thank you for the clarification. I've fixed our port flash_area_id_from_direct_image() and it's working as expected. I will close this PR.

PR #2822, opened 2026-08-05
boot: treat a slot with no valid image as having no dependencies

### LINKED ISSUE #2812: boot_verify_dependencies() cancels every image's pending upgrade because one unrelated image's slot was never provisioned
### Context

We're adding MCUboot support to [.NET nanoFramework](https://www.nanoframework.net/) as the
bootloader for our managed-code runtime (nanoCLR), using a 2-image config (`MCUBOOT_IMAGE_NUMBER
= 2`, which auto-enables `MCUBOOT_DEPENDENCY_CHECK`). Image 0 is the nanoCLR firmware; image 1 is
a "deployment" area that isn't independently executable - it's data the running CLR reads at its
own initiative.

On a fresh device, image 1's slots are simply erased flash (nothing has ever been deployed
there). While validating an upgrade of image 0 - with a perfectly valid, correctly pending image
in image 0's secondary slot - the upgrade never happened. `boot_swap_type_multi()` correctly
computed `Swap type: test` for image 0 on the initial pass, but by the time the update-dispatch
step ran, it had been silently overridden back to `none`.

### Root cause

`boot_verify_dependencies()` runs once per boot whenever any image has an upgrade pending, and
walks **every** configured image's active slot (secondary if that image is upgrading, otherwise
its primary) looking for `IMAGE_TLV_DEPENDENCY` entries:

```c
IMAGES_ITER(BOOT_CURR_IMG(state)) {
    ...
    rc = boot_verify_slot_dependencies(state, slot);
    if (rc == 0) {
        BOOT_CURR_IMG(state)++;
    } else {
        for (int idx = 0; idx < BOOT_IMAGE_NUMBER; idx++) {
            BOOT_CURR_IMG(state) = idx;
            BOOT_SWAP_TYPE(state) = BOOT_SWAP_TYPE_NONE;   // <- every image, not just this one
        }
        break;
    }
}
```

Image 1 has no pending upgrade, so it's checked via its *primary* slot - which, on a fresh
device, is erased flash. `bootutil_tlv_iter_begin()` correctly rejects that header (a stray
`ih_protect_tlv_size` byte reads back nonzero, and the following read doesn't find either TLV
info magic at the computed offset), and that `rc != 0` propagates up through
`boot_verify_slot_dependencies()` as "dependencies not satisfied" - which isn't what actually
happened; there was no image there to have dependencies at all.

`boot_verify_dependencies()` can't tell the two cases apart, and cancels the pending upgrade of
**every** image, including image 0, which had nothing to do with image 1's empty slot.

### Evidence

```
[INF] boot_verify_slot_dependencies: image 0 slot 1 fa_id 2
[INF] boot_verify_slot_dependencies: image 0 slot 1 returning rc=0
[INF] boot_verify_slot_dependencies: image 1 slot 0 fa_id 3
[INF] boot_verify_slot_dependencies: tlv_iter_begin failed rc=-1
[INF] boot_verify_slot_dependencies: image 1 slot 0 returning rc=-1
[INF] boot_verify_dependencies: image 1 slot 0 rc=-1, forcing swap type NONE for all images
```

After the fix, image 1's empty slot correctly resolves as "nothing to satisfy", and image 0's
upgrade proceeds:

```
[INF] boot_verify_slot_dependencies: image 1 slot 0 fa_id 3
[INF] boot_verify_slot_dependencies: no valid image at slot 0, trivially satisfied
[INF] boot_verify_slot_dependencies: image 1 slot 0 returning rc=0
[INF] Image 0 upgrade secondary slot -> primary slot
```

### Proposed fix

A slot with no valid image has no TLV area to read a dependency from, so it can't have violated
one. Check the image magic before attempting to read the TLV area in
`boot_verify_slot_dependencies()`, and treat an absent image the same as an image with zero
`IMAGE_TLV_DEPENDENCY` entries - trivially satisfied:

```c
/* A slot with no valid image has no dependencies to satisfy or violate. */
if (boot_img_hdr(state, slot)->ih_magic != IMAGE_MAGIC) {
    rc = 0;
    goto done;
}
```

We have this fix on a branch against `upstream/main`, tested on hardware.

josesimoes:
boot_verify_dependencies() runs once per boot whenever any image has an upgrade pending, and walks every image's active slot -- the secondary if that image is upgrading, otherwise its primary -- looking for IMAGE_TLV_DEPENDENCY entries. If reading any single slot's dependencies fails, it cancels the pending upgrade for every image, not just the one whose slot failed.

On a device where a given image has never been provisioned, that image's primary slot is erased flash: the header fields are garbage, and bootutil_tlv_iter_begin() correctly rejects it (ih_protect_tlv_size reads back nonzero, the following read doesn't find IMAGE_TLV_PROT_INFO_MAGIC or IMAGE_TLV_INFO_MAGIC at the computed offset). That rc gets propagated up by boot_verify_slot_dependencies() as "dependencies not satisfied", which is not what happened -- there was no image to have dependencies at all -- but boot_verify_dependencies() cannot tell the two apart, and cancels the upgrade of every other, unrelated image as a result.

Concretely: image 0 (already correctly resolved to swap type TEST) never gets swapped in, because image 1's primary slot -- which was never written to, and has no pending upgrade of its own -- fails this check.

A slot with no valid image can't have violated a dependency, since it has no TLV area to read one from. Check the image magic before attempting to read the TLV area, and treat an absent image the same as an image with no IMAGE_TLV_DEPENDENCY entries at all: trivially satisfied.

Resolves #2812

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -309,6 +309,13 @@ boot_verify_slot_dependencies(struct boot_loader_state *state, uint32_t slot)
     assert(fap != NULL);
 
     BOOT_LOG_DBG("boot_verify_slot_dependencies");
+
+    /* A slot with no valid image has no dependencies to satisfy or violate. */
+    if (boot_img_hdr(state, slot)->ih_magic != IMAGE_MAGIC) {
+        rc = 0;
+        goto done;
+    }
+
 #if defined(MCUBOOT_SWAP_USING_OFFSET)
     it.start_off = boot_get_state_secondary_offset(state, fap);
 #endif


nordicjm (commented):
this does seem dangerous, it means you can bypass this check if you can interrupt NVM reading of an image that has dependencies. The system in MCUboot is not ideal, so if you have 2 images and they both depend upon the same version of each other, only 1 of the can have that version information and the other must rely upon that image's check, but with this you could time a fault injection to coincide with reading the header and get it to bypass this check

josesimoes:
That's a valid concern and I can understand how this would open the described security flaw.
I'll widraw this PR. Anyway, the condition that raised it from our testing has been already addressed with suggestions from related issues. We're now deploying a "blank image" to the deploy partition with the header + zero-length payload + TLV, which allows the verification to pass.

PR #2824, opened 2026-08-05
boot: bootutil: feed watchdog in bootutil_img_hash

tpambor:
If the image is large and/or stored on slow memory, e.g. external flash, reading the image to calculate its hash can take a significant amount of time. Feed the watchdog while hashing is in progress to prevent a reset from happening under these conditions.

### DIFF
--- a/boot/bootutil/src/bootutil_img_hash.c
+++ b/boot/bootutil/src/bootutil_img_hash.c
@@ -187,6 +187,8 @@ bootutil_img_hash(struct boot_loader_state *state,
         }
 #endif
         bootutil_sha_update(&sha_ctx, tmp_buf, blk_sz);
+
+        MCUBOOT_WATCHDOG_FEED();
     }
 #endif /* MCUBOOT_RAM_LOAD */
 #endif /* MCUBOOT_HASH_STORAGE_DIRECTLY */


boot/bootutil/src/bootutil_img_hash.c
```diff
@@ -187,6 +187,8 @@ bootutil_img_hash(struct boot_loader_state *state,
         }
 #endif
         bootutil_sha_update(&sha_ctx, tmp_buf, blk_sz);
+
+        MCUBOOT_WATCHDOG_FEED();
```
de-nordic, line 191: Shouldn't the feed happen on top of for, in front of `flash_read` then between read and `bootutil_sha_update` and then after the loop?

tpambor: I aligned this with how it's done for `boot_copy_region`, where the feed is placed at the end of the loop, after `flash_area_read` and `flash_area_write`. https://github.com/mcu-tools/mcuboot/blob/5098d8eef1139abd3d821d8138a6321d656dcd3e/boot/bootutil/src/loader.c#L906

I think one feed per iteration is enough, but if it is preferred I can change it.

nordicjm (approved):

boot/bootutil/src/bootutil_img_hash.c
tpambor: @de-nordic Could you take another look? Can it stay as-is?

de-nordic: > I aligned this with how it's done for `boot_copy_region`, where the feed is placed at the end of the loop, after `flash_area_read` and `flash_area_write`.
> 
> https://github.com/mcu-tools/mcuboot/blob/5098d8eef1139abd3d821d8138a6321d656dcd3e/boot/bootutil/src/loader.c#L906
> 
> I think one feed per iteration is enough, but if it is preferred I can change it.

The iteration does not matter, The operation does. The feed is time based, so my question was whether the feed should not be updated prior and after operations that may not stuck but take longer time.

de-nordic (approved):
Seems ok.

PR #2829, opened 2026-08-11
boot: zephyr: board: Add M5stack Unit C6L overlay

marekmatej:
The PI4IOE5V6408 expander driver selects I2C, overriding MCUboot CONFIG_I2C=n, and its polling work queue fails to link without multithreading. Disable the expander node, and the unused I2C buses with it. Related board PR https://github.com/zephyrproject-rtos/zephyr/pull/113221

### DIFF
--- a/boot/zephyr/boards/m5stack_unitc6l_esp32c6_hpcore.overlay
+++ b/boot/zephyr/boards/m5stack_unitc6l_esp32c6_hpcore.overlay
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2026 Espressif Systems (Shanghai) Co., Ltd.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* MCUboot builds without multithreading, so the I2C buses and the I2C GPIO
+ * expander (none of which the bootloader needs) must be kept out of the image.
+ */
+
+&i2c0 {
+	status = "disabled";
+};
+
+&gpio_exp {
+	status = "disabled";
+};
+
+&grove_i2c {
+	status = "disabled";
+};


nordicjm (commented):
whilst the changes are OK, bit puzzled why they are needed since I2C should not be enabled, so why would any of the others?

marekmatej:
> whilst the changes are OK, bit puzzled why they are needed since I2C should not be enabled, so why would any of the others?

The GPIO expander, which is enabled by default, is pulling the I2C. But I agree with you, but I didn't find a better solution to the linker error (if sysbuild is used) than doing what the other M5stack board `m5stack_cores3` does.
Anyway, I added a link to the board PR in question to the description.

nordicjm:
> > whilst the changes are OK, bit puzzled why they are needed since I2C should not be enabled, so why would any of the others?
> 
> The GPIO expander, which is enabled by default, is pulling the I2C. But I agree with you, but I didn't find a better solution to the linker error (if sysbuild is used) than doing what the other M5stack board `m5stack_cores3` does. Anyway, I added a link to the board PR in question to the description.

why is the GPIO port expander enabled by default?

marekmatej:
> why is the GPIO port expander enabled by default?

It is used to connect the radio.

nordicjm:
> > why is the GPIO port expander enabled by default?
> 
> It is used to connect the radio.

right but MCUboot does not need a radio :) is it just because `CONFIG_GPIO` is on which enables it and the other thing selects `CONFIG_I2C`?

marekmatej:
> > > why is the GPIO port expander enabled by default?
> > 
> > It is used to connect the radio.
> 
> right but MCUboot does not need a radio :) is it just because `CONFIG_GPIO` is on which enables it and the other thing selects `CONFIG_I2C`?

You're right, this doesn't belong to MCUboot. Will fix dependencies and remove this PR.

marekmatej:
Closing as solved on a board level.

PR #2840, opened 2026-08-18
boot: zephyr: Allow a custom do_boot implementation

VynDragon:
Let a platform define do_boot for itself.

### Suggested implementation in zephyr, in modules/Kconfig.mcuboot:

```
# Toggle BOOT_CUSTOM_BOOT to indicate Zephyr platform code provides do_boot
config MCUBOOT_CUSTOM_BOOT
	bool
	select BOOT_CUSTOM_BOOT
```

### Example implementation for a soc:

soc/thesoc/CMakeLists.txt:
```
if (CONFIG_MCUBOOT)
  zephyr_include_directories(${ZEPHYR_MCUBOOT_MODULE_DIR}/boot/bootutil/include)
  zephyr_sources(do_boot.c)
endif()
```
soc/thesoc/Kconfig.defconfig:
```
if MCUBOOT

configdefault MCUBOOT_CUSTOM_BOOT
	default y

endif # MCUBOOT
```

ESP32 likely should use that method if this is merged.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -73,7 +73,9 @@ target_sources(app PRIVATE
   watchdog.c
 )
 
-if(CONFIG_ARM)
+if(CONFIG_BOOT_CUSTOM_BOOT)
+  # No implementation of do_boot
+elseif(CONFIG_ARM)
   target_sources(app PRIVATE arch/arm.c)
 elseif(CONFIG_SOC_FAMILY_ESPRESSIF_ESP32)
   target_sources(app PRIVATE arch/esp32.c)
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -137,6 +137,12 @@ endif # BOOT_USE_PSA_CRYPTO
 
 menu "MCUBoot settings"
 
+config BOOT_CUSTOM_BOOT
+	bool
+	help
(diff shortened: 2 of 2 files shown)


nordicjm (changes requested):
Nothing needed in tree for this, add to your module:
```
set_source_files_properties(
  ${ZEPHYR_MCUBOOT_MODULE_DIR}/boot/zephyr/arch/arm.c
  DIRECTORY ${ZEPHYR_MCUBOOT_MODULE_DIR}/boot/zephyr
  PROPERTIES HEADER_FILE_ONLY ON
)
```
and replace `arm.c` with the file you want to exclude

VynDragon:
Would you allow that in upstream zephyr in the future?

nordicjm:
> Would you allow that in upstream zephyr in the future?

zephyr will not allow that in zephyr, direct MCUboot code should not live in zephyr either. If you're adding a new architecture then it can be added directly to MCUboot

VynDragon:
The issue is that a lot of SoCs are not as homogeneous as ARM, this is basically a esp32.c (or more) for each soc family , is that ok?

nordicjm:
> The issue is that a lot of SoCs are not as homogeneous as ARM, this is basically a esp32.c (or more) for each soc family , is that ok?

I don't see why not, perhaps @d3zd3z can comment

VynDragon:
Was just to confirm it was fine, thanks.

PR #2842, opened 2026-08-20
boot: serial: charge the whole read loop against the timeout

### LINKED ISSUE #2841: boot_serial: serial recovery wait lasts several times the configured timeout
### Summary

Under `CONFIG_BOOT_SERIAL_WAIT_FOR_DFU`, the wait lasts several times longer
than the configured timeout. Measured 5.5x on a Cortex-M0+ and 3.2x on a
Cortex-M23. The countdown is proportional, so it is not a hang, but a
`BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT` of 5000 is a 22 s window in practice.

### Cause

`boot_serial_read_console()` in `boot/boot_serial/src/boot_serial.c`:

```c
MCUBOOT_WATCHDOG_FEED();
uint32_t start = k_uptime_get_32();
rc = f->read(in_buf + off, sizeof(in_buf) - off, &full_line);
...
check_timeout:
elapsed_in_ms = (k_uptime_get_32() - start);
timeout_in_ms -= elapsed_in_ms;
```

`start` is sampled after the watchdog feed and after the previous iteration's
own measurement, so only `f->read()` is ever charged to the timeout. Everything
else in the iteration is free, including the two `k_uptime_get_32()` calls that
do the measuring.

On a core without hardware divide those calls are the bulk of the loop. Zephyr
converts ticks to milliseconds with a 64-bit division, so the sampling itself
costs more than the work being sampled. Sampling the PC over SWD through one
whole recovery window on an STM32C071 (Cortex-M0+), 166 samples inside the
bootloader:

| share | symbol |
|---|---|
| 47.0% | `__udivmoddi4` |
| 10.8% | `__udivsi3` |
| 8.4% | `sys_clock_tick_get` |
| 4.8% | `console_read` |
| 4.8% | `__clzdi2` |
| 3.6% | `__aeabi_uldivmod` |

Over 72% in the 64-bit division helpers and the tick read, against 4.8% in the
console read that is the only part the timeout is charged for.

### Measurements

STM32C071 (Cortex-M0+), MCUmgr over a 9600 baud serial line, timeout 30000 ms:
recovery ended 165 s after the last byte, with the bus fully silent and the
exit detected over SWD rather than on the wire. A 5000 ms
`WAIT_FOR_DFU_TIMEOUT` window on the same board measured 22.6 s.

PIC32CM5112GC00100 (Cortex-M23), same configuration: 3.2x.

The factor is stable enough to predict a run: a 30000 ms timeout followed by a
second boot paying a 5000 ms window again predicts 165 + 20 = 185 s, measured
184.8 s.

### Suggested fix

Charge the whole iteration by carrying one timestamp across iterations instead
of re-sampling inside it:

```c
uint32_t last = k_uptime_get_32();
while (...) {
    ...
    uint32_t now = k_uptime_get_32();
    timeout_in_ms -= (now - last);
    last = now;
}
```

That leaves one `k_uptime_get_32()` per iteration and accounts for the feed, the
loop overhead and the measurement itself. Happy to send a patch if the shape
looks right.

Found while measuring #2830, which is what made the figure worth trusting in the
first place. It is not caused by that PR and reproduces without it.

TheArkadiuszGrzelka:
Fixes #2841.

`boot_serial_read_console()` sampled a timestamp just before `f->read()` and took a second one right after it, so the only thing subtracted from `timeout_in_ms` was the time spent inside the read. Everything else in the iteration went uncharged: the watchdog feed, the decode and dispatch of a received line, the loop overhead, and the two `k_uptime_get_32()` calls themselves. On a core without a hardware divider the uptime read is a 64-bit software division, and that pair of calls is what dominates a loop that mostly finds no input.

The result is a window that runs several times longer than the configured timeout, and the overshoot grows the slower the core is.

This patch samples the timestamp once before the loop and rolls it forward at the bottom of each iteration. The iteration is charged in full, including the timekeeping, and one uptime read per iteration replaces two. `timeout_in_ms -= elapsed_in_ms;` is left in place, so this does not conflict with #2830.

### Measured

STM32C071 (Cortex-M0+, 48 MHz), same image before and after, boot time about 1 s. Elapsed is measured from the host as the time until the application answers on its own bus.

| scenario | before | after | expected |
| --- | --- | --- | --- |
| `CONFIG_BOOT_SERIAL_WAIT_FOR_DFU_TIMEOUT=5000`, no serial activity at all | 22.6 s | 6.0 s | ~6 s |

The other two cases need `CONFIG_BOOT_SERIAL_INACTIVITY_TIMEOUT` from #2830, which is not merged, so they are listed separately rather than as evidence for this patch on its own. Both used a 30000 ms inactivity timeout:

| scenario | before | after | expected |
| --- | --- | --- | --- |
| boot mode entered, session abandoned | 165.2 s | 34.6 s | ~33 s |
| session started in the every-boot window, then silence | 184.8 s | 36.1 s | ~38 s |

Both configurations above were also built and run on hardware with this patch applied. A build with `MCUBOOT_SERIAL_WAIT_FOR_DFU` undefined was not exercised, but that path is unchanged: `start` lives in the same `#ifdef` as its uses, and `elapsed_in_ms` stays 0 exactly as before.

### DIFF
--- a/boot/boot_serial/src/boot_serial.c
+++ b/boot/boot_serial/src/boot_serial.c
@@ -1620,6 +1620,13 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
     int full_line;
     int max_input;
     int elapsed_in_ms = 0;
+#ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
+    /* Start of the iteration charged against the timeout; rolled forward at
+     * the loop's end so every ms in the loop counts, not just f->read().
+     */
+    uint32_t start = k_uptime_get_32();
+    uint32_t now;
+#endif
 #ifdef MCUBOOT_SERIAL_RAW_PROTOCOL_INPUT_TIMEOUT
     uint32_t raw_input_start = 0;
 #endif
@@ -1653,9 +1660,6 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
         }
 #endif
         MCUBOOT_WATCHDOG_FEED();
-#ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
-        uint32_t start = k_uptime_get_32();
-#endif
         rc = f->read(in_buf + off, sizeof(in_buf) - off, &full_line);
         if (rc <= 0 && !full_line) {
 #ifndef MCUBOOT_SERIAL_WAIT_FOR_DFU
@@ -1714,7 +1718,9 @@ check_timeout:
 #endif
         /* Subtract elapsed time */
 #ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
-        elapsed_in_ms = (k_uptime_get_32() - start);
+        now = k_uptime_get_32();
+        elapsed_in_ms = (int)(now - start);
+        start = now;
 #endif
         timeout_in_ms -= elapsed_in_ms;
     }


nordicjm (approved):

boot/boot_serial/src/boot_serial.c
```diff
@@ -1714,7 +1720,12 @@ boot_serial_read_console(const struct boot_uart_funcs *f,int timeout_in_ms)
 #endif
         /* Subtract elapsed time */
 #ifdef MCUBOOT_SERIAL_WAIT_FOR_DFU
-        elapsed_in_ms = (k_uptime_get_32() - start);
+        {
```
nordicjm: whilst the fix is OK this is not in the coding style of MCUboot, put the variable above (top of scope, where other variables are declared, but only declared inside of `#ifdef`) and do not have the `{ }` here

nordicjm (changes requested):

boot/boot_serial/src/boot_serial.c
TheArkadiuszGrzelka: Done. `now` is declared at the top of the function inside the existing `MCUBOOT_SERIAL_WAIT_FOR_DFU` block next to `start`, and the braces at the use site are gone. Repushed on current main.

nordicjm (approved):

PR #2844, opened 2026-08-24
boot: bootutil: factor the single-slot loader into bootutil

### LINKED ISSUE #2826: Single-slot boot path is duplicated in ports; consider factoring it into `boot/bootutil`
The single-slot (`MCUBOOT_SINGLE_APPLICATION_SLOT`) boot path exists twice, in two
ports, and not at all in the shared library:

| File | Lines |
| --- | --- |
| `boot/zephyr/single_loader.c` | 255 |
| `boot/mynewt/src/single_loader.c` | 136 |
| `boot/bootutil/src/` | — none |

Both provide `boot_go()` plus a local `boot_image_validate()`; Zephyr's
additionally provides `boot_get_loader_state()`, `boot_get_image_max_sizes()`,
`boot_open_all_flash_areas()`, `boot_close_all_flash_areas()` and
`app_max_size()`, of which the first four are only referenced by
`boot_fetch_slot_state_sizes()` in `bootutil_misc.c`, inside
`#if defined(MCUBOOT_SERIAL_IMG_GRP_SLOT_INFO)`.

Neither implementation appears to depend on its host OS. In
`boot/zephyr/single_loader.c` I could find no Zephyr API call, no `CONFIG_*`
reference and no devicetree use; its includes are `assert.h`, `bootutil/*`,
`bootutil_priv.h` and `mcuboot_config/mcuboot_config.h`, and the only
non-bootutil symbols it references are the `BOOT_LOG_*` macros, which are
already part of the port configuration contract. The same appears to hold for
the Mynewt version.

### Why it matters

`boot_go()` has two implementations — the multi-slot state machine in
`boot/bootutil/src/loader.c`, and the single-slot path — and the build system
selects between them (`boot/zephyr/CMakeLists.txt`, the
`CONFIG_SINGLE_APPLICATION_SLOT` branch). Because only the multi-slot
implementation lives in the shared library, an out-of-tree port that wants
single-slot support has to write the second one itself, even though it is
almost entirely calls into `bootutil`.

That also means the two in-tree copies can drift from each other and from
`bootutil`, and each has to be updated separately when the `bootutil` interface
changes.

### Proposal

Move the OS-independent part of the single-slot boot path into
`boot/bootutil/src/single_loader.c` and have the Zephyr and Mynewt ports use it,
leaving anything genuinely port-specific behind. This would be a
behaviour-preserving refactor rather than a feature change, and would give
out-of-tree ports the same choice of `boot_go()` implementation that the in-tree
ports have.

### Questions before attempting a PR

1. Is consolidating these two into `bootutil` a direction you would accept in
   principle, or is the duplication deliberate?
2. `sim/Cargo.toml` has features for `overwrite-only`, `swap-move`,
   `swap-offset` and so on, but none for single-slot, so as far as I can tell the
   single-slot path currently has no simulator coverage. For a
   behaviour-preserving move, would you want simulator coverage for single-slot
   added first (or as part of the same PR), or is the absence of coverage
   acceptable for a relocation?
3. If you would take it, do you have a preference on scope — move Zephyr's
   version and have Mynewt adopt it, or reduce to the smaller common subset and
   let each port keep what only it needs?

Happy to do the work if the direction is welcome; asking first per
`docs/contributing.md`.

danielriege:
Factors the OS-independent single-slot boot path into `boot/bootutil` and gives it simulator coverage. Closes #2826 — direction confirmed there: common code, Zephyr's version as the base, tests in the same PR.

### What moves

`MCUBOOT_SINGLE_APPLICATION_SLOT` lived twice — `boot/zephyr/single_loader.c` (255 lines) and `boot/mynewt/src/single_loader.c` (136) — and not in the shared library, so out-of-tree ports had to write their own `boot_go()` for a path that is almost all bootutil calls. Zephyr's copy, the superset, moves to `boot/bootutil/src/single_loader.c`; both ports build it. Like `swap_move.c` it guards its own body, so it compiles to nothing when unwanted and can sit in a flat source list.

### Bugs found on the way

All in configurations no CI job builds, all reproducible:

- `boot_get_max_app_size()` was defined here and in `bootutil_misc.c` — duplicate symbol with `SERIAL_IMG_GRP_SLOT_INFO` (fccd8905 moved it, leaving the port copy).
- `boot_get_image_max_sizes()` was guarded on `SERIAL_IMG_GRP_SLOT_INFO`, but its declaration and callers use `... || DATA_SHARING` — undefined reference with data sharing.
- The measured-boot and data-sharing error paths returned past `boot_go()`'s cleanup label, leaking the primary flash area and returning a raw `int` as a `fih_ret`.
- `bootutil_get_caps()` claimed `SWAP_USING_SCRATCH` for a one-slot build — a secondary slot and scratch area that do not exist.
- newt reads `pkg.ign_files` as unanchored regexes, so `"loader.c"` also matches `bootutil_loader.c` (silently dropped from single-slot Mynewt builds today) and would have eaten the relocated file.

### Testing

`single_loader.c` gains `context_boot_go()`, mirroring `loader.c`. That unblocked the simulator, which drives the bootloader through `context_boot_go()` with a per-invocation state because it runs tests as parallel threads; the single-slot loader had only `boot_go()` over a file-static state. Its scratch buffer becomes `TARGET_STATIC` for the same reason. `run.c` itself needed no changes.

New `single-slot` feature and two tests: a valid primary slot boots from the primary area leaving flash byte-identical; one failing validation is not booted. Adds a `single_slot` Mynewt CI target, since none existed — Zephyr already covers this via `sample.bootloader.mcuboot.single_slot`.

| verified locally | result |
| --- | --- |
| simulator, 12 feature configs | 27–29 tests each, pass |
| Mynewt, 10 existing targets + the new one | build |
| Zephyr twister, `nrf52840dk/nrf52840`, 14 scenarios | build |

### Behaviour changes

- Mynewt reports a failed `flash_area_open()` instead of asserting.
- Mynewt single-slot holds a `boot_loader_state` rather than one flash-area pointer (small RAM increase), and now compiles `bootutil_loader.c`.
- `bootutil_get_caps()` returns the new bit instead of `SWAP_USING_SCRATCH`; only the simulator reads it.


boot/bootutil/src/single_loader.c
```diff
@@ -222,21 +222,26 @@ boot_go(struct boot_rsp *rsp)
...
+            BOOT_LOG_ERR("Failed to add data to shared memory area.");
+            fih_rc = FIH_FAILURE;
+        }
     }
 #endif /* MCUBOOT_DATA_SHARING */
 
-    rsp->br_flash_dev_id = flash_area_get_device_id(BOOT_IMG_AREA(&state, BOOT_SLOT_PRIMARY));
-    rsp->br_image_off = flash_area_get_off(BOOT_IMG_AREA(&state, BOOT_SLOT_PRIMARY));
-    rsp->br_hdr = &_hdr;
+    if (rc == 0) {
```
nordicjm, line 242: line 239 and 244 are superfluous

boot/bootutil/src/single_loader.c
```diff
@@ -222,21 +222,26 @@ boot_go(struct boot_rsp *rsp)
...
+        fih_rc = FIH_FAILURE;
     }
 #endif /* MCUBOOT_MEASURED_BOOT */
 
 #ifdef MCUBOOT_DATA_SHARING
-    rc = boot_save_shared_data(&_hdr, BOOT_IMG_AREA(&state, BOOT_SLOT_PRIMARY), 0, NULL);
-    if (rc != 0) {
-        BOOT_LOG_ERR("Failed to add data to shared memory area.");
-        return rc;
+    if (rc == 0) {
```
nordicjm, line 233: line 230 also

boot/bootutil/src/single_loader.c
```diff
@@ -167,50 +167,53 @@ int app_max_size(struct boot_loader_state *state)
...
     int rc = -1;
     FIH_DECLARE(fih_rc, FIH_FAILURE);
 
-    BOOT_LOG_DBG("boot_go: Single loader");
+    BOOT_LOG_DBG("context_boot_go: Single loader");
 
-    rc = boot_open_all_flash_areas(&state);
+    hdr = boot_img_hdr(state, BOOT_SLOT_PRIMARY);
+
+    rc = boot_open_all_flash_areas(state);
```
nordicjm, line 189: nit. hdr should be set after opening flash areas successfully?

boot/bootutil/src/caps.c
```diff
@@ -41,7 +41,10 @@ uint32_t bootutil_get_caps(void)
 #if defined(MCUBOOT_SIGN_ED25519)
     res |= BOOTUTIL_CAP_ED25519;
 #endif
-#if defined(MCUBOOT_OVERWRITE_ONLY)
+#if defined(MCUBOOT_SINGLE_APPLICATION_SLOT) || \
+    defined(MCUBOOT_SINGLE_APPLICATION_SLOT_RAM_LOAD)
+    res |= BOOTUTIL_CAP_SINGLE_APPLICATION_SLOT;
+#elif defined(MCUBOOT_OVERWRITE_ONLY)
```
nordicjm, line 47: don't think the same capability should be used for both, they are different capabilities

PR #2853, opened 2026-09-05
boot: zephyr: use the generic mcuboot defaults on espressif socs

sylvioalves:
Drop the Espressif-only overrides that forced overwrite-only, no signature check and no primary slot validation, so these SoCs follow the same defaults as every other target: swap using offset, RSA signatures and slot 0 validation.


sylvioalves:
@nordicjm any ideas why CI is failing?

nordicjm:
Seems a legacy behaviour was removed https://github.com/mcu-tools/mcuboot/pull/2851

nordicjm (approved):

nordicjm:
Rebase will fix

PR #2855, opened 2026-09-07
boot: zephyr: add new boot_image_jump hook

PetrBuchtaNXP:
This new hook is called after a boot image has been selected and right before jumping to the image:

**boot_go() -> ... -> boot_image_jump_hook() -> do_boot()**

The hook receives a pointer to the boot_rsp structure so it can inspect or modify the boot response. This hook was introduced to be able to setup things like flash protection when DIRECT-XIP is used where the settings may vary based on slot number. It also allows to setup flash memory remapping for targets that support this feature.

There are targets that require setup of flash protection based on execution range. This is no problem when always running from slot0 but in DIRECT-XIP mode I need the boot_rsp structure to determine what slot will be used. This hook also attempts to make flash remapping possible by allowing to modify the boot response which in case of flash remapping is crucial.

Without this hook I have to use a combination of boot_go_hook and status_change_hook(image_found) but that doesn't give me information about selected image to be booted. I could hack around this by catching the rsp pointer from boot_go_hook handler and use it in the status_change_hook handler but that's far from "good enough".


boot/zephyr/main.c
```diff
@@ -422,6 +422,13 @@ int main(void)
     mcuboot_status_change(MCUBOOT_STATUS_BOOTABLE_IMAGE_FOUND);
 
     ZEPHYR_BOOT_LOG_STOP();
+
+    BOOT_HOOK_IMAGE_JUMP_CALL_FIH(boot_image_jump_hook, fih_rc, &rsp);
+    if (FIH_NOT_EQ(fih_rc, FIH_SUCCESS)) {
+        BOOT_LOG_ERR("boot_image_jump_hook failed");
```
nordicjm, line 426: how can you log after the logger is stopped?

PetrBuchtaNXP: That is a good question! I missed that. Not sure how I was getting all the logs from the hook then.

nordicjm:
Rebase will fix CI (be sure to fix log issue also)

nordicjm (approved):

de-nordic (approved):

PR #2856, opened 2026-09-08
boot: zephyr: fix image selection in flash write block size checks

WangHunZi:
Firstly, this pr don't matter the control flow, since the `swap_write_block_size_check` always return true.

Using BOOT_IMG_AREA ensures the flash_get_write_block_size calls follow the current image index in state->curr_img_idx.

```c
#if (BOOT_IMAGE_NUMBER > 1)
#define BOOT_CURR_IMG(state) ((state)->curr_img_idx)
#else
#define BOOT_CURR_IMG(state) 0
#endif

#define BOOT_IMG(state, slot) ((state)->imgs[BOOT_CURR_IMG(state)][(slot)])
#define BOOT_IMG_AREA(state, slot) (BOOT_IMG(state, slot).area)
```

### DIFF
--- a/boot/zephyr/flash_check.c
+++ b/boot/zephyr/flash_check.c
@@ -26,8 +26,7 @@ bool swap_write_block_size_check(struct boot_loader_state *state)
 #endif
 
 #ifdef MCUBOOT_SLOT0_EXPECTED_WRITE_SIZE
-    flash_write_block_size_pri = flash_get_write_block_size(
-                                              state->imgs[0][BOOT_SLOT_PRIMARY].area->fa_dev);
+    flash_write_block_size_pri = flash_get_write_block_size(BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY)->fa_dev);
 
     if (flash_write_block_size_pri != MCUBOOT_SLOT0_EXPECTED_WRITE_SIZE) {
         BOOT_LOG_DBG("Discrepancy, slot0 expected write block size: %d, actual: %d",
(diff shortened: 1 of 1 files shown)


boot/zephyr/flash_check.c
```diff
@@ -26,8 +26,7 @@ bool swap_write_block_size_check(struct boot_loader_state *state)
 #endif
 
 #ifdef MCUBOOT_SLOT0_EXPECTED_WRITE_SIZE
-    flash_write_block_size_pri = flash_get_write_block_size(
-                                              state->imgs[0][BOOT_SLOT_PRIMARY].area->fa_dev);
+    flash_write_block_size_pri = flash_get_write_block_size(BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY)->fa_dev);
```
nordicjm, line 29: this isn't valid, as the output below shows this is for slot0 specifically which would be image0. Though the check _could_ be expanded for all images

nordicjm (changes requested):
not valid

boot/zephyr/flash_check.c
WangHunZi: Yes, you're right. I hadn't noticed that the macro refers to slot 0.

Should I update these flash checks to cover all configured images, rather than a specific image index such as 0 or 1? My setup uses multiple images, and that's where I encountered this issue

WangHunZi:
> not valid

Should I make the flash_check to compatiable to multiply image?

> not valid

Would you like me to extend the write-block-size checks to cover all configured images?

boot/zephyr/flash_check.c
nordicjm: yep running on all images makes sense

PR #2858, opened 2026-09-08
boot: bootutil: add decentralized key revocation from primary slot

### LINKED ISSUE #2857: RFC: Decentralized key revocation for MCUboot (`MCUBOOT_KEY_REVOCATION_FROM_PRIMARY`)
**Status**: proposal, with a reference implementation ready for review (see linked PR).

**Related**: 
- issue #221 ("Support for revoking/invalidating keys," open since 2018)
- PR #2305 (closed, multi-signature/key-ID approach)
- PR #2701 (merged, second embedded signing key for Zephyr)

## 1. Summary

This is an optional, opt-in feature for MCUboot: a device can reject an image signed with an older key, once a newer key has already produced a confirmed image in the primary slot. The device needs no server, no revocation list, and no new persistent state. The feature is off by default; a build that does not enable it behaves exactly as it does today.

We've written a reference implementation of this design (linked PR) and are opening this issue first to get feedback on the concept and design before asking for a full review of the code.

## 2. Problem statement

A signing key can leak. Today, MCUboot has no supported way to react to this on a device already in the field, other than shipping a whole new bootloader with a different key set. Two related, long-standing gaps back this up:

- Issue #221, "Support for revoking/invalidating keys," has been open since 2018. It asks for exactly this capability. It carries the label "someday." No implementation has landed.
- PR #2701 recently added support for a second embedded signing key to the Zephyr port, to separate production and development signing custody. During review, maintainer @de-nordic wrote: "There is no mechanism to disable any of the keys, so the chances of losing any of these basically doubles without means to mitigate the losses." This is a direct, current statement that the gap is real.

We also looked at PR #2305, which attempted a related but different feature (multi-signature verification with a key-ID TLV, aimed at PSA use cases). It was closed. We read the review discussion carefully and tried not to repeat the same mistakes; section 8 lists what we took from that review.

Key rotation without a bootloader update is not supported today either. We would be equally satisfied with a supported key-rotation mechanism, but as far as we can tell none exists in MCUboot today. This proposal is the closest practical alternative we found: revocation, not rotation, built on infrastructure MCUboot already has.

## 3. Goals

- Let a device reject images signed with a key older than the key that produced the currently active, confirmed image.
- Do this without a server, without a network connection, and without a revocation list stored anywhere.
- Do this without new persistent (NV) state and without a new image field.
- Make the feature strictly opt-in. A build that does not enable it must be bit-for-bit unaffected.
- Make the feature strictly narrowing. It must never cause MCUboot to accept an image it would otherwise reject today. It may only cause MCUboot to reject an image it would otherwise have accepted.
- Reuse existing MCUboot infrastructure (the embedded key array, the existing key-matching code, the existing "reject as bad image" path) instead of adding new mechanisms.

## 4. Non-goals

- This is not a general key-rotation mechanism. The active key cannot be "rolled forward" without also producing a new, confirmed image.
- This is not a revocation-list mechanism. There is no list of revoked keys anywhere, on the device or on a server.
- This is not a replacement for version-based or security-counter-based downgrade prevention. Both stay active, unchanged, at all times.
- This does not change the on-disk image format. No new TLV is needed.
- This does not target `MCUBOOT_BUILTIN_KEY` or `MCUBOOT_HW_KEY` builds (see section 7.1); both are rejected at compile time.

## 5. Background: how MCUboot handles multiple keys today

MCUboot already supports embedding more than one verification key in the bootloader, through the `bootutil_keys[]` array and `bootutil_key_cnt`. `bootutil_find_key()` already returns the array index of the key that matches an image's key TLV. This exists to let a bootloader accept images from more than one signer (for example: production and development, as landed in PR #2701).

Today, all embedded keys are permanently and equally valid. There is no concept of key age, key generation, or key priority. Any embedded key can sign any accepted image, forever, until the bootloader itself is rebuilt and reflashed. This proposal adds exactly one new concept on top of this existing infrastructure: the key's position in the array also serves as a generation number, and the bootloader can optionally compare generation numbers.

### 6.1 New build option: `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY`

Off by default. With the option off, MCUboot behaves exactly as it does today. This section describes the behavior when the option is on.

### 6.2 The trust anchor: the primary slot

In swap mode (test + confirm), only the bootloader ever writes to the primary slot. This makes the primary slot's content trustworthy: whatever image sits there has already passed signature validation and the swap-confirm sequence at some point in the past.

When a candidate image is present in the secondary slot and a swap decision is pending, the bootloader determines the key index of the currently active image in the primary slot, using the same key-matching mechanism it already uses for the candidate image (`bootutil_find_key()`).

Two outcomes:

- **No key match for the primary slot.** This happens on first boot with a factory image, during recovery flows, or if the primary slot was signed with a key no longer embedded in the current bootloader build. There is no floor to enforce. The bootloader falls back to today's behavior: it accepts any candidate that passes the existing signature check. The device stays recoverable.
- **A key match is found.** This key's index becomes the floor for the comparison in 6.3.

This reuses the signature-verification machinery MCUboot already runs for the candidate image. No new persistent state, no new NV counter, no new TLV.

### 6.3 The new check

The bootloader accepts an upgrade only when all of the following hold. The first two are existing MCUboot behavior:

1. The candidate's version, or security counter, is equal to or higher than the stored value.
2. The candidate's signature is valid against one of the embedded keys.
3. **New.** If a key floor was found for the primary slot (6.2), the candidate's key index must be equal to or higher than that floor.

Check 3 only runs while a swap decision is pending (a candidate image exists in the secondary slot). A normal boot with no candidate image never triggers it.

A failed check 3 is treated exactly like a failed check 2: the image is rejected through the existing "bad image" path, and — like a genuinely invalid signature — the rejected candidate is erased from the secondary slot, since it can never become valid on its own.

### 6.4 Key ordering is a release-process contract, not a technical mechanism

The order of keys in the embedded key array defines the generation order. Index 0 is the oldest key; the highest index is the newest. This ordering is fixed at bootloader-build time. It must be documented, versioned, and treated with the same care as the signing policy itself. Getting the order wrong causes wrong revocation: a key inserted at the wrong position revokes every key at a lower index in one step. This is a process risk, not a code risk, and no code mechanism enforces correct ordering beyond documentation.

### 6.5 Flow

```mermaid
flowchart TD
    A[New image in secondary slot] --> B{Version / security counter valid?}
    B -- No --> X1[["Exit: rejected (existing behavior)"]]
    B -- Yes --> C{Signature valid?}
    C -- No --> X2[["Exit: rejected (existing behavior)"]]
    C -- Yes --> K{MCUBOOT_KEY_REVOCATION_FROM_PRIMARY enabled?}
    K -- No --> F[Swap starts as test]
    K -- Yes --> D[Read key index of active image from primary slot]
    D --> M{Key match found?}
    M -- No, no floor --> F
    M -- Yes --> E{Candidate key index >= active key index?}
    E -- No --> X3[["Exit: rejected, key too old"]]
    E -- Yes --> F
    F --> G{App confirms before next reset?}
    G -- No --> X4[["Exit: automatic revert (existing behavior)"]]
    G -- Yes --> H[Swap becomes permanent; primary slot is the new anchor]

    classDef new fill:#9f6,stroke:#2b7a2b,stroke-width:2px,color:#000;
    class K,D,M,E,X3 new
```

### 7.1 Key backends

MCUboot supports three key backends. We reviewed how each one determines "which key validated this image," in `boot/bootutil/src/image_validate.c` and `boot/bootutil/src/bootutil_find_key.c`:

| Backend | Key index available? | Works with this proposal? |
|---|---|---|
| Default embedded key array (`bootutil_keys[]`, KEYHASH TLV) | Yes — `bootutil_find_key()` returns the array index. | Yes. |
| `MCUBOOT_HW_KEY` (full public key TLV, hash checked against a single hardware/KMU-provided value) | Formally yes, `bootutil_find_key(image_index, ...)` returns an `int`, but in practice **no**: this backend verifies the embedded public key against one hardware-provided hash and always returns index `0` on success (see `boot/bootutil/src/bootutil_find_key.c`). There is no array of candidate keys to index into at the `bootutil` level — a KMU/OTP-backed key slot is out of scope for this proposal, not an instance of it. | No. Every image would report the same key index, collapsing the floor to a constant and making the check a silent no-op. The implementation raises a compile-time `#error` if both options are enabled together. |
| `MCUBOOT_BUILTIN_KEY` | No. `bootutil_find_key()` is not even compiled in this configuration. `key_id` is set to `image_index` directly — that is, which image slot, not which of several candidate keys. The underlying crypto backend (for example PSA) owns one opaque key per image slot; there is no array of candidate keys at the `bootutil` level. | No. The implementation raises a compile-time `#error` if both options are enabled together. |

### 7.2 Upgrade strategy

This design requires swap mode with test and confirm. It requires that only the bootloader can write the primary slot, which does not hold in overwrite-only mode, direct-XIP, or RAM-load (there is no bootloader-owned trust anchor to compare against). The implementation raises a compile-time `#error` if `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` is enabled together with one of these upgrade strategies.

### 7.3 Multi-image builds

With multiple images (for example an application core and a network core, each independently updatable), each image has its own key array context and its own `image_index`. The key floor is computed and enforced per image; there is no shared or global floor across images.

### 7.4 Optional hardening

`MCUBOOT_VALIDATE_PRIMARY_SLOT` already exists and re-validates the primary image's full integrity on every boot, not only at swap-decision time. This option is orthogonal to our proposal and can be enabled independently for additional hardening.

### 7.5 Multi-signed images (transition images)

If an image carries more than one valid `(key, signature)` TLV pair — for example a key-rotation transition image deliberately signed with both the old and the new key so it remains verifiable by bootloaders that only know one of the two — the *strongest* (highest-index) key among all validly-verified signatures is used as the floor, regardless of the order the TLV pairs appear in the image.

### 7.6 Encrypted images (`MCUBOOT_ENC_IMAGES`)

This proposal is compatible with image encryption, unchanged. We checked this against `docs/encrypted_images.md` and `boot/bootutil/src/bootutil_img_hash.c`. Two facts make this straightforward:

- Encryption only ever covers the firmware payload. The header and all TLVs, including the key-hash TLV and the signature TLV, are always stored and processed in plain form, by design. Reading the key index from the primary slot (section 6.2) only needs the plain key-hash TLV; it needs no decryption step, with or without `MCUBOOT_ENC_IMAGES` enabled.
- MCUboot treats an encrypted payload found in the primary slot as an error condition. The payload is decrypted once, during the copy from the secondary to the primary slot; from then on, the primary slot holds a plain payload. This means that even a full re-validation of the primary slot's signature (as an optional hardening step, alongside `MCUBOOT_VALIDATE_PRIMARY_SLOT`) needs no live decryption for the primary slot either.

We see no interaction, and no additional implementation cost, between this proposal and `MCUBOOT_ENC_IMAGES`.

### 8.1 Fault-injection hardening

MCUboot's existing security-relevant comparisons (for example the security-counter check in `image_validate.c`) use the project's fault-injection-hardening (FIH) primitives (`fih_int`, `fih_ret_encode_zero_equality`, `FIH_CALL`/`FIH_EQ`), not a plain C comparison. The review of PR #2305 raised exactly this concern about a different key-related comparison: a single, unprotected `if (key == expected)` is a fault-injection (glitching) target. The reference implementation carries this all the way down: the candidate's and the primary slot's winning key index are themselves `fih_int` values (not plain `int`) from the point they leave `bootutil_img_validate()`, decoded only at the point of comparison via `fih_int_decode()`, with the comparison result itself wrapped in `fih_ret_encode_zero_equality()` — the same pattern already used for the security-counter check, applied to the key id end-to-end rather than only at the final comparison.

### 8.2 What this does not protect against

- It does not protect a key from leaking in the first place.
- It does not protect the primary slot's trustworthiness against a debug port or other means to write flash outside the bootloader. Locking or disabling JTAG/SWD in production builds is a standard, separate mitigation, independent of this proposal.
- A leaked key at the highest embedded index cannot be revoked by this mechanism, because there is no higher key to replace it with. Operational key management (holding a reserve, unprovisioned key for exactly this case) is a deployment decision, not something this proposal can enforce in code.
- This mechanism narrows what MCUboot accepts. It never widens it. In the worst case (misconfiguration, or no key match found for the primary slot), behavior falls back to today's unmodified acceptance rules — it does not fail open into accepting something it otherwise would have rejected, nor does a bug in this feature bypass the existing signature or version/security-counter checks, which run first and are unmodified.
- **This only starts rejecting an older key once the primary slot has itself been upgraded to, and confirmed, an image signed with a newer key.** There is no independently-provisioned "latest known key" outside of what is currently installed and confirmed. Until that first rotation happens on a given device, images signed with any previously-accepted key are still treated as valid upgrades; a device that is never upgraded past its factory key never benefits from the check at all. This is a progressive, software-only key revocation *based on the currently installed image*, not an equivalent of a hardware/OTP-backed key-revocation mechanism (e.g. a KMU with an independent monotonic key-generation counter): it can also be undone by any means that writes the primary slot outside the normal, bootloader-mediated upgrade path (see the JTAG/SWD point above).

### 8.3 Relationship to existing anti-rollback mechanisms

Version-based and security-counter-based downgrade prevention are unrelated, unmodified, and remain fully active. They protect against replaying an older, still-validly-signed image. This proposal protects against a different threat: a currently-valid signature from a key that should no longer be trusted, independent of the image's version or security-counter value. The two mechanisms are complementary, not overlapping, and this proposal should be documented and adopted as an addition to anti-rollback counters, never as a substitute for them.

## 9. New configuration surface

| Name | Type | Default | Effect |
|---|---|---|---|
| `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` (Zephyr: `CONFIG_BOOT_KEY_REVOCATION_FROM_PRIMARY`) | build option | off | Enables the check described in section 6.3. |

No new NV storage, no new image TLV, no new error code. A rejection is logged under the existing bad-image path (message text only, not a new return code) to aid field debugging.

## 10. Compile-time error conditions

The reference implementation rejects the following invalid configurations at build time, not at runtime:

- `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` together with `MCUBOOT_BUILTIN_KEY` — no key index is available (section 7.1).
- `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` together with `MCUBOOT_HW_KEY` — every image reports key index 0 under this backend, which would collapse the floor to a constant and silently turn the check into a no-op (section 7.1).
- `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` together with overwrite-only, direct-XIP, or RAM-load upgrade strategies — the primary slot is not a trust anchor in these modes (section 7.2).
- `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` together with `MCUBOOT_BYPASS_KEY_MATCH` — with a single key always matching, there is no floor to compute; this combination is a no-op at best and a silent misconfiguration at worst, so it is rejected at compile time rather than accepted as a feature flag that does nothing.

## 11. Reference implementation

We've implemented this design; see the linked pull request for the full diff. Summary of what changed:

- `bootutil_img_validate()` / the loader now report the winning key index for both the primary and the secondary slot (previously an internal-only detail of `image_validate.c`).
- The new check lives next to `check_downgrade_prevention()` in `boot/bootutil/src/loader.c`, alongside the existing version/security-counter comparison, and reuses the FIH comparison pattern.
- The compile-time guards from section 10 are implemented in `bootutil_priv.h` / Kconfig.
- No changes to `imgtool`, no new TLV, no changes to the on-disk image format.
- Zephyr Kconfig option `CONFIG_BOOT_KEY_REVOCATION_FROM_PRIMARY`, documented in `docs/design.md` and a release note.
- Simulator (`sim/`) test coverage for: same-key upgrade (unaffected), forward rotation (allowed), stale-key rejection (rejected + candidate erased), no floor when the primary key is unrecognized or its signature is invalid, bootstrap/first-boot coexistence, multi-image independent floors, and multi-signed "transition" images regardless of TLV order.

## 12. Alternatives considered

- **A revocation list (stored on-device or fetched from a server).** Rejected: requires new persistent state and, for a server-fetched list, a network dependency and trust in a delivery channel. Contradicts the goal of a fully offline, decentralized mechanism.
- **A key-ID TLV plus an explicit "revoke" signal in the image, as attempted in PR #2305.** Rejected for the same reasons the community rejected it: added complexity, a design coupled to one crypto backend (PSA), and a fault-injection-fragile comparison. This proposal reuses the existing key-hash matching mechanism instead of adding a new TLV.
- **N-of-M multi-signature policy.** Out of scope. It solves a different problem (requiring multiple simultaneous signers) and was explicitly identified during the PR #2305 review as not something MCUboot implements or needs to implement for this purpose.
- **Hardware-backed revocation (OTP fuses, secure element counters).** Complementary, not a replacement. It requires specific hardware support that not all target platforms have. This proposal works on plain flash and is a fallback/lower bound, not a competitor, for products that do have such hardware.
- **Key delegation / certificate chains.** Embed a single root public key in the bootloader. Each image carries a short-lived "signing-key certificate": a firmware-signing public key, signed by the root key. The bootloader first verifies the certificate against the embedded root key, then uses the certified firmware key to verify the image itself. Issue #221 already anticipated this direction in 2019: "key revocation becomes a lot more important when we allow key chains instead of just individual keys." We see this as a valid, larger alternative, but not one we propose here: it does not solve revocation by itself (the root key still cannot be revoked without a bootloader update), and it needs a new certificate format plus changes to the signing pipeline and `imgtool` — a materially larger and higher-risk change. A delegation/chain model is worth its own, separate proposal, and is compatible with (not blocked by) accepting this one first.

---

We're happy to discuss this further and would appreciate any feedback, questions, or pushback from the maintainers and community — whether on the overall direction or on specific details. Looking forward to a constructive discussion.

Note for transparency: parts of this write-up and parts of the reference implementation were drafted with AI assistance and then reviewed and edited by us.

michaelcontento:
Closes #2857 — see that issue for the full design discussion, rationale, and alternatives considered. This PR is the reference implementation for the proposal described there.

## Changes

- `boot/bootutil/src/loader.c`, `bootutil_priv.h`, `bootutil_loader.{c,h}`: core key-floor check, next to the existing `check_downgrade_prevention()` logic. Candidate/primary key indices are carried as `fih_int`, decoded only at the comparison point, matching the FIH treatment used for the security-counter check.
- `boot/bootutil/src/image_validate.c`, `include/bootutil/image.h`: expose the winning key index (`fih_int`) from primary/secondary validation.
- `boot/zephyr/Kconfig`, `mcuboot_config.h`, `samples/mcuboot_config/mcuboot_config.template.h`: new `CONFIG_BOOT_KEY_REVOCATION_FROM_PRIMARY` / `MCUBOOT_KEY_REVOCATION_FROM_PRIMARY` option.
- `bootutil_priv.h` / Kconfig: compile-time guards against `MCUBOOT_BUILTIN_KEY`, `MCUBOOT_BYPASS_KEY_MATCH`, `MCUBOOT_HW_KEY` (always resolves to key index 0), and non-swap upgrade strategies — all incompatible with a meaningful key floor.
- `docs/design.md`: new "Key revocation" section (mechanism, limits, key-ordering process responsibility, release-process checklist).
- `docs/release-notes.d/key-revocation-from-primary.md`: release note.
- `sim/`: simulator support (new `key-revocation` / `sig-second-key` Cargo features, `image.rs`, `tlv.rs`) plus `sim/tests/core.rs` coverage for: same-key upgrades (unaffected), forward rotation (allowed), stale-key rejection (rejected + candidate erased), no floor when the primary key is unrecognized/invalid, bootstrap/first-boot coexistence, per-image independent floors in multi-image builds, and multi-signed transition images regardless of TLV order.
- Minor mechanical fallout from the `image_validate.c` signature change (`boot_serial*.c`, `single_loader.c`, `firmware_loader.c`, `hooks.c`) to thread the new key-index output parameter (`fih_int *`) through existing call sites.

## Testing

- `sim/tests/core.rs`: new `key_revocation` test module (see above). All existing simulator tests continue to pass.
- `cargo test --features "sig-ed25519,sig-second-key,swap-move,key-revocation"` (plus `bootstrap` and `multiimage` variants).
- CI: extended `.github/workflows/sim.yaml` to exercise the new feature combinations.

---

We're happy to discuss this further and would appreciate any feedback, questions, or pushback — happy to iterate on this. Looking forward to a constructive discussion.

Note for transparency: parts of this write-up and parts of the implementation were drafted with AI assistance and then reviewed and edited by us.


de-nordic:
Not going to bother to read nor maintain code nobody bothered to write. I am out.

michaelcontento:
This PR was created with the help of AI and I've been open about that. If that's not welcome, that's totally fine.

Feel free to treat this PR as a pure proof-of-concept! Nonetheless I'd really appreciate any feedback on the general solution in the RFC-style issue #2857 :-)

PR #2860, opened 2026-09-10
Fix several formats in log output

josesimoes:
- Now are using proper format specifiers.
- This was producing wrong ouput in log message e.g. `[INF] Image 0 copying the secondary slot to the primary slot: 0xzx bytes`


boot/boot_serial/src/boot_serial.c
```diff
@@ -887,8 +887,8 @@ static off_t erase_range(const struct flash_area *fap, off_t start, off_t end)
     }
 
     size = flash_sector_get_off(&sect) + flash_sector_get_size(&sect) - start;
-    BOOT_LOG_DBG("Erasing range 0x%jx:0x%jx", (intmax_t)start,
-		 (intmax_t)(start + size - 1));
+    BOOT_LOG_DBG("Erasing range 0x%lx:0x%lx", (unsigned long)start,
+		 (unsigned long)(start + size - 1));
```
nordicjm: I think intmax here is correct, whilst unsigned long is correct for a 32-bit device, it is incorrect for a 64-bit device

boot/bootutil/src/bootutil_area.c
```diff
@@ -221,7 +221,7 @@ boot_erase_region(const struct flash_area *fa, uint32_t off, uint32_t size, bool
     int rc = 0;
 
     BOOT_LOG_DBG("boot_erase_region: flash_area %p, offset %" PRIu32 ""
-                 ", size %" PRIu32 ", backwards == %" PRIu8,
+                 ", size %" PRIu32 ", backwards == %d",
                  fa, off, size, (int)backwards);
```
nordicjm: isn't the fix to change `(int)backwards` to `(uint8_t)backwards`?

boot/bootutil/src/loader.c
```diff
@@ -1009,8 +1009,8 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     }
 #endif
 
-    BOOT_LOG_INF("Image %d copying the secondary slot to the primary slot: 0x%zx bytes",
-                 image_index, size);
+    BOOT_LOG_INF("Image %d copying the secondary slot to the primary slot: 0x%lx bytes",
```
nordicjm: size_t is 64-bit on 64-bit platforms so this seems wrong. It's doubtful that a 64-bit platform is going to load an image that is so large, but it is possible

nordicjm:
> This was producing wrong ouput in log message e.g. [INF] Image 0 copying the secondary slot to the primary slot: 0xzx bytes

Those parameters are C99 and above, which I would say should be required for building MCUboot since even GCC version 4 supports it (and gcc4 is the minimum supported GCC version for building MCUboot)

josesimoes:
Thanks, agreed. I've doubled checked the parameter types across the affected code and paying attention to your review notes.

The `%lu` replacements are not portable: both `slot` and the image copy `size` are `size_t`, and `start` is `off_t`, so they may be wider than `unsigned long` on 64-bit targets. I’ll revert those three changes and retain `%zu`/`%zx`/`%jx` as appropriate.

For `backwards`, I suggest keeping `PRIu8` and change the argument cast to `(uint8_t)backwards`. The ECDSA `key_id` change to `%d` with an explicit `int` cast is valid and consistent with the existing Ed25519 logging code.

nordicjm (commented):
please squash changes for review :)

josesimoes:
@nordicjm squashed!

boot/bootutil/src/image_ecdsa.c
```diff
@@ -80,7 +80,7 @@ bootutil_verify_sig(uint8_t *hash, uint32_t hlen, uint8_t *sig, size_t slen,
     bootutil_ecdsa_context ctx;
     FIH_DECLARE(fih_rc, FIH_FAILURE);
 
-    BOOT_LOG_DBG("bootutil_verify_sig: ECDSA embedded key %hhd", key_id);
+    BOOT_LOG_DBG("bootutil_verify_sig: ECDSA embedded key %d", (int)key_id);
```
nordicjm: since key_id is uint8_t should this be changed to `PRIu8` and kept as uint8_t?

josesimoes: `%d` with the explicit `(int)` cast is intentional and avoids potential portability issues across the supported logging backends, so I think the current form is preferable.

nordicjm: but doesn't PRIu8 match exactly what is in bootutil_area.c in the above file? Not following it https://github.com/mcu-tools/mcuboot/pull/2860/files#diff-b4c03a5751bb1a0117c5d39fdab401243219207eb274f596b86b1c0ad49aea27R224

josesimoes: You’re right; since `PRIu8` is already used for the same kind of value in `bootutil_area.c`, I’ll follow that convention here as well.

nordicjm (approved):

